PR #20, opened 2022-11-17
Add cache option to run.toml which serializes/deserializes the problem.

### LINKED ISSUE #19: serialize the problem to disk
To reduce the latency. We should support this in the API / TOML file.

Some earlier experiments serialized the ODEProblem to disk, see https://github.com/Deltares/Ribasim.jl/tree/main/compile

evetion:
We prefix the cache file with the version of the package, as code changes will invalidate the cache. Note that other changes, such as changing the lsws in the input are not (yet) detected, and thus produce invalid results.

Fixes #19


visr (approved):
This is a great feature to have. Eventually we should see if we can make this automatic, such that it always caches the model, and automatically uses it if it sees that the model is the same.

For safety and to avoid confusion I turned it off in the run.toml now.

PR #32, opened 2022-12-05
Rename `ribasim_cli` to `ribasim`

Hofer-Julian:
This name should fit better to user's expectations.

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -172,7 +172,7 @@ of course, more detail requires more parameters.
 
 ## Installing the compiled executable
 
-Binaries of Ribasim CLI can be downloaded from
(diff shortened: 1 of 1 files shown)


docs/index.qmd
```diff
@@ -172,17 +172,18 @@ of course, more detail requires more parameters.
 
 ## Installing the compiled executable
 
-Binaries of `ribasim_cli` can be downloaded from
+Binaries of Ribasim CLI can be downloaded from
```
visr, line 175:
```suggestion
Binaries of Ribasim CLI (command-line interface) can be downloaded from
```

Might be good to introduce the term CLI at first appearance in the docs.

Hofer-Julian: Sounds good

PR #42, opened 2023-01-18
Deploy docs with CI

### LINKED ISSUE #31: Deploy docs with CI
Documentation is currently build and deployed from @visr's PC.
We should automize that and do that from GitHub Actions instead.

Depends on https://github.com/Deltares/Ribasim.jl/issues/30

Hofer-Julian:
- Adds a `docs` workflow which renders & executes docs on every commit, but only publishes on main
- Move `test/data` to `data`
- Move data helper script into its own file under `utils/testdata.jl`
- Publish isn't tested yet, but we'll see if it works as soon as this PR is merged

Fixes #31


Hofer-Julian:
@evetion As discussed today, I hope that caching could be improved.
The cache step also spits out a lot of warnings.
Is this something you could pick up as a follow-up?

evetion:
Looking at https://github.com/julia-actions/cache, it seems most of Julia is cached by default. I think we want to setup Python and/or Quarto caching, which could indeed be a followup.

Hofer-Julian:
Python and Quarto is irrelevant when it comes to build time.

I find it hard to believe that everything is cached. On my local machine the render takes 5 min, not 20.

evetion:
It will only cache after a successful build, maybe only the next ones are fast?

Hofer-Julian:
> It will only cache after a successful build, maybe only the next ones are fast?

That would be great :)

visr:
Awesome, this will be great to have. Is there a reason demo.ipynb is now committed with output (embedden images)? It's quite large and I'd like to keep the repo mostly code-only to keep it small.

Will it still work if you run "clear outputs" on the notebook, and amend the current commit to make sure the doesn't end up in .git?

Hofer-Julian:
> Is there a reason demo.ipynb is now committed with output (embedden images)? 

No, that was an oversight.
Good that you looked more closely than me.
I've now pushed the notebook without output :)

visr (approved):

PR #43, opened 2023-01-19
Added Legolas with first node/edge schemes.

### LINKED ISSUE #24: validation of inputs
Enable validation of input:
- [ ] rules
- [ ] API
- [ ] code implementation
- [ ] documentation

evetion:
Fixes #24. Legolas only really does validation on a schema level, but the defined struct allows for per row validation. Next up is to validate the graph itself.


src/Ribasim.jl
```diff
@@ -23,6 +25,7 @@ using Serialization: serialize, deserialize
 
 export interpolator, Register, ForwardFill
 
+include("validation.jl")
```
visr, line 28: This file still needs to be checked in.

src/Ribasim.jl
```diff
@@ -14,6 +14,8 @@ using Dictionaries
 using Graphs
 using DataInterpolations: LinearInterpolation
 using DiffEqCallbacks
+import Legolas
+using Legolas: @schema, @version, validate
```
visr, line 18:
```suggestion
using Legolas: Legolas, @schema, @version, validate
```

evetion:
I've added a load of TODOs, not all of them need to be fixed in this PR.

Creating a JSON scheme can be done, but getting all the little rules in there (connectors) will be a pain and there's no automation for it in the current ecosystem.

I had to fix the tests to adhere to the Schema we defined, so that's great already.

src/bmi.jl
```diff
@@ -59,11 +59,19 @@ end
...
+_read_table(entry) = entry
+
+function read_table(entry; schema = nothing)
+    table = _read_table(entry)
+    @assert Tables.istable(table)
+    if !isnothing(schema)
+        sv = schema()
+        validate(Tables.schema(table), sv)
+        R = Legolas.record_type(sv)
+        foreach(R, Tables.rows(table))  # construct each row
```
visr, line 72: Do you know roughly how long this takes for the national schematization? Especially the forcing can contain millions of rows, and sometimes only a fraction of them is used for short calculations.

Also, the result of this loop is not kept, that means that we only get errors, but any changes like [`clamp` in the Legolas tour](https://github.com/beacon-biosignals/Legolas.jl/blob/b22e0ce2ed7e2846c1d62b893e0e1bf63eb1f509/examples/tour.jl#L151-L155) is never done, or not?

src/construction.jl
```diff
@@ -1,18 +1,24 @@
 "Load all Arrow input data to SubDataFrames that are filtered for used IDs"
 function load_data(config::Dict, starttime::DateTime, endtime::DateTime)
-    node = DataFrame(read_table(config["node"]))
-    edge = DataFrame(read_table(config["edge"]))
-    state = DataFrame(read_table(config["state"]))
-    static = DataFrame(read_table(config["static"]))
-    profile = DataFrame(read_table(config["profile"]))
-    forcing = DataFrame(read_table(config["forcing"]))
+    # Load data and validate schema + rows for required field tyopes and values
+    node = read_table(config["node"]; schema = NodeV1SchemaVersion)
```
visr, line 4: Is there any reason to pass the type, and not its instance, `NodeV1SchemaVersion()`? The latter feels a bit more natural. And the `schema` keyword can then be set to `::Union{SchemaVersion, Nothing}`.

src/construction.jl
```diff
@@ -1,18 +1,24 @@
 "Load all Arrow input data to SubDataFrames that are filtered for used IDs"
 function load_data(config::Dict, starttime::DateTime, endtime::DateTime)
-    node = DataFrame(read_table(config["node"]))
-    edge = DataFrame(read_table(config["edge"]))
-    state = DataFrame(read_table(config["state"]))
-    static = DataFrame(read_table(config["static"]))
-    profile = DataFrame(read_table(config["profile"]))
-    forcing = DataFrame(read_table(config["forcing"]))
+    # Load data and validate schema + rows for required field tyopes and values
```
visr, line 3: *types

src/construction.jl
```diff
@@ -1,18 +1,24 @@
...
+    node = read_table(config["node"]; schema = NodeV1SchemaVersion)
+    edge = read_table(config["edge"]; schema = EdgeV1SchemaVersion)
+    state = read_table(config["state"]; schema = StateV1SchemaVersion)
+    static = read_table(config["static"]; schema = StaticV1SchemaVersion)
+    profile = read_table(config["profile"]; schema = ProfileV1SchemaVersion)
+    forcing = read_table(config["forcing"]; schema = ForcingV1SchemaVersion)
+
+    # Validate consistency in the data
+    @assert is_consistent(node, edge, state, static, profile, forcing) "Data is not consistent"
```
visr, line 12: Legolas throws `ArgumentError`s. Perhaps we should not use assertions, see the docstring for `@assert`, "Warning: An assert might be disabled at various optimization levels."

I know I also used them already, but what do you think?

src/construction.jl
```diff
@@ -1,18 +1,24 @@
...
 
     if haskey(config, "ids")
         ids = config["ids"]::Vector{Int}
     else
         # use all ids in the node table if it is not given in the TOML file
         ids = Vector{Int}(node.id)
+        # ids = reduce(vcat, isolated_nodegroups(edge)[1000:end])
```
visr, line 19: can be removed

src/construction.jl
```diff
@@ -29,15 +35,16 @@ function load_data(config::Dict, starttime::DateTime, endtime::DateTime)
...
+    @assert issorted(forcing.time)
     startrow = searchsortedfirst(forcing.time, starttime)
     endrow = searchsortedlast(forcing.time, endtime)
     forcing = @view forcing[startrow:endrow, :]
     # then keep only IDs we use
     forcing = filter(:id => in(ids), forcing; view = true)
 
+    # TODO Is order required for ids?
     @assert issorted(profile.id)
     @assert issorted(static.id)
```
visr, line 47: For the config setting `ids`, it does not need to be ordered. In the static and profile tables, they do need to be sorted currently, otherwise these two lines will return incorrect results:

```julia
profile_rows = searchsorted(df.id, id)
rows = searchsorted(static.id, id)
```

For profile it is documented, but I see in https://deltares.github.io/Ribasim.jl/#static it is not mentioned, do you mind adding that as well to the docs?

src/construction.jl
```diff
@@ -29,15 +35,16 @@ function load_data(config::Dict, starttime::DateTime, endtime::DateTime)
     profile = filter(:id => in(ids), profile; view = true)
 
     # for forcing first get the right time range out
+    @assert issorted(forcing.time)
     startrow = searchsortedfirst(forcing.time, starttime)
     endrow = searchsortedlast(forcing.time, endtime)
     forcing = @view forcing[startrow:endrow, :]
     # then keep only IDs we use
     forcing = filter(:id => in(ids), forcing; view = true)
```
visr, line 43: Good catch that the assert must be moved up because of the searchsorted first.

In general in this function before I only validated the subset that we are using, and now we are validating the whole dataset. If it is cheap enough that is fine, but good to check the cost.

src/construction.jl
```diff
@@ -102,7 +109,7 @@ function create_nodes(node, state, profile, static)::Dictionary{Int, ODESystem}
     emptysys = ODESystem(Equation[], t, [], []; name = :empty)
     sysdict = Dictionary{Int, ODESystem}(node.id, fill(emptysys, nrow(node)))
     # create all node systems
-    for node in eachrow(node)
+    for node in Tables.rows(node)
```
visr, line 112: Does this also return a DataFrameRow, but with the advantage that it also supports other tables?

src/validation.jl
```diff
@@ -0,0 +1,98 @@
...
+    id::Int
+    time::DateTime
+    variable::String = isempty(variable) ? error("Empty variable") : variable
+    value::Float64
+end
+
+function is_consistent(node, edge, state, static, profile, forcing)
+
+    # Check that node ids exist
+    # TODO Do we need to check the reverse as well? All ids in use?
```
visr, line 74: I don't think all IDs need to be in use, such that we are able to test swapping out node types in the network, and use `nodes.arrow` more as a database of all nodes that we have created but are not neccesarily in use currently.

src/bmi.jl
evetion: That's correct, at the moment I only check the types and for empty strings. And I rather throw on error than clamp some values, but that's up for debate.

src/construction.jl
evetion: Because Legolas code requires the type, I passed the type. I can make it an instance and use typeof?

src/construction.jl
evetion: The slowest is indeed validation the forcing on 3M rows, which takes 1.7s.

```julia
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.node"), 1} for 29616 rows.
  0.019118 seconds (206.28 k allocations: 4.905 MiB)
┌ Warning: unsupported ARROW:extension:name type: "geoarrow.linestring", arrow type = Vector{Tuple{Float64, Float64}}
└ @ Arrow ~/.julia/packages/Arrow/QsQ3U/src/eltypes.jl:53
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.edge"), 1} for 39043 rows.
  0.024101 seconds (426.03 k allocations: 10.671 MiB)
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.state"), 1} for 8531 rows.
  0.028598 seconds (75.18 k allocations: 1.541 MiB, 76.58% gc time)
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.static"), 1} for 5464 rows.
  0.010356 seconds (52.27 k allocations: 1.375 MiB)
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.profile"), 1} for 46560 rows.
  0.029184 seconds (601.92 k allocations: 13.447 MiB)
[ Info: Validating Legolas.SchemaVersion{Symbol("ribasim.forcing"), 1} for 3088222 rows.
  1.741670 seconds (33.91 M allocations: 837.849 MiB, 4.59% gc time)
```

src/construction.jl
evetion: I think it's fine. The warning has been there for a long time now, but asserts are always active. There is a PR introducing a `@check` that will never be disabled (https://github.com/JuliaLang/julia/pull/41342), whereas `@assert` could.

src/construction.jl
evetion: Yes x2 👍🏻

src/bmi.jl
visr: Ok yeah right now we don't use something like clamp, and I can't directly think of a use case where it would be better than an error, so let's keep it like this.

Any idea about the performance of this line?

src/construction.jl
visr: Ok if Legolas works on the type we can keep it as is, but https://github.com/Deltares/Ribasim.jl/pull/43/files#diff-cd7159638f60b3d2b30ed4bbe7b622ffcdc834f0a2beafe12f46e219a4aa932aR69 (`sv = schema()`) suggests it doesn't.

src/construction.jl
visr: Ok let's keep the asserts, we can change them if it becomes an issue.

src/construction.jl
visr: Thanks for the numbers, I'd say 1.7s is no reason to worry for now, we don't have to optimize further. But perhaps later on if needed we will validate only what we use.

visr (approved):

PR #44, opened 2023-01-19
add file to reduce OrdinaryDiffEq precompile time

visr:
The defaults for both of these are true, but we don't use these algorithms, so we can save time precompiling.

https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep/

If we don't want any precompilation I think it's sufficient to add `PrecompileDefaultSpecialize = false` to this file.


evetion:
Funny, I came across Preferences.jl today and thought precompile was slow, great stuff! 👍

I think you should make these settings permanent in the Project.toml. This file could be renamed as .example and put in the .gitignore.

visr:
> I think you should make these settings permanent in the Project.toml.

How is that done, and why? Note that these preferences are only set in the `run`, i.e. dev environment.

evetion:
According to https://github.com/JuliaPackaging/Preferences.jl:
> Preferences are stored as TOML dictionaries and are, by default, stored within a (Julia)LocalPreferences.toml file next to the currently-active project. If a preference is "exported", it is instead stored within the (Julia)Project.toml. The intention is to allow shared projects to contain shared preferences, while allowing for users themselves to override those preferences with their own settings in the LocalPreferences.toml file, which should be .gitignored as the name implies.

visr:
Got it. It wasn't clear to me from the README so I made https://github.com/JuliaPackaging/Preferences.jl/pull/49. This should be better.

PR #49, opened 2023-01-26
resolve docs manifest after adding Legolas dep

visr:
Also removes the run Manifest since run more of a developer environment, that is not used in CI anymore.


docs/Manifest.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.5"
+julia_version = "1.9.0-beta3"
```
Hofer-Julian, line 3: That should probably stay 1.8.*.
At least that's what we use in the CI, and having a stable version there sounds reasonable to me.

Hofer-Julian (changes requested):

Hofer-Julian (approved):

PR #50, opened 2023-01-26
Add name to CI jobs

Hofer-Julian:


visr:
This looks good to me, but @evetion should confirm. And I think the branch protection rules have to be updated indicating which statuses are required after this is merged. This goes by name, and the name changed in #49 and this also changes it. But as far as I can tell the new action name first has to have run on main, before we can select it.

evetion (approved):

PR #51, opened 2023-01-30
Use QNDF and simplify Bifurcation ODESystem creation

visr:


evetion:
Can you comment, maybe in code, on why we need the pin for Domainsets and on the change of default solver?

The change to Bifurcation seems great, simplifying things and actually doing what the name implies. Could you add a test for a system with a bifurcation?

visr:
The DomainSets issue is https://github.com/JuliaApproximation/DomainSets.jl/issues/127. I think it only shows up in the `run` environment since there are more packages using it, like Makie. It just needs a nudge to prefer later DomainSets, should be temporary.

The solver seems to better handle larger systems.

For bifurcation, I just want to leave things in a manageble state before focusing on DiffEq. We can add tests if we go back to MTK.

PR #52, opened 2023-01-30
Add timers to integration tests and generate test reports.

evetion:
For Teamcity, the test report generation requires `julia --project -e "using Ribasim; using TestReports; TestReports.test(\"Ribasim\")"`. After that you should have a `testlog.xml` which Teamcity can parse.

Both the place of the `@timeit_debug`, as now timing the two sytems in equations.jl are up for debate and should be changed once we have a proper integration setup running. For now this should provide all the building blocks to have something running.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,4 +1,4 @@
-using Ribasim, Test, SafeTestsets, Aqua
+using Ribasim, Test, SafeTestsets, TimerOutputs, Aqua
 
 include("../utils/testdata.jl")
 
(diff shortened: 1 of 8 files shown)


test/runtests.jl
```diff
@@ -15,15 +15,20 @@ testdata("static.arrow", "lhm/static.arrow")
...
     end
     @safetestset "Configuration" begin
         include("config.jl")
     end
+
     # @safetestset "Water allocation" begin include("alloc.jl") end
     @safetestset "Equations" begin
+        include("../utils/testdata.jl")  # to include Teamcity specific utils
         include("equations.jl")
     end
```
visr, line 30: The way it's set up now is that all test files that are included in a safetestset, can be run by themselves, as long as the test data is available. It'd be nice to keep that property, so that these extra lines go into the test files that need them.

PR #53, opened 2023-01-30
Explain how to add developer dependency on Ribasim

Hofer-Julian:
Now that there is no run manifest anymore this has to be done by the user


visr (approved):
Right, thanks. I should've realized that removing the run Manifest means these docs had to be updated.

Issue #54, opened 2023-01-30
Setting the table


visr:
## TODO

- [x] https://github.com/Deltares/Ribasim.jl/pull/65
- [x] update the docs
- [x] see if we can avoid the DataFrames dependency, (groupby could be replaced by other queries for GPKG, relying on sorted IDs for Arrow, and `searchsorted(ids, id)`
- [x] generalize the forcing update code for non-LSW nodes (@visr)
- [x] "build back better" the validation (@evetion)
- [x] add "active" column to node table, and support using that to run sub-models

Last week @Huite and I talked about restructuring our tables, I'll try to summarize our ideas so others can comment.

## Forcing

Right now the forcing table is in long format, with 4 columns: time, id, variable, value. It needs to be sorted by time. It's nice to have just one table, but it is also a bit odd to have to all the the variables spliced together like that. A user working on a model is probably thinking from the perspective of a variable that it wants to update or visualize. To get direct access to a variable, it needs to be its own column. Another downside of putting variables together is that you might want to mix different types, like Int, Float32 and Float64. Although Arrow supports [unions](https://arrow.apache.org/docs/format/Columnar.html#union-layout), having a column per variable solves that as well.

To give an example, we would transform this current long layout:

time                |       id | variable      | value
------------------- |--------- |-------------- |-------
2019-01-01T00:00:00 |        6 | P             | 2.89352e-10
2019-01-01T00:00:00 |        7 | P             | 2.89352e-10
2020-12-31T00:00:00 |        6 | E\_pot        | 1.15741e-9
2020-12-31T00:00:00 |        7 | E\_pot        | 1.15741e-9

To this wide layout:

time                |       id | P             | E\_pot
------------------- |--------- |-------------- |-------
2019-01-01T00:00:00 |        6 | 2.89352e-10   | 1.15741e-9
2019-01-01T00:00:00 |        7 | 2.89352e-10   | 1.15741e-9
2019-01-02T00:00:00 |        6 | 2.89352e-10   | 1.15741e-9
2019-01-02T00:00:00 |        7 | 2.89352e-10   | 1.15741e-9

However, some variables get updated daily, and others yearly or irregularly. Having one shared time column limits flexibility here. To address that, variables can be grouped together in a table if they share the same time column. So a variable like `target_level` can be separate from the daily meteo forcing table, as shown below. It would also work to keep it together in the daily table with missing data, if that is more convenient for a particular use case.

time                |       id | target_level
[...]

PR #56, opened 2023-02-03
add script to create a GeoPackage from Arrow

visr:
This script is one-off, and not live, but is useful to share just like the other preprocessing scripts in the pre/ folder. Ideally we'd maintain a fully automated pipeline from original LHM data to Ribasim.jl input, but right now there are too many manual steps.

Related to #54

Some file size info:
Just spatial tables (node and edge): 9.7 MB
After adding non-spatial data excluding forcing: 11.5 MB
After adding forcing as well: 158.6 MB

If the forcing is written to a Parquet file, it is only 23.5 MB using the default snappy compression, or 33.0 MB uncompressed.
A forcing Arrow file is 16.5 / 20.9 / 42.3 MB for zstd / lz4 / uncompressed. Probably lz4 is a good default.


visr:
Failing docs are unrelated and fixed in #58.

Hofer-Julian (approved):

PR #60, opened 2023-02-07
Update libribasim environment

Hofer-Julian:
- Add PackageCompiler as dependency
- Add `Manifest.toml`


visr:
This is a bit confusing, but libribasim is the environment that we want to build. PackageCompiler is the tool we want to use to build it. Therefore it shouldn't be in the libribasim environment. There should probably be a new environmentin `create_lib`, just like `create_app` in https://github.com/Deltares/Ribasim.jl/tree/main/compile/ribasim_cli. That would include PackageCompiler. There is also a precompile file there that runs some Ribasim code. Because of this I also add Ribasim to the `create_*` projects, though perhaps it's good to check if it can do without, perhaps PackageCompiler already makes sure to execute it under the `libribasim` environment.

visr:
I also updated the manifest, since #59 changed the Ribasim dependencies a lot, and that wasn't reflected here yet.

Hofer-Julian:
Thanks for the clarification @visr 

> Because of this I also add Ribasim to the create_* projects, though perhaps it's good to check if it can do without, perhaps PackageCompiler already makes sure to execute it under the libribasim environment.

Seems like it also works without depending on Ribasim in the `create_*` environment.

run/run.toml
```diff
@@ -28,58 +28,3 @@ node = "../data/lhm/node.arrow"
 edge = "../data/lhm/edge.arrow"
 
 waterbalance = "../data/lhm/waterbalance.arrow"
```
Hofer-Julian, line 30: Precompile depends on `run`, but that one seems outdated.
The ribasim artifacts don't contain a `forcing-long.arrow` or `waterbalance.arrow`.
https://github.com/visr/ribasim-artifacts/releases/tag/v0.1.0

visr: Ah sorry about that. Forcing long is the old long format, so you can just rename it back to forcing.arrow.

visr:
Good to know Ribasim isn't needed in the create lib env!

run/run.toml
Hofer-Julian: What should I do with waterbalance? :)

visr: Oh that was output, that shouldn't be an issue.

Hofer-Julian: Ah I see, thanks :)

Hofer-Julian:
The PR should be fine to merge now.

visr (approved):

PR #64, opened 2023-02-10
Restructure compile folder

Hofer-Julian:
After a couple of iterations, this now seems to work.


compile/create_binaries/Project.toml
```diff
@@ -1,2 +1,3 @@
 [deps]
 PackageCompiler = "9b87118b-4619-50d2-8e1e-99f35a4d4d9d"
+Ribasim = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
```
visr, line 3: I think you found that the Ribasim dependency was not needed here, could you remove it and also commit the updated manifest after you do?

visr: Would be good to also add TOML and LibGit2 deps explicitly (it works since they are stdlibs).

visr (changes requested):
Thanks! Left some small comments. Also in the `libribasim/Project.toml`, could you add

```toml
[compat]
BasicModelInterface = "0.1"
julia = "1.8"

```

compile/create_binaries/Project.toml
Hofer-Julian: Thanks.
I've changed it locally for now.
If the output is valid, I will update this branch.

Hofer-Julian: Done

visr (approved):

PR #65, opened 2023-02-10
Use new file formats

### LINKED ISSUE #37: Use WKB (Well Known Binary) encoding in arrow/feather files
As far as I can tell, the GeoArrow spec supports a column with "logical" representation, which are nested arrays of coordinate pairs, as well as WKT and WKB representation.

The setup currently uses the logical representation. This is supported by GDAL and therefore QGIS, but not by e.g. geopandas. The R bindings likely do not either, as they explicitly mention WKB: https://wcjochem.github.io/sfarrow/index.html

This means the current files cannot easily be read with geopandas. You have to read the feather file with pandas instead, and then convert the nested arrays yourself into shapely/pygeos geometries.

Since GDAL supports WKB just fine, it seems an obvious choice to use WKB encoding instead of logical.

visr:
This PR will implemement the new file formats as outlined in #54.
Fixes #37


visr:
I'll stop here to give @evetion a chance to review ;)

evetion (approved):
Somehow a "refresh" button kept appearing above the PR.

PR #66, opened 2023-02-17
Make TOML Config into a struct.

evetion:
Note that I also tried:
- Using FilePaths(Base) for paths
- Using Second from Dates (or even Unitful) for timestepping

as I like them for correctness, but it slows down our development too much. Using Seconds is not supported by DiffEq, so we need to convert back. FilePaths don't support mixing Paths and Strings, and don't overload all methods like normpath with multiple arguments.


Hofer-Julian:
Nice.
Do I see this correctly, that `Configurations.jl` supports the relevant subset of pydantic functionality that we need here?

evetion:
>  Do I see this correctly, that Configurations.jl supports the relevant subset of pydantic functionality that we need here?

No that would actually be https://github.com/Roger-luo/KungIChi.jl, but it seems less mature. So for now, this mostly enables easy construction from Dict (or TOML directly) into nested structs with some defaults.

Hofer-Julian:
> > Do I see this correctly, that Configurations.jl supports the relevant subset of pydantic functionality that we need here?
> 
> No that would actually be https://github.com/Roger-luo/KungIChi.jl, but it seems less mature. So for now, this mostly enables easy construction from Dict (or TOML directly) into nested structs with some defaults.

So what happens if the dict doesn't match the struct?
Does it give helpful error messages?

evetion:
When putting a = 1 in the [forcing] group in the TOML:
```julia
julia> Ribasim.parsefile("run/run.toml")
ERROR: InvalidKeyError: invalid key a, possible keys are: Bifurcation, GeneralUser, GeneralUser_P, HeadBoundary, LevelControl, LevelLink, LSW, NoFlowBoundary... please check documentation for other valid keys
Stacktrace:
```

Or when removing endtime from the TOML:
```julia
julia> Ribasim.parsefile("run/run.toml")
ERROR: expect key: endtime
```

visr:
This seems quite nice. Could you flip the TOML section and keys to do this?

```toml
[LSW]
forcing = "forcing.arrow"
```

And perhaps also do `using Configurations: a, b` so it's easy to see where things are from, and a bit more robust as well. Do we still need to use TOML directly at all or can that be removed from Ribasim.jl and Project.toml? I see Configurations.jl depends on it.

visr:
Regarding the TOML section and keys flip, we'd want to (in the future) support custom table names for a given node type as well, e.g. they currently all support static and forcing, but only 2 support lookup, but it could be that a new node type requires some new table format. So it'd be fine to specify those options somewhere per node type.

evetion:
> This seems quite nice. Could you flip the TOML section and keys to do this?
> 
> ```toml
> [LSW]
> forcing = "forcing.arrow"
> ```

If the relevant issue is updated to document and reflect it ;)

visr:
Ok I updated #54, the flip could go in a separate PR. It could be done separately from loading the tables I think.

evetion:
Thanks! I think this can be merged.

> It could be done separately from loading the tables I think.

I'm overhauling table loading anyway in my next PR, making it more generic and getting the validation back in.

visr:
Sounds good. I think changing the name from `ribasim_forcing_lsw` to `ribasim_lsw_forcing` in the database won't be as much work as the TOML update. I'm working on a PR to turn the preprocessing into a single pipeline so we can more easily re-derive the LHM based dataset, and change it if we need to.

visr (approved):
This Config struct is quite nice. Might be interesting for Wflow as well.

Hofer-Julian:
> This Config struct is quite nice. Might be interesting for Wflow as well.

I think so too, maybe open an issue for Wflow?

PR #75, opened 2023-02-27
Use initial state, make transient forcing work

Huite:
In SQLite, there's no specific datetime type.

This would trip up with a dynamic forcing, as datetime2unix doesn't work for strings.
It seems convenient in general to have a `load_dataframe` function.


visr (approved):
Thanks, yes I only tried this Arrow forcing so far, this makes sense.

PR #88, opened 2023-03-02
write ribasim.cmd in root cli dir on Windows

visr:
This seems to work well for me, also if I put the cli in a path with spaces.

It probably would be good to do something similar for other OSes, not sure what the best approach elsewhere is though.

### DIFF
--- a/compile/create_binaries/create_app.jl
+++ b/compile/create_binaries/create_app.jl
@@ -22,3 +22,22 @@ create_app(
 
 include("add_metadata.jl")
 add_metadata(project_dir, license_file, output_dir, git_repo)
+
+# On Windows, write ribasim.cmd in the output_dir, that starts ribasim.exe.
+# Since the bin dir contains a julia.exe and many DLLs that you may not want in your path,
+# with this script you can put output_dir in your path instead.
+if Sys.iswindows()
+    cmd = raw"""
+    @echo off
+    setlocal
+
+    set bindir=%~dp0bin
+    set path=%bindir%;%path%
+    "%bindir%\ribasim.exe" %*
+
+    endlocal
+    """
+    open(normpath(output_dir, "ribasim.cmd"); write = true) do io
+        print(io, cmd)
+    end
+end


Hofer-Julian:
> It probably would be good to do something similar for other OSes, not sure what the best approach elsewhere is though.

Maybe a shell script that modifies `LD_LIBRARY_PATH`?

Alternatively, we could write a small executable in Rust that handles all three operating systems.

visr:
Perhaps a Rust CLI is not such a bad idea. We could even put our command line argument parsing there.
Feel free to close this if you'd like to pick that up, or merge and later remove.

compile/create_binaries/create_app.jl
```diff
@@ -22,3 +22,22 @@ create_app(
...
 include("add_metadata.jl")
 add_metadata(project_dir, license_file, output_dir, git_repo)
+
+# On Windows, write ribasim.cmd in the output_dir, that starts ribasim.exe.
+# Since the bin dir contains a julia.exe and many DLLs that you may not want in your path,
+# with this script you can put output_dir in your path instead.
+if Sys.iswindows()
+    cmd = raw"""
+    @echo off
+    setlocal
```
Hofer-Julian, line 32: Does Julia cut off the whitespace, or is it now also indented in the script?

Hofer-Julian (approved):
If I put the dir where `ribasim.cmd` lives, can I then just run `ribasim` and it works?
Or do I have to type `ribasim.cmd`?

Hofer-Julian:
> Perhaps a Rust CLI is not such a bad idea. We could even put our command line argument parsing there.

The argument parsing is a good idea, I didn't think of that advantage.

> Feel free to close this if you'd like to pick that up, or merge and later remove.

I am happy to pick this up, put let's merge this first :)

compile/create_binaries/create_app.jl
visr: It cuts it off, also the first newline.

visr:
You can just run `ribasim`, no need to type out `ribasim.cmd`.

compile/create_binaries/create_app.jl
Hofer-Julian: Oh nice.

PR #90, opened 2023-03-02
Create a CLI wrapper in Rust

Hofer-Julian:
The compiled executable works fine, when put in the parent folder of `bin`.
It can be found here: [cli_wrapper.zip](https://github.com/Deltares/Ribasim.jl/files/10872465/cli_wrapper.zip)

Things that still need to be discussed:

- Do we prefer a single source of truth in Rust over two scripts, one in bash one in cmd?
- If we go with Rust, do we want to offload more work to it than setting env vars and passing on arguments?


Hofer-Julian:
The compiled executable works fine.
It can be found here: [cli_wrapper.zip](https://github.com/Deltares/Ribasim.jl/files/10872465/cli_wrapper.zip)

Things that still need to be discussed:

- Do we prefer a single source of truth in Rust over two scripts, one in bash one in cmd?
- If we go with Rust, do we want to offload more work to it than setting env vars and passing on arguments?

Hofer-Julian:
Two more things that came to mind:

- Do we actually have to modify the PATH on Windows? I think I remember that's not necessary if DLLs are in the same directory as the executable
- Does PackageCompiler.jl modify the RPATH of the ribasim executable on Linux? If yes, we don't need a wrapper at all.

visr:
> Do we prefer a single source of truth in Rust over two scripts, one in bash one in cmd?

This would mainly depend on the complexity of the two scripts versus Rust. If the bash and cmd scripts are simple and all they do is to call an exe in a subdir, perhaps not.

If however we could more easily do model validation in Rust before starting the Julia process, in a way that JSON Schemas can be validated, it could be more interesting. Perhaps it could even be an option to have Python also use the Rust code for this, or for other things that it would benefit from.

You answered you Windows PATH question in #92, and about RPATH, [it looks like it](https://github.com/JuliaLang/PackageCompiler.jl/blob/df3543443b8b5740150e28a5638f7c635c2c61eb/docs/src/devdocs/relocatable_part_3.md?plain=1#L177), though the best approach would probably to try it out using the Linux build.

Hofer-Julian:
I think this can be closed for now

PR #95, opened 2023-03-05
organize source files

visr:
This also renames Register to Model.


src/create.jl
```diff
@@ -1,4 +1,4 @@
-function create_connectivity(db::DB)::Connectivity
+function Connectivity(db::DB)::Connectivity
```
Hofer-Julian, line 1: Why aren't these snake case?

src/io.jl
```diff
@@ -45,13 +45,95 @@ DateTime. This is used to convert between the solver's inner float time, and the
...
+    return normpath(config.toml_dir, config.output_dir, path)
+end
+
+parsefile(config_path::AbstractString) =
+    from_toml(Config, config_path; toml_dir = dirname(normpath(config_path)))
+
+# Read into memory for now with read, to avoid locking the file, since it mmaps otherwise.
+# We could pass Mmap.mmap(path) ourselves and make sure it gets closed, since Arrow.Table
+# does not have an io handle to close.
+_read_table(entry::AbstractString) = Arrow.Table(read(entry))
```
Hofer-Julian, line 115: Why prefix with an underscore?

src/create.jl
visr: They are constructors for the structs in solve.jl.

src/io.jl
visr: Sometimes it's done to signify internal methods. Perhaps it could be renamed though, since we don't use leading underscores elsewhere in the code. Although this particular function is currently not used, it will become relevant again for the validation, when that is hooked up again.

Hofer-Julian: Too bad that we have to rely on python-like private conventions instead simply not exporting functions from Julia modules :)

Hofer-Julian (approved):

Issue #97, opened 2023-03-06
merge repos


visr:
Merge
https://github.com/Deltares/ribasim-python
into
https://github.com/Deltares/Ribasim.jl

This will make it easier to keep them synchronized.

visr:
Fixed in #100

Issue #96, opened 2023-03-06
Add ribasim-python and ribasim-qgis to the docs


Hofer-Julian:
- [ ] Example showing how to preprocess and run models with ribasim-python (Ribasim.jl will be run via Julia rather than via the compiled executable)
- [ ] Example showing how to use ribasim-qgis (from doc)

PR #100, opened 2023-03-06
merge ribasim-python in here

visr:
This merges all commits from https://github.com/Deltares/ribasim-python, fixing #97.

- `build`: PackageCompiler and other build tools
- `core`: the Julia package Ribasim.jl
- `docs`: combined Ribasim documentation
- `python`: the Python package ribasim
- `qgis`: the QGIS plugin Ribasim

There is also still `modflow`, `pre`, `run`, `utils`. Perhaps some of those can go in the future.
Merging the docs can be done separately, we have #96 for that.

With this PR I'd also like to propose to rename the repository to Ribasim, since it's not just the Julia package.


Hofer-Julian (approved):
TeamCity will need to be updated, both because of the folder restructuring and the repo rename

PR #103, opened 2023-03-10
add ribasim-python to docs

visr:
Still a draft, since I still need to update [docs.yml](https://github.com/Deltares/Ribasim/blob/main/.github/workflows/docs.yml).
This uses [quartodoc](https://github.com/machow/quartodoc/) for the API reference, and the [include-code-files](https://github.com/quarto-ext/include-code-files) Quarto extension directly include the example Python scripts into the docs.

Some screenshots:

image
image
image
image


docs/_quarto.yml
```diff
@@ -1,23 +1,52 @@
 project:
   type: website
+  resources:
+    - objects.json
```
Hofer-Julian, line 4: What does this do?

docs/python/index.qmd
```diff
@@ -0,0 +1,19 @@
...
+
+The Ribasim Python package aims to make it easy to build, update and analyze Ribasim models
+programmatically.
+
+The Ribasim QGIS plugin allows users to construct a model from scratch without programming.
+For specific tasks, like adding observed rainfall timeseries, it can be faster to use
+Python instead.
+
+One can also use Ribasim Python to build entire models from base data, such that your model
+setup is fully reproducible.
```
Hofer-Julian, line 13: That is much better :)

Hofer-Julian (approved):
Good job!

PR #106, opened 2023-03-13
docs: create hybrid navigation

visr:
Meaning a top navbar for the major chapters, and a sidebar for the sections. Right now only the Python package docs have multiple pages. We still need to re-arrange to have chapters for the core and the QGIS plugin as well.

There seems to be a quarto bug that if I don't add a second sidebar entry, it shows the contents for all chapters, hence I added "To be continued".

Based on:
- https://quarto.org/docs/websites/website-navigation.html#hybrid-navigation
- https://deltares.github.io/iMOD-Documentation/


Hofer-Julian:
CI is still failing

Hofer-Julian (approved):

PR #111, opened 2023-03-14
remove DifferentialEquations.jl

visr:
This sheds some weight that we don't need, perhaps 10-20 indirect dependencies.

https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep/#Common-Example:-Using-only-OrdinaryDiffEq.jl


Hofer-Julian (approved):
Looks good!

PR #112, opened 2023-03-14
restructure docs under main topics

visr:
As discussed with @Hofer-Julian, this follows the lead of the [iMOD documentation](https://deltares.github.io/iMOD-Documentation/), to use the top navigation for the main topics, and uses the sidebar for organizing the main topics.

These are the main topics proposed in this PR:

image

Naming suggestions welcome. We could remove the Julia and Python part and just use "Core" and "Tooling".

The main diff here is that I moved the usage docs from the main page to a separate page under the "Julia core" topic. More work is still needed on the docs as for instance the equations page is getting quite outdated.

Each topic has an index pages that you land on when you click the topic. I wrote some text for those, aiming for a brief overview text with (introduced) links to the other pages in the topics.


Hofer-Julian:
> Naming suggestions welcome. We could remove the Julia and Python part and just use "Core" and "Tooling".

IMO it’s fine as is

Hofer-Julian (approved):

PR #114, opened 2023-03-15
Refactor Ribasim Python

Hofer-Julian:
Python:
- Added CI
- Added dev docs
- Extended README
- Fixed ruff lints
- Removed tox
- Extended environment.yml

Rest:
 - Renamed CI workflows
 - Limited cases where certain workflows are run


docs/python/developer.qmd
```diff
@@ -0,0 +1,80 @@
...
+    "[python]": {
+        "editor.formatOnSave": true,
+        "editor.codeActionsOnSave": {
+            "source.fixAll": true
+        }
+    },
+    "python.formatting.provider": "black",
+    "autoDocstring.docstringFormat": "numpy"
+}
+```
```
visr, line 80: Shall we check these in, `git add -f .vscode/settings.json`? With also `"julia.environmentPath": "core"` I'd say, since the root of the repo is no longer a Julia environment.

python/ribasim/__init__.py
```diff
@@ -1,12 +1,14 @@
...
-from ribasim.basin import Basin
-from ribasim.edge import Edge
-from ribasim.fractional_flow import FractionalFlow
-from ribasim.level_control import LevelControl
-from ribasim.linear_level_connection import LinearLevelConnection
-from ribasim.model import Model
-from ribasim.node import Node
-from ribasim.tabulated_rating_curve import TabulatedRatingCurve
+from ribasim import utils as utils
+from ribasim.basin import Basin as Basin
```
visr, line 5: Did `black` add `as Basin`? I don't particularly mind, but I don't see why it would be preferred.

visr (changes requested):
Some minor comments

python/ribasim/__init__.py
Hofer-Julian: Linters like mypy and ruff complain about unused symbols otherwise.

I saw this pattern in other projects like e.g. https://github.com/tiangolo/fastapi/blob/master/fastapi/__init__.py

visr: I see, thanks, so let's keep it as is.

docs/python/developer.qmd
Hofer-Julian: That way the settings.json gets tracked, which means that also unrelated changes by developers get easily included.
I now added a template instead.

visr: Right, I didn't think of that. Too bad that is not possible, a template file seems like a good alternative.

visr (approved):

Issue #118, opened 2023-03-16
set solver settigns from TOML


visr:
Right now the solver algorithm and settings are hardcoded. We should use a reasonable default and enable users to choose a different one from the TOML file. Probably some related solver settings should also be exposed.

We could start simple with just making a few mappings to solvers and settings that we need.
After that it would be good to see if we can automatically create mappings for all available solvers and settings.

https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/

Issue #116, opened 2023-03-16
Create example models using python in core test suite


visr:
Right now https://github.com/Deltares/Ribasim/blob/main/core/test/basin.jl runs pre-prepared models that it downloads in runtests.jl. This creates an extra step when changing things. Preferably we run the python code before running the tests.

Hofer-Julian:
Yeah, that is not correct

visr:
Let's keep this open after #149 so we can look into further improvements later, meaning the PythonCall solution mentioned in https://github.com/Deltares/Ribasim/pull/149#issue-1654524866.

visr:
Closed in favor of #244

Issue #40, opened 2022-12-12
implicit scheme


Huite:
Here's a rough sketch:

```julia
"""
This is the parameters struct used in the DiffEq, which contains all variables except u & du.
"""
struct Parameters
    level_connections::Vector{LevelConnection?}
    outflow_tables::Vector{OutflowTable?}
    precipitation::Vector{Precipitation}
end

struct LevelConnection
   a::Int
   b::Int
   conductance::Float
end

struct Precipitation
    index::Int
    value::Float
end

function formulate!(p::Precipitation, u, du)
    i = p.index
    du[i] += p.value
    return
end

function formulate!(lc::LevelConnection, u, du)
    a = lc.a
    b = lc.b
    diff = u[b] - u[a]
    flow = lc.conductance * diff
    du[a] = -flow
    du[b] = flow
    return
end

function ode_func!(du, u, p, t)
    @unpack level_connections, outflow_tables, precipitation = p

    for p in precipitation
        formulate!(p, u, du)
    end

    for connection in level_connection
       formulate!(level_connection, u, du)
    end
    
    return
end


u0 = zeros(n_reservoir)
tspan = (0.0, 100.0)
problem = ODEProblem(ode_func!, u0, tspan, parameters)
sol = solve(problem, callback=callbacks)
```

Huite:
For an educational example: https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/
As far as I can tell, we should be able to help a little with the Jacobian if we're choosing an implicit solution.

Huite:
Another thought: 

We know the state variables exactly in advance, i.e. the storage volume. So maybe a nice approach is to filter reservoir nodes from the graph, and enumerate them. This then creates a lookup from the "global node number" to the index in the state variable.

To setup the connections is then a matter of looking up the global node numbers.

In general, I believe it's worth to always do the translation step to internal indices, because the `ode_func` may be called quite a lot for adaptive timestepping.

gijsber:
extend implementation to implicit scheme

gijsber:
set to open because of need to implement implicit scheme

gijsber:
~4d
need a function to create derivative per struct
for every boundary condition: create_jacobian
if the user chooses a implicit solver we provide the jacobian

visr:
[...]

PR #119, opened 2023-03-16
allow setting some solver options for some solvers

visr:
Fixes https://github.com/Deltares/Ribasim/issues/118.
This removes the `saveat` key from the config root, and moves it to the new `solver` section, together with `algorithm` and some other options. The `saveat` Union type is extended, since in TOML `[]` becomes a `Union{}[]` which cannot convert to `Float64[]`.

The algorithms were chosen to cover some different approaches. This switches the default to QNDF, but for that to work we need to set `autodiff=false`, probably until https://github.com/Deltares/Ribasim/issues/40 is fixed. @Huite that means implicit already works, though autodiff support or providing jacobians is probably still worthwhile.

```TOML
[solver]
autodiff = false
```

This still needs docs and proper Python support, I hardcoded `autodiff=false` in the TOML writing for now.


Hofer-Julian:
@visr did you change the ribasim artifacts?

It now fails because it doesn't recognize the solver key.

visr:
Yeah I added this to the TOML files in https://github.com/visr/ribasim-artifacts:

```
[solver]
autodiff = false
```

So you need to remove the TOMLs and let the test suite redownload them, or add this yourself. Another reason to fix #116.

Hofer-Julian:
> Yeah I added this to the TOML files in https://github.com/visr/ribasim-artifacts:
> 
> ```
> [solver]
> autodiff = false
> ```
> 
> So you need to remove the TOMLs and let the test suite redownload them, or add this yourself. Another reason to fix #116.

I am surprised that this still says three weeks ago: https://github.com/visr/ribasim-artifacts/releases/tag/v0.5.0

You must have done this just recently.

visr:
The tag is old, but I updated the TOMLs 11 hours ago, that is indicated. Although I probably should've created a new tag, sorry.

core/src/config.jl
```diff
@@ -74,3 +85,32 @@ function Base.show(io::IO, c::TableOption)
...
+"Create an OrdinaryDiffEqAlgorithm from solver config"
+function algorithm(solver::Solver)::OrdinaryDiffEqAlgorithm
+    algotype = get(algorithms, solver.algorithm, nothing)
+    if isnothing(algotype)
+        options = join(keys(algorithms), ", ")
+        error("Given solver algorithm $(solver.algorithm) not supported.\n\
+            Available options are: ($(options)).")
+    end
+    # not all algorithms support this keyword
+    return try
```
Hofer-Julian, line 111:
```suggestion
    try
```

python/ribasim/model.py
```diff
@@ -31,6 +31,16 @@
 )
 
 
+class Solver(BaseModel):
+    algorithms: Optional[str]
```
Hofer-Julian, line 35: Does it make sense to make all fields optional?
I feel like at least `algorithms` should be required.

Also, why `algorithms` and not `algorithm`?

Hofer-Julian (approved):
Looks good apart from two small comments

python/ribasim/model.py
visr: Thanks, that should be singular, fixed.

They are all optional in the sense that in the julia code we pick reasonable defaults in the Solver struct. I could copy those defaults here, but that would mean they are always written out to the TOML file if a model is generated from Python. Do we want big explicit TOMLs with everything spelled out, or do we want to omit optional keys and leave the core to fill them in?

Hofer-Julian: Fair enough.

visr: Why do you think algorithm should be required? I think most users don't know or don't care about this, I don't want to force them to pick one, we can pick one for them that works well in most scenarios (though perhaps not the best in all).

gijsber: Agree, we pick a default that works in most situations but allow the modeller to change  as needed

Hofer-Julian: I think the status quo is fine for me as is.

But making algorithms required wouldn't change that Solver as a whole is optional.
It would just force people who mess with solver settings to also specify the algorithm

visr: Right, I understand. Though only the autodiff setting directly relates to the algorithm (i.e. is a keyword of).

Issue #126, opened 2023-03-17
Devdocs: how to add a new node type


visr:
Split from #21.
Use this PR as a guide: #127 
Possibly directly do some light refactoring if it is possible to make it slightly easier, perhaps there are too many node lists in the code.

evetion:
Closed by #222

PR #127, opened 2023-03-17
add Pump

### LINKED ISSUE #21: add pump/pumpstation object (pomp/gemaal)
Add component for a pumpstation with one or more pumps

- [x] conceptual design: simple pump with a capacity from the forcing table that reduces extractions on an empty basin
- [ ] implementation
- [ ] unit tests (needs #116 first)
- [ ] input validation
- [ ] updated example model
- [ ] updated tutorial/example model building script
- [x] devdocs: how to add a new node type (moved to #126)
- [ ] include as nodetype in QGIS

### LINKED ISSUE #116: Create example models using python in core test suite

visr:
Fixes #21. Except for the devdocs (#126) and unit tests. Since there are no proper unit tests for node types yet, I prefer to fix #116 first before addressing that.

A pump is added to the basic model which runs in the docs, and based on the output from there it seems to work ok, stopping to pump when the upstream basin is empty.

### DIFF
--- a/docs/core/usage.qmd
+++ b/docs/core/usage.qmd
@@ -248,7 +248,7 @@ discharge | Float64 | $m^3 s^{-1}$ | non-negative
 
 Pump water from a source node to a destination node.
 The set flow rate will be pumped unless the intake storage is less than 10m3,
-in which case the flow rate will be linearly reduced to 0 m3/s.
+in which case the flow rate will be linearly reduced to $0~m^3/s$.
 A negative flow rate means pumping against the edge direction.
 Note that the intake must always be a Basin.
 
--- a/docs/core/usage.qmd
+++ b/docs/core/usage.qmd
@@ -247,7 +247,7 @@ discharge | Float64 | $m^3 s^{-1}$ | non-negative
 ### Pump
 
 Pump water from a source node to a destination node.
-The set flow rate will be pumped unless the intake storage is less than 10m3,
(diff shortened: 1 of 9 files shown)


docs/core/usage.qmd
```diff
@@ -244,6 +244,19 @@ node_id   | Int     | -            | sorted
 storage   | Float64 | $m^3$        | non-negative, start at 0
 discharge | Float64 | $m^3 s^{-1}$ | non-negative
 
+### Pump
+
+Pump water from a source node to a destination node.
+The set flow rate will be pumped unless the intake storage is less than 10m3,
```
Hofer-Julian, line 250:
```suggestion
The set flow rate will be pumped unless the intake storage is less than 10 $m^3$,
```

docs/core/usage.qmd
```diff
@@ -244,6 +244,19 @@ node_id   | Int     | -            | sorted
...
+in which case the flow rate will be linearly reduced to 0 m3/s.
```
Hofer-Julian, line 251:
```suggestion
in which case the flow rate will be linearly reduced to 0 $m^3/s$.
```

Hofer-Julian (approved):
Looks good apart from minor comments

PR #128, opened 2023-03-21
use regular Dict, remove Dictionaries

visr:
From the Dictionaries README:

> The three main difference to Dict are that it preserves the order of elements, it iterates much faster, and it iterates values rather than key-value pairs.

In the current code we have only one `Dictionary`, the `u_index` (basin node ID to state index), and we don't iterate over it. So it seems to me the advantages of Dictionaries don't apply anymore, which means we can drop a dependency.


evetion (approved):

core/src/create.jl
```diff
@@ -5,7 +5,7 @@ function Connectivity(db::DB)::Connectivity
     nonzeros(flow) .= 0.0
 
     basin_id = get_ids(db, "Basin")
-    u_index = Dictionary(basin_id, 1:length(basin_id))
+    u_index = Dict(k => v for (v, k) in enumerate(basin_id))
```
visr, line 8:
```suggestion
    u_index = Dict(id => i for (i, id) in enumerate(basin_id))
```

PR #136, opened 2023-03-24
TabulatedRatingCurve: use level instead of storage

visr:
Also included here:
- don't plot the axis box and ticks by default, only the network
- seed the precipitation for reproducability
- don't write model to tempdir, but to gitignored local dir, for easier development


Huite (commented):
lgtm

Huite (approved):

PR #140, opened 2023-03-27
libribasim: support getting state and time info

Hofer-Julian:


visr:
This doesn't pass the libribasim tests from Python in TeamCity yet, but that is still being set up.
This PR is probably already big enough, plus it is tested from the Julia side as well.

build/libribasim/src/libribasim.jl
```diff
@@ -71,10 +91,61 @@ Base.@ccallable function update_until(time::Cdouble)::Cint
...
+            julia_type_to_numpy(typeof(value))
+        else
+            error("Unsupported value type $(typeof(value))")
+        end
+
+        var_type_ptr = pointer(var_type)
+        for (i, char) in enumerate(ascii(dtype))
+            unsafe_store!(var_type_ptr, char, i)
+        end
+        unsafe_store!(var_type_ptr, '\0', length(dtype) + 1)
```
evetion, line 116: Might be nice to have an (maybe private) unsafe_store!(::Cstring, dtype) that handles this.

build/libribasim/src/libribasim.jl
```diff
@@ -71,10 +91,61 @@ Base.@ccallable function update_until(time::Cdouble)::Cint
...
+Base.@ccallable function get_var_type(name::Cstring, var_type::Cstring)::Cint
+    @try_c begin
+        value = BMI.get_value_ptr(model, unsafe_string(name))
+        dtype = if value isa Vector
+            julia_type_to_numpy(eltype(value))
+        elseif value isa Number
+            julia_type_to_numpy(typeof(value))
+        else
+            error("Unsupported value type $(typeof(value))")
+        end
```
evetion, line 110: This seems to be more nicely with multiple dispatch (even though it doesn't improve performance), or at least to this code within another function. It can yield roughly the same error here as in julia_type_to_numpy (only supporting double now).

build/libribasim/src/libribasim.jl
```diff
@@ -71,10 +91,61 @@ Base.@ccallable function update_until(time::Cdouble)::Cint
...
-        BMI.get_current_time(model)
+        value = BMI.get_value_ptr(model, unsafe_string(name))
+        n = ndims(value)
+        unsafe_store!(rank, n)
+    end
+end
+
+Base.@ccallable function get_value_ptr(name::Cstring, value_ptr::Ptr{Ptr{Cvoid}})::Cint
+    @try_c begin
+        value = BMI.get_value_ptr(model, unsafe_string(name))
```
evetion, line 130: Does Julia know the return type here? Is this always a vector? An annotation could help the reader as well.

build/libribasim/src/libribasim.jl
```diff
@@ -71,10 +91,61 @@ Base.@ccallable function update_until(time::Cdouble)::Cint
...
+end
+
+Base.@ccallable function get_value_ptr_double(
+    name::Cstring,
+    value_ptr::Ptr{Ptr{Cvoid}},
+)::Cint
+    get_value_ptr(name, value_ptr)
+end
+
+function julia_type_to_numpy(type)::String
```
evetion, line 144: Our C API is numpy specific? I'm not sure how BMI would work with other languages when one can retrieve the dtype.

python/ribasim_api/tests/test_bmi.py
```diff
@@ -1,9 +1,47 @@
...
+
 def test_update(ribasim_basic):
     libribasim, config_file = ribasim_basic
     libribasim.initialize(config_file)
     libribasim.update()
+    time = libribasim.get_current_time()
+    assert time > 0.0
+
+
+@pytest.mark.skip(reason="update_until not in xmipy, see issue #92")
```
evetion, line 26: Make this into a link to xmipy#92?

utils/testdata.jl
```diff
@@ -32,4 +32,31 @@ function teamcity_message(name, d::Dict)
     )
 end
 
+# Taken from Julia's testsuite
```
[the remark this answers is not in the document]
evetion: Nicely fixed with IOCapture 👍🏻

PR #148, opened 2023-04-04
draw edges with arrows

### LINKED ISSUE #147: QGIS: draw edges with arrows
Or make it clear that edges are directed in another way. Right now only a simple line is drawn.

https://deltares.github.io/Ribasim/qgis/#plot-time-series

visr:
Fixes #147

QGIS
image

`edge.plot()`
Code_agWsmWbAYx

### DIFF
--- a/python/ribasim/ribasim/edge.py
+++ b/python/ribasim/ribasim/edge.py
@@ -1,8 +1,10 @@
 from typing import Any
 
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 import pandera as pa
+import shapely
 from pandera.typing import DataFrame, Series
 from pandera.typing.geopandas import GeoSeries
 from pydantic import BaseModel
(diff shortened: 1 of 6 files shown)


visr:
@Huite I don't see an easy way to vectorize drawing the arrows, this doesn't work:

```
import matplotlib.pyplot as plt
plt.scatter([1.2,1.6],[0.4,0.2],marker=([3,3],[0,0],[90,90]))
plt.scatter([1.2,1.6],[0.4,0.2],marker=[(3,0,90),(3,0,90)])
```

Perhaps it's fine for now, I don't expect it to quickly become a performance issue.

Huite:
I've done a little research -- I figured I might as well, since I might add comparable functionality to xugrid, which could feature relatively large network.

Computing, x, y, and angle is easy to thanks to Shapely 2.0 functionality. However, you're right that `scatter` from matplotlib doesn't support rotation, one by one. This is because it roughly does the following thing:

* Check if the marker is already a Matplotlib marker, if not make it (from `(3, 0, 0)`.
* Next, "offsets" for every individual marker (the x, y locations).
* Create a PathCollection from the individual marker and the offsets.
* Add the PathCollection to the ax.

A path collection requires either a single matplotlib Path or a list of them, so there's no real gain there.

The most feasible alternative is likely using Quiver, which makes a PolyCollection in the background. To get the same effect, you'd have to set the arrow lengths to zero. Primary problem is that Quiver wants to scale by length of dx, dy (which could be normalized), and it seems tedious in general.

Huite (approved):

visr:
Thanks! I just realized that if the edges are more than one segment, the arrow doesn't fall on the line in matplotlib. In QGIS it draws one per segment. But that would be a good follow up.

PR #163, opened 2023-04-14
Store mapping of (src, dst) -> edge fid

Huite:


visr (commented):
Could you update the Edge and Flow output tables in usage.qmd as well?

PR #166, opened 2023-04-16
Enable validation by adding schemas for all IO layer rows. Added / static postfix.

### LINKED ISSUE #24: validation of inputs

### LINKED ISSUE #144: don't special case static tables
In the data model we have a set of possible tables per node type, like this:

```
Basin / static
Basin / forcing
Basic / profile
Basic / state
```

Right now in the QGIS plugin we use "Basin" to mean "Basin / static". It's better to remove this implicit assumption. It also doesn't align with how things are set up in the core and ribasim-python.

https://github.com/Deltares/Ribasim/blob/ac757a626deafd830004f3c6f6c32d422d1cc88e/qgis/core/nodes.py#L327-L339

Edit: this also applies to the core and ribasim-python, e.g. for TabulatedRatingCurve with `static` and `time` tables it uses `TabulatedRatingCurve` and `TabulatedRatingCurve / time`.

evetion:
Fixes #24 
Fixes #144 

Now one only needs to specify a new schema and record in `validation.jl`, from which both the Config as the validation will be triggered. This is completely discoupled from the compute types, the connection is only made on the call to load_data like `data = load_data(db, config, TabulatedRatingCurveStaticV1)`.

Note that I've taken the liberty to already append `static` to all base tables, so this requires new geopackages. I've also inversed the `Config` to have a [basin] key with forcing underneath instead of [forcing] with basin underneath.

One can generate jsonschema's with `julia --project=core scripts/gen_schema.jl`. These end up in the docs/schema folder, which will then be live on the website as well. In another PR I will automate the Python and QGIS code based on these schemas. Incremental changes to validation can be made now by a single line change, like `node_id::Int = clamp(node_id, 0, typemax(Int))`.


visr (approved):
Great stuff. I pushed a few changes/fixes, but it's good to have this merged.

Issue #171, opened 2023-04-18
Test output tables


visr:
This should probably be done in the core tests, since we run the example models there. The Python IO tests currently only test that the input was written to disk and can be written back in.

visr:
This done for many test models now. Can probably be better, but this issue is too broad.

PR #172, opened 2023-04-18
fix output IDs

visr:
The `sort` is important here since we use a regular unordered Dict. This resulted in the wrong `node_id` in the header of the basin output.

Also adds `edge_id` to the flow output, as intended in #163.

No tests right now since we don't have any tests on the output tables yet. I createad #171 for that.


Huite (approved):
lgtm!

PR #174, opened 2023-04-18
Associate edge output

Huite:


gijsber:
Nice, the only thing I do not really understand is: why do we see 3 and 4 as legend for the flows. I understand the outflow of 3, but I actually expected to outflow of 2 (instead of a reference to 4)

Huite:
Those are the edge_id's. Which is perhaps an argument for also showing the edge_id as a label? But maybe greyed out.

gijsber:
yes, edge_id as label in grey would probably help

visr:
The edge_ids were basically added to support this feature. Otherwise we use (from_node_id, to_node_id) everywhere. The nodes are already labeled. In the plot, is it possible to should this as a label?

PR #178, opened 2023-04-20
Generate Python schemas automatically.

evetion:
Update with 
`datamodel-codegen --use-title-as-name --input docs/schema/root.schema.json --output python/ribasim/ribasim/models.py`

Geometry is now dropped from the schema, so there's 1 column less, failing a test. Not sure what we want with these geometry columns.


visr:
We want a geometry only for the node and edge tables. The geometry for the other tables are a side effect of using geopandas to load them, but this should be fixed.

evetion:
Can someone reproduce the Quarto error?

Essentially what goes wrong is that `structtable.node_id` is a `Vector{Float64}`, where it should be a `Vector{Int}`. But I don't see why, and cannot reproduce it with running `julia --project=core -e 'using Ribasim; Ribasim.run("data/basic-transient/basic-transient.toml")'`
 [1] qh_interpolation(node_id::Int64, table::StructArrays.StructVector{Ribasim.TabulatedRatingCurveStaticV1, NamedTuple{(:node_id, :level, :discharge), Tuple{Vector{Float64}, Vector{Float64}, Vector{Float64}}}, Int64})

visr:
I can reproduce. The SQLite.Query for the "TabulatedRatingCurve / static" table shows `node_id` as `Union{Missing, Float64}`. The Float64 should be a union, and ideally missings are not allowed either.

All the `node_id` columns I see are REAL:
image

Why is that, given these schemas:

```
@version TabulatedRatingCurveStaticV1 begin
    node_id::Int
    level::Float64
    discharge::Float64
end

class TabulatedRatingCurveStatic(BaseModel):
    node_id: int = Field(..., description="node_id")
    discharge: float = Field(..., description="discharge")
    level: float = Field(..., description="level")
```

Huite:
So I guess something in pydantic coerces to float for some reason.

If you take a part of the examples:

```python
    seconds_in_day = 24 * 3600
    q1000 = 1000.0 * 0.01 / seconds_in_day
    rating_curve = ribasim.TabulatedRatingCurve(
        static=pd.DataFrame(
            data={
                "node_id": [2, 2],
                "level": [0.0, 1.0],
                "discharge": [0.0, q1000],
            }
        ),
        time=pd.DataFrame(
            data={
                "node_id": [3, 3, 3, 3, 3, 3],
                "time": [
                    pd.Timestamp("2020-01"),
                    pd.Timestamp("2020-01"),
                    pd.Timestamp("2020-02"),
                    pd.Timestamp("2020-02"),
                    pd.Timestamp("2020-03"),
                    pd.Timestamp("2020-03"),
                ],
                "level": [0.0, 1.0, 0.0, 1.1, 0.0, 1.2],
                "discharge": [0.0, q1000, 0.0, q1000, 0.0, q1000],
            }
        ),
    )
    print(rating_curve.dtypes)
```

You'll get:
```
node_id      float64
discharge    float64
level        float64
dtype: object
```

I'll run the debugger through this...

Huite:
I think we should hold off merging this for now. The first issue is that it doesn't seem to quite work right at the moment:

Compare:

```python
ribasim.pump.StaticSchema.to_schema()

<Schema DataFrameSchema(columns={}, checks=[], index=None, coerce=True, dtype=<class 'ribasim.models.PumpStatic'>, strict=False, name=StaticSchema, ordered=False, unique_column_names=False)>
```

Versus:

```python
import pandera as pa
from pandera.typing import DataFrame, Series


class StaticSchema(pa.SchemaModel):
    node_id: Series[int] = pa.Field(coerce=True)
    flow_rate: Series[float]

StaticSchema.to_schema()

<Schema DataFrameSchema(columns={'node_id': <Schema Column(name=node_id, type=DataType(int64))>, 'flow_rate': <Schema Column(name=flow_rate, type=DataType(float64))>}, checks=[], index=None, coerce=False, dtype=None, strict=False, name=StaticSchema, ordered=False, unique_column_names=False)>
```

It also seems to cast everything to float for me, even when I test with only ints:

```python
# ...
class PumpStatic(BaseModel):
    flow_rate: int = Field(..., description="flow_rate")
    node_id: int = Field(..., description="node_id")

# ...

pump = ribasim.Pump(
    static=pd.DataFrame(
        data={
            "node_id": [7],
            "flow_rate": [0.5 / 3600],
        }
    )
)

pump.static.dtypes

flow_rate    float64
node_id      float64
dtype: object

```

Glancing through the pandera docs, I'm not sure what the problem is. However, I did come across this rather relevant note [here](https://pandera.readthedocs.io/en/stable/pydantic_integration.html#using-pydantic-models-in-pandera-schemas):

> By combining dtype=PydanticModel(...) and coerce=True, pandera will apply the pydantic model validation process to each row of the dataframe, converting the model back to a dictionary with the BaseModel.dict() method.

That's extremely undesirable with regards to performance, but it means that we cannot easily rely on `datamodel-codegen` --> `pydantic` --> `pandera`.

However, pandera has their own JSON conventions as far as I can tell:

https://pandera.readthedocs.io/en/stable/schema_inference.html?highlight=json#write-to-json

If we generate those instead from Julia, we don't need codegen (which I find a bit iffy anyway) and we can just instantiate a schema `from_json`. There's another argument here: we're dealing with tabular data, and general datamodels might not be the richest/most convenient representation. The pandera JSON form obviously is fundamentally columnar. It's more obscure, but something we come up by ourselves is even more obscure, of course.

evetion:
Agreed on holding off for now, we need to guarantee correct types when writing from Python. And we don't understand the Pandera behavior here.

visr:
Another comment from @Hofer-Julian

>We should also ask ourselves how much we gain from this deduplication. I feel maintenance is fine with duplication as long as we have a good testsuite.

And from me, I haven't looked into this yet, but we are already using Arrow on both sides, and Arrow has Schemas. Would that be another option instead of the pandera JSON? The schemas can be serialized by themselves: https://stackoverflow.com/questions/75954819/is-there-a-way-to-deserialize-pyarrow-table-schemas/75956683#75956683. That would cover the biggest tables, which are most performance sensitive. For the GeoPackage we currently assume it's fine to read it all into memory anyway.

Hofer-Julian:
> Another comment from @Hofer-Julian
> 
> > We should also ask ourselves how much we gain from this deduplication. I feel maintenance is fine with duplication as long as we have a good testsuite.

Maybe to add to this.
I am not against this at all, I appreciate the effort and if all pieces work well together it is a clear improvement over the status quo.

I just wanted to point out, that if we find that the ecosystem is not there yet, duplicating a couple of things is acceptable at the moment.

evetion:
Reported the float coercion issue https://github.com/unionai-oss/pandera/issues/1184. The workaround, using Pandera JSON schemas doesn't seem possible, because one can't generate a SchemaModel from it (only a Schema): https://github.com/unionai-oss/pandera/issues/393.

visr (approved):

PR #181, opened 2023-04-21
Add time functions to libribasim

Hofer-Julian:
Also started using `pytest.approx`

### DIFF
--- a/python/ribasim_api/tests/test_bmi.py
+++ b/python/ribasim_api/tests/test_bmi.py
@@ -17,14 +17,14 @@ def test_get_start_time(libribasim, basic, tmp_path):
     config_file = str(tmp_path / f"{basic.modelname}.toml")
     libribasim.initialize(config_file)
     time = libribasim.get_start_time()
-    assert time == 0.0
+    assert time == pytest.approx(0.0)
 
 
 def test_get_current_time(libribasim, basic, tmp_path):
     basic.write(tmp_path)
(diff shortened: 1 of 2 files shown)


python/ribasim_api/tests/test_bmi.py
```diff
@@ -12,12 +12,28 @@ def test_initialize(libribasim, basic, tmp_path):
     libribasim.initialize(config_file)
 
 
+def test_get_start_time(libribasim, basic, tmp_path):
+    basic.write(tmp_path)
+    config_file = str(tmp_path / f"{basic.modelname}.toml")
+    libribasim.initialize(config_file)
+    time = libribasim.get_start_time()
+    assert time == pytest.approx(0.0)
```
visr, line 20:
```suggestion
    assert time == 0.0
```

Wouldn't any slight deviations here be unexpected? This 0 is not the result of any floating point calculations.

visr (approved):
Looks good to me, 1 optional comment.

Is `get_time_units` not needed?

python/ribasim_api/tests/test_bmi.py
Hofer-Julian: I am unsure about best practises for these cases.
My feeling is that it's fine to compare floats like `abs(a - b) < eps` for all tests.

But I am fine with changing it to your suggestion if you prefer that.
It is arguably more precise.

PR #184, opened 2023-04-21
model.write: deleting old geopackage files if they exist

### LINKED ISSUE #173: Python: model.write() will append to existing geopackages
It looks like geopandas will append layers to an existing geopackage. This is not desirable when calling `model.write`: you're writing the entire model. I think we should check if a file with the given name already exists, then delete it before writing.

visr:
Fix #173 for real this time; forgot to include this in #182.


Hofer-Julian (approved):
Ah that makes more sense.
Should have caught that during review :)

Issue #80, opened 2023-03-01
consider Chezy or Manning-based Level connector


gijsber:
the current LinearLevelConnector may be hard to configure/calibrate

a Manning or Chezy based connector using a wetted area (of the upstream storage) and a roughness factor may be easier given available data sources

gijsber:
rectangular profile + roughness
~3-4d

gijsber:
@Huite please rename to ManningResistance (to have naming consistency with the LinearResistance node)

evetion:
+ Investigate oscillations (flow) as part of testing this nodetype. @SouthEndMusic

SouthEndMusic:
It seems that the oscillatory behavior of the flow around the manning resistance disappears if the profiles of the connected basins are physical (which they are **not** in the example and test models), in the way discussed in [#255](https://github.com/Deltares/Ribasim/issues/225). This makes intuitive sense from the fact that the ManningResistance behaviour is rooted in physics.

visr:
Oh that is a nice and reassuring find!

visr:
> It seems that the oscillatory behavior of the flow around the manning resistance disappears if the profiles of the connected basins are physical.

In #280 they are physical, but I still get oscillations, these are the flows over the basic-transient model edges:
![image](https://github.com/Deltares/Ribasim/assets/4471859/bf462c5d-dda2-48b6-b6dc-96287d6d9038)

Versus current main:
![image](https://github.com/Deltares/Ribasim/assets/4471859/e43c48e7-70da-4e16-a9cc-cf87cb020f6c)

SouthEndMusic:
> In #280 they are physical, but I still get oscillations

I see that you now get the storages by integrating over the areas which is nice, but if I am not mistaken the interpolation you use between those storage values is still linear which is unphysical. In the example I made I kept the area constant over the level which makes linear interpolation correct.

![test_manning_model](https://github.com/Deltares/Ribasim/assets/74617371/5613f8ca-486a-4742-94b8-2e3fb3447eff)
![Constant_area_profile](https://github.com/Deltares/Ribasim/assets/74617371/94d27fef-ed81-4cbf-8379-0c1c23a10dad)
![test_manning_flow](https://github.com/Deltares/Ribasim/assets/74617371/c3c4cd5a-d7a7-48c9-9b44-eab3073b9dc7)

SouthEndMusic:
Agreed

PR #186, opened 2023-04-24
add ManningConnection

Huite:
I haven't updated everything in QGIS yet, and I need to setup the tests.

Fixes most of #80

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -1,5 +1,6 @@
 ## types and functions
 const Interpolation = LinearInterpolation{Vector{Float64}, Vector{Float64}, true, Float64}
+const TWO_G = 2 * 9.81
 
 """
 Store the connectivity information
@@ -120,14 +121,15 @@ Requirements:
 struct ManningConnection
     node_id::Vector{Int}
     length::Vector{Float64}
-    n_inverted::Vector{Float64}
-    profile_bottom::Vector{Float64}
+    n_squared::Vector{Float64}
     profile_width::Vector{Float64}
     profile_slope::Vector{Float64}
     profile_slope_unit_length::Vector{Float64}
+    contraction_coefficient:::Vector{Float64}
+    expansion_coefficient:::Vector{Float64}
 end
 
-ManningConnection() = ManningConnection(Int[], Float64[], Float64[], Float64[], Float64[], Float64[])
+ManningConnection() = ManningConnection(Int[], Float64[], Float64[], Float64[], Float64[], Float64[], Float64[], Float64[])
 
 """
 Requirements:
@@ -167,6 +169,7 @@ struct Parameters
     connectivity::Connectivity
     basin::Basin
     linear_level_connection::LinearLevelConnection
+    manning_connection::ManningConnection
     tabulated_rating_curve::TabulatedRatingCurve
     fractional_flow::FractionalFlow
     level_control::LevelControl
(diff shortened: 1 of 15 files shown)


core/src/solve.jl
```diff
@@ -199,6 +249,83 @@ function formulate!(
...
+    (; graph, flow, u_index) = connectivity
+    (; node_id, length, n_squared, profile_width, profile_slope, unit_length, contraction_c, expansion_c) = manning_connection
+    for (i, id) in enumerate(node_id)
+        basin_1_id = only(inneighbors(graph, id))
+        basin_2_id = only(outneighbors(graph, id))
+
+        h_1 = level[u_index[basin_1_id]]
+        h_2 = level[u_index[basin_2_id]]
+        bottom_1 = first(basin.level[basin_1_id].u)
+        bottom_2 = first(basin.level[basin_1_id].u)
```
visr, line 294: The state `u` and the vectors in Basin are aligned, same length and order. The ID cannot be used as an index, so here you also need to use `u_index[basin_1_id]`, so it's probably a good idea to save that as a variable. Also note that you use `basin_1_id` in the second line instead of `basin_2_id`.

core/src/solve.jl
```diff
@@ -199,6 +249,83 @@ function formulate!(
...
+        d_1 = h_1 - bottom_1
+        d_2 = h_2 - bottom_2
+        d_mean = 0.5 * (d_1 + d_2)
+
+        # Determine direction of flow
```
visr, line 299:
```suggestion
        # Determine direction of flow, it goes from a to b
```

core/src/solve.jl
```diff
@@ -199,6 +249,83 @@ function formulate!(
...
+
+        # Use water depth to compute either contraction or expansion
+        C = (d_b >= d_a) ? expansion_c : -1.0 * contraction_c
+        C_v = (1.0 + C) / TWO_G
+        C_f = n_squared * length
+
+        # Compute cross sectional areas and hydraulic radius
+        A_a = (profile_width * d_a + profile_slope * d_a^2)
+        A_b = (profile_width * d_b + profile_slope * d_b^2)
+        A2 = (profile_width * d_mean + profile_slope * d_mean^2)
```
visr, line 318:
```suggestion
        A_a = profile_width * d_a + profile_slope * d_a^2
        A_b = profile_width * d_b + profile_slope * d_b^2
        A2 = profile_width * d_mean + profile_slope * d_mean^2
```

docs/core/equations.qmd
```diff
@@ -245,6 +245,185 @@ $$
 \mathit{Q_{in}} = -\mathit{Q_{out}}
 $$ {#eq-flushing}
 
+## Manning connection
+
+Ribasim is capable of simulating steady flow between basins through a reaches
+described by a trapezoidal profile and a Manning roughness coefficient.
+
+We describe the discharge from basin $a$ for basin $b$ solely as a function of
```
visr, line 253:
```suggestion
We describe the discharge from basin $a$ to basin $b$ solely as a function of
```

docs/core/equations.qmd
```diff
@@ -245,6 +245,185 @@ $$
...
+The discharge $Q$ is defined as:
+
+$$
+Q = Av
+$$
+
+where $A$ is cross-sectional area.
+  
+We use conservation of energy to relate the total head at $a$ to $b$, with $H_a
+> H_b$ as follows:
```
visr, line 287:
```suggestion
We use conservation of energy to relate the total head at $a$ to $b$,
with $H_a> H_b$ as follows:
```

python/ribasim/ribasim/linear_level_connection.py
visr, line 1: These are the schemas of the linear level connection, looks like your edited the wrong file.

python/ribasim/ribasim/model.py
visr, line 1: Nit, but almost everywhere you put the ManningConnection after the LinearLevelConnection, except for 3x in this file. Perhaps keep the order the same?

visr (commented):
Great to have this. Made some initial comments.

PR #187, opened 2023-04-26
sort the tables

visr:
Tables require a particular sorting order. Sometime the core assumed it was sorted, and sometimes it sorted it itself.

This PR sorts things out. The required sorting order is documented for all tables. Before writing tables, ribasim-python sorts them in the needed order.

The core now sorts it when it can, and asserts it is sorted if it can't. It can sort when the StructVector is in memory (comes from GeoPackage). It cannot sort memory mapped Arrow tables. At least not without loading them in memory, which is what we generally want to avoid when using Arrow tables.


Hofer-Julian (approved):
Looks good!

Issue #84, opened 2023-03-01
Accommodate all variations of predefined boundary conditions


gijsber:
New node types:

- [x] Level (initial or terminal)
- [ ] Flow (initial or terminal)
- [ ] Terminal (only terminal)

Wait with Outlet with/without Salinity (terminal node) user request shows the need

SouthEndMusic:
@visr and I had a discussion about this. We came to the following conclusions:
1. What is now the `LevelControl` node can be turned into a `Level` node, by removing the `conductance` field and renaming the `target_level` field to `level`. `LevelControl` is then equivalent to the combination of a `LinearResistance` node and a `Level` node.
2. The flow node is essentially a pump with only a single connection, so this could either be added as a new node or be added to the `Pump` node.

A flow node is also equivalent to the current `Pump` node and a `Level` node, but then the role of the `Level` node is a bit vague.

gijsber:
I am not sure about the Pump. In my opinion we currently miss the maximum flow capacity in the pump which tells the controller what its pump range is. However, I can imagine that we have a Flow node (or a SetFlow node) which has an optional maximum flow rate and acts like a SetFlowRate in the model. But the direction of flow being set should be clear and I had the impression the current pump implementation looks at waterlevels to decide in which direction the flow rate has to go.

gijsber:
The Level node definitely has a role as a downstream boundary condition and possibly upstream as well. I have not made my mind up whether it is wise from a modeller's perspective to split LevelControl in two nodes, since level control is so important with Dutch water management. But if it becomes logical within the entire control concept that is under development, I would probably agree. Some face to face discussion with sketch board would help (ie next week :-))

visr:
> But the direction of flow being set should be clear and I had the impression the current pump implementation looks at waterlevels to decide in which direction the flow rate has to go.

The pump looks at the edge direction for the flow direction, a positive flow is along the edge direction.

https://github.com/Deltares/Ribasim/blob/30691e8e065119404dd6b2219b55d51c9131eb10/core/src/solve.jl#L229-L230

Similarly for other directional nodes, like TabulatedRatingCurve. Hence #148, since it is important that users see the direction. Also all flows on edges are defined to be positive in the edge direction.

SouthEndMusic:
After a discussion with @gijsber and @visr we came to the conclusion that we need 3 new node types as adjusted in the first comment above. I won't touch the LevelControl for now.

SouthEndMusic:
[...]

PR #196, opened 2023-05-08
Naive addition of Level node

SouthEndMusic:
I made an implementation of the level node (#84) analogous to Basin/LevelControl. I have not yet added this new node to a test model.


visr:
Looks like a good start! We still need to fix #126 to document this process.

The main things I can think of now that are still missing:
- `formulate!` function for the Level node, and calling that in `water_balance!`
- create a new module `python/ribasim/ribasim/level.py`
- add new Level class in `qgis/nodes.py`

@gijsber is correct that we'd like to support dynamic levels as well. Though there are a few other node types that still need this as well, so we can also do that together with Pump etc. in a separate PR.

visr (approved):

PR #209, opened 2023-05-12
move to julia 1.9, drop 1.8 support

visr:
Technically we could keep maintaining 1.8 support, but it's not really worth the effort at this point. This PR:

- moves CI over to julia 1.9
- updates the manifests to 1.9
- adds julia 1.9 requirement to Project.toml files
- relies on julia 1.9's new pkgversion function so we can remove our own
- some minor cleanups

This should only be merged when 1.9 is available on TeamCity, and our configuration is updated to default to that.

### DIFF
--- a/core/src/utils.jl
+++ b/core/src/utils.jl
@@ -1,16 +1,4 @@
-# Backported from Julia 1.9
-# TODO Doesn't support dev'd modules
-function pkgversion(m::Module)
-    rootmodule = Base.moduleroot(m)
-    pkg = Base.PkgId(rootmodule)
-    pkgorigin = get(Base.pkgorigins, pkg, nothing)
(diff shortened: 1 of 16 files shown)


core/src/utils.jl
```diff
@@ -1,16 +1,4 @@
-# Backported from Julia 1.9
-# TODO Doesn't support dev'd modules
-function pkgversion(m::Module)
-    rootmodule = Base.moduleroot(m)
-    pkg = Base.PkgId(rootmodule)
-    pkgorigin = get(Base.pkgorigins, pkg, nothing)
-    return pkgorigin === nothing ? nothing : pkgorigin.version
-end
-
-"
```
evetion, line 10: Please add back the docstring (maybe with more ")

Hofer-Julian (approved):

PR #214, opened 2023-05-15
Add ribasim_testmodels

### LINKED ISSUE #210: Add `ribasim_testmodels` Python package
One common task we have across our Ribasim packages is to preprocess models in order to use them for tests.
At the moment, we duplicate that work or call `conftest.py` of Ribasim Python's testbench.
As discussed with @visr, we should instead create a separate Python package `ribasim_testmodels` which we can use across these testbenches.

Tasks:
- [x] Create a Python package `ribasim_testmodels`
- [x] Move functionality of [python/ribasim/tests/conftest](https://github.com/Deltares/Ribasim/blob/main/python/ribasim/tests/conftest.py) into it
- [x] Let `python/ribasim/tests` depend on it
- [x] Let `python/ribasim_api/tests` depend on it

Hofer-Julian:
Closes #210 

This also updates the manifest to fix the failure on teamcity.


visr (approved):
Nice to have this in a separate package!

Hofer-Julian:
And the tests on macOS are now running for over an hour.
I think we should just merge, what do you say @visr?

PR #228, opened 2023-05-17
Ribasim python graph plot consistent with QGIS

SouthEndMusic:
I made sure that the node representation in the ribasim python graph plot and the QGIS plugin are consistent (in terms of marker and color) and added an option for a legend in the ribasim python plot:

ribasim_python_graph

### DIFF
--- a/python/ribasim/ribasim/node.py
+++ b/python/ribasim/ribasim/node.py
@@ -106,7 +106,7 @@ class Node(InputMixin, BaseModel):
         handles = []
         legend_labels = []
 
-        markers = {
+        MARKERS = {
             "Basin": "o",
             "FractionalFlow": "^",
             "LevelControl": "*",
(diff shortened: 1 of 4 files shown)


python/ribasim/ribasim/node.py
```diff
@@ -25,9 +25,23 @@
     "ManningResistance": "D",
     "TabulatedRatingCurve": "D",
     "Pump": "h",
+    "Terminal": "s",
     "": "o",
 }
 
+_COlORS = {
```
Hofer-Julian, line 32: 1. I assume the `l` became lowercase by accident
2. How about you name it `COLORS` (without the `_` prefix) and inline it in the `plot` function, since it is only used there?

Hofer-Julian (approved):
That's a nice improvement, Bart!
The code itself feels like we are working around limitations of `Dataframe.plot` here, but I also don't see a better way without refactoring the surrounding code.
After the comment is fixed, it is good to be merged from my side.

PR #229, opened 2023-05-19
Fix mypy errors and add github actions job

Hofer-Julian:
Mypy found a couple of mistakes, plus a bit of unused code.

Sidenote: `python/ribasim` is currently a bit too dynamic for my taste

### DIFF
--- a/.github/workflows/python_lint.yml
+++ b/.github/workflows/python_lint.yml
@@ -96,11 +96,11 @@ jobs:
           pip install python/ribasim_api
       - name: Run mypy on python/ribasim
         run: |
-          mypy --ignore-missing-imports python/ribasim
+          mypy --ignore-missing-imports python/ribasim/ribasim
       - name: Run mypy on python/ribasim_testmodels
         run: |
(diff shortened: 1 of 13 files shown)


.github/workflows/python_lint.yml
```diff
@@ -28,20 +26,68 @@ jobs:
...
+      - name: Setup Micromamba
+        uses: mamba-org/provision-with-micromamba@main
+        with:
+          python_version: ${{ matrix.python_version }}
+          cache-downloads: true
+          cache-env: true
+          environment-file: ./environment.yml
+      - name: Run mypy
+        run: |
+          mypy --ignore-missing-imports python/ribasim
```
SouthEndMusic, line 93: Should this run only on the ribasim subdirectory? Also add newline at the end.

Hofer-Julian: Unfortunately, it is hard to do that locally because of this: https://github.com/pypa/setuptools/issues/3518
However, we don't hit that on the CI so I've added it.

PR #232, opened 2023-05-19
fix config.toml in docs, saner default for toml_dir

visr:


core/src/config.jl
```diff
@@ -45,7 +45,7 @@ end
     update_timestep::Float64 = 60 * 60 * 24.0
 
     # optional, default is the path of the TOML
-    toml_dir::String = pwd()
+    toml_dir::String = "."
```
Hofer-Julian, line 48: I am not sure that this makes it better :)
If its name is `toml_dir` it should always be the path to the TOML, not just the default

Hofer-Julian (changes requested):

core/src/config.jl
Hofer-Julian: I think this should be a required parameter.

visr: That would only work if Config is always loaded from a TOML file. Because then we can set this field from the file path. We also want to be able to create a Config in Julia code (used in testing).

Note that the `toml_dir` is not really a configuration option and is not documented as such. Its only purpose is to be able represent all configuration information, which includes the `toml_dir`, in one object.

visr: https://github.com/Deltares/Ribasim/blob/32d348bcb00a2264688d7823cab8a2961efd03fc/core/src/io.jl#L104-L106

The "." acts like I would expect here, `nothing` would error:

```
julia> normpath(nothing, "asdf")
ERROR: MethodError: no method matching normpath(::Nothing, ::String)

Closest candidates are:
  normpath(::AbstractString, ::AbstractString...)
   @ Base path.jl:411

Stacktrace:
 [1] top-level scope
   @ REPL[1]:1

julia> normpath(".", "asdf")
"asdf"
```

Hofer-Julian: Yeah it would require code changes.
You would have to check if toml_dir is not nothing and if yes proceed accordingly.

Alternatively, you can just rename toml_dir to relative_dir and keep your code.

visr: Good point, `relative_dir` is a more accurate name for how it behaves. Renamed it, with an updated comment, in a200e7eb3b5b32bac85004b75f26402db411a10e.

core/src/config.jl
```diff
@@ -44,8 +44,8 @@ end
     # [s] Δt for periodic update frequency, including user horizons
     update_timestep::Float64 = 60 * 60 * 24.0
 
-    # optional, default is the path of the TOML
-    toml_dir::String = pwd()
+    # optional, when Config is created from a TOML file, this is its directory
+    relative_dir::String = "."
     input_dir::String = "."
     output_dir::String = "."
```
Hofer-Julian, line 50: Final nitpick: I think we should use pwd for all three dirs instead of ".". If someone does weird stuff with cwd while running ribasim (for example while coupling) that might otherwise be annoying to debug.

Hofer-Julian: Mmmh, actually only relative_dir.
so the behavior from before.

visr: Right, I see what you mean: 235a8c2cdad5002917a07192ee4a899c9c430399

Hofer-Julian (approved):

PR #234, opened 2023-05-19
Remove LevelControl node

visr:
Since it is exactly the same as the combination of a LinearResistance and LevelBoundary, as shown here.
This was brought up in #84. If we implement level control using a range, it would be done quite different than this, so I'm quite sure this is safe to remove.


SouthEndMusic (approved):
Looks like a all tracks of the level_control node were erased (:

Issue #200, opened 2023-05-09
Validation network connectivity


gijsber:
We need to define and validate the network connectivity rules.
Basins need some connector nodetypes inbetween, some non-basin nodetypes are not connectors. Could be related to additional characterization needs such as #199 

Some inpiration: a flexible schema definition from the network editor of Ribasim 8:
[NetworkValidationRules.zip](https://github.com/Deltares/Ribasim/files/11432230/NetworkValidationRules.zip)
http://fews.wldelft.nl/schemas/version1.0/topology-editor-schemas/networkValidationRules.xsd

visr:
The place in the code has been reserved :)

https://github.com/Deltares/Ribasim/blob/6a8183d439000c75914113b342a6c07574337af0/core/src/solve.jl#L17-L30

evetion:
- Also include the information network (control)

evetion:
These will be hardcoded rules in Julia, no external file parsing/generation yet.

PR #235, opened 2023-05-19
Add flow boundary

SouthEndMusic:
As discussed in issue https://github.com/Deltares/Ribasim/issues/84.


SouthEndMusic:
Don't forget to add to documentation:

The (unique) edge connecting to a FlowBoundary node should point away from that no so that positive flow always corresponds to introducing water to the model and negative flow corresponds to removing water from the model.

edge altijd van de flow boundary af; dan betekent positieve flow op die edge dat water geïntroduceerd wordt

core/src/solve.jl
```diff
@@ -346,6 +356,27 @@ function formulate!(level_control::LevelControl, p::Parameters)::Nothing
...
+    (; graph, flow) = connectivity
+    (; node_id, flow_rate) = flow_boundary
+
+    for (id, rate) in zip(node_id, flow_rate)
+        # Convention: edge points away from the flow boundary
+        dst_id = only(outneighbors(graph, id))
+
+        # Adding water is always possible
+        if rate >= 0
+            flow[node_id, id] = rate
```
visr, line 370:
```suggestion
            flow[id, dst_id] = rate
```

core/src/solve.jl
```diff
@@ -346,6 +356,27 @@ function formulate!(level_control::LevelControl, p::Parameters)::Nothing
...
+        else
+            storage = u[dst_id]
+            reduction_factor = min(storage, 10.0) / 10.0
+            q = reduction_factor * rate
+            flow[node_id, id] = q
```
visr, line 375:
```suggestion
            flow[id, dst_id] = q
```

python/ribasim/ribasim/models.py
```diff
@@ -11,106 +11,117 @@
 
 
 class Edge(BaseModel):
-    remarks: Optional[str] = Field("", description="a hack for pandera")
-    fid: int = Field(..., description="fid")
-    to_node_id: int = Field(..., description="to_node_id")
-    from_node_id: int = Field(..., description="from_node_id")
+    remarks: Optional[str] = Field('', description='a hack for pandera')
```
visr, line 14: Looks like you still need to run Black after generating this file.
https://deltares.github.io/Ribasim/contribute/addnode.html#finishing-up

core/src/solve.jl
```diff
@@ -346,6 +356,27 @@ function formulate!(level_control::LevelControl, p::Parameters)::Nothing
...
+
+    for (id, rate) in zip(node_id, flow_rate)
+        # Convention: edge points away from the flow boundary
+        dst_id = only(outneighbors(graph, id))
+
+        # Adding water is always possible
+        if rate >= 0
+            flow[id, dst_id] = rate
+        else
+            storage = u[dst_id]
```
visr, line 372: This uses and ID instead of an index into the storage vector. You can use the function `id_index` like used in `formulate!` for Pump, to translate the ID to an index.

core/src/solve.jl
```diff
@@ -346,6 +356,30 @@ function formulate!(level_control::LevelControl, p::Parameters)::Nothing
     return nothing
 end
 
+function formulate!(flow_boundary::FlowBoundary, p::Parameters, u)::Nothing
+    (; connectivity, basin) = p
+    (; graph, flow) = connectivity
+    (; node_id, flow_rate) = flow_boundary
+
+    for (id, rate) in zip(node_id, flow_rate)
+        # Convention: edge points away from the flow boundary
```
visr, line 365: Here, and in the docs below you call it a convention, which suggests that either way would work. But looking at the function below, at best you get a "FlowBoundary intake not a Basin" error, and at worst we remove water from basins below 0 m3. So we should probably call it a requirement, and in #200 add validation checks for it.

docs/python/examples.ipynb
visr, line 1: Perhaps we should make the default for legend True? I feel that without it, it will be hard to read for most.

visr (commented):
Some last comments, otherwise this is good to go.

docs/python/examples.ipynb
Hofer-Julian: Agreed

Hofer-Julian (approved):

PR #241, opened 2023-05-23
Python ribasim test coverage

SouthEndMusic:


SouthEndMusic:
@Hofer-Julian the tests now cover 97%

python/ribasim/tests/test_node.py
```diff
@@ -0,0 +1,40 @@
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from matplotlib import axes
+from ribasim.node import Node
+
+
+def test(tmp_path):
```
Hofer-Julian, line 8: It would be good to have a descriptive name for the function.
The only requirement is that its name starts with `test`

python/ribasim/tests/test_node.py
```diff
@@ -0,0 +1,40 @@
...
+            crs="EPSG:28992",
+        )
+    )
+
+    # Plotting
+    ax = node.plot(legend=True)
+    assert isinstance(ax, axes._axes.Axes)
+    assert ax.get_legend() is not None
+
+    return
```
Hofer-Julian, line 32:
```suggestion
```

I know we have that in other parts of the codebase too, but it is untypical coding style

python/ribasim/tests/test_node.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+if __name__ == "__main__":
+    from pathlib import Path
+
+    tmp_path = Path()
+
+    test(tmp_path)
```
Hofer-Julian, line 40: Let's remove that before merging

python/ribasim/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from numpy.testing import assert_array_equal
+from ribasim import utils
+from ribasim.node import Node
+from shapely import LineString
+
+
+def test():
```
Hofer-Julian, line 10: Let's also give that one a better name

python/ribasim/tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    assert lines[1].equals(LineString([xy[from_id[1] - 1], xy[to_id[1] - 1]]))
+
+    from_id_, to_id_ = utils.connectivity_from_geometry(node, lines)
+
+    assert_array_equal(from_id, from_id_)
+    assert_array_equal(to_id, to_id_)
+
+
+if __name__ == "__main__":
+    test()
```
Hofer-Julian, line 40:
```suggestion
if __name__ == "__main__":
    test()
```

Remove before merging

Hofer-Julian: If you added this in order to debug it, we can have a look at this today

python/ribasim/tests/test_io.py
```diff
@@ -18,28 +23,61 @@ def assert_equal(a, b):
...
+    # Type cast if possible is allowed (myfield_b int -> float)
+    assert (
+        repr(dummy_1)
+        == "<ribasim.Dummy>\n   myfield_a: 1\n   myfield_b: 1.0\n   myfield_c: DataFrame(rows=1) (col_x, col_y)"
+    )
+
+    with pytest.raises(ValidationError) as exc_info:
+        Dummy(myfield_a=1, myfield_b=1, myfield_c=bad_c)
+
+    assert exc_info.value.errors()[0]["loc"] == ("myfield_c",)
```
Hofer-Julian, line 79: My gut feeling is that this test is a bit too low level.

Instead of creating a new class that implements `InputMixin`, let's use an existing class and check it's `repr`.

python/ribasim/tests/test_edge.py
```diff
@@ -16,6 +17,10 @@ def test():
     edge = Edge(static=df)
     assert isinstance(edge, Edge)
 
+    # Plotting
```
Hofer-Julian, line 20: Let's give this test a better name so that it is clear that we test whether validation works.

Then make a separate test which tests whether plotting works.

Hofer-Julian (commented):
That's a good start!
I've added a couple of comments.

Hofer-Julian (approved):

PR #242, opened 2023-05-23
Move to setup-micromamba

Hofer-Julian:
See https://github.com/mamba-org/provision-with-micromamba/releases/tag/v16.
Supersedes #240


visr (approved):
Looks good to me!

PR #250, opened 2023-05-26
Add python model validation

### LINKED ISSUE #245: Basic Model validation in Ribasim python
This is more basic than #200; when you instantiate objects in Python, some basic checks should be done on the validity of the input:
- `Node`: passed node types are valid
- Every node ID is used in exactly one node object

These are ones I noticed were missing but more could be helpful. Misspelling a node type made QGIS not display the node properly without throwing an error, and mistakes with the node IDs only come to light in the Julia core with errors with long tracebacks.

SouthEndMusic:
Fixes #245.


python/ribasim/ribasim/model.py
```diff
@@ -148,6 +149,47 @@ def _write_tables(self, directory: FilePath) -> None:
...
+    def validate_model(self):
+        """
+        Checks:
+        - Whether all node types in the node field are valid
+        - Whether the node IDs of the node_type fields are valid
+        """
+
+        node_names_all, node_cls_all = list(
+            zip(*inspect.getmembers(node_types, inspect.isclass))
+        )
```
SouthEndMusic, line 161: Maybe this should be a separate function/method somewhere?

python/ribasim/ribasim/model.py
```diff
@@ -148,6 +149,47 @@ def _write_tables(self, directory: FilePath) -> None:
...
+        node_IDs_all = np.concatenate(node_IDs_all)
+        node_IDs_unique = np.unique(node_IDs_all)
+
+        assert (
+            len(node_IDs_unique) == n_nodes
+        ), "Invalid number of unique node IDs in node type fields"
+        assert np.min(node_IDs_unique) == 1, "The smallest node ID should be 1"
+        assert (
+            np.max(node_IDs_unique) == n_nodes
+        ), "The largest node ID should be equal to the number of nodes in the node field"
```
SouthEndMusic, line 191: Maybe it is better to raise a more specific error type?

python/ribasim/ribasim/model.py
```diff
@@ -148,6 +149,47 @@ def _write_tables(self, directory: FilePath) -> None:
                 input_entry.write(directory, self.modelname)
         return
 
+    def validate_model(self):
+        """
+        Checks:
+        - Whether all node types in the node field are valid
+        - Whether the node IDs of the node_type fields are valid
+        """
```
SouthEndMusic, line 157: Should there also be a check for whether node IDs from model fields correspond to node types from node object? Maybe we should go to a single source of truth here.

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,41 @@
...
+        {"flow_rate": 1, "node_id": -1, "remarks": ""}, ignore_index=True
+    )
+
+    with pytest.raises(AssertionError) as exec_info:
+        model.validate_model()
+
+    assert (
+        exec_info.value.args[0]
+        == "Invalid number of unique node IDs in node type fields"
+    )
```
SouthEndMusic, line 41: Better to make 2 separate tests?

python/ribasim/ribasim/model.py
SouthEndMusic: `inspect.getmembers` gives a list of tuples, I want two iterables with in one the names and in the other the classes

Hofer-Julian: I see.
Up to you if you want to put it into a function or not

visr:
Don't forget to remove the output from the notebook at the end.

Hofer-Julian:
> Don't forget to remove the output from the notebook at the end.

Yeah, we should really add a check and ideally a commit hook for that.

python/ribasim/ribasim/model.py
Hofer-Julian: Yes, that would be a good addition.
Up to you if you want to create an issue for that or add it within this PR

python/ribasim/ribasim/model.py
```diff
@@ -148,6 +149,61 @@ def _write_tables(self, directory: FilePath) -> None:
                 input_entry.write(directory, self.modelname)
         return
 
+    def validate_model(self):
+        """
+        Checks:
+        - Whether all node types in the node field are valid
+        - Whether the node IDs of the node_type fields are valid
```
Hofer-Julian, line 156: It's good to have one method `validate_model`, but let's split its subtasks into separate methods

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,70 @@
...
+    model.node.static = model.node.static._append(
+        {"type": "InvalidNodeType", "geometry": Point(0, 0)}, ignore_index=True
+    )
+
+    with pytest.raises(TypeError) as exec_info:
+        model.validate_model()
+
+    assert exec_info.value.args[0].startswith(
+        "InvalidNodeType is not a valid node type, choose from:"
+    )
```
Hofer-Julian, line 23: You can use `pytest.raises` for that directly. It has a match parameter which takes a regex.
https://docs.pytest.org/en/4.6.x/reference.html#pytest.raises

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,70 @@
...
+
+    # Add entry with invalid node ID
+    model.pump.static = model.pump.static._append(
+        {"flow_rate": 1, "node_id": -1, "remarks": ""}, ignore_index=True
+    )
+
+    with pytest.raises(ValueError) as exec_info:
+        model.validate_model()
+
+    assert exec_info.value.args[0] == "Node IDs must be positive integers, got [-1]."
```
Hofer-Julian, line 37: Use `pytest.raises`

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,70 @@
...
+
+    # Add duplicate node ID
+    model.pump.static = model.pump.static._append(
+        {"flow_rate": 1, "node_id": 1, "remarks": ""}, ignore_index=True
+    )
+
+    with pytest.raises(ValueError) as exec_info:
+        model.validate_model()
+
+    assert (
```
Hofer-Julian, line 51: Use `pytest.raises`

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,70 @@
...
+    model = basic
+
+    # Add entry in node but not in pump
+    model.node.static = model.node.static._append(
+        {"type": "Pump", "geometry": Point(0, 0)}, ignore_index=True
+    )
+
+    with pytest.raises(ValueError) as exec_info:
+        model.validate_model()
+    assert (
```
Hofer-Julian, line 67: Use `pytest.raises`

Hofer-Julian (changes requested):
Looks mostly good, after the comments are addressed it is ready to be merged.

python/ribasim/ribasim/model.py
```diff
@@ -204,6 +206,40 @@ def validate_model(self):
                 f"Expected node IDs from 1 to {n_nodes} (the number of rows in self.node.static), but these node IDs are missing: {node_IDs_missing}."
             )
 
+    def validate_model_node_IDs(self):
+        """
+        Checks whether the node IDs in the node field correspond to the node IDs on the node type fields
+        """
+
+        _, node_cls_all = list(zip(*inspect.getmembers(node_types, inspect.isclass)))
```
Hofer-Julian, line 214:
```suggestion
        node_cls_all = [members[1] for member in inspect.getmembers(node_types, inspect.isclass)]
```

python/ribasim/ribasim/model.py
```diff
@@ -204,6 +206,40 @@ def validate_model(self):
...
+
+        node_names_all_snake_case = [cls.get_toml_key() for cls in node_cls_all]
+
+        for name in self.fields():
+            if name in node_names_all_snake_case:
+                node_field = getattr(self, name)
```
Hofer-Julian, line 220:
```suggestion
                node_field = self.name
```

python/ribasim/ribasim/model.py
```diff
@@ -148,6 +149,111 @@ def _write_tables(self, directory: FilePath) -> None:
...
 
+    def validate_model_node_types(self):
+        """
+        Checks whether all node types in the node field are valid
+        """
+
+        node_names_all = [
+            node_name
+            for (node_name, _) in inspect.getmembers(node_types, inspect.isclass)
+        ]
```
Hofer-Julian, line 160: Now that this expression is used three times, it would indeed make sense to make a method out of it.

python/ribasim/tests/test_model.py
```diff
@@ -1,3 +1,74 @@
...
+    model = basic
+
+    # Add duplicate node ID
+    model.pump.static = model.pump.static._append(
+        {"flow_rate": 1, "node_id": 1, "remarks": ""}, ignore_index=True
+    )
+
+    with pytest.raises(
+        ValueError,
+        match=re.escape("These node IDs were assigned to multiple node types: [1]."),
```
Hofer-Julian, line 49: I didn't know of `re.escape`.
Nice!

Hofer-Julian (approved):
One small comment, but otherwise good to go from my side

python/ribasim/ribasim/model.py
```diff
@@ -149,15 +149,19 @@ def _write_tables(self, directory: FilePath) -> None:
                 input_entry.write(directory, self.modelname)
         return
 
+    @staticmethod
+    def get_node_types():
```
Hofer-Julian, line 153: Would be good to add type hints for the return value here.

python/ribasim/ribasim/model.py
```diff
@@ -149,15 +149,19 @@ def _write_tables(self, directory: FilePath) -> None:
...
+        node_names_all, node_cls_all = list(
+            zip(*inspect.getmembers(node_types, inspect.isclass))
+        )
+        return node_names_all, node_cls_all
+
     def validate_model_node_types(self):
```
Hofer-Julian, line 159: Same here

PR #253, opened 2023-05-30
Activate mypy plugins for pydantic, numpy and pandera

Hofer-Julian:


SouthEndMusic:
I looked a bit into what this does, and it seems good to me.

Issue #243, opened 2023-05-23
Refactor `Node.plot` of Ribasim Python


Hofer-Julian:
@SouthEndMusic added a legend to `Node.plot`, but he had to workaround the current implementation to make it happen.
https://github.com/Deltares/Ribasim/blob/d986d7c565621a7e8e13bf3410e2b002e5668162/python/ribasim/ribasim/node.py#L141

I think this an indication that we should stop using `Dataframe.plot` and solely rely on matplotlib instead.

Additionally, `kwargs` should be removed and replaced with an optional parameter `ax`.

PR #254, opened 2023-05-30
Node.plot refactor

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/243.

### DIFF
--- a/python/ribasim/ribasim/geometry/node.py
+++ b/python/ribasim/ribasim/geometry/node.py
@@ -78,28 +78,22 @@ class Node(InputMixin, BaseModel):
 
         return kwargs
 
-    def plot(self, **kwargs) -> Any:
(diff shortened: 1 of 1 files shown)


Hofer-Julian (approved):
Looks good!

python/ribasim/ribasim/geometry/node.py
```diff
@@ -78,28 +78,22 @@ def _kwargs_from_geopackage(cls, path: FilePath) -> Dict:
 
         return kwargs
 
-    def plot(self, **kwargs) -> Any:
+    def plot(self, ax=None, zorder=None) -> Any:
```
Hofer-Julian, line 81: Can you try to use `matplotlib.Axes` instead of `Any`?

PR #259, opened 2023-05-31
Allow missing nodetypes in python model validation

SouthEndMusic:
Also some better error messages in the model validation (i.e. identify all of the same problems in stead of raising an exception at the first).

### DIFF
--- a/python/ribasim_testmodels/ribasim_testmodels/basic.py
+++ b/python/ribasim_testmodels/ribasim_testmodels/basic.py
@@ -346,7 +346,7 @@ def tabulated_rating_curve_model() -> ribasim.Model:
     rating_curve = ribasim.TabulatedRatingCurve(
         static=pd.DataFrame(
             data={
-                "node_id": [2, 2],
+                "node_id": [2, 3],
                 "level": [0.0, 1.0],
                 "discharge": [0.0, q1000],
             }
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -163,12 +163,18 @@ class Model(BaseModel):
 
         node_names_all, _ = Model.get_node_types()
 
+        invalid_node_types = set()
+
         # Check node types
         for node_type in self.node.static["type"]:
             if node_type not in node_names_all:
-                raise TypeError(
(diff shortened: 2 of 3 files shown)


Hofer-Julian (approved):

python/ribasim_testmodels/ribasim_testmodels/basic.py
```diff
@@ -346,7 +346,7 @@ def tabulated_rating_curve_model() -> ribasim.Model:
     rating_curve = ribasim.TabulatedRatingCurve(
         static=pd.DataFrame(
             data={
-                "node_id": [2, 2],
+                "node_id": [2, 3],
```
visr, line 349: This doesn't seem right. Node 3 is in the time table.

visr: The static table is basically a collection of lookup tables stacked on top of each other. That is why multiple lines have the same node ID, because they describe one discharge/level relation. For node ID 2 this relation is fixed in time and described by two rows, and for node ID 3 the relation varies with time and is in the `time` table.

https://deltares.github.io/Ribasim/core/usage.html#tabulatedratingcurve

SouthEndMusic: It should also be noted that this went under the radar because the the `tabulated_rating_curve` testmodel does not in fact appear in any test ^^

visr: That model serves mainly to test updating tables over time. There is a test using that here: https://github.com/Deltares/Ribasim/blob/d208665dce98d35671f00f18d97636686b4e7a04/core/test/basin.jl#L27. Though without #258 it can still easily go under the radar.

> I assumed that every node type class has a mandatory static field with all node IDs for that node type appearing at least once

No, at least I didn't assume this so far. If you have a time series for some parameter and want to use that, it seems unnecessary extra work to ask users to provide a static value for that parameter even though it won't be used.

> Is there a way to obtain the node IDs from the node type objects without making a specific case for each node type?

Not really, no. So far I always use the node table for this:

https://github.com/Deltares/Ribasim/blob/d208665dce98d35671f00f18d97636686b4e7a04/core/src/io.jl#L6

Although I guess you could take the union of all `node_id` columns.

PR #271, opened 2023-06-02
always close the GeoPackage

### LINKED ISSUE #269: Release Geopackage file
I've noticed that when the Ribasim core is run from a REPL (e.g. `Ribasim.run("path/to/model.toml")`, the geopackage file is not released after this call. This is a bit of a nuisance when you want to make small changes to the model and write it again, since you cannot do this when the Julia REPL session is still active. Ideally the geopackage file would even be released when the Julia core throws an error.

visr:
Also if there is an error during initialization.
Fixes part of #269.


SouthEndMusic:
I tried it but now I found that I could not delete a generated geopackage file because of the Python REPL from which I wrote it.

visr:
Ah ok. Probably this is still good to have though. We can keep #269 open for the python issue.

PR #273, opened 2023-06-05
Validate nodetypes in graph.

evetion:
Solves issue #200 

# Description
To determine valid connections between nodetypes on an edge in the graph, one defines:

```julia
# Allowed types for downstream (to_node_id) nodes given the type of the upstream (from_node_id) node
neighbortypes(::Val{:Pump}) = Set((:Basin,))
```

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -12,22 +12,39 @@ struct Connectivity
     graph::DiGraph{Int}
     flow::SparseMatrixCSC{Float64, Int}
     edge_ids::Dictionary{Tuple{Int, Int}, Int}
-    function Connectivity(graph, flow, edge_ids)
-        if is_valid(graph, flow, edge_ids)
-            new(graph, flow, edge_ids)
+    edge_types::Dictionary{Int, Tuple{Symbol, Symbol}}
+    function Connectivity(graph, flow, edge_ids, edge_types)
+        if is_valid(graph, flow, edge_ids, edge_types)
+            new(graph, flow, edge_ids, edge_types)
         else
             error("Invalid graph")
         end
     end
 end
 
-# TODO Add actual validation
 function is_valid(
     graph::DiGraph{Int},
     flow::SparseMatrixCSC{Float64, Int},
     edge_ids::Dictionary{Tuple{Int, Int}, Int},
+    edge_types::Dictionary{Int, Tuple{Symbol, Symbol}},
 )
-    return true
+    rev_edge_ids = dictionary((v => k for (k, v) in pairs(edge_ids)))
+    errors = String[]
+    for (edge_id, (from_type, to_type)) in pairs(edge_types)
+        if !(to_type in neighbortypes(from_type))
+            a, b = rev_edge_ids[edge_id]
+            push!(
+                errors,
+                "Edge #$edge_id from $from_type (#$a) to $to_type (#$b) is not allowed.",
+            )
+        end
+    end
+    return if isempty(errors)
+        true
+    else
+        @error join(errors, "\n")
+        false
+    end
 end
 
 """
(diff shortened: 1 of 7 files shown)


core/src/validation.jl
```diff
@@ -23,6 +23,28 @@ isnode(sv::Type{SchemaVersion{T, N}}) where {T, N} = length(split(string(T), "."
 nodetype(sv::Type{SchemaVersion{T, N}}) where {T, N} = Symbol.(split(string(T), ".")[2:3])
 nodetype(sv::SchemaVersion{T, N}) where {T, N} = Symbol.(split(string(T), ".")[2:3])
 
+# Allowed types for downstream (to_node_id) nodes given the type of the upstream (from_node_id) node
+neighbortypes(nodetype::Symbol) = neighbortypes(Val(nodetype))
```
visr, line 27: Is it really worth it to do all of this in the type domain? Why not just one function for symbols?

core/src/solve.jl
```diff
@@ -12,22 +12,39 @@ struct Connectivity
...
     edge_ids::Dictionary{Tuple{Int, Int}, Int}
-    function Connectivity(graph, flow, edge_ids)
-        if is_valid(graph, flow, edge_ids)
-            new(graph, flow, edge_ids)
+    edge_types::Dictionary{Int, Tuple{Symbol, Symbol}}
+    function Connectivity(graph, flow, edge_ids, edge_types)
+        if is_valid(graph, flow, edge_ids, edge_types)
+            new(graph, flow, edge_ids, edge_types)
         else
             error("Invalid graph")
```
visr, line 20:
```suggestion
            error("Invalid network")
```

core/src/validation.jl
```diff
@@ -23,6 +23,28 @@ isnode(sv::Type{SchemaVersion{T, N}}) where {T, N} = length(split(string(T), "."
...
+neighbortypes(::Val{:pump}) = Set((:basin,))
+neighbortypes(::Val{:basin}) = Set((
+    :linearresistance,
```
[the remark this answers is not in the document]
evetion: So ideally we change the namespace of the validation as well? Because these are based on it, aligning with the nodetypes used also for the jsonschema.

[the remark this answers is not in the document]
visr: Yes exactly, that'd be great.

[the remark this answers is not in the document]
evetion: Well namespaced can't be changed (Legolas rules), but I've changed all the Val{:type} to be camelcased like everywhere else.

core/src/validation.jl
```diff
@@ -23,6 +23,26 @@ isnode(sv::Type{SchemaVersion{T, N}}) where {T, N} = length(split(string(T), "."
 nodetype(sv::Type{SchemaVersion{T, N}}) where {T, N} = Symbol.(split(string(T), ".")[2:3])
 nodetype(sv::SchemaVersion{T, N}) where {T, N} = Symbol.(split(string(T), ".")[2:3])
 
+# Allowed types for downstream (to_node_id) nodes given the type of the upstream (from_node_id) node
+neighbortypes(nodetype::Symbol) = neighbortypes(Val(nodetype))
+neighbortypes(::Val{:Pump}) = Set((:Basin,))
```
evetion, line 28: @visr Can you check these rules? And otherwise add/merge?

visr: Pushed in d62323e823eb0308e81aad20fa0736a0944eabb1. I think that covers it. Fun to think about.

visr (approved):
Needs an update of addnode.qmd as well. Otherwise it looks good to me.

core/src/validation.jl
evetion: And fun commit message as well!

Issue #266, opened 2023-06-01
Implement data structure for (boolean) control


SnippenE:
Part of #13

SouthEndMusic:
Here is a concept for how to implement control, feel free to criticize.

## Relevant node categories

There are 3 categories of node that are relevant for control:
- The nodes with state (e.g. `Basin`), whose state forms the input of the control logic;
- A newly introduced node type called `Control`, which implements the control logic;
- The nodes which are to be controlled, say `TabulatedRatingCurve` or `Pump`.

## Tables

We introduce a new table called `control` for each node type that can be controlled. This table is similar to the [time table](https://deltares.github.io/Ribasim/core/usage.html#tabulatedratingcurve-time) currently available for `TabulatedRatingCurve`, but instead of the column `time` it has the column `control_state`, containing a descriptive identifier string for each control state.

The `Control` node has these associated tables:
- `conditions`, with columns:
  -  `node_id` (`int`), 
  - `variable` (`str`),
  - `bigger_than` (`float`).
This table defines conditions like 'the level in the basin with ID `n` is bigger than `L`'.
- `logic_mapping`, with columns:
   - `node_id` (`int`), 
   - `true_conditions` (not sure),
   - `false_conditions` (not sure), 
   - `control_state` (`str`).
This table states that the node with the given node ID is in the given control state if the given conditions are true and false respectively.
- `minimal_time`, with columns:
   - `control_state` (`str`)
   - `Dt` (`float`). 
This table states for each control state what the minimal time is that a node spends in this control state, to avoid oscillation between control states.

## Implementation in Julia core

Say we implement a `Control` struct, with the following fields:
- `minimal_time`, `conditions`, `logic_mapping`: see above.
- `condition_value`: a boolean vector stating the truth value of each condition at the current time.
- `control_state`: a dictionary from node ID to a tuple '(current control state, start time of current control state)'.

We add a `control` field of type `Control` to the `Parameters` struct. We update the control states using a `ContinuousCallback` (see in the docs [here](https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/)). The condition value is updated in the `condition!` function:

```julia
function condition!(u,t,integrator)
   p = integrator.p
   
   out = 1.0
   condition_index = 0   

   for node_id, var, bigger_than in p.control.conditions
      condition_index += 1

      # get value of variable
      value = get_value(node_id,var)
      diff = value - bigger_than
      out *= diff
[...]

PR #274, opened 2023-06-05
Start of control implementation

SouthEndMusic:
Solves issue #266.


docs/python/example_control.ipynb
visr, line 1: Don't forget to remove the output when you're done :)

core/src/bmi.jl
```diff
@@ -74,10 +102,112 @@ function create_callbacks(
...
+        for target_node_id in outneighbors(control.graph, control_node_id)
+            set_control_params!(p, target_node_id, control_state_new)
+        end
+
+        control.control_state[control_node_id] = (control_state_new, integrator.t)
+    end
+end
+
+function set_control_params!(p::Parameters, node_id::Int, control_state::String)
+    if node_id in p.pump.node_id
```
evetion, line 202: We probably need to add validation here.

evetion:
We need to implement a Julia test for control to merge this, plus some documentation. The rest can be discussed in the refinement and added later.

core/src/bmi.jl
```diff
@@ -54,9 +54,28 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
         config.solver.maxiters,
     )
 
+    set_initial_controlled_parameters(parameters.control, u0, integrator)
+
     return Model(integrator, config, saved_flow)
 end
 
+function set_initial_controlled_parameters(
```
evetion, line 62: This seems a mutating function, might be good to add ! to the function name. If so, you might want to reorder the arguments too (mutating arguments first).

core/src/bmi.jl
```diff
@@ -81,10 +100,122 @@ function create_callbacks(
...
+"""
+Get a value for a condition. Currently only supports getting levels from basins.
+"""
+function get_value(p::Parameters, node_id::Int, variable::String, u)
+    # TODO: Add support for getting flow values
+
+    if variable == "level"
+        basin = p.basin
+
+        # NOTE: Getting the level with get_level does NOT work
```
evetion, line 144: Either remove or explain why (root finding needs interpolated values?)

core/src/create.jl
```diff
@@ -85,7 +85,29 @@ end
 
 function Pump(db::DB, config::Config)::Pump
     static = load_structvector(db, config, PumpStaticV1)
-    return Pump(static.node_id, static.flow_rate)
+
+    control_mapping = Dict{Tuple{Int, String}, NamedTuple}()
+
+    if length(static.control_state) > 0 && !any(ismissing.(static.control_state))
+        # Starting flow_rates are first one found (can be updated by control initialisation)
```
evetion, line 92: So the default state (normally missing) is not necessarily the one used, if the order of the static table is incorrect?

core/src/create.jl
```diff
@@ -127,6 +149,42 @@ function Basin(db::DB, config::Config)::Basin
...
+    rows = execute(db, "select from_node_id, edge_type from Edge")
+    for (; from_node_id, edge_type) in rows
+        if edge_type == "control"
+            control_state[from_node_id] = ("undefined_state", 0.0)
+        end
+    end
+
+    logic = load_structvector(db, config, ControlLogicV1)
+
+    logic_mapping::Dict{Tuple{Int, String}, String} = Dict()
```
evetion, line 167: I think you refactored this back, why not
`logic_mapping = Dict{Tuple{Int, String}, String}()`?

core/src/create.jl
evetion: I think the previous code making an `ismissing` mask (or a defaultmask, once we define an other default value), makes more sense, also in readability.

core/src/io.jl
```diff
@@ -86,6 +86,10 @@ function load_structvector(
         nt = merge(nt, (; time = DateTime.(nt.time, dateformat"yyyy-mm-dd HH:MM:SS.s")))
     end
 
+    if hasfield(T, :control_state) && !hasfield(typeof(nt), :control_state)
```
evetion, line 89: Is this required for backwards compatibility? Ideally we don't add hardcoded fields to this generic function. (The :time check above is already a bit ugly).

core/test/control.jl
```diff
@@ -0,0 +1,18 @@
...
+
+    pump_control_mapping::Dict{Tuple{Int64, String}, NamedTuple} =
+        Dict((4, "off") => (flow_rate = 0.0,), (4, "on") => (flow_rate = 1.0e-5,))
+
+    logic_mapping::Dict{Tuple{Int64, String}, String} =
+        Dict((5, "TT") => "on", (5, "TF") => "off", (5, "FF") => "on", (5, "FT") => "off")
+
+    @test p.pump.control_mapping == pump_control_mapping
+    @test p.control.logic_mapping == logic_mapping
+end
```
evetion, line 18: EOF

evetion (commented):
Is the python code coverage something we can test with a small example, or do we wait for the larger use cases?

core/src/io.jl
SouthEndMusic: This is indeed needed otherwise `table = StructVector{T}(nt)` below throws an error.

Issue #279, opened 2023-06-06
Extrapolating the LinearInterpolation


visr:
We use the LinearInterpolation from DataInterpolations for Q(h), A(S) and h(S) relations, the last two provided by users as A(h) once #225 is fixed.

It's important to control the extrapolation behavior, e.g. what happens if h is below the lowest h in the table, or above the largest. We currently have the default behavior of LinearInterpolation, which is linear based on the slope of the last segment. Visually:

![image](https://github.com/Deltares/Ribasim/assets/4471859/337e6c2d-9091-40c0-81d2-4c794aaf682b)

I think the linear extrapolation on the high end is fine, since generally the trend continues there, rather than stop. On the low end however we have a different situation, where we should keep it constant instead. So I think this is desired:

![image](https://github.com/Deltares/Ribasim/assets/4471859/af97147a-5d5a-4dcb-b6f2-77c1bce4802b)

There is no direct option for this in DataInterpolations, but it can be achieved by prepending an entry in the table, using `prevfloat` on the input, and the same output to get a horizontal segment.

**Q(h)**
- [x] First Q must be 0 #1469
- [x] If h is lower than minimum, keep it 0. #1469

**A(S) and h(S), provided by a user as A(h)**
- [x] First A must be >0. #438 undone in #472
- [x] Second A must be larger than 0. #438
- [x] If S is lower than 0, do constant extrapolation. In principle the solver should prevent S < 0, but good to do this for safety. This way the water level can't be below the bottom, and the area not below 0. Superseded by #1425.

The interpolation input, level, gets sorted. We currently don't enforce that the interpolation output, Q or A, are increasing (or staying constant). In most cases you want this, but it's possible to image scenarios where you don't. A bit contrived, but:

![image](https://github.com/Deltares/Ribasim/assets/4471859/22946159-cb5a-4419-b0d0-d6a61fee7387)

A decreasing surface water area is possible if you think about fishbowl or cave like profiles. And for Q(h) naturally this typically won't happen, but it could be of interest if you want to encode steering behavior without separate control rules.

- [x] Since it is quite rare, and in most cases a decreasing Q or A is a user error, I think we should start enforcing this. If there is a need for it in the future, we could add config options to enable decreasing Q or A, to allow the user to say, I know what I'm doing, don't get in my way.

- [x] For Q(h) a single row is probably fine, but for A(h) we need at least two, where the h is increasing, otherwise you cannot have storage.

gijsber:
[...]

PR #280, opened 2023-06-06
calculate storage from Basin / profile

### LINKED ISSUE #225: remove storage from Basin / profile
This week we were discussing the "Basin / profile" table with @Hofer-Julian and @SouthEndMusic, and @SouthEndMusic mentioned that having all 3 of storage, area and level is overdetermined. This can lead to profiles that don't have any physical representation. Like the example we use:

https://github.com/Deltares/Ribasim/blob/a10e0d0e2bcb95905d53b04da75eea73f6bc18cc/python/ribasim_testmodels/ribasim_testmodels/basic.py#L70-L72

There a storage of 1000 means an area of 1000 and level of 1. When the level goes to 0, it is not possible that both the area and the storage go down to 0 linearly. Either the area stays constant (vertical walls), such that the storage can go to 0 linearly, or if the area goes linearly to 0 then the storage changes more per centimeter level change near the top since the area is larger there.

In the code we look up both the level and area based on the storage. The storage of the profile can be removed from the table, since we can calculate it ourselves during model initialization by integrating the area (m2) over the level (m) to get storage (m3). I think storage is the most logical one to leave out since level and area are more easily measured, and easier to imagine.

I think it's good to keep supporting models that don't use level at all (by not using node types that use level). For those use cases they can interpret level as a relative number only used to describe the profile and start the bottom at 0 or any other number.

https://deltares.github.io/Ribasim/core/usage.html#basin-profile

visr:
Fixes #225

This removes the need to provide the storage for a profile. We can calculate it ourselves by integrating over A(h) to get A(S) and h(S).

First this updates the test models to a correct storage in the profile. Then we calculate the profile and get the same results. Then we remove the storage from the schemas and docs.

I hoped this would help with #80 but based on the plot I added there it doesn't seem to. This does not yet add all the validation and extrapolation for lookup tables as described in #279, that can be done separately.


SouthEndMusic:
I'll try adding the proper interpolation.

SouthEndMusic:
I suggest something like this for the interpolation:

```julia
function get_area_and_level(basin::Basin, state_idx::Int, storage::Float64)
   storage_discrete = basin.storage[state_idx]
   area_discrete = basin.storage[state_idx]
   level_discrete = basin.store[state_idx]
  
   # storage_idx: smallest index such that storage_discrete[storage_idx] >= storage
   storage_idx = searchsortedfirst(storage_discrete,storage)

   if storage_idx == 1
      # Handle extrapolation below given storage values

   elseif storage_idx <= length(storage_discrete)

      area_higher = area_discrete[storage_idx]
      area_lower = area_discrete[storage_idx-1]
      level_higher = level_discrete[storage_idx]
      level_lower = level_discrete[storage_idx-]
      storage_lower = storage_discrete[storage_idx-1]
        
      area_diff = area_higher - area_lower
      level_diff = level_higher - level_lower # This should never be 0, validate?

      if area_diff == 0
      # Constant area means linear interpolation
         area = area_lower

         storage_higher = storage_discrete[storage_idx]
         level = level_lower + level_diff*(storage-storage_lower)/(storage_higher-storage_lower)
         
      else
         area = sqrt(area_lower^2 + 2*area_diff/level_diff*(storage-storage_lower))
         level = level_lower + level_diff*(area-area_lower)/area_diff
      end
 
   else # storage_idx = length(storage_discrete)+1
      # Handle extrapolation above given storage values
  
   return area, level
   end

end
```



Here I assume that the discrete storage variables are normal vectors. Should we also require that the profile is given by at least 2 height values?

SouthEndMusic (approved):

Issue #281, opened 2023-06-07
Create minimal, tailor-made MODFLOW6 and Ribasim model


Hofer-Julian:
Initially, we will focus on MODFLOW 6 and Ribasim coupling.
For that, we need to create a model involving those two components.

PR #295, opened 2023-06-09
Add a trivial model (only one basin) as first coupled test case

Huite:
This is part of #281.

I will work on a minimal modflow6 fixture in imod_coupler next.


core/test/basin.jl
```diff
@@ -3,6 +3,15 @@ using Ribasim
 import BasicModelInterface as BMI
 using SciMLBase
 
+
+@testset "trivial model" begin
+    toml_path = normpath(@__DIR__, "../../data/trivial/trivial.toml")
+    @test ispath(toml_path)
+    model = Ribasim.run(toml_path)
+    @test model isa Ribasim.Model
+    @test model.integrator.sol.retcode == Ribasim.ReturnCode.Success
```
Hofer-Julian, line 12: Does it make sense to test results too?
Like "there is no less water in the basin then there was in the beginning?"

Hofer-Julian (commented):
Looks good so far.

After the comment as resolved and the unrelated typo fixed, it is ready to be merged from my side.

Hofer-Julian (approved):

Issue #299, opened 2023-06-12
Create output file of control


SouthEndMusic:
The Julia core should produce an output file (say `output/control.arrow`) with the following columns:

- time
- control node id
- truth state
- control state

A new entry should be added in the 

```julia
if control_state_now != control_state_new
```

block in `control_affect!`.

PR #303, opened 2023-06-13
Control output file

SouthEndMusic:
Solves issue https://github.com/Deltares/Ribasim/issues/299.


SouthEndMusic:
Prints a control overview like this:

```
0. At 2020-01-01 00:00:00 the control node with ID 5 reached truth state TF:
	For node ID 1 (Basin): level > 0.8
	For node ID 3 (Basin): level < 0.4

   This yielded control state "off":
	For node ID 4 (pump): flow_rate = 0.0

1. At 2020-01-30 12:07:44.413000 the control node with ID 5 reached truth state FF:
	For node ID 1 (Basin): level < 0.8
	For node ID 3 (Basin): level < 0.4

   This yielded control state "on":
	For node ID 4 (pump): flow_rate = 1e-05

2. At 2020-02-16 04:47:34.025000 the control node with ID 5 reached truth state FT:
	For node ID 1 (Basin): level < 0.8
	For node ID 3 (Basin): level > 0.4

   This yielded control state "off":
	For node ID 4 (pump): flow_rate = 0.0
``` 

Might be hard to test since this has to happen in Python after the Julia core has run

gijsber:
Hoi @SouthEndMusic,

het lijkt erop alsof dit een text file is.
Als we het (tevens) in een digitale file vastleggen (bijv. CSV) kunnen we er veel makkelijker een analyse mee uitvoeren.
De table is niet optimaal genromailseerd maar de header zou kunnen zijn.
Time, control_node, truth_state, control_state, condition_node, condition_state

core/src/io.jl
```diff
@@ -165,3 +166,18 @@ function write_flow_output(model::Model)
...
+    if length(record.time) > 0
+        time = convert.(Arrow.DATETIME, datetime_since.(record.time, config.starttime))
+
+        table = (; time, record.control_node_id, record.truth_state, record.control_state)
+
+        path = output_path(config, config.output.control)
+        mkpath(dirname(path))
+        Arrow.write(path, table; compress = :lz4)
+    end
+end
```
evetion, line 183: EOF

core/src/io.jl
```diff
@@ -165,3 +166,18 @@ function write_flow_output(model::Model)
     mkpath(dirname(path))
     Arrow.write(path, table; compress = :lz4)
 end
+
+function write_control_output(model::Model)
+    config = model.config
+    record = model.integrator.p.control.record
+
+    if length(record.time) > 0
```
evetion, line 174: I'm not sure what the behaviour of the other arrow writing code is. Do we write empty tables when asked? We should do the same here.

docs/core/usage.qmd
```diff
@@ -42,6 +42,9 @@ geopackage = "model.gpkg"  # required
 flow = "output/flow.arrow"    # optional, default "output/flow.arrow"
 basin = "output/basin.arrow"  # optional, default "output/basin.arrow"
 
+# This output file is only written if the model involves control
```
evetion, line 45: Yeah, so we could have a control table with rules that are never triggered right? This comment implies that we should write an empty table.

evetion (commented):
Can you document the Python utility print function somewhere? Or add a TODO/issue that we want to test it?

Otherwise LGTM, apart from some minor comments.

core/src/io.jl
SouthEndMusic: I think basin and flow tables are always written, but if they are empty the model is rather trivial.

evetion (approved):

PR #309, opened 2023-06-13
add ManningResistance to QGIS node type dropdown

visr:
This one was missing:

image

All others are there, and I verified that this step is listed in the dev docs on adding node types.


evetion (approved):
Can you make a todo or (link to an) issue that we should automate this code? We have schemas lying around somewhere that we could at least parse for the nodetypes, but I also believe that just the classes used to read the proper tables in QGIS should be enough?

PR #314, opened 2023-06-14
More accurate interpolation

SouthEndMusic:
I have implemented the physical interpolation (as proposed in https://github.com/Deltares/Ribasim/issues/225). I have not checked whether this implementation is exactly what we want for extrapolation (https://github.com/Deltares/Ribasim/issues/279), but this implementation assumes that the storage is 0 at the lowest level and does linear extrapolation of the area as a function of the level based on the last 2 values.

This does not solve the problem of high oscillating flows produced by the ManningResistance node as hoped (https://github.com/Deltares/Ribasim/issues/80). I think the problem there is that the calculated flows are way to large, and therefore the equilibrium is overshooted at every time step.

I cannot wrap my head around the current ManningResistance implementation. I understand that flow occurs due to a difference in level, but I do not understand why that means that all the water comes in motion, not just the bit at the top (which is why I find LinearResistance more intuitive). Is anyone familiar with a ManningResistance being used for flow that can go both directions, or is it only used for flow that goes one direction due to gravity? In that last case I understand better why all water is in motion.

Also, the current implementation assumes the presence of a lot of water in the channel itself, which fluctuates with the level in the upstream basin with no regard for (local) water preservation (by which I do not mean that the ManningResistance node is not preserving but it is less physical).

### DIFF
--- a/core/src/utils.jl
+++ b/core/src/utils.jl
@@ -34,23 +34,103 @@ function profile_storage(levels::Vector{Float64}, areas::Vector{Float64})::Vecto
     return storages
 end
 
-"Read the Basin / profile table and return all A(S) and h(S) functions"
+"Read the Basin / profile table and return all area and level and computed storage values"
 function create_storage_tables(
     db::DB,
     config::Config,
-)::Tuple{Vector{Interpolation}, Vector{Interpolation}}
+)::Tuple{Vector{Vector{Float64}}, Vector{Vector{Float64}}, Vector{Vector{Float64}}}
     profiles = load_structvector(db, config, BasinProfileV1)
-    area = Interpolation[]
-    level = Interpolation[]
+    area = Vector{Vector{Float64}}()
+    level = Vector{Vector{Float64}}()
+    storage = Vector{Vector{Float64}}()
+
     for group in IterTools.groupby(row -> row.node_id, profiles)
         group_area = getproperty.(group, :area)
         group_level = getproperty.(group, :level)
         group_storage = profile_storage(group_level, group_area)
-        area_itp = LinearInterpolation(group_area, group_storage)
-        level_itp = LinearInterpolation(group_level, group_storage)
-        push!(area, area_itp)
-        push!(level, level_itp)
+        push!(area, group_area)
+        push!(level, group_level)
+        push!(storage, group_storage)
     end
+    return area, level, storage
+end
+
+function get_area_and_level(
+    basin::Basin,
+    state_idx::Int,
+    storage::Float64,
+)::Tuple{Float64, Float64}
+    storage_discrete = basin.storage[state_idx]
+    area_discrete = basin.area[state_idx]
+    level_discrete = basin.level[state_idx]
+
+    return get_area_and_level(storage_discrete, area_discrete, level_discrete, storage)
+end
+
+function get_area_and_level(
+    storage_discrete::Vector{Float64},
+    area_discrete::Vector{Float64},
+    level_discrete::Vector{Float64},
+    storage::Float64,
+)::Tuple{Float64, Float64}
+    # storage_idx: smallest index such that storage_discrete[storage_idx] >= storage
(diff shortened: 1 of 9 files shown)


SouthEndMusic:
I restricted this PR to the profile interpolation since fixing the ManningResistance turns out to be a separate problem.

core/src/utils.jl
```diff
@@ -208,19 +279,21 @@ end
 
 "Return the bottom elevation of the basin with index i"
 function basin_bottom_index(basin::Basin, i::Int)::Float64
-    # get level(storage) interpolation function
-    itp = basin.level[i]
-    # and return the first level in the underlying table, which represents the bottom
-    return first(itp.u)
+    # TODO: Is this function still needed now that it is so simple?
```
visr, line 282: Fine by me to replace with `bottom = basin.level[i][1]` in formulate.

core/src/utils.jl
```diff
@@ -208,19 +279,21 @@ end
...
+    # get discrete level values
+    level_discrete = basin.level[i]
+    # and return the first level in this vector, representing the bottom
+    return first(level_discrete)
 end
 
 "Return the bottom elevation of the basin with index i"
 function basin_bottom(basin::Basin, node_id::Int)::Float64
+    # TODO: Is this function still needed now that it is so simple?
```
visr, line 291: It's still about 8 lines of code, making it easier to read: `bottom_a = basin_bottom(basin, basin_a_id)`

core/src/utils.jl
```diff
@@ -34,23 +34,94 @@ function profile_storage(levels::Vector{Float64}, areas::Vector{Float64})::Vecto
...
+        area_higher = area_discrete[end]
+        level_lower = level_discrete[end - 1]
+        level_higher = level_discrete[end]
+        storage_lower = storage_discrete[end - 1]
+        storage_higher = storage_discrete[end]
+
+        area_diff = area_higher - area_lower
+        level_diff = level_higher - level_lower
+
+        if area_diff == 0
```
visr, line 86:
```suggestion
        if area_diff ≈ 0
```

I think that's better to allow for floating point inaccuracies.

core/src/utils.jl
```diff
@@ -34,23 +34,94 @@ function profile_storage(levels::Vector{Float64}, areas::Vector{Float64})::Vecto
...
+        area_higher = area_discrete[storage_idx]
+        level_lower = level_discrete[storage_idx - 1]
+        level_higher = level_discrete[storage_idx]
+        storage_lower = storage_discrete[storage_idx - 1]
+        storage_higher = storage_discrete[storage_idx]
+
+        area_diff = area_higher - area_lower
+        level_diff = level_higher - level_lower # This should never be 0, validate?
+
+        if area_diff == 0
```
visr, line 111:
```suggestion
        if area_diff ≈ 0
```

core/src/utils.jl
```diff
@@ -34,23 +34,94 @@ function profile_storage(levels::Vector{Float64}, areas::Vector{Float64})::Vecto
...
+    else
+        area_lower = area_discrete[storage_idx - 1]
+        area_higher = area_discrete[storage_idx]
+        level_lower = level_discrete[storage_idx - 1]
+        level_higher = level_discrete[storage_idx]
+        storage_lower = storage_discrete[storage_idx - 1]
+        storage_higher = storage_discrete[storage_idx]
+
+        area_diff = area_higher - area_lower
+        level_diff = level_higher - level_lower # This should never be 0, validate?
```
visr, line 109: Perhaps it's good to add this comment to https://github.com/Deltares/Ribasim/issues/279

visr (changes requested):
It looks good to me overall, I left some comments. Let's add some tests for `get_area_and_level`. For that I think you can create a Basin directly (not an entire test model), and then make sure you check all 3 main branches, including the area_diff 0 branches.

visr (approved):

PR #315, opened 2023-06-14
move config from qmd to separate toml

### LINKED ISSUE #304: Setting custom path for output files fails
Say  in `basic.toml` you add the line `basin = "myawesomefolder/basin.arrow"`, as indicated in [usage](https://deltares.github.io/Ribasim/core/usage.html). When running the Julia core on this TOML file, I now get this error:

```
ERROR: expect an AbstractDict or Ribasim.basin, got String
```

visr:
Fixes #304

The separate toml is included into the docs, so it looks the same as before. This way we can test this configuration to make sure it doesn't go out of date.

This checks in https://github.com/quarto-ext/include-code-files as suggested there.

### DIFF
--- a/docs/_extensions/quarto-ext/include-code-files/include-code-files.lua
+++ b/docs/_extensions/quarto-ext/include-code-files/include-code-files.lua
@@ -1,0 +1,62 @@
+--- include-code-files.lua – filter to include code from source files
+---
+--- Copyright: © 2020 Bruno BEAUFILS
+--- License:   MIT – see LICENSE file for details
+
+--- Dedent a line
+local function dedent (line, n)
+  return line:sub(1,n):gsub(" ","") .. line:sub(n+1)
+end
+
+--- Filter function for code blocks
+local function transclude (cb)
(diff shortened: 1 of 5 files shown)


docs/_extensions/quarto-ext/include-code-files/include-code-files.lua
```diff
@@ -0,0 +1,62 @@
+--- include-code-files.lua – filter to include code from source files
```
Hofer-Julian, line 1: I assume there is no other way than vendoring at the moment?

evetion (approved):
Seems like a good improvement, although the lua includement is sad. Maybe merge this and make a tech-debt labeled issue to find a way to not have to include quarto extensions?

visr:
The extension docs state:

> If you're using version control, you will want to check in this directory.

I guess we could avoid this, by adding `quarto add quarto-ext/include-code-files` to the dev docs and CI. But given that this was suggested and it is only a tiny bit of code and is simpler, I went for the suggested way.

PR #327, opened 2023-06-16
allow one-directional outflow to LevelBoundary

visr:
This came up in a model from @d2hydro (https://github.com/d2hydro/lhm-ribasim/issues/5#issuecomment-1593707186).

FractionalFlow, FlowBoundary and TabulatedRatingCurve should be able to flow out into a LevelBoundary. This is slightly odd since  in these cases the level of the level boundary does not matter, since the flow is always into the LevelBoundary as determined by the other side. So using a Terminal might make more sense there.

Still, validation shouldn't block this since you may want a single boundary node that connects to multiple node types, some directed like this, some with resistance for which you need a level.


SouthEndMusic (approved):
Looks good to me!

Issue #323, opened 2023-06-16
Start of P(ID) controller


SouthEndMusic:
As motivated in https://github.com/Deltares/Ribasim/issues/322. First steps:

- [ ] Introduce the PID controller node which for now only supports P
- [ ] Have it control a pump which has associated minimum and maximum flow rates

gijsber:
Example controller settings from Landelijk Sobek model (LSMKRW) 
node: PKN_pop_108SASA
Pump capacity 22.5m3/s
Suction side switch on 6.5mNAP, suction side switch off 6.4mNAP
Setpoint is a time series
![image](https://github.com/Deltares/Ribasim/assets/14976756/d4f01ace-1abe-4cec-8254-6377f0ccafb9)

node ZWLM_27
Pump capacity 4.5m3/s
Suction side switch on 17.25mNAP, suction side switch off 17.15mNAP

![image](https://github.com/Deltares/Ribasim/assets/14976756/30c4b738-51d2-45e5-96a6-217c283d5ede)

SouthEndMusic:
> Example controller settings from Landelijk Sobek model (LSMKRW) node: PKN_pop_108SASA Pump capacity 22.5m3/s Suction side switch on 6.5mNAP, suction side switch off 6.4mNAP Setpoint is a time series
> node ZWLM_27 Pump capacity 4.5m3/s Suction side switch on 17.25mNAP, suction side switch off 17.15mNAP

@gijsber thanks for the data. Do I understand correctly that this is not PID control but discrete control (e.g. the pumps are either on or off), or does the second one interpolate between the values in the table? Does the Sobek model also contain profile data?

At this moment I do not know how to incorporate the deadband (which I assume is the tolerance around the setpoint), I can image that to be a criterion for optimization.

gijsber:
@SouthEndMusic I'll have to ask Geert Prinsen for that, but this were the pump-related control types I could identify. they look discrete indeed. and yes, profile data is available. I could check whether the gate or weir controllers have settings that are more PID like. Maybe @visr has been digging into D-RTC more with some example.

For the deadband: this indeed is the lower/upperbound which indicates when the pump is turned on.

gijsber:
Maybe this one is useful as well: the user manual of D-RTC Tools: https://usermanual.wiki/Pdf/DRTCUserManual.1356732301.pdf

gijsber:
[...]

Issue #262, opened 2023-06-01
Testmodels to represent automatic weir behaviour which fluctuates Q to keep H at a predefined level


gijsber:
We need functionality to represent a very common situation in NL where an automatic were adjusts Q to keep H at a predefined level.

Preferably it is defined in an easy to use way, e.g. a specific type of connector node with some embedded control or a specific type of control node

Refinement 16/6/2023 Bart, Martijn, Peter

- [x] https://github.com/Deltares/Ribasim/issues/346
- [x]  https://github.com/Deltares/Ribasim/issues/347
- [x]  https://github.com/Deltares/Ribasim/issues/348
- [x]  https://github.com/Deltares/Ribasim/issues/323 (within the control layer a new PID-control is introduced, with as a first implementation only the P (proportional) part. In future his may be extended to PI and then PID. Links to https://github.com/Deltares/Ribasim/issues/322
- [x]  #300 make tests for usecases above

gijsber:
Use case A: free flow situation with abundant upstream inflow
a fluctuating timeseries at a FlowBoundary, followed by 2 basins with a automatic weir inbetween. The weir has a target_waterlevel for the upstream basin as setpoint and a minimum and maximum waterlevel within which it should stay.
The Q_out is the controlled variable.

gijsber:
Use case A.2: free flow situation with abundant upstream volume and level control on the downstream basin
a fluctuating timeseries at a FlowBoundary, followed by 2 basins with an automatic weir inbetween. The weir has a target_waterlevel for the downstream basin as setpoint and a minimum and maximum waterlevel within which it should stay. the downstream basin has a huge loss (evaporation or seepage) which needs to be compensated by inflow from the upstream basin.
The Q_in of the downstream basin (or the Q_out of the upstream basin) is the controlled variable.

gijsber:
Use case A.3: free flow situation with limited upstream inflow and level control on both the upstream and the downstream basin
a fluctuating timeseries at a FlowBoundary, followed by 2 basins with an automatic weir inbetween. The weir has a target_waterlevel for the downstream basin as setpoint and a minimum and maximum waterlevel within which it should stay. The downstream basin has a huge loss (evaporation or seepage) which needs to be compensated by inflow from the upstream basin. The weir has a target_waterlevel for the upstream basin as setpoint and a minimum and maximum waterlevel within which it should stay.
Keeping the minimum water level at the upstream basin has priority over keeping the minimum water level at the downstream basin.
[...]

PR #329, opened 2023-06-16
P(id)controller

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/323. This is how far I got before my vacation. I've encountered a weird problem: 

I've added an optional column `target_level` to `Basin \ static`. When for instance the basic test model is initiated in Python, the column is automatically added with `None` values. However, if I then write the model data to a geopackage file, the `target_level` column is no longer present. I find this especially strange since this works fine with the current optional column `control_state` in `Pump \ static`.


core/src/create.jl
```diff
@@ -189,15 +189,18 @@ function Control(db::DB, config::Config)::Control
     )
 end
 
-function PIDControl(db::DB, config::Config)::PIDControl
-    static = load_structvector(db, config, PIDControlStaticV1)
+function PidControl(db::DB, config::Config)::PidControl
```
visr, line 192: Why did you rename from PIDControl to PidControl? PID controllers are typically spelled with uppercase PID since it is an acronym.

SouthEndMusic: That is because of the automatic conversion between snake case and upper camel case.

SouthEndMusic:
I am working on the following example of PID control (based on [this](https://github.com/Deltares/Ribasim/issues/262#issuecomment-1582759721) use case).

PID_model

I'm having difficulties using the PID controller to get the basin to a target level, even if the flow of the `flow_boundary` and the `target_level` of the basin are fixed. Simply setting the `flow_rate` of the `pump` proportional to the level error (`target_level` - basin level) does not work, because then the system reaches an equilibrium if the flow boundary flow rate and the pump flow rate sum to 0 for the basin, at some arbitrary level of the basin.

When trying to come up with a solution I keep coming back to the controller having to know what the net storage change of the basin is (without the flow of the pump).

core/src/create.jl
```diff
@@ -131,7 +131,7 @@ function Basin(db::DB, config::Config)::Basin
     check_no_nans(table, "Basin")
 
     # If not specified, target_level = 0
```
Hofer-Julian, line 133: You will want to adapt the comment too :)

SouthEndMusic:
One way to approach this is to not have the pump flow rate as the controlled parameter directly, but *the rate of change* of the flow rate. Then having this rate of change proportional to the storage error would lead to sinusoidal oscillations around the target level, which could be mitigated with a damping derivative term. This does require a computation of the rate of change of the target level.

This I think would require quite some refactoring since this way the flow rate would be a non-storage state variable.

If we let $p(t)$ be the flow rate of the pump, $s(t)$ the storage of the basin and $s_T(t)$ be the target level of the basin. Then the above method yields

$$
   \dot{p}(t) = A(s_T(t) - s(t)) + B(\dot{s}_T(t) - \dot{s}(t))
$$

for given $A,B > 0$.

However, integrating the above equation over $[0,t]$ yields

$$
   p(t) = p(0) + A\int_0^t (s_T(t') - s(t')) \text{d}t' + B(s_T(t) - s(t)).
$$

(Throwing away the controbution for $t=0$ of the last term). Now we are back to controlling the pump directly but now we need the error integral so now this has to be a state (or maybe there is a julia package that can handle this automatically outside the current state vector in the Julia core).

core/src/create.jl
```diff
@@ -173,14 +187,27 @@ function Control(db::DB, config::Config)::Control
...
-        graph,
         record,
     )
 end
 
-function Parameters(db::DB, config::Config)::Parameters
+function PidControl(db::DB, config::Config)::PidControl
+    static = load_structvector(db, config, PidControlStaticV1)
+
+    proportional = coalesce.(static.proportional, NaN)
```
visr, line 197: Not for in this PR, but I'm wondering whether we should move this to `load_structvector`. We probably need a separate issue to test our handling of missing data in tables.

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+        du.storage[i] += precipitation - evaporation + drainage - infiltration
     end
     return nothing
 end
 
+function get_error(pid_control::PidControl, p::Parameters)::Vector{Float64}
+    (; basin) = p
+    (; node_id, listen_node_id) = pid_control
+
+    error = zero(pid_control.derivative)
```
visr, line 313: Should we put this vector in the PidControl struct, so we can keep re-using the same memory rather than allocating a new vector every timestep?

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+
+    for (i, id) in enumerate(node_id)
```
visr, line 315: The `id` is not used currently, perhaps use `for i in eachindex(node_id)` or similar.

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+        @assert has_index "Listen node $listened_node_id is not a basin."
+        target_level = basin.target_level[listened_node_idx]
+        error[i] = target_level - basin.current_level[listened_node_idx]
+    end
+
+    return error
+end
+
+function continuous_control!(
+    du::AbstractVector,
```
visr, line 327:
```suggestion
    du::ComponentVector{Float64},
```

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+    pid_control::PidControl,
+    p::Parameters,
+    integral_::Vector{Float64},
```
visr, line 330: It is not directly clear to me how `integral_` and `integral` are different. Could you explain, perhaps with a comment in the code?

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+
+        if !isnan(integral[i])
+            # coefficient * current value of integral
+            flow_rate += integral[i] * integral_[i]
+        end
+
+        flow_rate = max(flow_rate, 0.0)
+
+        controlled_node_id = only(outneighbors(graph_control, id))
+        controlled_node_idx = findfirst(pump.node_id .== controlled_node_id) # TODO: support the use of id_index
```
visr, line 371: Best to put comments like this on the line above, otherwise the line becomes quite long.

core/src/solve.jl
```diff
@@ -256,21 +301,88 @@ function formulate!(du::AbstractVector, basin::Basin, u::AbstractVector, t::Real
...
+        pump.flow_rate[controlled_node_idx] = flow_rate
+    end
```
visr, line 373:
```suggestion
    end
    return nothing
```

core/src/solve.jl
```diff
@@ -463,30 +586,39 @@ function water_balance!(du, u, p, t)::Nothing
         fractional_flow,
         flow_boundary,
         pump,
+        pid_control,
     ) = p
 
+    storage = collect(u.storage)
+    integral = collect(u.integral)
+
+    dstorage_old = copy(collect(du.storage))
```
visr, line 595: Why do you need `collect` here?

Regarding the `copy`, if possible we should avoid allocating arrays in our time loop. If we need it we should put it in one of the structs so we can allocate it once and reuse.

core/test/bmi.jl
```diff
@@ -45,7 +45,7 @@ end
 @testset "get_value_ptr" begin
     model = BMI.initialize(Ribasim.Model, toml_path)
     u0 = BMI.get_value_ptr(model, "volume")
-    @test u0 == ones(4)
+    @test collect(u0.storage) == ones(4)
```
visr, line 48: Hmm this seems like it should be fixed in `get_value_ptr` rather than the test. Don't we want only the storage part of u if we are asking for "volume"?

python/ribasim/tests/conftest.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import ribasim
 from ribasim_testmodels import (
+    PID_control_model_1,
```
visr, line 6: Shall we keep the functions lowercase? `pid_control_model_1`.

visr (commented):
Left some comments, none major however. Nice work on this!

SouthEndMusic:
@visr regarding the use of `collect`: say `u.storage` is is `SubArray{Float64}`, `collect` turns it into a `Vector{Float64}`.

visr:
Where did you run into issues regarding SubArray versus Array? Both are dense arrays and should work for most cases. Probably we are too strict in some function signatures where now we say `f(a::Vector{Float64}) = ...` but `f(a::AbstractVector{Float64}) = ...` would allow either.

SouthEndMusic:
> Where did you run into issues regarding SubArray versus Array? Both are dense arrays and should work for most cases. Probably we are too strict in some function signatures where now we say `f(a::Vector{Float64}) = ...` but `f(a::AbstractVector{Float64}) = ...` would allow either.

With the function signatures indeed. I also had some other problems but I cannot reproduce those anymore.

visr (approved):

PR #330, opened 2023-06-16
ManningResistance can't connect to LevelBoundary

visr:
ManningResistance doesn't just need the level on either side, but also the botton elevation, which LevelBoundary doesn't have.

@Huite do you think we can easily support LevelBoundary? Otherwise we should merge this PR.

Came up in https://github.com/d2hydro/lhm-ribasim/issues/5#issuecomment-1595142706


visr:
Converted to draft since we can probably support this

https://github.com/Deltares/Ribasim/issues/334#issuecomment-1596774193

PR #332, opened 2023-06-19
Expose cached "level" via `get_value_ptr`

### LINKED ISSUE #353: Expose "level" via `get_value`
Currently, the BMI exposes "level" by directly accessing the `Vector{Interpolation}`.
That doesn't make a lot of sense, it should return `Vector{Float64}` instead.
Since that means computation is involved, `get_value` rather than `get_value_ptr` should be used.

Hofer-Julian:
Fixes #353

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -296,9 +296,10 @@ function BMI.get_value_ptr(model::Model, name::AbstractString)
     if name == "volume"
         model.integrator.u
     elseif name == "level"
-        # FIXME: stop hard-coding "0"
-        hasindex, basin_idx = id_index(basin.node_id, 0)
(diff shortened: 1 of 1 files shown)


visr:
This is supposed to return a `Vector{Float64}` of all the current levels in the Basins, right? If so you can also use `model.integrator.p.basin.current_level`.

Hofer-Julian:
> This is supposed to return a `Vector{Float64}` of all the current levels in the Basins, right? If so you can also use `model.integrator.p.basin.current_level`.

Oh nice, I already wondered if that is cached somewhere.
Is that guaranteed to always point at the same memory location?

visr (approved):

PR #350, opened 2023-06-20
draw edges without edge_type

visr:
I noticed there was one small regression in #335. When drawing new edges, they were invisible, until you stopped editing. At that point the plugin would fill in the current default edge_type = "flow" and it would show up in blue.

With this PR the newly created edges are visible in black, until they are turned into flow edges and become blue again.
The node markers already have a similar "other / unknown / new" marker for this purpose, it makes sense to have that for edges as well.
image


evetion (approved):
Good catch!

PR #351, opened 2023-06-20
saveat is allowed to be a scalar float as well

visr:
Meaning an interval at which to save.

We support this option from [DifferentialEquations](https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/#CommonSolve.solve-Tuple{SciMLBase.AbstractDEProblem,%20Vararg{Any}}):
> If saveat is given a number, then it will automatically expand to `tspan[1]:saveat:tspan[2]`.

This is already tested in Julia, but pydantic didn;t allow this yet. Also added tests.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -1,7 +1,7 @@
 import datetime
 import inspect
 from pathlib import Path
-from typing import Any, List, Optional, Type, cast
+from typing import Any, List, Optional, Type, Union, cast
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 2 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -33,7 +33,7 @@
 class Solver(BaseModel):
     algorithm: Optional[str]
     autodiff: Optional[bool]
-    saveat: Optional[List[float]]
+    saveat: Optional[Union[float, List[float]]]
```
Hofer-Julian, line 36:
```suggestion
    saveat: float | List[float] | None
```

Yes, this requires Python 3.10 but this is just so much more readable

visr: I'm fine with requiring 3.10, but shouldn't the pyproject.toml and other things then also be updated to reflect that?

Hofer-Julian: Good point, let's look into that for a different PR

visr: Does that mean you approve this PR? ;)

Hofer-Julian (approved):

PR #358, opened 2023-06-22
don't add data tables index to GeoPackage

### LINKED ISSUE #344: data tables from ribasim-python have index column
The index column wasn't there before and I think it's not needed. It's not in the schema nor the docs. We probably have to explicitly exclude the pandas index when doing `.to_sql`.

image

visr:
Fixes #344

Passing required columns as an index always led to validation errors, so this should be safe.


evetion:
You also need to fix line 127 in the same file

evetion (approved):

PR #365, opened 2023-06-27
Write files during finalize

Hofer-Julian:
Before, there was no way of writing output files when using the BMI. Now the files are written during finalize

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -59,10 +59,11 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
     return Model(integrator, config, saved_flow)
 end
 
-function BMI.finalize(model::Model)
+function BMI.finalize(model::Model)::Model
     write_basin_output(model)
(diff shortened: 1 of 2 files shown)


core/src/bmi.jl
```diff
@@ -59,6 +59,12 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
     return Model(integrator, config, saved_flow)
 end
 
+function BMI.finalize(model::Model)
+    write_basin_output(model)
+    write_flow_output(model)
+    write_control_output(model)
+end
```
visr, line 66: Let's return model just like the other BMI functions, rather than whatever `write_control_output` returns.

```suggestion
function BMI.finalize(model::Model)::Model
    write_basin_output(model)
    write_flow_output(model)
    write_control_output(model)
    return model
end
```

Hofer-Julian: Sounds good!

visr (approved):

Issue #320, opened 2023-06-15
Test QGIS plugin


Hofer-Julian:
We have an ever-growing of QGIS issues, and we typically notice them weeks after it broke in the code.
The data loading was broken for three months and nobody noticed: https://github.com/Deltares/Ribasim/issues/291
That is because we don't have any kind of automatic tests for the QGIS plugin at the moment.
This is not only the case because it was never prioritized, but also because it is exceptionally hard to do.
This is the case because of two reasons:
- GUIs are harder to test than CLI applications or libraries
- We don't control the whole stack

To elaborate on the second point: We don't ship QGIS with our plugin, but expect users to install the plugin on top of their existing QGIS application. This means that we are confined by QGIS' API, making testing (and many other things) much harder.

I think this issue needs further discussion, mostly because it is unclear to me if the QGIS plugin is valuable enough to justify that kind of investment. https://github.com/Deltares/Ribasim/issues/321

evetion:
Let's test whether it works now (with QGIS bugs fixed)

gijsber:
Manual checks needed on:
- [ ] adding nodes and edges
- [ ] inspect and change the attributes for the tables
- [ ] visualize the output in timeseries graphs
- [ ] visualize the output in animated maps

PR #367, opened 2023-06-27
Add Github action to run QGIS tests.

### LINKED ISSUE #231: CI for QGIS plugin
Currently there is none. We currently only manually check the plugin, see #202.

We should see if we can test at least a part of the plugin automatically. QGIS is available in Conda so perhaps that can be used, for instance to check if it loads properly.

Also Black and Ruff are only run on the python directory, not the qgis one. (EDIT: fixed)

evetion:
Fixes #231.

Relates to #320.

## Description
Use Docker Compose to setup a QGIS environment in which we can test our QGIS plugin.

Running this locally requires:

```bash
docker compose -f .docker/compose.yml up -d
docker exec -t qgis sh -c "qgis_setup.sh ribasim_qgis"
docker exec -t qgis sh -c "cd /tests_directory && qgis_testrunner.sh ribasim_qgis.tests"
```

Note that with the above, you can make changes to the test files, your local files are synced into the QGIS plugin directory in the image.

## Background
Few QGIS plugins have CI. Of the popular plugins that do are several by 3liz, I used https://github.com/3liz/QuickOSM for inspiration, without taking a dependency on their common toolchain https://github.com/3liz/qgis_plugin_tools. I hope to be as generic as possible, so this setup is easily shared by our other QGIS plugins. Any new tests named `qgis/tests/test_*.py`, will be run automatically, without needing to change the CI workflow (scripts).


evetion:
Not sure what's happening with the codecov decreasing each time I make a PR.

evetion:
```
Running test ribasim_qgis.tests.test_io ...
Warning: QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
ERROR: Status 2: File /tests_directory/ribasim_qgis.tests.test_io could not be found
QGIS Test Runner Inside - starting the tests ...
QGIS Test Runner - Trying to import ribasim_qgis.tests.test_io
QGIS Test Runner Inside - executing function <function run_all at 0x7feaf52a7640>
test_open_geopackage (ribasim_qgis.tests.test_io.TestIO) ... ok
test_passes (ribasim_qgis.tests.test_io.TestIO) ... ok

----------------------------------------------------------------------
Ran 2 tests in 0.001s

OK
Finished running test ribasim_qgis.tests.test_io (codes: IS_DEAD=1 IS_FAILED=1 IS_PASSED=0).
```

.docker/start.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
Hofer-Julian, line 1:
```suggestion
#!/usr/bin/env bash
set -euxo pipefail
```

Please set the typical bash options: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

.docker/stop.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
Hofer-Julian, line 1: Same here

.docker/test.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
```
Hofer-Julian, line 1: Same here

qgis/tests/test_io.py
```diff
@@ -0,0 +1,22 @@
...
+from qgis.testing import unittest
+
+
+class TestIO(unittest.TestCase):
+    def test_passes(self):
+        self.assertTrue(True)
+
+    # TODO Open actual geopackage
+    def test_open_geopackage(self):
+        pass
```
Hofer-Julian, line 12: I think we should wait with merging this PR until it actually tests something :)

evetion: This actually does test that the plugin is installable and can load in QGIS.

Hofer-Julian: I see.

Does that also work if you remove the dummy tests?

evetion: And, besides being timeboxed, this does provide the quite generic boilerplate that our other X-qgis packages can use.

Hofer-Julian (commented):
Two nitpicks, then it's fine to go from my side

Hofer-Julian (approved):

evetion:
@JoerivanEngelen FYI

PR #368, opened 2023-06-27
Make Config Nodetype groups CamelCased.

### LINKED ISSUE #316: TOML node type sections
We [probably](https://github.com/Deltares/Ribasim/issues/170) support this in TOML

```toml
[tabulatedratingcurve]
static = "path/to/some/file.arrow"
```

I assumed we had TabulatedRatingCurve (the node type) as a section name, but we use this, extracted from all the schemas:

```julia
julia> Ribasim.nodetypes
10-element Vector{Symbol}:
 :tabulatedratingcurve
 :linearresistance
 :basin
 :fractionalflow
 :control
 :levelboundary
 :pump
 :flowboundary
 :manningresistance
 :terminal
```

I'd like to avoid this spelling since it is hard to read, and use `flow_boundary` for instances, and `FlowBoundary` for the type.

However we cannot use capitals or underscores in the Legolas schema names, so I'm not sure how to fix this. It's hard to automatically go from `flowboundary` to `FlowBoundary` (or `flow_boundary`).

https://github.com/Deltares/Ribasim/blob/15f4499a955378fa82f7c13e8049cb8d62f0b834/core/src/validation.jl#L2

evetion:
Fixes #316 

### Description
So now you have to do this (note `Basin` in `core/test/docs.toml`), I'm not sure if I like it.

```toml
# can also be set to a date-time like 1979-05-27T07:32:00
starttime = 2019-01-01  # required
endtime = 2021-01-01    # required

# all timesteps are in seconds
update_timestep = 86400.0  # optional, default 1 day

# input files
geopackage = "model.gpkg"  # required


# Specific tables can also go into Arrow files rather than the GeoPackage.
# For large tables this can benefit from better compressed file sizes.
# This is optional, tables are retrieved from the GeoPackage if not specified in the TOML.
[Basin]
forcing = "forcing.arrow"


[solver]
algorithm = "QNDF"  # optional, default "QNDF"
# autodiff can only be set to true for implicit solvers, but is not yet supported
autodiff = false  # optional, default false
# saveat can be a number, which is the saving interval in seconds, or
# it can be a list of numbers, which are the times in seconds since start that are saved
saveat = []  # optional, default [], which will save every timestep
dt = 0.0  # optional, default 0.0
abstol = 1e-6  # optional, default 1e-6
reltol = 1e-3  # optional, default 1e-3
maxiters = 1e9  # optional, default 1e9
```

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -4,9 +4,8 @@ import IterTools
 import BasicModelInterface as BMI
 
 using Arrow: Arrow, Table
-using Configurations: Configurations, Maybe, @option, from_toml
+using Configurations: from_toml
 using DataInterpolations: LinearInterpolation
-using DataStructures: DefaultDict
 using Dates
 using DBInterface: execute, prepare
 using Dictionaries: Indices, Dictionary, gettoken, gettokenvalue, dictionary
@@ -28,6 +27,12 @@ TimerOutputs.complement!()
 include("validation.jl")
 include("solve.jl")
 include("config.jl")
+
+# namespaced config.jl to prevent clashes
+const Config = C.Config
+const Solver = C.Solver
+const algorithm = C.algorithm
+
 include("utils.jl")
 include("lib.jl")
 include("io.jl")
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -27,12 +27,7 @@ TimerOutputs.complement!()
 include("validation.jl")
 include("solve.jl")
 include("config.jl")
-
-# namespaced config.jl to prevent clashes
-const Config = C.Config
-const Solver = C.Solver
-const algorithm = C.algorithm
-
+using .config: Config, Solver, algorithm
 include("utils.jl")
 include("lib.jl")
 include("io.jl")
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -1,4 +1,4 @@
-module C
+module config
 
 using Configurations: Configurations, Maybe, @option, from_toml, @type_alias
 using DataStructures: DefaultDict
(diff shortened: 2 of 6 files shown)


core/src/config.jl
```diff
@@ -1,3 +1,12 @@
+module C
```
Hofer-Julian, line 1: How do you feel about calling the module `config` instead of `C`?

core/src/Ribasim.jl
```diff
@@ -28,6 +27,12 @@ TimerOutputs.complement!()
 include("validation.jl")
 include("solve.jl")
 include("config.jl")
+
+# namespaced config.jl to prevent clashes
+const Config = C.Config
```
Hofer-Julian, line 32: I don't think I fully understand why you are adding an alias instead of `using .C: Config, Solver, algorithm`

Hofer-Julian (commented):
> I'm not sure if I like it

I am very happy that this will get rid of `tabulatedratingcurve`.
Is there anything specific that you don't like here?

core/src/Ribasim.jl
evetion: That requires the package C (or config) to be added to the Project.toml (and Manifest) I believe, and I don't want to go down that rabbithole.

Hofer-Julian: Do you mind if I take over this PR and try?

evetion:
> I am very happy that this will get rid of `tabulatedratingcurve`. Is there anything specific that you don't like here?

Everything in that file is lowercased, and suddenly groups have to be CamelCased? I tried to be backwards compatible (also allowing the lowercased alias), but couldn't get that to work quickly.

But if you and Martijn are happy with this, we should merge :)

Hofer-Julian:
> Everything in that file is lowercased, and suddenly groups have to be CamelCased? I tried to be backwards compatible (also allowing the lowercased alias), but couldn't get that to work quickly.

Yeah, snake_case would be more natural for the TOML file.
But that would mean that the type would be snake_case too, right?

Hofer-Julian (approved):

Issue #300, opened 2023-06-12
Add control tests


SouthEndMusic:
Add tests for control behavior based on control output file (https://github.com/Deltares/Ribasim/issues/299).

What should be tested?

See #262 for multiple testcases

gijsber:
- [ ] does the control_state change at the right moment in time ?
- [ ] has the proper control_action been taken which is associated with this control state

PR #384, opened 2023-07-03
Add tests for discrete control times

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/300.

### DIFF
--- a/core/src/io.jl
+++ b/core/src/io.jl
@@ -116,7 +116,7 @@ function parsefile(config_path::AbstractString)::Config
     return from_toml(Config, config_path; relative_dir = dirname(normpath(config_path)))
 end
 
-function write_basin_output(model::Model)
+function get_storages_and_levels(model::Model)
     (; config, integrator) = model
     (; sol, p) = integrator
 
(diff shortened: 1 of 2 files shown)


core/src/io.jl
```diff
@@ -125,16 +125,27 @@ function write_basin_output(model::Model)
...
-    node_id = repeat(basin_id; outer = ntsteps)
-
     storage = reshape(vec(sol), nbasin, ntsteps)
     level = zero(storage)
     for (i, basin_storage) in enumerate(eachrow(storage))
         level[i, :] =
             [get_area_and_level(p.basin, i, storage)[2] for storage in basin_storage]
     end
 
+    return storage, level, tsteps, ntsteps
```
visr, line 135: Shall we remove `ntsteps` from this function, and just calculate it from tsteps in the other function?
Also I think I'd prefer to return a single NamedTuple of Vectors here. That is also a valid Tables.jl table type.

`(; time=tsteps, storage, level)`

visr (changes requested):

visr (approved):

Issue #346, opened 2023-06-20
Extend pump object with min. and max. pump capacity.


SnippenE:
To provide the operating range to the PID controller #323. The pump is used to bring water from downstream (the lower level) to upstream (opposing gravity)

PR #393, opened 2023-07-07
Add pump bounds

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/346. Should the lower bound of the pump default to 0?


gijsber:
yes, default of lower bound=0 m3/s

visr (approved):

PR #399, opened 2023-07-07
disallow multiple outflows from TabulatedRatingCurve

visr:
It is fine to split flows from a Basin by connecting two separate TabulatedRatingCurve nodes to one Basin. But if we allow two outflows from one TabulatedRatingCurve, we create water. So disallow it.

An alternative would be to split the Q into equal parts, but that isn't very useful or generic, and we have FractionalFlow for that.


visr:
Nevermind, I forgot that we actually use this when we let one TabulatedRatingCurve flow via FractionalFlow to different Basins.

Issue #206, opened 2023-05-12
Test node equations


visr:
We are currently lacking unit tests that verify the correct functioning of individual node types. All node types are run by their inclusion into the "basic" model with all node types. For that we check the final states. But we need a more granular level of tests. Like, did we implement the Manning resistance correctly? What about the different edge cases for the TabulatedRatingCurve? If we set this up for one node type, it will be easy to copy the approach for the other node types.

The simplest would be to directly test the `formulate!` methods, so we don't need to create an entire model for each test. We might need some testing utilities that make this easy.

evetion:
Should be one PR per nodetype

visr:
@SouthEndMusic thanks for starting to tackle this in #388!

Do you think it's simpler/better to create a test model to test the equation right now?
Looking at the function with the equation:

https://github.com/Deltares/Ribasim/blob/ed4fbefdc23d0c62ca84939a7f9e6b055f5a1f0b/core/src/solve.jl#L267-L279

It does need an entire parameter struct, but from that it only needs the graph and the flow. Perhaps if we add constructors for Parameters that make the node types optional it may make it easier. To test these functions directly.

SouthEndMusic:
> Do you think it's simpler/better to create a test model to test the equation right now?

I'm not sure what's better. If you just test the function, do you not have the risk of rewriting the function in a test to test the function? With minimal test models you can in many cases compare against an analytical solution.

visr:
I agree that comparing against an analytical solution is a great test. Indeed with a single `formulate!` call you cannot do this, so you quickly end up in the direction of using a test model. That is fine as long as it is easy enough to do, and doesn't make it harder to maintain.

SouthEndMusic:
Are `LinearResistance`, `TabulatedRatingCurve` and `ManningResistance` sufficient or do we need to test more of them?

visr:
Let's also do the Pump, FractionalFlow and Boundaries

PR #402, opened 2023-07-10
Adding Manning resistance equation test

SouthEndMusic:
Fixes part of https://github.com/Deltares/Ribasim/issues/206.


SouthEndMusic:
Not sure why all those previous commits are included here?

Hofer-Julian:
> Not sure why all those previous commits are included here?

I assume because your PR targets `Test-node-equations-feature` rather than `main`.
Was that intended?

SouthEndMusic:
Yes that was intended because this is the feature  branch for equation testing (I'm giving that up after this). My chances still need review though, although it doesn't say so.

Hofer-Julian:
> Yes that was intended because this is the feature branch for equation testing (I'm giving that up after this). My chances still need review though, although it doesn't say so.

What you can try is to merge main into the feature branch.
If that doesn't help with the big diff then we have to look for alternatives.

Hofer-Julian (approved):
One minor comment, apart from that looks good to me!

PR #406, opened 2023-07-10
Add pre-commit

### LINKED ISSUE #394: Add pre-commit
We have vscode settings template, but it still happens quite often that one of us commits code that fails on of our linters.
Let's add [pre-commit](https://pre-commit.com/) following @savente93's presentation.

There are a many useful plugins, feel free to modify this issue to add some(https://pre-commit.com/hooks.html)

- [x] [nbstripout](https://github.com/kynan/nbstripout/#using-nbstripout-as-a-pre-commit-hook) to remove Jupyter notebook output
- [x] trailing-whitespace
- [x] end-of-file-fixer
- [x] check-added-large-files
- [x] github.com/charliermarsh/ruff-pre-commit 
- [x] github.com/psf/black 

We should also use pre-commit in the CI then. How can we cache the virtual environment that's set up by pre-commit?

Hofer-Julian:
Fixes #394

- Add pre-commit
- Fix files
- Add to GitHub Actions
- Add GitHub Actions workflow to update pre-commit repos
- Remove redundant GitHub Actions workflows
- Extend docs


.teamcity/Ribasim_Ribasim/buildTypes/BuildRibasimCliWindows.xml
visr, line 1: Earlier I excluded these files since they are autogenerated. Let's see, I hope this will not cause an edit war between two bots.

.github/workflows/python_lint.yml
visr, line 1: Shouldn't we leave Black and Ruff running in CI? What if people contribute that don't have pre-commit set up?

Hofer-Julian: I don't think I understand your point.
They can still run black and ruff locally. Why would they care how we ensure black and ruff conformance on the CI side?

visr: Oh so we still check on CI, with pre-commit-check.yml? That was what I was wondering.

visr (approved):

.github/workflows/python_lint.yml
Hofer-Julian: I hope so, otherwise I wouldn't know what this action does :P

PR #407, opened 2023-07-10
run ruff hook with autofix before black

visr:
Follow up on #406.

On https://github.com/astral-sh/ruff-pre-commit it states:

> Ruff's pre-commit hook should be placed after other formatting tools, such as Black and isort, unless you enable autofix, in which case, Ruff's pre-commit hook should run before Black, isort, and other formatting tools, as Ruff's autofix behavior can output code changes that require reformatting.


Hofer-Julian (approved):
Thanks!

PR #411, opened 2023-07-11
Other nodes

SouthEndMusic:
Fixes the final part of part of https://github.com/Deltares/Ribasim/issues/206 (close issue when feature branch is merged into main).

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -521,19 +521,21 @@ function formulate!(
 
     for (id, rate) in zip(node_id, flow_rate)
         # Requirement: edge points away from the flow boundary
-        dst_id = only(outneighbors(graph_flow, id))
+        # TODO: Check that only multiple outneighbours exist if these all go to fractionalflow
+        for dst_id in outneighbors(graph_flow, id)
 
-        # Adding water is always possible
-        if rate >= 0
-            flow[id, dst_id] = rate
-        else
-            hasindex, basin_idx = id_index(basin.node_id, dst_id)
-            @assert hasindex "FlowBoundary intake not a Basin"
+            # Adding water is always possible
+            if rate >= 0
+                flow[id, dst_id] = rate
+            else
+                hasindex, basin_idx = id_index(basin.node_id, dst_id)
+                @assert hasindex "FlowBoundary intake not a Basin"
 
-            s = storage[basin_idx]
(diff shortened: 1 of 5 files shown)


core/src/solve.jl
```diff
@@ -521,19 +521,21 @@ function formulate!(
...
-            flow[id, dst_id] = rate
-        else
-            hasindex, basin_idx = id_index(basin.node_id, dst_id)
-            @assert hasindex "FlowBoundary intake not a Basin"
-
-            s = storage[basin_idx]
-            reduction_factor = min(s, 10.0) / 10.0
-            q = reduction_factor * rate
-            flow[id, dst_id] = q
+        # TODO: Check that only multiple outneighbours exist if these all go to fractionalflow
```
Hofer-Julian, line 524: If this is the final part of the issue, then I assume this TODO will stay.
Can you make an issue instead?

core/src/solve.jl
```diff
@@ -521,19 +521,21 @@ function formulate!(
...
+        for dst_id in outneighbors(graph_flow, id)
+
```
Hofer-Julian, line 526:
```suggestion
```

core/test/equations.jl
```diff
@@ -145,3 +145,25 @@ end
 
     @test all(isapprox.(LHS, RHS; rtol = 0.005)) # Fails with '≈'
 end
+
+# Simple solutions: 
+# storage1 = storage1(t0) + (t-t0)*(frac*q_boundary - q_pump)
+# storage2 = storage2(t0) + (t-t0)*q_pump
+# Note: uses Euler algorithm
```
Hofer-Julian, line 152: I feel we should discuss if we find the errors of the default solver acceptable.
If yes, that's also the solver we should use for this test. Agreed @visr?

python/ribasim_testmodels/ribasim_testmodels/equations.py
```diff
@@ -5,7 +5,7 @@
 
 
 def linear_resistance_model():
-    """Set up a minimal model which uses a linear resistance node."""
+    """Set up a minimal model which uses a linear_resistance node."""
```
Hofer-Julian, line 8: I think using natural language would be fine here, but up to you.

Hofer-Julian (approved):

Issue #396, opened 2023-07-07
Introduce discrete flow condition


SouthEndMusic:
Flow is not associated with a node, so the idea is to rename `listen_node_id` in `Control / condition` to `listen_feature_id`. Then the variable can be called something like `edge_flow` so that from this variable name it can be understood that the `listen_feature_id` in this special case is not a node ID but an edge ID.

PR #421, opened 2023-07-12
Flow conditions

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/396.


visr (approved):
Looks good! Would be interesting at some point to evaluate the performance on large models, given that the flows need to be re-formulated.

PR #430, opened 2023-07-17
Avoid pandera 16.0

Hofer-Julian:
See https://github.com/unionai-oss/pandera/issues/1267

That bug made the iMOD Coupler testbench fail: 
https://dpcbuild.deltares.nl/buildConfiguration/iMOD6_Coupler_TestbenchCouplerWin64_2/2568077?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true

### DIFF
--- a/python/ribasim/pyproject.toml
+++ b/python/ribasim/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "geopandas",
     "matplotlib",
     "pandas",
-    "pandera",
+    "pandera != 16.0",
     "pyarrow",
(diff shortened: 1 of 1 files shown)


visr:
Should this not also be set in the environment.yml?

Hofer-Julian:
> Should this not also be set in the environment.yml?

I assume that this release will never hit conda forge. But we can still add it just to be sure.

visr (approved):
Ok then I'm fine with no environment.yml.

PR #431, opened 2023-07-17
Use Deltares blue for navbar

Hofer-Julian:
This way the Ribasim docs also fit the style of the iMOD Documentation

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -7,6 +7,7 @@ website:
   title: "Ribasim"
   page-navigation: true
   navbar:
+    background: "#080c80"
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
> Did you render the docs with this? I've tried this before but I found that it didn't look good with our fonts, making it harder to read. Also I'm not quite sure about this being "Deltares blue", e.g. the Ribasim logo is a lighter shade of blue (which looks bad combined with this PR). I think the PMT is looking into unifying this, so perhaps we should hold off for now.

I found it was okay, but you make good points, let's wait for PMT here

PR #432, opened 2023-07-17
Run update jobs less often

Hofer-Julian:
As discussed with @visr


visr (approved):
Love my monday morning coffee with a pull request.

Issue #419, opened 2023-07-11
Speed up large model


visr:
As far as I can see this is now a valid large model with only static inputs. It should run quickly, but currently takes many hours. Depending on the solver settings it reaches an instability. So we need to investigate why this happens, and do some profiling to see where it spends most time.

Old version: https://github.com/d2hydro/lhm-ribasim/tree/main/models/ribasim_model_nederland
Modified version (profile starts at area 0, Manning become Linear):
- https://github.com/visr/ribasim-artifacts/releases/download/v0.6.0/nl.toml
- https://github.com/visr/ribasim-artifacts/releases/download/v0.6.0/nl.gpkg

visr:
Some initial things I learned:
- taking the median in dudt is a bad idea, fix in #437
- writing output seems to take a long time, in particular the LZ4 compression, we should look into that
- the lookups in the TabulatedRatingCurve are frequent and relatively expensive, look into that
- Euler steps along fine with small dt until it reaches an instability
- ImplicitEuler is very slow, a full Ribasim.run (with output writing) takes 192s for a 1s simulation with 1s timestep.

For the latter, a profile shows almost all time is in LU factorization on a fresh cache, during a linear solve. I assume during successive solves this becomes cheaper. A 10s run (10x longer) takes 6.4 times longer, so partially true. The rest of the time is more running `water_balance!` as part of `jacobian_finitediff_forward!`. We should consider providing [jacobians](https://docs.sciml.ai/DiffEqDocs/stable/features/performance_overloads/#performance_overloads) or looking into autodiff, and looking into linear solver [options](https://docs.sciml.ai/DiffEqDocs/stable/features/linear_nonlinear/#linear_nonlinear).

visr:
The [advanced_ode_example](https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/) tutorial is a good guide here. I tried `Symbolics.jacobian_sparsity` but our `water_balance!` does not like Symbolics at all, see https://github.com/Deltares/Ribasim/commit/c319c6ff9c00db2faaa44d47c8dc9f270a75da90 for the attempt.

But since our problem is so sparse that is a good avenue to explore. We did the same using ModelingToolkit.

Huite:
Some thoughts / speculation:

[...]

PR #437, opened 2023-07-18
use the maximum area for precipitation

visr:
Calculating the median area at runtime showed up as a big performance issue when looking into #419. Both the maximum and medium are quite arbitrary, but the maximum is cheaper and I think easier to explain, since all water that falls within your profile gets directly added to your Basin.

This does mean some tests have to be updated since the precipitation inflows go up.


SouthEndMusic (approved):
One test still needs to be fixed, but otherwise this looks good

Issue #376, opened 2023-06-30
Validate interpolation data


SouthEndMusic:
There have been several problems with models lately that boil down to having invalid data that is interpolated during runtime. These problems can be solved much faster with a validation step with comprehensible error message in the Julia core pre-simulation. The problem is I think always that there is duplication in the independent variable e.g. multiple Q for one h in a Q(h) relationship, or multiple areas for one level in a profile. I think successive zero areas in a profile should also raise an error.

visr:
Closely related to #279 which lists most validation rules. But it's good to have this to separate it from the extrapolation issue.

SnippenE:
Split issue into two:
QH - relation 
Basin geometry (Profile)

PR #438, opened 2023-07-18
Interpolation validation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/376.

The bucket model fails the new validation but I don't know what it is used for so I didn't want to change it.


SouthEndMusic:
@visr I think it is best if you take another look at `function TabulatedRatingCurve`, maybe also change `qh_interpolation` so that the validation is more easily testable.

SouthEndMusic:
I'll have a look at it now

visr (approved):

Issue #439, opened 2023-07-19
Time-dependent flow boundary


SouthEndMusic:
Currently only static flow boundaries are supported. A time-dependent flow boundary is required for https://github.com/Deltares/Ribasim/issues/428.

PR #443, opened 2023-07-21
Time dependent flow boundary

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/439.


core/src/create.jl
```diff
@@ -6,7 +6,7 @@ function parse_static(
 )::NamedTuple
     static_type = eltype(static)
     columnnames_static = collect(fieldnames(static_type))
-    mask = [symb ∉ [:node_id, :control_state] for symb in columnnames_static]
+    mask = [!(symb in [:node_id, :control_state]) for symb in columnnames_static]
```
Hofer-Julian, line 9: I think ∉ was more readable to be honest.
Do you agree @visr?

core/src/create.jl
```diff
@@ -225,13 +210,47 @@ end
...
-        static_parsed.flow_rate,
-    )
+    time = load_structvector(db, config, FlowBoundaryTimeV1)
+
+    static = load_structvector(db, config, FlowBoundaryStaticV1)
+    time = load_structvector(db, config, FlowBoundaryTimeV1)
+
+    static_node_ids = Set(static.node_id)
+    time_node_ids = Set(time.node_id)
+    msg = "Flowboundary cannot be in both static and time tables"
```
Hofer-Julian, line 220:
```suggestion
    msg = "FlowBoundary cannot be in both static and time tables"
```

core/src/create.jl
```diff
@@ -225,13 +210,47 @@ end
...
+    @assert isdisjoint(static_node_ids, time_node_ids) msg
+    static_node_ids = collect(static_node_ids)
+    time_node_ids = collect(time_node_ids)
```
Hofer-Julian, line 223: Why do you need to collect here?

core/src/create.jl
```diff
@@ -225,13 +210,47 @@ end
...
+    node_ids = get_ids(db, "FlowBoundary")
+
+    msg = "FlowBoundary node IDs don't match"
+    @assert issetequal(node_ids, union(static_node_ids, time_node_ids)) msg
+
+    control_mapping = Dict{Tuple{Int, String}, NamedTuple}()
```
Hofer-Julian, line 229: Where is this variable used?

core/src/create.jl
```diff
@@ -225,13 +210,47 @@ end
...
+    active = BitVector()
+    flow_rate = Float64[]
```
Hofer-Julian, line 231: Nice, didn't know this syntax before.

core/src/solve.jl
```diff
@@ -232,10 +232,11 @@ end
 node_id: node ID of the FlowBoundary node
 flow_rate: target flow rate
```
Hofer-Julian, line 233: Document `time` here

docs/core/usage.qmd
```diff
@@ -321,6 +322,19 @@ node_id       | Int     | -            | sorted
 active        | Bool    | -            | (optional, default true)
 flow_rate     | Float64 | $m^3 s^{-1}$ | -
 
+### FlowBoundary / time
+
+This table is the transient form of the `FlowBoundary` table.
+The only difference is that a time column is added.
+The table must by sorted by time, and per time it must be sorted by node_id.
```
Hofer-Julian, line 329:
```suggestion
The table must by sorted by time, and per time it must be sorted by `node_id`.
```

python/ribasim/ribasim/node_types/flow_boundary.py
```diff
@@ -15,6 +17,14 @@ class Config:
         dtype = PydanticModel(models.FlowBoundaryStatic)
 
 
+class TimeSchema(pa.SchemaModel):
+    class Config:
+        """Config with dataframe-level data type."""
+
+        dtype = PydanticModel(models.FlowBoundaryTime)
+        coerce = True  # this is required, otherwise a SchemaInitError is raised
```
Hofer-Julian, line 25: Was that the solution for the empty dataframe problem we encountered yesterday?

Hofer-Julian (commented):
Looks good to me. Maybe @visr can have another look as well.

core/src/create.jl
SouthEndMusic: Because I need to convert from `Set` to `Vector` for other operations later to work, I think `findfirst(static_node_ids .== node_id)`. It's a bit silly though, I'll use `static.node_id` and `time.node_id` directly.

core/src/create.jl
SouthEndMusic: Nowehere :^)

Hofer-Julian: We really need some kind of static analyzer for that

python/ribasim/ribasim/node_types/flow_boundary.py
SouthEndMusic: No it wasn't, I wanted an empty `FlowBoundary / static` but I added a dummy node in there. In the end that was fine since I needed to test having both static and dynamic nodes. The problem persists though, I will post about it in the dev channel.

core/src/create.jl
```diff
@@ -225,13 +210,44 @@ end
...
-    static_parsed = parse_static(static, db, "FlowBoundary", defaults)
-    return FlowBoundary(
-        static_parsed.node_id,
-        static_parsed.active,
-        static_parsed.flow_rate,
-    )
+    time = load_structvector(db, config, FlowBoundaryTimeV1)
+
+    static = load_structvector(db, config, FlowBoundaryStaticV1)
+    time = load_structvector(db, config, FlowBoundaryTimeV1)
```
visr, line 216: Looks like static and time tables are loaded twice?

core/src/create.jl
```diff
@@ -225,13 +210,44 @@ end
...
+
+    msg = "FlowBoundary node IDs don't match"
+    @assert issetequal(node_ids, union(static_node_ids, time_node_ids)) msg
+
+    active = BitVector()
+    flow_rate = Float64[]
+
+    for node_id in node_ids
+        if node_id in static_node_ids
+            static_idx = findfirst(static.node_id .== node_id)
```
visr, line 232: This line is correct but can be improved in two ways. `a .== b` creates an intermediate boolean vector that we can avoid, e.g. by `findfirst(==(node_id), static.node_id)`. `==(node_id)` creates a function that takes one argument that checks if it is equal to `node_id`, similar to `x -> x == node_id`.

But we also know from `load_structvector` that `static.node_id` is sorted, so we can use binary search to speed things up, which will help if there are many FlowBoundaries.

```suggestion
            static_idx = searchsortedfirst(static.node_id, node_id)
```

docs/core/usage.qmd
```diff
@@ -321,6 +322,19 @@ node_id       | Int     | -            | sorted
 active        | Bool    | -            | (optional, default true)
 flow_rate     | Float64 | $m^3 s^{-1}$ | -
 
+### FlowBoundary / time
+
+This table is the transient form of the `FlowBoundary` table.
+The only difference is that a time column is added.
+The table must by sorted by time, and per time it must be sorted by `node_id`.
+With this the rating curves can be updated over time.
```
visr, line 330:
```suggestion
With this the flow rates can be updated over time.
```

python/ribasim/ribasim/models.py
```diff
@@ -161,6 +168,7 @@ class BasinStatic(BaseModel):
 class Root(BaseModel):
     DiscreteControlLogic: Optional[DiscreteControlLogic] = None
     Edge: Optional[Edge] = None
+    FlowBoundaryTime: Optional[FlowBoundaryTime] = None
```
visr, line 171: Nit, but it's easier to read if this is next to the FlowBoundaryStatic. Perhaps we should sort all these lines alphabetically (your editor can do that).

core/test/time.jl
```diff
@@ -0,0 +1,13 @@
+using Ribasim
+
+@testset "Time dependent flow boundary" begin
+    toml_path = normpath(@__DIR__, "../../data/flow_boundary_time/flow_boundary_time.toml")
+    @test ispath(toml_path)
+    model = Ribasim.run(toml_path)
+
+    t = model.saved_flow.t
+    flow = [flows[1] for flows in model.saved_flow.saveval]
+    flow_expected = 1 .+ sin.(0.5 * π * model.saved_flow.t ./ t[end]) .^ 2
```
visr, line 10: It's fine either way, but FYI if you get tired of the dots, you can use the `@.` macro: `@. y = x + 3 * sin(x)` does everything elementwise.

visr (changes requested):
Some small improvements, doc fixes, but otherwise this looks good to me!

core/src/create.jl
SouthEndMusic: Neat, I found some more instances of this

core/test/time.jl
SouthEndMusic: It seems that your usage of the macro is not quite right, it should be `y = @. x + 3 * sin(x)`.

SouthEndMusic:
@visr how do I properly handle 'Changes requested'?

Hofer-Julian:
> @visr how do I properly handle 'Changes requested'?

You wait until @visr gives an approval. Let's see if I can do that too.

Hofer-Julian (approved):

Hofer-Julian:
No, seems like Martijn has to approve himself.

core/test/time.jl
visr: `@. y = x + 3 * sin(x)` also works but only if you already have a vector y, then it fills it in place, if you don't, then indeed your version is the right one to use.

visr (approved):

Issue #428, opened 2023-07-14
Discrete Control tests based on realistic model


SouthEndMusic:
At the time of writing controlling node types other than `Pump` has been merged (https://github.com/Deltares/Ribasim/pull/424), and making nodes active and inactive is on its way (https://github.com/Deltares/Ribasim/pull/425). These features do not have proper tests yet, but they came about form a discussion about a model of the Netherlands. So, this issue is for

- [ ] Constructing said model;
- [ ] Testing discrete control on non-pump nodes and node activity.
- [ ] #482 

For the details of this model see below.

visr:
![20230717_105250](https://github.com/Deltares/Ribasim/assets/4471859/b43d74b9-a165-4eba-8917-b353509be607)

SouthEndMusic:
@gijsber thanks for the information. I have some questions:
- Are the areas equal to the totalwidth tomes the length?
- What are the parameters of LinearResistance (3, 4, 11), TabulatedRatingCurve (8, 13), PidControl (20), LevelBoundary (16), Pump (14)?

Other points:
- Time dependent flow boundaries are not yet supported: https://github.com/Deltares/Ribasim/issues/439
- See here for the current draft script of the model: https://github.com/Deltares/Ribasim/blob/8c1d126ee9be52bd27b8dd7a7be74c1e0b6fd1a1/python/ribasim_testmodels/ribasim_testmodels/dutch_waterways.py

gijsber:
Are the areas equal to the totalwidth times the length?
yes
Here are some best guesses (actually this is a task for Hydrologic):
LinearResistance: 5000 (assumption)
LevelBoundary: 0 (assumption)
Pump: min 0, max 50, default setpoint 25 (assumption)
PID: Kp: -0.005, Ki:0, Kd: -0.002 (from SBK3 model for gate)
QH: 7.45m Q=418m3/s
elke centimeter stijgt Q met 2.15m3/s

SouthEndMusic:
@gijsber I have distilled the discrete control of nodes 17 and 18 as follows.

**Notation**
- `q`: Flow at edge `(1,2)`, flow of flow boundary with id 1
- `w`: Flow at edge `(6,8)`, used to indicate whether node 8 is active

**Conditions**
1. `q > 250`
2. `q > 750`
3. `q > 800`
4. `w > 0`

**Control states**
- "pump_low":
  - pump (node 9) active
  - pump (node 9) flow rate 15
[...]

PR #446, opened 2023-07-21
Dutch waterways model

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/428.


SouthEndMusic:
The model now runs but does not properly converge yet

SouthEndMusic:
One thing to mention here is that the truth states U/D were removed from the docs (`usage.qmd`) but they are still in use (also as user input), so maybe they have to be brought back to the docs?

python/ribasim/ribasim/geometry/node.py
```diff
@@ -43,6 +44,34 @@ def _layername(cls, field) -> str:
     def hasfid(cls):
         return True
 
+    @staticmethod
+    def get_node_ids_and_types(*nodes):
+        node_type = pd.DataFrame(columns=["node_id", "node_type"])
```
visr, line 49: Best to initialize this with typed columns, something like:
```
data_types = {'column1': int, 'column2': float, 'column3': str}
empty_df = pd.DataFrame({col: pd.Series(dtype=dtype) for col, dtype in data_types.items()})
```

python/ribasim/tests/test_model.py
visr, line 1: Was this meant to be deleted?

SouthEndMusic: Yes, the validation function that is tested here was modified (in a previous PR) and is now tested in `test_node_ids_unsequential`.

Issue #440, opened 2023-07-19
Add truth state wildcard for DiscreteControl


SouthEndMusic:
As motivated in https://github.com/Deltares/Ribasim/issues/428#issuecomment-1642378964. The number of required entries in the logic table for `DiscreteControl` can be drastically reduced by introducing a wildcard `A` which means that a condition can be either true or false.

SouthEndMusic:
Refinement needs:
- Is it the best to convert these wildcards into explicit truth states pre-simulation or can they be used as-is and dealt with in the control functionality on the Julia core?
- Is it the best solution to do this processing in Ribasim Python or in the Julia core?
- If it is done in Ribasim Python, how do we handle these processing steps? E.g. is it done when calling `Model.write`, or in a separate method that has to be called beforehand?

SouthEndMusic:
Prototype function for expanding the logic table:

```python
    def expand_logic(logic):
        """
        Expand truth states by creating rows with all possible substitution combinations
        of 'F' and 'T' for 'A'.
        """
        logic_new = pd.DataFrame(columns=("node_id", "truth_state", "control_state"))

        for i, row in logic.iterrows():
            truth_state = row.truth_state
            n_substitutions = truth_state.count("A")

            truth_states_expanded = []

            for substitution in product("TF", repeat=n_substitutions):
                truth_state_expanded = ""
                index_s = 0

                for truth_value in truth_state:
                    if truth_value == "A":
                        truth_state_expanded += substitution[index_s]
                        index_s += 1
                    else:
                        truth_state_expanded += truth_value

                truth_states_expanded.append(truth_state_expanded)

            rows_new = pd.DataFrame(
                data={
                    "node_id": row.node_id,
                    "truth_state": truth_states_expanded,
                    "control_state": row.control_state,
                }
            )

            logic_new = pd.concat([logic_new, rows_new])

        return logic_new
```

SouthEndMusic:
Wildcard symbol is changed to '*'

PR #449, opened 2023-07-24
Control wildcard

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/440.

I mistakenly first thought I had to apply this to `control_mapping` instead of `logic_mapping`.

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -4,7 +4,6 @@ import IterTools
 import BasicModelInterface as BMI
 
 using Arrow: Arrow, Table
-using Base.Iterators
 using Configurations: from_toml
 using ComponentArrays: ComponentVector
 using DataInterpolations: LinearInterpolation
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -4,6 +4,7 @@ import IterTools
 import BasicModelInterface as BMI
 
 using Arrow: Arrow, Table
+using Base.Iterators
 using Configurations: from_toml
 using ComponentArrays: ComponentVector
 using DataInterpolations: LinearInterpolation
--- a/core/src/utils.jl
+++ b/core/src/utils.jl
@@ -329,10 +329,10 @@ function basin_bottoms(
 end
 
 """
-Replace all truth states in the control mapping containing wildcards with
+Replace the truth states in the logic mapping which contain wildcards with
 all possible explicit truth states.
 """
-function expand_control_mapping!(control_mapping::Dict{Tuple{Int, String}, String})::Nothing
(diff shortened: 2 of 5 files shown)


core/src/Ribasim.jl
```diff
@@ -4,6 +4,7 @@ import IterTools
 import BasicModelInterface as BMI
 
 using Arrow: Arrow, Table
+using Base.Iterators
```
visr, line 7: I don't think this is needed? `Iterators` is already available, `Iterators.product` works without this.

core/src/utils.jl
```diff
@@ -327,3 +327,39 @@ function basin_bottoms(
...
+        pattern = r"^[ $(TF*)* ]+$"
+        msg = "Truth state \'$truth_state\' contains illegal characters or is empty."
+        @assert occursin(pattern, truth_state) msg
+
+        n_wildcards = count(==('*'), truth_state)
+
+        if n_wildcards > 0
+            control_state = logic_mapping[(node_id, truth_state)]
+            delete!(logic_mapping, (node_id, truth_state))
+            for substitution in Iterators.product(fill(['T', 'F'], n_wildcards)...)
```
visr, line 348: I'm struggling to understand how this part works, could use a comment.

Since strings are immutable, they are not very convenient for these type of operations. You could also use a `Vector{UInt8}` after checking that the truth_state is valid:

```
julia> UInt8['T', 'F', '*']
3-element Vector{UInt8}:
 0x54
 0x46
 0x2a

julia> String(ans)  # note: consumes the vector
"TF*"
```

core/src/utils.jl
```diff
@@ -327,3 +327,39 @@ function basin_bottoms(
...
+
+"""
+Replace the truth states in the logic mapping which contain wildcards with
+all possible explicit truth states.
+"""
+function expand_logic_mapping!(logic_mapping::Dict{Tuple{Int, String}, String})::Nothing
+    keys_old = collect(keys(logic_mapping))
+
+    for (node_id, truth_state) in keys_old
+        pattern = r"^[ $(TF*)* ]+$"
```
visr, line 339: Doesn't this also allow spaces? Do we want that? Doesn't `^[TF\*]+$` also capture the options?

visr (commented):
What do we do if a user supplies both TF and T* for instance? Can we easily prevent that?

SouthEndMusic:
> What do we do if a user supplies both TF and T* for instance? Can we easily prevent that?

If the user supplies the same control state for both that should be allowed, otherwise it should throw an error. I'll make an issue for this.

visr (approved):

PR #451, opened 2023-07-24
use julia-actions/cache and cache more

visr:
This will hopefully help to speed up CI. [julia-actions/cache](https://github.com/julia-actions/cache) is a wrapper around actions/cache.

I added `cache-compiled: "true"`, and I know what I'm doing (or rather trust people on Slack).


evetion:
Nice! This is something we can re-use in other packages (Wflow, JuliaGeo as well?).

evetion (approved):

visr:
Yes for sure, let's have a look at the CI times to track its effects. `cache-compiled: "true"` might become the default soon.

Issue #405, opened 2023-07-10
Unreasonably high manning flows


SouthEndMusic:
Looking at the flow result of the basic model, the problems with the Manning resistance node have not been resolved yet. Could it be due to a unit mismatch?

![image](https://github.com/Deltares/Ribasim/assets/74617371/4da840ff-712d-4e80-b570-ef243fee7401)

Currently in  `equations.py` the Manning resistance model uses a way to large `manning_n` to get reasonable behavior for testing.

SouthEndMusic:
This may well reveal the cause of the problems with the Manning resistance node:

https://www.desmos.com/calculator/ms3nmwcp8y

As the level difference between the connected basins tends to 0, the corresponding terms in the Jacobian tend to $\pm \infty$. At $\Delta h = 0$ the Jacobian is not defined at all, but this is exactly the state the `basic` models start in. 

I'm quite sure that the root of the problem is with the function

$$
f(x)=\text{sign}(x)\sqrt{|x|}
$$

having the problems described above at $x=0$. There are several approximations of $f$ that do not have this problem, for instance

$$
\begin{align}
g(x) &= \text{sign}(x)\left(\sqrt{|x|+p}-\sqrt{p}\right), \\
h(x) &= \text{sign}(x)\sqrt{\frac{2}{\pi}x\arctan{kx}},
\end{align}
$$

but these have parameters to tune.

With this script you can see how the Manning flow depends on both storages in the `Basic` models (sending a still image doesn't do it justice):

```python
import numpy as np
import matplotlib.pyplot as plt

n = 0.04
L = 900.0
w = 1.0
s = 3

b_a = 0.0
b_b = 0.0

def level(storage):
    return np.sqrt(storage/500)

N = 100
s_a = np.linspace(0,10000,N)
s_b = np.linspace(0,10000,N)

s_a, s_b = np.meshgrid(s_a, s_b)

h_a = level(s_a)
h_b = level(s_b)

d_a = h_a - b_a
d_b = h_b - b_b

d = (d_a + d_b)/2

A_a = w*d + s*d_a**2
A_b = w*d + s*d_b**2
A = (A_a + A_b)/2

P_a = w + 2*d_a * np.sqrt(1 + s**2)
P_b = w + 2*d_b * np.sqrt(1 + s**2)

R_h_a = A_a/P_a
R_h_b = A_b/P_b
R_h = (R_h_a+R_h_b)/2

Dh = h_a - h_b
[...]

PR #464, opened 2023-07-31
Smooth out manning

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/405.

@Huite with the new expression for the Manning flow an analytical solution seems unfeasible. Your test in `run_models.jl` seems good enough, although I don't know how to modify it to make it work again.

### DIFF
--- a/.teamcity/Ribasim_Ribasim/buildTypes/Ribasim_Ribasim_BuildPythonWheels.xml
+++ b/.teamcity/Ribasim_Ribasim/buildTypes/Ribasim_Ribasim_BuildPythonWheels.xml
@@ -37,4 +37,3 @@ conda run -p "%conda_env_path%mm" micromamba run -p %conda_env_path% pip wheel p
     <cleanup />
   </settings>
 </build-type>
-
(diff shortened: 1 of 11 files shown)


Huite (approved):

Huite:
This all looks reasonable and so I've approved. Unfortunately, due to other responsibilities these two days, I don't have time to run the tests myself.

> an analytical solution seems unfeasible

An analytical solution wasn't really feasible before either? (In case you're looking for literature, you'd want to search for "Gradually Varying Flow".) Since the 1950's, these computations have been done with either a graphical or numerical method.

> although I don't know how to modify it to make it work again.

Just increase the tolerance? It's not like my choices were especially carefully chosen, rather that a few centimeters off seems reasonable given slightly different choices of approximation. How much are they off now?

SouthEndMusic:
> An analytical solution wasn't really feasible before either? (In case you're looking for literature, you'd want to search for "Gradually Varying Flow".) Since the 1950's, these computations have been done with either a graphical or numerical method.

I did find an implicit analytical solution for a special case using Wolfram Alpha. Now that I set $k=1000$ the tests work again.

PR #466, opened 2023-07-31
replace basic-transient with basic_transient

visr:
This is the only test model with a dash rather than an underscore, making some things unnecessarily difficult.
In this case, getting the model name from a python variable name that can only have underscores.


SouthEndMusic (approved):
LGTM

PR #468, opened 2023-08-01
Truncate timestamps to milliseconds

### LINKED ISSUE #444: Julia only accepts up to 3 decimals in the seconds in timestamps
I encountered the following error when trying to use timestamps with nanosecond accuracy:

`Unable to parse date time. Expected directive DatePart(s) at char 21`

After some research @Hofer-Julian found that only millisecond accuracy is supported by our implementation: https://docs.julialang.org/en/v1/stdlib/Dates/#Dates.DateFormat, in the following line.

```julia
nt = merge(nt, (; time = DateTime.(nt.time, dateformat"yyyy-mm-dd HH:MM:SS.s")))
```

Millisecond accuracy for the timestamps is fair enough, but the error should be more informative. I would suggest that having more decimals does not throw an error but these get truncated in the Julia core, and only a warning is shown that this has happened.

evetion:
Fixes #444

### DIFF
--- a/core/src/io.jl
+++ b/core/src/io.jl
@@ -84,7 +84,15 @@ function load_structvector(
     if table isa Query && haskey(nt, :time)
         # time has type timestamp and is stored as a String in the GeoPackage
         # currently SQLite.jl does not automatically convert it to DateTime
-        nt = merge(nt, (; time = DateTime.(nt.time, dateformat"yyyy-mm-dd HH:MM:SS.s")))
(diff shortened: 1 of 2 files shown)


SouthEndMusic:
Milliseconds still alowed?

visr (approved):

visr:
Yes. Title is wrong, fractional seconds are truncated to 3 decimals (ms).

Issue #473, opened 2023-08-01
Equation test for PID control


SouthEndMusic:
There is not yet a test with an analytical solution using the `PidControl` node.

PR #474, opened 2023-08-02
Pid equation test

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/473.

There are some complications with the PID derivative term that make it rather complex to implement. The core of the problem is that the derivative term makes the storage change of the listened to basin dependent on itself:

$$
\frac{\text{d}u}{\text{d}t} = \sum F + \sum Q - \phi(u)\left[ K_pe + K_iI + K_d\left(\dot{\text{SP}} - \frac{\text{d}u}{\text{d}t}\right) \right]
$$

where $\phi(u)$ is defined the same as here: https://github.com/Deltares/Ribasim/pull/465.

This computation is further complicated by the clipping of the pump flow value to its minimum or maximum. I don't think it is possible to create an analytical solution of the above equation that converges to the setpoint with only positive flow without using the derivative term (of course that shouldn't be the main argument to keep the derivative term). Here is what the solution looks like (assuming this is correct): https://www.desmos.com/calculator/jqbpnlv8pi.

There is already a `PidControl` test that converges to the setpoint without the derivative term with the pump turning of when the PID control dictates negative flow. I could attempt to formulate an analytical solution of that one.

The reason I started reformulating the derivative term of PID control is that currently the right hand side $\frac{\text{d}u}{\text{d}t}$ of the above equation was taken from not the current but the previous call to `water_balance!`, which I thought could lead to problems (since I got ill behaved solutions).

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -86,8 +86,6 @@ struct Basin{C} <: AbstractParameterNode
     target_level::Vector{Float64}
     # data source for parameter updates
     time::StructVector{BasinForcingV1, C, Int}
-    # Storage derivative for use in PID controller
-    dstorage::Vector{Float64}
 
     function Basin(
         node_id,
@@ -102,7 +100,6 @@ struct Basin{C} <: AbstractParameterNode
         storage,
         target_level,
         time::StructVector{BasinForcingV1, C, Int},
-        dstorage,
     ) where {C}
         errors = valid_profiles(node_id, level, area)
         if isempty(errors)
@@ -119,7 +116,6 @@ struct Basin{C} <: AbstractParameterNode
                 storage,
                 target_level,
                 time,
-                dstorage,
             )
         else
             @error join(errors, "\n")
@@ -252,6 +248,7 @@ struct Pump <: AbstractParameterNode
     min_flow_rate::Vector{Float64}
     max_flow_rate::Vector{Float64}
     control_mapping::Dict{Tuple{Int, String}, NamedTuple}
+    is_pid_controlled::BitVector
 end
 
 """
@@ -461,7 +458,6 @@ function continuous_control!(
     # TODO: also support time varying target levels
     (; connectivity, pump, basin) = p
     (; min_flow_rate, max_flow_rate) = pump
-    (; dstorage) = basin
     (; graph_control) = connectivity
     (; node_id, active, proportional, integral, derivative, listen_node_id, error) =
         pid_control
@@ -483,26 +479,31 @@ function continuous_control!(
 
         listened_node_id = listen_node_id[i]
         _, listened_node_idx = id_index(basin.node_id, listened_node_id)
+        storage_listened_basin = u.storage[listened_node_idx]
+        phi = storage_listened_basin < 10.0 ? storage_listened_basin / 10.0 : 1.0
 
         flow_rate = 0.0
 
         if !isnan(proportional[i])
-            flow_rate += proportional[i] * error[i]
+            flow_rate += phi * proportional[i] * error[i]
+        end
+
+        if !isnan(integral[i])
+            # coefficient * current value of integral
+            flow_rate += phi * integral[i] * integral_value[i]
         end
 
         if !isnan(derivative[i])
             # dlevel/dstorage = 1/area
             area = basin.current_area[listened_node_idx]
-            level_derivative = dstorage[listened_node_idx] / area
-            target_level_derivative = 0.0
-
-            error_deriv = target_level_derivative - level_derivative
-            flow_rate += derivative[i] * error_deriv
-        end
-
-        if !isnan(integral[i])
-            # coefficient * current value of integral
-            flow_rate += integral[i] * integral_value[i]
+            dlevel = du.storage[listened_node_idx] / area
+            dtarget_level = 0.0
+            du_listened_basin_old = du.storage[listened_node_idx]
+            du_listened_basin =
+                (du_listened_basin_old - flow_rate - phi * derivative[i] * dtarget_level) /
+                (1 - phi * derivative[i] / area)
+            du.storage[listened_node_idx] = du_listened_basin
+            flow_rate = du_listened_basin - du_listened_basin_old
         end
 
         # Clip values outside pump flow rate bounds
@@ -512,7 +513,6 @@ function continuous_control!(
             flow_rate = min(flow_rate, max_flow_rate[i])
         end
 
-        println(flow_rate)
         pump.flow_rate[controlled_node_idx] = flow_rate
     end
     return nothing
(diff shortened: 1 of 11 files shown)


SouthEndMusic:
I may have done did the thing though

core/src/solve.jl
```diff
@@ -434,14 +431,19 @@ end
...
     for i in eachindex(listen_node_id)
         listened_node_id = listen_node_id[i]
         has_index, listened_node_idx = id_index(basin.node_id, listened_node_id)
         @assert has_index "Listen node $listened_node_id is not a Basin."
         target_level = basin.target_level[listened_node_idx]
-        error[i] = target_level - basin.current_level[listened_node_idx]
+        if isnan(target_level)
+            error("No target level specified for listen basin #$listened_node_id.")
+        end
+        Error[i] = target_level - basin.current_level[listened_node_idx]
```
visr, line 446:
```suggestion
    pid_error = pid_control.error

    for i in eachindex(listen_node_id)
        listened_node_id = listen_node_id[i]
        has_index, listened_node_idx = id_index(basin.node_id, listened_node_id)
        @assert has_index "Listen node $listened_node_id is not a Basin."
        target_level = basin.target_level[listened_node_idx]
        if isnan(target_level)
            error("No target level specified for listen basin #$listened_node_id.")
        end
        pid_e
```suggestion
    Error = pid_control.error

    for i in eachindex(listen_node_id)
        listened_node_id = listen_node_id[i]
        has_index, listened_node_idx = id_index(basin.node_id, listened_node_id)
        @assert has_index "Listen node $listened_node_id is not a Basin."
        target_level = basin.target_level[listened_node_idx]
        if isnan(target_level)
            error("No target level specified for listen basin #$listened_node_id.")
        end
        Error[i] = target_level - basin.current_level[listened_node_idx]
```rror[i] = target_level - basin.current_level[listened_node_idx]
```

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
 
         listened_node_id = listen_node_id[i]
         _, listened_node_idx = id_index(basin.node_id, listened_node_id)
+        storage_listened_basin = u.storage[listened_node_idx]
+        phi = storage_listened_basin < 10.0 ? storage_listened_basin / 10.0 : 1.0
```
visr, line 483: Does this need to be the same as the `reduction_factor` for the controlling node? Perhaps it's best to use the same syntax: `reduction_factor = min(s, 10.0) / 10.0`.

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
             area = basin.current_area[listened_node_idx]
-
-            error_deriv = -dstorage[listened_node_idx] / area
-            flow_rate += derivative[i] * error_deriv
-        end
-
-        if !isnan(integral[i])
-            # coefficient * current value of integral
-            flow_rate += integral[i] * integral_value[i]
+            dlevel = du.storage[listened_node_idx] / area
```
visr, line 499: Not used?

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
+            dtarget_level = 0.0
+            du_listened_basin_old = du.storage[listened_node_idx]
```
visr, line 501: Perhaps a comment about the need for an implicit approach would be useful.

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
+            flow_rate = min_flow_rate[i]
+        end
 
         if !isnan(max_flow_rate[i])
-            flow_rate = min(flow_rate, max_flow_rate[i])
+            if flow_rate > max_flow_rate[i]
+                was_clipped = true
+                flow_rate = max_flow_rate[i]
+            end
         end
```
visr, line 521: I don't see `was_clipped` being used?

It might be easier to use `clamp(x, lo, hi)` here. That also doesn't need special NaN handling:
```julia
julia> clamp(-4.0, 0.0, NaN)
0.0
```

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
 
         pump.flow_rate[controlled_node_idx] = flow_rate
+        du.storage[listened_node_idx] -= flow_rate
+
+        # Set flow for connected edges
+        src_id = only(inneighbors(graph_flow, controlled_node_id))
+        dst_id = only(outneighbors(graph_flow, controlled_node_id)) # TODO: also take fractional flow into account
```
visr, line 528:
```suggestion
        # TODO: also take fractional flow into account
        dst_id = only(outneighbors(graph_flow, controlled_node_id))
```

core/src/solve.jl
SouthEndMusic: Well that's interesting....

SouthEndMusic: Ah `du.storage[listened_node_idx]` is used lateron

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
+
+        flow[src_id, controlled_node_id] = flow_rate
+        flow[controlled_node_id, dst_id] = flow_rate
+
+        has_index, dst_idx = id_index(basin.node_id, dst_id)
```
visr, line 533: Perhaps add a comment that this is normally only done when formulating `du` in `formulate!(du, connectivity, basin)`, but since we modified the flows we also need to update `du`.

core/src/solve.jl
```diff
@@ -478,34 +479,76 @@ function continuous_control!(
...
+        if has_index
+            du.storage[dst_idx] += flow_rate
+        end
+
+        for id in outneighbors(graph_flow, controlled_node_id)
```
visr, line 538: A PID-controlled Pump flowing to a FractionalFlow is not yet tested, right? Shall we either add a test or disallow it in the validation until we get a chance to add a test? We could create a separate issue for that.

core/src/utils.jl
```diff
@@ -531,30 +531,33 @@ function update_jac_prototype!(
...
 
-        # The basin upstream of the pump
-        has_index, idx_out_in = id_index(basin.node_id, id_out_in)
+        # The basin downstream of the pump
+        has_index, idx_out_out = id_index(basin.node_id, id_pump_out)
 
         if has_index
-            jac_prototype[pid_state_idx, idx_out_in] = 1.0
-            jac_prototype[idx_out_in, pid_state_idx] = 1.0
+            # The basin downstream of the pu,p PID control integral state
```
visr, line 556: typo

visr (commented):
Left some comments

core/src/solve.jl
SouthEndMusic: Nice, that indeed suffices here. In the Jac computations I do need to now whether the flow rate was actually changed here.

core/src/solve.jl
SouthEndMusic: I'll make an issue

visr (approved):

PR #475, opened 2023-08-03
require profile areas to start positive

### LINKED ISSUE #472: Avoid basin areas becoming 0
Basin areas of 0 (which currently occur when a basin is empty) yield division by 0 in the analytical Jacobian (https://github.com/Deltares/Ribasim/pull/465). To avoid this we need some way to make sure that the basin area never becomes 0.

One way to do this is to add a validation for the input areas always being positive. There might also be some parts of the code that already rely on the area of the basin being 0 on the bottom.

visr:
Fixes #472

Since we change the validation from requiring profile areas to start at 0 to requiring a positive area, this breaks all existing models. So therefore all testmodels also had to be updated. I changed 0 to 0.01 m2 (10 x 10cm), sufficiently small not to have to update test states. Or for the cases where we added a sharply widening profile from level 0 to 1 followed by a vertical profile, I just make the profile start at the intended value.

`get_area_and_level` did not need updating since 0 was not hardcoded.

There is one unrelated change with regard to the logging of a vector of error messages, I change that since I think it now looks clearer:

```julia
julia> errors = ["this is wrong", "that is wrong"]

julia> @error join(errors, "\n")
┌ Error: this is wrong
│ that is wrong
└ @ Main REPL[2]:1

julia> foreach(x -> @error(x), errors)
┌ Error: this is wrong
└ @ Main REPL[3]:1
┌ Error: that is wrong
└ @ Main REPL[3]:1
```

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -118,7 +118,7 @@ struct Basin{C} <: AbstractParameterNode
                 time,
             )
         else
-            @error join(errors, "\n")
+            foreach(x -> @error(x), errors)
             error("Errors occurred when parsing Basin data.")
         end
     end
@@ -349,7 +349,7 @@ function valid_n_neighbors(p::Parameters)::Bool
     if isempty(errors)
         return true
     else
-        @error join(errors, "\n")
+        foreach(x -> @error(x), errors)
         return false
     end
 end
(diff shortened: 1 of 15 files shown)


SouthEndMusic (approved):
Looks good!

Issue #413, opened 2023-07-11
Validation of FractionalFlow


SouthEndMusic:
There are two connectivity rules that have to be added to preserve water when using fractional flow nodes:

- [ ] Partition of unity: the fractions of the fractional flow nodes connected to a particular node should be non-negative and sum to 1.
- [ ] If multiple nodes are 'down-edge' of a node that prescribes flow, all these down-edge nodes have to be fractional flow nodes.

PR #477, opened 2023-08-03
Fractional flow validation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/413.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -21,7 +21,7 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
             error("Invalid number of connections for certain node types.")
         end
 
-        (; pid_control, connectivity, basin, pump) = parameters
+        (; pid_control, connectivity, basin, pump, fractional_flow) = parameters
         if !valid_pid_connectivity(
             pid_control.node_id,
             pid_control.listen_node_id,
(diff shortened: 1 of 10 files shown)


SouthEndMusic:
Why does that only fail on Ubuntu 🤔

visr (approved):
The Ubuntu fail is quite strange. I restarted it but still the same issue. It didn't write the new test model this PR adds, though it should since it is written in conftest.py.

The PR seems good to me, nice to have these extra checks in place.

One thing about the logging of errors in `valid_fractional_flow`. I know this pattern is used elsewhere as well, but I see no benefit over collecting error messages in a vector and logging them later, versus logging them directly, and setting a boolean to true, to mark the invalid model. The function `valid_pid_connectivity` above uses this. When an error is logged by default it notes where in the source code this is coming from. This will then also be a more useful location as well.

```
┌ Error: this is wrong
└ @ Main REPL[3]:1
┌ Error: that is wrong
└ @ Main REPL[3]:1
```

SouthEndMusic:
@visr isn't it bad to have this fail on main?

visr:
Yes it is. We could merge it though to double check if it also errors on main. If it does, we can skip the test on Ubuntu while we figure out the issue.

Issue #471, opened 2023-08-01
Flow condition only on interpolated flow boundary


SouthEndMusic:
As https://github.com/Deltares/Ribasim/issues/456 demonstrates, the current implementation of a dynamic flow boundary from a time series with forward fill does not work. @visr and I came to the conclusion that the best solution for now (in particular for the use case https://github.com/Deltares/Ribasim/issues/428) is to:

- [ ] Refactor dynamic flow boundaries so that they are linearly interpolated (for now just with a `LinearInterpolation` object, later it can be decided whether it is better to add callbacks and whether it yields a significant speedup if the code does not have to compute the interpolation slope each time);
- [ ] Have flow conditions only on flow boundaries. Flow conditions on edges are in general still a problem since discrete control actions can create discontinuities in flow in various ways.

SouthEndMusic:
An implementation was done of linear interpolation of the flow without using callbacks (https://github.com/Deltares/Ribasim/pull/478/commits/4526100c4b5a3ec46623bc0785afe26239701488). The problem with this, at least in a minimal test model, that the solver misses a lot; in a test where the the flow boundary is constant at the beginning and only changes after a few months, the solver assumes a fully constant flow which resulted in literally only 2 time points in the entire simulation. Therefore I am now switching to linearly interpolated transient flow boundaries combined with callbacks.

PR #478, opened 2023-08-04
Flow condition on flow boundary

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/471.


visr (approved):
Thanks, this looks good to me.
Since negative flow_rate is no longer supported, didn't you intend for FlowBoundary to only add water to the system? If so, I guess Basin going to FlowBoundary should be disallowed from `neighbortypes`?

Issue #347, opened 2023-06-20
Add weir object with opposite characteristics of pump


SnippenE:
See #322

A weir object is introduced in the physiscal layer with the same characteristics as the pump (min.capacity, max.capacity, static flow rate(s), transient flow rate, controllable flowrate), but an opposite flow direction (from upstream to downstream following gravity)

gijsber:
Basically this could be a naming discussion. If we rename the current Pump object to something more universal, we do not need a second implementation. 
Let's pause this one for the moment

Issue #482, opened 2023-08-04
Tests for PID controlled pump


SouthEndMusic:
New cases for PID control to be tested emerged due to refactoring (https://github.com/Deltares/Ribasim/pull/474, https://github.com/Deltares/Ribasim/pull/465). These test cases are:

- [ ] The PID controlled pump flowing out into `FractionalFlow` nodes;
- [x] The PID controlled basin influencing itself trough being connected to e.g. a `LinearResistance`. This could be added to the equation test by connecting a `LevelBoundary` via `LinearResistance` to the PID controlled basin. This was first trought to be hard to support (https://github.com/Deltares/Ribasim/issues/481) but it seems to be supported now;
- [ ] The PID controlled basin getting in the 'reduction factor regime' (although this is probably quite rare in practice).
- [x] Upstream of the PID controlled weir is a `LevelBoundary` node (done in https://github.com/Deltares/Ribasim/pull/485).

SouthEndMusic:
Regarding the test of the controlled pump flowing out into fractional flow nodes: this possibility is also not covered yet in 

```julia
function update_jac_prototype!(
    jac_prototype::SparseMatrixCSC{Float64, Int64},
    p::Parameters,
    node::PidControl,
)::Nothing
```

SouthEndMusic:
Regarding the following:

> The PID controlled basin influencing itself trough being connected to e.g. a `LinearResistance`

In order to do this, the $\frac{\partial E}{\partial u_i}$ (see `equations.qmd` in https://github.com/Deltares/Ribasim/pull/485) are needed. Most of these are $0$; a nice method has to be found to figure out which aren't.

visr:
Paused this for now, we'll probably work on #152 first. Until then `jac = false` can be used for any models for which the Jacobian is not complete.

PR #485, opened 2023-08-08
Add weir

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/347.


Huite (changes requested):
While I'm no longer intimately familiar with the code base, this looks good to me -- the change is also somewhat minor as the weir is arguably just a Pump alias.

I think "weir" is maybe not the best term: it's explicitly a weir that maintains the level. I don't think the name has to be a blocker, but maybe it's worth keeping an issue for it.

What would be good to change before merging is to resolve the auto-generated code coverage remarks. Ideally the output is tested as well, but this is often difficult (hence 100% coverage doesn't mean 100% reliable). However, given that Julia is a dynamic language, it would be good that (nearly) all lines run during a test.

core/src/solve.jl
```diff
@@ -739,13 +814,21 @@ function formulate!(flow_boundary::FlowBoundary, p::Parameters, t::Float64)::Not
...
+    (; connectivity, basin) = p
     (; graph_flow, flow) = connectivity
-    (; node_id, active, flow_rate, is_pid_controlled) = pump
+    (; node_id, active, flow_rate, is_pid_controlled) = node
     for (id, isactive, rate, pid_controlled) in
         zip(node_id, active, flow_rate, is_pid_controlled)
-        @assert rate >= 0 "Pump flow rate must be positive, found $rate for Pump #$id"
+        if rate < 0
+            error(
+                "$(typeof(node)) flow rate must be non-negative, found $rate for Pump #$id.",
```
Huite, line 829: This error message is generated for both Weir and Pump, but only mentions "Pump".

SouthEndMusic: I refactored the flow rate sign test, because:
1. It is better validate this before the simulation than find out during
2. It was very fiddly to test the previous version

SouthEndMusic:
Many of the changes in this PR also affect `Pump` (mainly in `jac.jl`). Not all of it is tested, but this is covered in https://github.com/Deltares/Ribasim/issues/482.

core/src/jac.jl
```diff
@@ -391,18 +390,40 @@ function formulate_jac!(
             continue
         end
 
-        id_pump = only(outneighbors(graph_control, id))
+        controlled_node_id = only(outneighbors(graph_control, id))
+        controls_pump = (controlled_node_id in pump.node_id)
```
visr, line 394: The brackets can be removed here. But looking at the if-else block below, it'd be better to replace the else with a check if it is a weir, and add an else that throws an error, to be a bit more defensive.

And rather than checking all pump nodes with `x in y`, you could use `isempty(searchsorted([2,3], 4))`, which we could perhaps add as a utility function besides `findsorted`, with a name like `insorted`. Ha, actually I just found out that `insorted` already exists in Base.

SouthEndMusic: I added an extra check. In a case like `controls_pump = (controlled_node_id in pump.node_id)` I do prefer to write the brackets otherwise I find the syntax confusing.

Issue #460, opened 2023-07-26
Validate control states


SouthEndMusic:
At this point there is no validation for checking that all control states that can be reached by a control node are actually defined for the nodes it controls.

PR #488, opened 2023-08-10
Validate control states

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/460.

### DIFF
--- a/python/ribasim_testmodels/ribasim_testmodels/invalid.py
+++ b/python/ribasim_testmodels/ribasim_testmodels/invalid.py
@@ -64,6 +64,7 @@ def invalid_qh_model():
     basin = ribasim.Basin(profile=profile, static=static)
 
     rating_curve_static = pd.DataFrame(
+        # Invalid: levels must not be repeated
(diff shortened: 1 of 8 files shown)


python/ribasim_testmodels/ribasim_testmodels/invalid.py
```diff
@@ -195,3 +195,119 @@ def invalid_fractional_flow_model():
...
+
+    # Setup pump:
+    pump = ribasim.Pump(
+        static=pd.DataFrame(
+            data={
+                "node_id": [2],
+                "flow_rate": [0.5 / 3600],
+            }
+        )
+    )
```
evetion, line 267: Is the error here that this Pump has no state at all? Maybe nicer to set a state to "bar" here?

evetion: And it would be helpful for the future to indicate (in a comment) where the error in the code is.

evetion (approved):
Nice improvement to document the failing models.

Issue #454, opened 2023-07-25
Deceptive crossings of condition thresholds


SouthEndMusic:
The problem described here can be summarized as *"what starts as an up-crossing does not have to end as an up-crossing"*.

### Example

In the example below we look at the level of a basin trough time.

![control_problem](https://github.com/Deltares/Ribasim/assets/74617371/125576d9-aa58-4e57-8ce3-4891c2e903dd)

As the level crosses the upper threshold, a `Pump` flow rate is changed to bring the level to the setpoint (middle threshold). The `VectorContinuousCallback` detects the crossing of the upper threshold as an up-crossing, and so it is assumed that the condition 

`level > upper_threshold`

is true just after the time when `level = upper_threshold`:

```julia
function discrete_control_affect_upcrossing!(integrator, condition_idx)
    discrete_control = integrator.p.discrete_control
    discrete_control.condition_value[condition_idx] = true

    discrete_control_affect!(integrator, condition_idx)
end
```

However, due to the change induced by control, this turns out not to be an up-crossing at al, but merely a 'touch' of the threshold.

This is a non-trivial problem, because only after the control change and a call of `water_balance!` is the direction of this crossing really clear, by the sign of the `du` of this basin.

### Potential solution

At the moment it is quite expensive to get the `du` of a single basin (or the flow over a single edge for that matter, needed for flow conditions), because the implementation only supports computing all of them in one function call. It would be nice to have a way to only calculate the `du` of a single basin after a control state change (or the flow of a single edge), to have a relatively cheap way to infer the crossing type.

SouthEndMusic:
Using [DiscreteCallback](https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#discrete_callback) does not have this problem but is less precise on the time of the control change, dependent on `dt`.

Issue #447, opened 2023-07-22
Up- and down-crossing aware truth states


SouthEndMusic:
### Motivation

The line below shows a range of a to-be-controlled parameter/state `s`, with indicated upper and lower bounds (`ub`, `lb`) and upper and lower setpoints (`usp`, `lsp`).

```
lb   *   lsp            usp    *    ub      
|---------|--------------|----------|

s ->
````

Quite common control use cases are of these types:
- Once `s < lb`, `s` must rise up to `lsp`;
- Once `s > ub`, `s` must fall down to `usp`.

Up to now, we used to view this problem as 'when the parameter `s` is in one of these starred (`*`) regions, there must be some memory that says whether `s` is on its way to a setpoint or not'.

However, `DiscreteControl` only takes action at changes of conditions. Therefore, it is useful to consider how the parameter `s` entered the starred regions. More specifically:
- `s` is on its way to `lsp` if it did an up-crossing at `lb`;
- `s` is on its way to `usp` if it did a down-crossing at `ub`.

### Proposal by example

Say we have the following conditions for the situation described above:

1. `s > lb`
2. `s > lsp`
3. `s > usp`
4. `s > ub`

and also the control states 'rise' and 'fall', 'free' for making `s` rise, fall, and change freely respectively.

We introduce truth values in addition to `T` and `F`:
- `U` for an up-crossing, which is a special case of `T`; 
- `D` for a down-crossing, which is a special case of `F`.

Then for the upper set point case, the control logic can be as follows (using the wildcard 'A' from https://github.com/Deltares/Ribasim/issues/440):

- AAAT -> 'fall'
- AAAD -> 'fall'
- AATF -> 'free' 
- AAFA -> 'free'

Note that the second case implies the third case. So, we say that *crossing-type specific truth states take precedence*. 

### Shortcoming

The above does not work when an extra condition on `s` is added within the relevant starred region.

Issue #487, opened 2023-08-09
Formulating the `du` of a single basin


SouthEndMusic:
There are several internal use cases for computing the `du` of a single basin in the Julia core:
- The `du` of a PID controlled basin is needed for the Jacobian if the derivative term (D) is used*. The Jacobian is computed outside the scope of `water_balance!`, so `water_balance!` is called in `water_balance_jac!` to get this specific `du`, which is needlessly expensive for large models;
- In the context of `DiscreteControl` there is the idea of 'crossing aware' truth states (see https://github.com/Deltares/Ribasim/issues/447) but these (and also the current implementation without this feature) suffers from deceptive crossings of level thresholds (see https://github.com/Deltares/Ribasim/issues/454). The crossing direction after a `DiscreteControl` parameter change could be derived from the sign of the `du` of the basin on which the level condition is set.

`water_balance!` is currently not structured in a way that easily facilitates this. Here is how I envision this to be implemented:
1. Have some function, say `get_basin_dependency_node_ids(basin_node_id::Int, p::Parameters)`, which gives you for a specified basin the ID's of the surrounding nodes that contribute to the `du` of this basin (note in particular the role of `FractionalFlow` nodes) per node type;
2. Figure out some way to use the `formulate!` functions only for this basin and the node ID's given by `get_basin_dependency_node_ids`.

*. That is, the `du` without the contribution of the regulating `Pump`/`Weir`.

visr:
Since this issue is about performance, let's hold off until we see this being a bottleneck in real models.

visr:
No longer needed since we switched to autodiff.

PR #492, opened 2023-08-11
Level control with setpoint and min/max bounds

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/447, fixes https://github.com/Deltares/Ribasim/issues/454.

To demonstrate the result of this PR we consider this model:

setpoint_with_bounds_example_model

The basin has a net forcing coming from the level boundary and infiltration. The basin has a setpoint and minimum and maximum bounds for its level. We want to activate one of the pumps when the level gets outside these bounds and bring it back to the setpoint. This has indeed been achieved:

setpoint_with_bounds_example

This has been implemented using crossing aware truth states (https://github.com/Deltares/Ribasim/issues/447), and the 'deceptive crossing' problem (https://github.com/Deltares/Ribasim/issues/454) has been solved by reformulating the `du` of the level condition basin after the parameter change. This is now done in a very inefficient way, but this can be improved (https://github.com/Deltares/Ribasim/issues/487).

Some remaining points:
- The setpoint in this case is implemented as a level condition, not as a `target_level` in `Basin \ static` which is only used for PID control. This might be confusing for users;
- Since this is a much more realistic example of control, I think this would be much more informative as the control example in `examples.ipynb`. The only possible drawback is that I constructed the net forcing in a bit of a hacky way, by starting with the forcing I wanted to have and distributing the positive and negative values over the flow boundary and precipitation respectively;
- I haven't tested this yet, I'm not sure what the best way would be to test this.


python/ribasim_testmodels/ribasim_testmodels/time.py
```diff
@@ -102,3 +102,157 @@ def flow_boundary_time_model():
...
+        flow_boundary=flow_boundary,
+        discrete_control=discrete_control,
+        starttime="2020-01-01 00:00:00",
+        endtime="2021-01-01 00:00:00",
+    )
+
+    return model
+
+
+if __name__ == "__main__":
```
Huite, line 253: You probably want to remove this part.

python/ribasim_testmodels/ribasim_testmodels/time.py
```diff
@@ -102,3 +102,157 @@ def flow_boundary_time_model():
     )
 
     return model
+
+
+def conditions_on_discrete_flow_model():
```
Huite, line 107: Could you add a docstring on what this model exactly represents (and tests)?

(As far as I can tell, it always listens to the flow of node 1, checks the flow, and as different values are exceeded, the pump is given a higher rate?)

SouthEndMusic: I'll ad a docstring (and give it a better name). What the model does is explained in https://github.com/Deltares/Ribasim/issues/447, and demonstrated above. We discussed this at length a few weeks back, about going back to the setpoint once the level is outside an accepted range.

Huite:
I don't think I quite comprehend the changes, and so I think it's best to discuss this with @visr and you next week.

This adds additional truth states of "U" and "D". I'm afraid this is too hard a sell for a lot of users. 

Regarding your remarks:

> The setpoint in this case is implemented as a level condition, not as a target_level in Basin \ static which is only used for PID control. This might be confusing for users;

This is indeed unfortunate, and maybe it suggests that we should refactor. If different control mechanisms require different shapes of input, they just require separate inputs. I.e. the target level should be a property of the PID controller, not of the basin. 

> Since this is a much more realistic example of control, I think this would be much more informative as the control example in examples.ipynb. The only possible drawback is that I constructed the net forcing in a bit of a hacky way, by starting with the forcing I wanted to have and distributing the positive and negative values over the flow boundary and precipitation respectively;

We can make the example (the forcing) a lot simpler by including an additional node, e.g. a TabulatedRatingCurve, which causes the basin to slowly empty instead of the flow boundary. Then we would like to see the next phases:

1. The initial basin water level is too high. The pump is turned on. At the same time, the rating curve generates a trickle of water out of the basin (but is dwarfed by the pump).
2. The water basin level reaches the setpoint. The pump is turned off. Now the water level starts decreasing slowly due to the rating curve trickle.
3. The water basin level drops below the minimum value. The other pump is turned on.
4. The water basin level rises again and reaches the setpoint. The other pump is turned off.
5. The water basin level drop again.

This doesn't require any time dependent input, and the output level is much easier to interpret than the effect of the sinusoidal forcing.

> I haven't tested this yet, I'm not sure what the best way would be to test this.

I think the essential feature is that the model aims for a setpoint, but doesn't activate until another boundary value is crossed.
So if you use constant precipitation or constant evaporation, you expect to see a triangle wave. We can easily generate the expected values of the triangle wave we expect to see.

python/ribasim/ribasim/geometry/edge.py
```diff
@@ -65,7 +65,9 @@ def write(self, directory: FilePath, modelname: str) -> None:
             gdf = gdf.set_geometry("geometry")
         else:
             gdf["geometry"] = None
-        gdf.to_file(directory / f"{modelname}.gpkg", layer=name)
+        gdf.to_file(
+            directory / f"{modelname}.gpkg", layer=name, driver="GPKG"
+        )  # , engine="pyogrio"
```
visr, line 70:
```suggestion
        )
```

python/ribasim/ribasim/geometry/node.py
```diff
@@ -63,7 +63,11 @@ def write(self, directory: FilePath, modelname: str) -> None:
         gdf = gpd.GeoDataFrame(data=dataframe)
         gdf = gdf.set_geometry("geometry")
 
-        gdf.to_file(directory / f"{modelname}.gpkg", layer=name)
+        gdf.to_file(
+            directory / f"{modelname}.gpkg",
+            layer=name,
+            driver="GPKG",  # engine="pyogrio"
```
visr, line 69:
```suggestion
            driver="GPKG",
```

visr (approved):
Looks good to me. Two minor suggestions.

Issue #494, opened 2023-08-14
Making target level a property of the `PidControl` node


SouthEndMusic:
As discussed in https://github.com/Deltares/Ribasim/pull/492, there are currently several ways to implement a target level for a basin. To avoid confusion for users, it is better to move the `target_level` which is currently a column of the `Basin \ static` table to the `PidControl \ static` table (as this `target_level` is currently only used by the `PidControl` node). This must be done before https://github.com/Deltares/Ribasim/issues/397.

PR #498, opened 2023-08-14
target level to PID control

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/494.


docs/core/usage.qmd
```diff
@@ -466,6 +465,7 @@ column         | type     | unit     | restriction
 node_id        | Int      | -        | sorted
 active         | Bool     | -        | (optional, default true)
 listen_node_id | Int      | -        | -
+target_level   | Float64  | $m$      | -
```
evetion, line 468: Is optional

docs/schema/PIDControlStatic.schema.json
```diff
@@ -49,12 +51,17 @@
...
+            "format": "double",
+            "description": "target_level",
+            "type": "number"
         }
     },
     "required": [
         "node_id",
         "listen_node_id",
-        "proportional"
+        "target_level"
```
evetion, line 64: Why the change from proportional to target_level? You've used coalesce elsewhere, so that implies it's not always required?

evetion:
@Huite If we move this parameter `target_level` of a Basin to the PID control, I propose to rename to just `target` (or similar generic name), as the PID controller ideally should just know about values (y-intercept and their changes). Do you agree?

docs/core/usage.qmd
SouthEndMusic: How can it be optional? The PID control cannot function without a setpoint. For the Basin it was optional because not all basins have a target level (so before it was required that the PID control node was connected to a basin with a target_level).

docs/schema/PIDControlStatic.schema.json
SouthEndMusic: It's not a change from one to the other. Proportional was removed as a required input so that they are all optional. This is handled nicely in the Julia core by skipping the computations for the terms where the coefficient is not defined (NaN). Target level is required as discussed above.

gijsber:
Agree. Use either `target` or `setpoint` as they may be levels or flows

SouthEndMusic:
> @Huite If we move this parameter `target_level` of a Basin to the PID control, I propose to rename to just `target` (or similar generic name), as the PID controller ideally should just know about values (y-intercept and their changes). Do you agree?

@evetion I'll change it to 'target'. 

The implementation of the PID controller is not easily made that generic, variables each have their own algebraic and implementation challenges. For instance if you control flow you run into trouble with the derivative term because flow can change discontinuously over time (e.g. due to `DiscreteControl`). Also, in general certain Jacobian terms are solutions to implicit equations with dependencies on nodes in the vacinity, see https://deltares.github.io/Ribasim/core/equations.html#sec-PID.

visr (approved):

PR #504, opened 2023-08-15
Allow setting verbosity in the config toml

### LINKED ISSUE #370: Progress bar and verbosity at Julia core runtime
As discussed at the 28-6-2023 samenwerkdag, it would be nice to receive from feedback from the Julia core during a successful run. 

- A progress bar of the simulation time would be nice. It seems to be already activated (see below), but for some reason it does not show.
```julia
    @timeit_debug to "Setup integrator" integrator = init(
        prob,
        alg;
        progress = true,
        progress_name = "Simulating",
        callback,
        config.solver.saveat,
        config.solver.dt,
        config.solver.abstol,
        config.solver.reltol,
        config.solver.maxiters,
    )
``` 

Of course we would never get an uniformly moving progress bar, because of changing timestep etc.

- It would be nice to add some verbosity options, for instance for control.

evetion:
Fixes #370 

#### Description
One can now set the verbosity in the toml to `debug`, `info`, `warn`, and `error`. It only takes a little bit of piracy. Also, one can set the `timings` to trigger `@debug_timings` statements for debugging.

It is only triggered on `BMI.run` though, as we don't have a parsed config much earlier than that, certainly not in the ribasim_cli.

This was slightly harder to do than anticipated, and required LoggingExtras.jl to work somewhat intuitively. On the plus side, it should become really easy now to add logging to file (with timestamps) on top of this.


evetion:
Looking over the Python code, I think we should add a technical debt issue to improve it, it's a mix of config/nodes, so it doesn't easily generalize.

Ideally we can rewrite most of it in a style like https://github.com/Deltares/HYDROLIB-core/blob/bf75f024e48b515d4cd7c93cd3f997de4fc5b1ef/hydrolib/core/dimr/models.py#L80 (full-disclosure, I probably wrote that).

core/src/config.jl
```diff
@@ -97,6 +98,19 @@ end
     compression_level::Int = 6
 end
 
+function Base.convert(::Type{LogLevel}, level::AbstractString)
```
visr, line 101: Can we not avoid piracy by overloading `Configurations.from_dict`?

https://configurations.rogerluo.dev/stable/convert/

core/src/bmi.jl
```diff
@@ -494,9 +508,26 @@ BMI.get_time_step(model::Model) = get_proposed_dt(model.integrator)
...
+
 function run(config::Config)::Model
-    model = BMI.initialize(Model, config)
-    solve!(model.integrator)
-    BMI.finalize(model)
-    return model
+    logger = current_logger()
+
+    # Reconfigure the logger if necessary with the correct loglevel
+    # but make sure to only log from Ribasim
```
visr, line 519: Why did you only want logging from Ribasim? For debugging I guess sometimes you might also want all the info from DifferentialEquations.

evetion: Will make a new issue for this.

core/src/config.jl
```diff
@@ -134,6 +128,18 @@ end
...
 end
 
+function Configurations.from_dict(::Type{Logging}, ::Type{LogLevel}, level::AbstractString)
+    level == "debug" && return LogLevel(-1000)
+    level == "info" && return LogLevel(0)
+    level == "warn" && return LogLevel(1000)
+    level == "error" && return LogLevel(2000)
+    throw(
+        ArgumentError(
+            "verbosity $level not supported, choose one of: debug info warn error.",
```
visr, line 138: For consistency maybe also use this, like above:
`$(join(instances(Compression), " "))`

evetion: There is no such thing for LogLevel I believe

visr (approved):

Issue #294, opened 2023-06-08
Supply levels instead of storages as initial condition


SouthEndMusic:
As discussed here: https://github.com/Deltares/Ribasim/issues/288#issuecomment-1581987597.

PR #506, opened 2023-08-16
Input initial levels

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/294.


core/src/utils.jl
```diff
@@ -55,6 +55,56 @@ function create_storage_tables(
...
+        # If the level is equal to the bottom then the storage is 0
+        if level_lower_index == 0
+            push!(storages, 0.0)
+            continue
+        elseif level_lower_index == length(level_discrete)
+            level_lower_index -= 1
+        end
+
+        darea =
+            (area_discrete[level_lower_index + 1] - area_discrete[level_lower_index]) /
```
Huite, line 91: I think I'd prefer to see this packaged into a separate function, it's a nice logical unit. E.g. something like "interpolate_storage" or something:

```julia
function interpolate_storage(basin, level)
...
return storage
end

```

python/ribasim_testmodels/ribasim_testmodels/pid_control.py
```diff
@@ -77,7 +77,7 @@ def pid_control_model():
     state = pd.DataFrame(
         data={
             "node_id": [2],
-            "storage": [500.0],
+            "level": [0.5],
         }
     )
```
Huite, line 83: Is the QGIS change missing? https://github.com/Deltares/Ribasim/blob/b1a695c1d7abaffa38b18a43d09823064bc577a5/qgis/core/nodes.py#L326

Huite (changes requested):

core/test/equations.jl
```diff
@@ -96,12 +90,12 @@ end
     t = Ribasim.timesteps(model)
     storage = Ribasim.get_storages_and_levels(model).storage[1, :]
     basin_area = p.basin.area[1][2]
-    storage_min = 50.0
+    storage_min = 50.005
```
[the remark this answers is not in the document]
SouthEndMusic: It turns out for the test to pass it is not necessary. This stems from when the area of the profile had to start at 0, which means that towards the bottom the area cannot be constant. So I made these tests so that after this minimum storage the area is constant, which makes it easier to obtain an analytical solution.

core/src/utils.jl
SouthEndMusic: I refactored it

python/ribasim_testmodels/ribasim_testmodels/pid_control.py
SouthEndMusic: Yep, thanks

Issue #491, opened 2023-08-11
Renaming `Weir`


SouthEndMusic:
@Huite argued that the current node `Weir` is not aptly named:

> I think "weir" is maybe not the best term: it's explicitly a weir that maintains the level.

SouthEndMusic:
Rename to `Outlet`. This seems logical in control, because it lets water out of a controlled basin. Complements `FlowBoundary`.

PR #507, opened 2023-08-16
Rename weir to outlet

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/491.


visr (approved):
Could you search for "a outlet" and replace with "an outlet"? Otherwise looks good to me.

Issue #331, opened 2023-06-19
Anticipatory setpoint adjustment for control


gijsber:
When a dry period is coming or ongoing, water boards and Rijkswaterstaat adjust some of their operating setpoints such target water-levels in a boezemsystem or Lake IJsselmeer. These setpoints guide the PID control.

This adjustment is based on anticipated crossing of a threshold of a boundary condition (e.g. flow forecast of Lobith) or even an external condition which is not feed directly to the model (e.g. deviation of net precipitation shortage from the normal shortage).

Characteristics for the condition:
- timeshift (look ahead)
- boundary condition of external condition

evetion:
Based on PID control. We leave the preprocessing (such as deviation of net precipitation shortage) up to the user

- [ ] New condition based on time lookup with dT.

gijsber:
An initial check (befor running the simulation) is needed to evrify that the simulation length (e.g.365 days) and the external boundary provided matches the timeshift that is needed to check the condition.
If this check fails, an error needs to be raised (at this moment).

SouthEndMusic:
The look-ahead condition on boundary nodes could be implemented by adding an optional $\Delta t$ column to the `condition` table which defaults to $0$.

Then validation is needed:
- To check whether sufficient data is available to look as far ahead as you need even towards the end of the simulation as mentioned by @gijsber;
- Whether the condition of that row is indeed a (transient) boundary node for which look-ahead is possible.

PR #516, opened 2023-08-18
Conditions with lookahead

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/331 (the part on time shift). I suggest the external conditions get a separate issue.


SouthEndMusic:
There is a small uncovered edge case where you can put a condition with look ahead on a static flow boundary, and this will yield an error because for static flow boundaries a trivial interpolation object is generated with time bounds equal to the simulation time bounds, so the lookahead will look outside that. This error message is not very comprehensive for this particular situation though.

gijsber:
No problem to raise an error/exception if you can catch it and log that the user is not allowed to use a lookahaead condition on a static flow boundary.
From a hydrological perspective it is an unrealistic use case

visr (approved):

PR #523, opened 2023-08-22
don't save flow for every solver timestep

visr:
This led to huge output files with larger test models.
We now also avoid making an extra copy of the data in memory, for which we use the FlatVector wrapper type, see individual commit messages.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -105,7 +105,14 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
     end
     @debug "Setup ODEProblem."
 
-    callback, saved_flow = create_callbacks(parameters)
+    # expand scalar saveat for SavingCallback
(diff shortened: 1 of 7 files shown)


core/src/bmi.jl
```diff
@@ -105,7 +105,14 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
     end
     @debug "Setup ODEProblem."
 
-    callback, saved_flow = create_callbacks(parameters)
+    # expand scalar saveat for SavingCallback
+    saveat = if config.solver.saveat isa Number
+        range(timespan...; step = config.solver.saveat)
+    else
+        config.solver.saveat
```
SouthEndMusic, line 112: Should there be a check that this is indeed a list of floats if it is not a number?

SouthEndMusic (approved):

Issue #528, opened 2023-08-23
Test DiscreteControl of PidControl


SouthEndMusic:
Mainly setting the target of a PidControl node via DiscreteControl is a realistic use case.

PR #534, opened 2023-08-23
Discrete control of pid control

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/528.

This PR uses https://github.com/Deltares/Ribasim/pull/527 so that one should be reviewed first.

@gijsber This demonstrates what we discussed a few days ago, about having `DiscreteControl` (`IntervalControl`) set the target for the pid controller (in this case a level target). The pid controller doesn't work very nicely with jumps in the target, which can be seen from the overshoot.

Discrete_target_level_change_graph
Discrete_target_level_change

It's a nice flipped square root sign though.


core/src/create.jl
```diff
@@ -564,6 +564,18 @@ function PidControl(db::DB, config::Config)::PidControl
...
+    for (key, params) in parsed_parameters.control_mapping
+        (; proportional, integral, derivative) = params
+
+        times = params.proportional.t
+        K_p = proportional.u
+        K_i = integral.u
+        K_d = derivative.u
+        itp = LinearInterpolation(collect.(zip(K_p, K_i, K_d)), times)
+        parsed_parameters.control_mapping[key] =
+            (target = params.target, active = params.active, pid_params = itp)
```
visr, line 576:
```suggestion
        pid_params = LinearInterpolation(collect.(zip(K_p, K_i, K_d)), times)
        parsed_parameters.control_mapping[key] =
            (; params.target, params.active, pid_params)
```

FYI, the `(; a)` syntax for `(; a=a)` also works from `(; x.a)` to `(; a=x.a)`.

visr (approved):

core/src/create.jl
SouthEndMusic: Nice

visr:
Test failures look related:

```
Set PID target with DiscreteControl: Error During Test at /home/runner/work/Ribasim/Ribasim/core/test/control.jl:172
  Got exception outside of a @test
  MethodError: Cannot `convert` an object of type Vector{Float64} to an object of type DataInterpolations.LinearInterpolation{Vector{Vector{Float64}}, Vector{Float64}, true, Vector{Float64}}
```

SouthEndMusic:
> Test failures look related:

Yep, something was already called `pid_params`.

PR #538, opened 2023-08-24
fix cli test

visr:
I needed `!successful_retcode(model)` instead of `successful_retcode(model)` to get retcode 0 on success...


SouthEndMusic (approved):
LGTM

PR #539, opened 2023-08-24
julia_main: rely less on implicit conversion

visr:
Follow up to #538.


Hofer-Julian (approved):

build/ribasim_cli/src/ribasim_cli.jl
```diff
@@ -33,7 +33,11 @@ function julia_main()::Cint
         global_logger(TerminalLogger())
         model = Ribasim.run(arg)
         println(model.integrator.sol.retcode)
-        return !successful_retcode(model)
+        return if successful_retcode(model)
+            0
```
Hofer-Julian, line 37: Makes me a bit sad that there are no constants mapping 0 to SUCCESS and 1 to FAILURE

PR #544, opened 2023-08-24
Fix progressbar.

### LINKED ISSUE #542: progress bar doesn't show
Since #504. Both in the CLI and for interactive use. In both cases we relied on setting the global logger, but that is now overruled here:
https://github.com/Deltares/Ribasim/blob/3d2f7dcc866b8d24be146b35d0c7ffd5ce901aab/core/src/bmi.jl#L544-L549

In the REPL I was using
```julia
using Logging: global_logger
using TerminalLoggers: TerminalLogger
# show progress bar in terminal
global_logger(TerminalLogger())
```

To show the progress bar in VSCode. The same is done in [ribasim_cli.jl](https://github.com/Deltares/Ribasim/blob/3d2f7dcc866b8d24be146b35d0c7ffd5ce901aab/build/ribasim_cli/src/ribasim_cli.jl#L33).

evetion:
Fixes #542 
Note that in a interactive VS Code environment, you need to run

```julia
with_logger(TerminalLogger()) do 
    Ribasim.run(toml_path)
end
```

As VS Code attaches a task logger, overriding any global logger set.

### DIFF
--- a/build/ribasim_cli/src/ribasim_cli.jl
+++ b/build/ribasim_cli/src/ribasim_cli.jl
@@ -1,6 +1,6 @@
 module ribasim_cli
 
-using Logging: global_logger
+using Logging: global_logger, with_logger
 using TerminalLoggers: TerminalLogger
 using SciMLBase: successful_retcode
 using Ribasim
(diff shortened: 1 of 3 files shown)


visr (approved):
Merci beaucoup

Issue #378, opened 2023-06-30
run all test models with ribasim_cli


visr:
This would be a good test for both the correct working of the ribasim_cli, and to see if the test models don't error. It would be done on TeamCity.

https://github.com/Deltares/Ribasim/blob/a0e04c66a8678b6da35c5bd3b22be97f891b4220/python/ribasim_testmodels/ribasim_testmodels/__init__.py#L13-L18

Related: #339

visr:
Closed in favor of #339

PR #546, opened 2023-08-25
update manifests

visr:
This was missing from #536. Therefore the binaries were still built with a `DiffEqCallbacks < "2.28"`.


Hofer-Julian (approved):
Time to tackle https://github.com/Deltares/Ribasim/issues/378 then I guess

PR #547, opened 2023-08-25
add model time to cli output

visr:
Right now users of the CLI get a return code, but don't see until when the model ran, like in interactive use. This provides that information.

### DIFF
--- a/build/ribasim_cli/src/ribasim_cli.jl
+++ b/build/ribasim_cli/src/ribasim_cli.jl
@@ -33,7 +33,9 @@ function julia_main()::Cint
         model = with_logger(TerminalLogger()) do
             Ribasim.run(arg)
         end
-        println(model.integrator.sol.retcode)
+        t = Ribasim.datetime_since(model.integrator.t, model.config.starttime)
+        println("Model time:  ", t)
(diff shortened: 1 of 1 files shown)


build/ribasim_cli/src/ribasim_cli.jl
```diff
@@ -33,7 +33,9 @@ function julia_main()::Cint
         model = with_logger(TerminalLogger()) do
             Ribasim.run(arg)
         end
-        println(model.integrator.sol.retcode)
+        t = Ribasim.datetime_since(model.integrator.t, model.config.starttime)
+        println("Model time:  ", t)
+        println("Return code: ", model.integrator.sol.retcode)
```
Hofer-Julian, line 38: Those are these, right? https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#retcodes

Maybe don't print something like "The model finished successfully" if `successful_retcode`.
If not, maybe print "The model exited with return code {symbol}" and link to the docs.

visr: Yes, indeed. How about c4bedb534230c8780997f756dfcbae0d006309c4?

Hofer-Julian (approved):

Issue #548, opened 2023-08-25
Let `outlet` only flow from higher to lower level


SouthEndMusic:
Since the outlet works by gravity, it should only produce flow if the level of its inneighbor is higher than the level of its outneighbor.

Huite:
A minimum crest level seems like a good addition to the Outlet. The level then has to exceed this minimum crest level before checking the outneighbor level.

PR #549, opened 2023-08-25
Only gravity driven `Outlet` flow

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/548.

Currently the outlet switches on and off in a discrete way due to a switch in sign of the level difference. This can result in the outlet rapidly turning on and off:

flipperende_outlet

To solve this problem, a reduction factor should be applied to the outlet flow, based on the level difference. This reduction factor will have to be taken into account in the implementation of the analytical Jacobian (with an unsatisfying effort to improvement ratio). This I think will be a recurring problem until we are able to switch to an AD Jacobian.


SouthEndMusic:
I made this plot of the smooth reduction factor:

reduction_factor_plot

how can I add it to `equations.qmd`?

SouthEndMusic:
I found something peculiar when working on the test model for the `Outlet`:

outlet_model

outlet_constraints

In the current state it behaves as expected, that is: there is no outlet flow until the level boundary reaches the minimum crest level, and the basin level converges to the level boundary level. However, when the basin starts completely empty instead of with a tiny amount of water, it doesn't fill:

outlet_constraints_noflow

I haven't been able to figure out why this happens.

visr (approved):

PR #550, opened 2023-08-25
ForwardDiff.jl automatic Jacobian

SouthEndMusic:
*Edit: this initial message is outdated, see the discussion below.*

It turned out to be quite easy to get `ForwardDiff.jl` to work for the Jacobian of `water_balance!`, it was mainly a matter of replacing some instances of `Float64` with `Number` in function signatures or removing `Float64` altogether (I did this only where necessary).

See the example below for POC.

```julia
using ForwardDiff
using SQLite
using Ribasim


toml_path = "C:/Users/konin_bt/Ribasim_development/Ribasim/data/basic_transient/basic_transient.toml"

cfg = Config(toml_path)
gpkg_path = input_path(cfg, cfg.geopackage)
db = SQLite.DB(gpkg_path)

t = 0.0

state = load_structvector(db, cfg, BasinStateV1)
storage, _ = get_storages_from_levels(p.basin, state.level)
integral = zeros(length(p.pid_control.node_id))
u0 = ComponentVector{Float64}(; storage, integral)

function water_balance(u::ComponentVector)

    du = zero(u)
    water_balance!(du,u,p,t)
    return du
end

jac_AD = ForwardDiff.jacobian(water_balance, u0)
jac_AN = zero(jac_AD)
water_balance_jac!(jac_AN,u0,p,t)

println("AD Jacobian:")
display(jac_AD)
println("\nAN Jacobian:")
display(transpose(jac_AN))
```

output:
```
AD Jacobian:
4×4 Matrix{Float64}:
 -3.06216e-12   0.0          0.0           0.0
  0.0          -2.25533e-7   0.0           0.0
  0.0           2.48017e-8  -3.06216e-12   0.0
  0.0           4.96033e-8   0.0          -8.26725e-7

AN Jacobian:
4×4 transpose(::Matrix{Float64}) with eltype Float64:
 0.0   0.0         0.0   0.0
 0.0  -2.2553e-7   0.0   0.0
 0.0   2.48017e-8  0.0   0.0
 0.0   4.96033e-8  0.0  -8.26722e-7
```

~~There is a remaining problem where the AD Jacobian contained NaNs unless I turned on NaN-safe mode as suggested [here](https://juliadiff.org/ForwardDiff.jl/latest/user/advanced.html#Fixing-NaN/Inf-Issues-1).~~

Regarding the current erroring test; why does `Vector{Float64}` not satisfy `Vector{Number}`?


visr:
Great stuff! Regarding your last question:

```
julia> Float64[] isa Vector{Number}
false

julia> Float64[] isa Vector{<:Number}
true
```

Search also for "covariant" here for more expanation: https://docs.julialang.org/en/v1/manual/types/#man-parametric-composite-types

The changes of the struct to
```
struct A
    v::Vector{Number}
end
```
will be detrimental to performance, since the compiler cannot make any guarantees about what is inside. You want to make that a parametric type (see same link), where the element type can be any Number.

core/Project.toml
```diff
@@ -15,6 +15,7 @@ Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Dictionaries = "85a47980-9c8c-11e8-2b9f-f7ca1fa99fb4"
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
```
visr, line 18: Good to add a compat entry, `pkg> compat ForwardDiff 0.10`

SouthEndMusic:
Most tests pass now, only the comparsions with sparse/dense and AD / Finite diff Jacobains fail. I'm quite sure this is because the Jacobian prototype (i.e. the Jacobian sparsity) is not correct (at least this explains the failing dense vs sparse AD Jacobian). 

Since it seems that we'll keep using the Jacobian prototype, this has to work properly. Part of what is missing for certain PID control cases is also mentioned in https://github.com/Deltares/Ribasim/issues/482. 

The following warning now pops up often when using `DiffCache`:

DiffCache_warning

core/src/solve.jl
```diff
@@ -79,11 +82,8 @@ struct Basin{C} <: AbstractParameterNode
     drainage::Vector{Float64}
     infiltration::Vector{Float64}
     # cache this to avoid recomputation
-    current_level::Vector{Float64}
-    current_area::Vector{Float64}
-    # The derivative of the area with respect to the level
-    # used for the analytical Jacobian
-    current_darea::Vector{Float64}
+    current_level::Union{Vector{Float64}, DiffCache{Vector{Float64}}}
+    current_area::Union{Vector{Float64}, DiffCache{Vector{Float64}}}
```
visr, line 86: For these and other Unions, a type parameter wasn't working?

core/src/solve.jl
```diff
@@ -479,15 +487,17 @@ end
 
 function set_current_basin_properties!(
     basin::Basin,
-    storage::AbstractVector{Float64},
-    t::Real,
+    current_area,
+    current_level,
```
visr, line 491: What is the reason for not taking the `current_area` and `current_level` out of the already passed `basin` argument? Same for `formulate!` directly below.

core/src/utils.jl
```diff
@@ -110,12 +110,14 @@ end
...
 )::Tuple{Vector{Float64}, Bool}
-    storages = Float64[]
+    storage = get_storage_from_level(basin, 1, levels[1])
+    storages = [storage]
 
-    for (i, level) in enumerate(levels)
-        push!(storages, get_storage_from_level(basin, i, level))
+    for (i, level) in enumerate(levels[2:end])
+        storage = get_storage_from_level(basin, i + 1, level)
+        push!(storages, storage)
```
visr, line 120: Is there a reason for this change? It looks less readable to me.

core/src/bmi.jl
```diff
@@ -80,6 +80,7 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
     else
         storages, errors =
             get_storages_from_levels(parameters.basin, state.level)
+        storages = real.(storages)
```
visr, line 83: Is this to avoid dual numbers in your initial state?

core/src/bmi.jl
```diff
@@ -432,13 +432,15 @@ function set_control_params!(p::Parameters, node_id::Int, control_state::String)
...
+            vec = preallocation_dispatch(getfield(node, field), 0)
+            vec[idx] = value
         end
     end
 end
 
 "Copy the current flow to the SavedValues"
-save_flow(u, t, integrator) = copy(nonzeros(integrator.p.connectivity.flow))
+save_flow(u, t, integrator) =
+    copy(nonzeros(preallocation_dispatch(integrator.p.connectivity.flow, u)))
```
visr, line 443: If it doesn't fit on a single line I generally prefer to switch to the long function syntax:

```
function f()
    ...
end
```

visr (commented):
This is great. There are still some possible performance issues with e.g. using `preallocation_dispatch` and the `Union` fields. So it'd be good to see how the performance compares to finite diff for a realistic model.

core/src/utils.jl
```diff
@@ -899,3 +897,7 @@ function Base.getindex(fv::FlatVector, i::Int)
     v = fv.v[d + 1]
     return v[r + 1]
 end
+
+function preallocation_dispatch(var, value)
+    return isa(var, DiffCache) ? get_tmp(var, value) : var
+end
```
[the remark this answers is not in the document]
SouthEndMusic: That dispatch is indeed nicer. I'm having a hard time coming up with a docstring for this because I do not understand exactly what `get_tmp` does.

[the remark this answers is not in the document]
visr: It looks like instead of `preallocation_dispatch` you can just use `get_tmp` directly. It already has a fallback if the first argument is not a DiffCache: https://github.com/SciML/PreallocationTools.jl/blob/0c41450830881f87daedea946512eb06f956d5f2/src/PreallocationTools.jl#L143

core/src/create.jl
```diff
@@ -202,12 +202,18 @@ function Connectivity(db::DB)::Connectivity
...
+        flow = get_tmp(flowd, flow)
+    end
+
     nonzeros(flow) .= 0.0
 
     return Connectivity(
         graph_flow,
         graph_control,
-        flow,
+        config.solver.autodiff ? flowd : flow,
```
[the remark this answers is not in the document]
SouthEndMusic: `flow` is used in line 211; that operation cannot be performed on `flowd`. Line 208 might be unnecessary, I'm not sure.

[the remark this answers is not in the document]
visr: Right, I missed that. Haven't tried, but based on my understanding of `get_tmp` line 208 is indeed not needed, since the input and output `flow` should be the same.

core/src/solve.jl
SouthEndMusic: This is again about me not knowing much about `get_tmp`, and whether it is an expensive call.

core/src/solve.jl
SouthEndMusic: Yes It is, I changed it to parameterized now. I'm not sure it solves the performance issue though.

core/src/bmi.jl
SouthEndMusic: Yes, but it works without this

core/src/utils.jl
SouthEndMusic: Turns out I don't need this change (:

core/src/create.jl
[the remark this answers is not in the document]
SouthEndMusic: It works fine without line 208

docs/core/usage.qmd
```diff
@@ -54,10 +54,8 @@ For instance, `saveat = 86400.0` will save output after every day that passed.
 The Jacobian matrix provides information about the local sensitivity of the model with respect to changes in the states.
 For implicit solvers it must be calculated often, which can be expensive to do.
 There are several methods to do this.
-By default Ribasim uses an analytical Jacobian function written by the Ribasim developers.
-This is enabled by default, with the solver setting `jac = true`.
-If this is not used, the Jacobian is calculated with a finite difference method, which can be less accurate and more expensive.
-In the future we want to additionally support `autodiff = true` to calculate the Jacobian with automatic differentiation.
+By default Ribasim uses a Jocobian derived automatically using [ForwardDiff.jl](https://juliadiff.org/ForwardDiff.jl/stable/) with memory management provided by [PreallocationTools.jl](https://docs.sciml.ai/PreallocationTools/stable/).
```
Huite, line 57: Jocobian -> Jacobian

SouthEndMusic:
I can no longer reproduce the non-deterministic Jacobian problem, maybe it had something to do with my `get_tmp` wrapper 🤷‍♂️

core/src/Ribasim.jl
```diff
@@ -13,12 +13,15 @@ using DataInterpolations: LinearInterpolation, derivative
...
 using Dictionaries: Indices, Dictionary, gettoken, gettokenvalue, dictionary
+using ForwardDiff: pickchunksize
 using DiffEqCallbacks
 using Graphs: DiGraph, add_edge!, adjacency_matrix, inneighbors, outneighbors
 using Legolas: Legolas, @schema, @version, validate, SchemaVersion, declared
 using Logging: current_logger, min_enabled_level, with_logger
 using LoggingExtras: EarlyFilteredLogger, LevelOverrideLogger
 using OrdinaryDiffEq
+using PreallocationTools: DiffCache, get_tmp
+using Preferences: set_preferences!
```
visr, line 24: You no longer use this right? Then it can be removed from here and the project.

Also, since the dependencies changed, could you `pkg> resolve` the `libribasim` and `ribasim_cli` manifests to reflect this change?

visr (approved):

core/src/Ribasim.jl
SouthEndMusic: Did I do this correctly?

PR #555, opened 2023-08-28
enable compression of output Arrow files

visr:
I noticed changing the compression settings lead to identical Arrow files. It turns out nothing was being compressed.

Previously `get_compressor` always returned `nothing`, which was accepted by Arrow to mean: don't use compression.

This also depends on the Codec packages directly rather than relying on Arrow depending on them. That should catch breaking changes as well.


SouthEndMusic (approved):
Looks good to me

PR #560, opened 2023-08-29
Add ribasim_cli tests

### LINKED ISSUE #339: Add integration tests for `ribasim_cli`
After building ribasim.exe, we don't really test if it works as expected.
We should run a couple of integration tests on TeamCity.

Hofer-Julian:


Hofer-Julian:
This is ready for review from my side.
I couldn't get the tests to run on TeamCity for reasons that are completely non-obvious to me.
The tests fail with some Julia error, my theory was that it somehow installs an old version of Ribasim Python, but I wouldn't know how. Locally, the tests work fine. I've now disabled the TeamCity tests to be revisited later.

visr (approved):
Looks good to me. The `__dict__` trick saves some repetition.

PR #562, opened 2023-08-29
add reduction_factor to TabulatedRatingCurve

visr:
If the rating curve prescribes large Q even when the upstream Basin is nearly empty, this leads to an incorrect water balance.


visr:
replaced by #587

PR #580, opened 2023-09-07
Refactor Python code

### LINKED ISSUE #512: Refactor Python Model
There's some technical debt in there, as it mixes all nodetypes and the required config options.

evetion:
Fixes #512 

- [x] Add JSON Schema of Config and generate Python code
- [x] Autogenerate Pandera Schemas
- [x] Update docs to include the above
- [x] ~Refactor Python model into split Config & Root submodels~


evetion:
I started with splitting the Model into separate tables and config, but didn't get far without hitting the limits of the current design.

The problem is, there's no clear design, it's mix of different abstractions. Pandera DataFrames are high-level abstract things, but represent low-level tables (which can come from arrow or geopackage). There's a high-level abstraction of a nodetype in between, but it doesn't represent something low-level. The current Model is a mix of a representation of these nodetypes with a (partial) low-level config. Will make a separate issue about this, which needs a refinement.

core/src/config.jl
```diff
@@ -65,9 +65,13 @@ for (T, kinds) in pairs(nodekinds)
 end
 const nodetypes = collect(keys(nodekinds))
 
+# A [] in a TOML is parsed as Vector{Union{}}
+Base.convert(::Type{Union{Float64, Vector{Float64}}}, x::Vector{Union{}}) =
+    convert(Vector{Float64}, x)
+
 @option struct Solver <: TableOption
     algorithm::String = "QNDF"
-    saveat::Union{Float64, Vector{Float64}, Vector{Union{}}} = Float64[]
```
Hofer-Julian, line 70: What does `Vector{Union{}}` even mean? 😅

core/src/config.jl
```diff
@@ -109,15 +113,15 @@ end
...
+@option @addnodetypes struct Config <: TableOption
     starttime::DateTime
     endtime::DateTime
 
     # [s] Δt for periodic update frequency, including user horizons
     update_timestep::Float64 = 60 * 60 * 24.0
 
     # optional, when Config is created from a TOML file, this is its directory
-    relative_dir::String = pwd()
+    relative_dir::String = "."  # ignored(!)
```
Hofer-Julian, line 124: What is the "ignored" comment good for?
You mean because we always override in the current setup?

docs/gen_schema.jl
```diff
@@ -60,7 +80,7 @@ end
 
 os_line_separator() = Sys.iswindows() ? "\r\n" : "\n"
 
-function gen_schema(T::DataType, prefix = prefix)
+function gen_schema(T::DataType, prefix = prefix; pandera = true)
```
Hofer-Julian, line 83: Why is `pandera` optional?

Hofer-Julian (commented):
A few, first comments.
I also see that mypy is still unhappy with you.

Will documentation still be part of this PR?

core/src/config.jl
evetion: Yep, its value is always unused.

docs/gen_schema.jl
evetion: Because I hacked the remarks field into each type specifically for pandera. But the generation of the Config doesn't need it.

Frankly, this whole file is a hack, but it works. Doing it right requires a new package that actually implements the whole of JSONSchema.

evetion:
> I also see that mypy is still unhappy with you.

Yeah, not sure if I can fix that nicely, I dynamically add things to a module namespace. Will try though.

> Will documentation still be part of this PR?

Yep, at least about the generation of the config.py files.

core/src/config.jl
```diff
@@ -65,9 +65,13 @@ for (T, kinds) in pairs(nodekinds)
 end
 const nodetypes = collect(keys(nodekinds))
 
+# A [] in a TOML is parsed as Vector{Union{}}
+Base.convert(::Type{Union{Float64, Vector{Float64}}}, x::Vector{Union{}}) =
+    convert(Vector{Float64}, x)
+
```
visr, line 71: I saw your comment in #gripes. I wonder if perhaps that's just a bug that needs to be filed?

In any case, this is type piracy. Can't you do https://github.com/Deltares/Ribasim/pull/504#discussion_r1296835149 again?

python/ribasim/tests/test_model.py
```diff
@@ -14,8 +14,8 @@ def test_repr(basic):
 
 def test_solver():
     solver = Solver()
-    assert solver.algorithm is None
-    assert solver.saveat is None
+    assert solver.algorithm == "QNDF"  # default
```
visr, line 17: Nice that these defaults are now in the schema, and therefore visible in Python.

python/ribasim/ribasim/config.py
```diff
@@ -0,0 +1,117 @@
...
+    condition: Optional[str] = Field(None, description="condition")
+
+
+class Solver(BaseModel):
+    reltol: Optional[float] = Field(0.001, description="reltol")
+    saveat: Optional[Union[List[float], float]] = Field([], description="saveat")
+    maxiters: Optional[int] = Field(1000000000, description="maxiters")
+    autodiff: Optional[bool] = Field(True, description="autodiff")
+    adaptive: Optional[bool] = Field(True, description="adaptive")
+    algorithm: Optional[str] = Field("QNDF", description="algorithm")
```
visr, line 42: Is `Optional` still needed? Before when no value was given it was None, but now that it has defaults, isn't it just `str`?

python/ribasim/ribasim/config.py
```diff
@@ -0,0 +1,117 @@
...
+    flow_boundary: Optional[FlowBoundary] = None
+    pid_control: Optional[PidControl] = None
+    fractional_flow: Optional[FractionalFlow] = None
+    relative_dir: Optional[str] = Field(".", description="relative_dir")
+    endtime: datetime = Field(..., description="endtime")
+    manning_resistance: Optional[ManningResistance] = None
+    tabulated_rating_curve: Optional[TabulatedRatingCurve] = None
+    logging: Optional[Logging] = None
+    outlet: Optional[Outlet] = None
+    geopackage: str = Field(..., description="geopackage")
```
visr, line 114: What is `...`?

And what is the use for the description, if it is the same as the field name? And why is it sometimes there and sometimes not?

python/ribasim/ribasim/node_types/level_boundary.py
```diff
@@ -1,29 +1,16 @@
 from typing import Optional
 
-import pandera as pa
-from pandera.engines.pandas_engine import PydanticModel
 from pandera.typing import DataFrame
 
-from ribasim import models
 from ribasim.input_base import TableModel
+from ribasim.schemas import (  # type: ignore
```
visr, line 6: Is this the right place for the comment to apply to both imported schemas? See also other files in this directory if not.

evetion: It is, otherwise mypy would still complain.

core/src/config.jl
evetion: I'm ok with this piracy, as the convert to Vector{Float64} is already defined in the same way, so this is just a custom union by us that we want to behave in the same way. Can't really see this having a negative impact.

python/ribasim/ribasim/config.py
evetion: This file is automatically generated, we allow nothing in Julia, thus this is Optional here.

python/ribasim/ribasim/config.py
evetion: It's how Pydantic sets a Field to be required. I can probably remove the description by changing a command line flag, but it's not used anyway.

core/src/config.jl
visr: But `Configurations.from_dict` is meant for this right? It doesn't seem more complex.
I see Aqua tries to detect simple piracy these days, wonder why this isn't picked up.

python/ribasim/ribasim/config.py
visr: We don't directly allow nothing right? Or is that because the solver section itself is optional?

https://github.com/Deltares/Ribasim/blob/f503acb8674a6c59153298f655d036d82e1d2107/core/src/config.jl#L73

But so you agree that not having Optional here does not serve any purpose?

python/ribasim/ribasim/config.py
visr: Ah ok. If it's not used and you can find a flag to remove it, I'd prefer that.

visr:
Is it possible to remove the timestamp from the generated python files?

It creates merge conflicts:

```diff
<<<<<<< HEAD
#   timestamp: 2023-09-05T14:00:34+00:00
||||||| 1ad1340
#   timestamp: 2023-08-21T14:05:19+00:00
=======
#   timestamp: 2023-09-04T11:45:02+00:00
>>>>>>> main
```

And with #594 I expect we will run it more often as part of tasks, even if it is not strictly needed. In that case all it would do is update the timestamp.

visr (approved):

PR #586, opened 2023-09-11
remove outdated plots from README

visr:
Mainly want to see if CI is ok, seeing some GeoPackage time related issues locally.


visr:
It failing locally but not on CI. I just added a test in bf6ba43b076688fa22ccc8a7391462377a5f3b41, that locally gives:

```
AssertionError: assert Timestamp('2020-01-01 00:00:00') == Timestamp('2019-12-31 23:00:00')
```

Meaning that when we write the basic_transient model to disk, then read it back in again, the forcing shifted to one hour earlier. Note that the tests only improves the error message, the following test always covered this. So why is this suddenly an issue? Is it a daylight savings thing?

evetion (approved):
I will miss those pictures, it has a nice concrete feel to it. Maybe we need a new (QGIS based?) picture that shows what you can do with Ribasim.

visr:
Ha ok you are right, let's wait with removing until we have something else, see #590.
Now this is only adding a test.

PR #596, opened 2023-09-13
migrate to python tools extensions

visr:
When I opened my settings.json it warned me these settings will be deprecated soon, with this link for more info:

https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions


Hofer-Julian (approved):
Yeah, the went from pre-release to “the current way is deprecated” in one go 😅

PR #597, opened 2023-09-14
Sort DataFrames inplace.

evetion:
Fixes the performance of saving a Ribasim model in Python.

When we sorted, we assigned the returned dataframe to our model, triggering validation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     rev: v0.0.286
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
+        args: [--fix, --exit-non-zero-on-fix, --ignore=PD002]
   - repo: https://github.com/psf/black
     rev: 23.7.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     rev: v0.0.286
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix, --ignore=PD002]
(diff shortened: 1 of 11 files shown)


visr (approved):

.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
     rev: v0.0.286
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
+        args: [--fix, --exit-non-zero-on-fix, --ignore=PD002]
```
Hofer-Julian, line 20: Please modify ruff.toml instead

Hofer-Julian (commented):
Ruff claims:
> Further, in many cases, inplace=True does not provide a performance benefit, as pandas will often copy DataFrames in the background.

Would be good to benchmark if that actually improve things.

.pre-commit-config.yaml
visr: Fixed in 53310f27f7e1942da0ad9c7efe9d74b37ec2f680

visr:
Can confirm that this no longer triggers validation on write.

visr:
Turns out the Ubuntu CI failure was due to a missing custom sort method. Brought them all up to date with our docs in dc0c288bf04750e64837d7afc40039f7d42f85d5. @evetion already mentioned that the sort algorithm on Linux could be slightly different, and we just luckily ended up with a sorted GeoPackage on Windows.

On top of that, we sort data when reading it into the core as well, so not sorting in Python should not lead to test failures. But those were similarly outdated, fixed in fe642699f099027c919f0ce70a284e664b824407.

PR #603, opened 2023-09-18
Stop running `pixi install` in CI

Hofer-Julian:
See https://github.com/prefix-dev/pixi/pull/340#issuecomment-1722436539


visr (approved):

visr:
Forgot, but we should also remove this from the dev docs or not?

Hofer-Julian:
> Forgot, but we should also remove this from the dev docs or not?

No, it should be still needed when running locally.

PR #604, opened 2023-09-18
Add more pixi tasks

Hofer-Julian:
See #598


docs/schema/DiscreteControlCondition.schema.json
```diff
@@ -27,10 +27,10 @@
             "format": "default",
             "anyOf": [
                 {
-                    "type": "number"
+                    "type": "null"
```
Hofer-Julian, line 30: Are these schema changes expected?

visr: This is strange, it seems my laptop likes a different order than @evetion and yours: https://github.com/Deltares/Ribasim/pull/574/commits/54273a0c2176ed461c2bb8515d561262e5c7bf65

Hofer-Julian: Mmmh, this is annoying.
Let's hope that it was just some version mismatch and moving to pixi solves that.

utils/generate-testmodels.py
```diff
@@ -1,9 +1,12 @@
+import shutil
 from pathlib import Path
 
 import ribasim_testmodels
 
 if __name__ == "__main__":
     datadir = Path("data")
+    if datadir.is_dir():
+        shutil.rmtree(datadir)
```
visr, line 9: Oh this would delete a lot of useful stuff for me. I could move my stuff to another folder and gitignore that. Good to announce this change though.

visr (approved):

utils/generate-testmodels.py
visr: I agree it's nice to have a clean folder. Perhaps having a more dedicated name for it, like testmodels, perhaps with a README explaining it's use and that it is automatically emptied, would be better.

Hofer-Julian: Sounds good to me.

Hofer-Julian: Done in https://github.com/Deltares/Ribasim/pull/604/commits/91590e09c08037f9c360579682c63ef8974fc5dd

PR #607, opened 2023-09-19
add more julia docstrings

visr:
I started this off simply wanting to switch to Documenter.jl version 1.0 that recently got released. But what wasn't clear from #470 is that the [DocumenterMarkdown](https://github.com/JuliaDocs/DocumenterMarkdown.jl) backend we use is not actively maintained, meaning we are stuck for now on the old Documenter 0.27.

In any case this addresses a few Documenter 0.27 warnings about missing docstrings. And when I started adding docstrings I figured the key functions really need some docstrings. I also added some more convenient methods, e.g. extending `solve!` from SciML on our Model, and adding `model = Ribasim.Model(config)` instead of only the more verbose BMI interface method `model = BMI.initialize(Ribasim.Model, config)`.


Hofer-Julian:
> But what wasn't clear from https://github.com/Deltares/Ribasim/pull/470 is that the [DocumenterMarkdown](https://github.com/JuliaDocs/DocumenterMarkdown.jl) backend we use is not actively maintained, meaning we are stuck for now on the old Documenter 0.27.

Can you open an issue for this?

Hofer-Julian (approved):
Nice addition to our current docstrings.

The convenience functions also make sense to me

PR #609, opened 2023-09-20
Pid control and sparsity

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/482.

The above issue is quite outdated, but I picked it up because something was missing for the sparsity when using PID control. However, I have not been able to (re)produce the problem, so the current sparsity seems fine.


SouthEndMusic:
As the problem this PR tries to solve cannot be (re)produced, I close this PR.

Issue #462, opened 2023-07-27
Regression testing


SouthEndMusic:
There are a few use cases for testing which are quite expensive:
- Run models with various solvers and compare their results;
- Run models with a sparse Jacobian (https://github.com/Deltares/Ribasim/pull/459) and a dense one and compare the results.

These could possibly also involve large models like the one mentioned in https://github.com/Deltares/Ribasim/issues/419#issue-1799111085.

@evetion suggested regression testing for this.

evetion:
So we want:
- [ ] Storage of output values
- [ ] Comparison of output values (~)
- [ ] Running this on TeamCity every week or so
- [ ] Running specific Julia tests only on TC

Open for discussion, is this a Python tool, or do we do this in pure Julia?

evetion:
Check in with the coupler, as they already do some reference checks.

Jingru923:
Regression test in Ribasim should have following content which I think it's most important for now. I will build benchmark for these content.

I selected 4 representative small test models: basic, pid_control, trivial, subnetwork_with_sources. With 7 implicit different solvers, Jacobian densities, autodiff on-off, the following output value will be check:

- The start and the  end of basin storages
- Basin storage convergences
- The start and the end of basin levels
- Basin level convergences
- `balance_error` of basins
- `relative_error` of basins
- any idea for subgrid_leve?

For the larger models, their performance is also important. I see overlapping with issue #1471. So they will be worked on in coming sprint.

Jingru923:
When testing different solver for `trivial` model (a simple model) all solver can give very approximate result (results of different solver can ≈ each other).

However, this is not the case for the more complex model `basic` and `pid_control`. I believe same will happen for the `subnetwork_with_sources` model. Due to the time limit, I will search for a good margin for all the different solvers to pass the regression tests. For long-term goal, we should revise if all the solvers are actually suitable and necessary for Ribasim, and which one is better to keep

PR #611, opened 2023-09-22
add runstats utility

visr:
This addresses a part of #462. It is a way to run and log performance related characteristics into a table. That table can then be used to do further analysis on what the effects of e.g. code changes or solver settings are.

This does not yet hook it up to anything, so there is the possibility for bitrot. But I've had this locally for a while, and I wanted to share it. Even if this doesn't run in TeamCity every week it can be useful for local analysis.

There are many columns, and one row per run, so a small part of the table looks like:

image

Columns are about solver settings, host machine, julia version, git commit, timing and GC.

### DIFF
--- a/utils/runstats.jl
+++ b/utils/runstats.jl
@@ -1,0 +1,136 @@
+# This runs all testmodels and logs many performance related variables into a table.
+# It can be used to track performance over time, or determine the most efficient solver
+# settings for a particular model.
+# Right now it runs all the generated_testmodels, however running larger models that are
+# more typical of real usage is probably more useful.
+
+# TODO look into
+# https://github.com/SciML/SciMLBenchmarks.jl
+# https://github.com/JuliaCI/PkgBenchmark.jl
+
+import Arrow
+import Tables
+using Ribasim
+using Configurations: to_dict
+using DataStructures: OrderedDict
+using Dates
+using LibGit2
+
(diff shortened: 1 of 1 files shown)


utils/runstats.jl
```diff
@@ -0,0 +1,136 @@
...
+    Ribasim.run(config)
+    timed = @timed Ribasim.run(config)
+    model = timed.value
+    dict = run_dict(toml_path, config, timed)
+    push!(runs, dict)
+end
+
+tbl = Tables.columntable(runs)
+
+# Arrow.append("runs.arrow", tbl)
```
Hofer-Julian, line 135: Maybe remove this comment?

Hofer-Julian (commented):
Looks good. Only found one stray commented piece of code.

Regarding TeamCity, we might get away with modifying the ribasim_cli tests a bit.
TeamCity already records the duration for each test:

grafik

utils/runstats.jl
visr: The way I currently expect it to be used it by making some edits to the script as needed by your application. In some cases you want to re-run it and append runs to the table after e.g. checking out a different commit. In that case you'd want this line.

visr:
> TeamCity already records the duration for each test

Good to know. By running from Julia we do have access to many more outputs, which we store here. That was the main reason for writing this in Julia.

PR #622, opened 2023-09-25
rename Basin / forcing to Basin / time

visr:
All other node types use the table name "time" to represent time-varying data, as opposed to the static table. I've been wanting to rename this for a while, best not to postpone it more, because of the breaking change.


SouthEndMusic (approved):
LGTM

Issue #626, opened 2023-09-26
Call `get_tmp` when needed


SouthEndMusic:
When working on the performance of the incorporation of `ForwardDiff.jl`, calls to `get_tmp` were minimized to increase performance. However, several improvements have been made so that `get_tmp` is now basically free to call. 

Now variables obtained with `get_tmp` are retrieved at the beginning of `water_balance!` and passed to functions. We should change this to obtaining these variables with `get_tmp` where they are needed, which will make the function calls in `water_balance!` cleaner.

PR #628, opened 2023-09-27
get_tmp and formulate_flow! refactor

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/626.

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -1075,19 +1075,6 @@ function formulate_flow!(
     return nothing
 end
 
-function formulate_flow!(
-    node::AbstractParameterNode,
-    p::Parameters,
-    storage::AbstractVector,
-    t::Float64,
-)::Nothing
-    node_type = nameof(typeof(node))
-    if !isa(node, Union{Basin, LevelBoundary, DiscreteControl, PidControl, Terminal})
-        error("'formulate_flow!' not defined for nodes of type $node_type.")
-    end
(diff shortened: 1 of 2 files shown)


core/src/solve.jl
```diff
@@ -1080,6 +1075,19 @@ function formulate_flow!(
...
 end
 
+function formulate_flow!(
+    node::AbstractParameterNode,
+    p::Parameters,
+    storage::AbstractVector,
+    t::Float64,
+)::Nothing
+    node_type = nameof(typeof(node))
+    if !isa(node, Union{Basin, LevelBoundary, DiscreteControl, PidControl, Terminal})
```
visr, line 1085: What is the advantage of this approach over explicitly listing the `formulate_flow!` methods we want to call?

If I understand correctly, this PR switches to looping over all nodetypes, with the exception that FractionalFlow need to be last and this method that does nothing for certain allowed node types.

The only benefit I see is that you get an error when you add a node type but forgot to call `formulate_flow!`. Though that wouldn't really happen unless we add untested node types. The old method is more explicit and easier to read in my opinion, and more flexible as well in case we want to change the arguments for one of the methods.

visr (commented):
Great, this cleans up nicely. I left one comment on how this changes how `formulate_flow!` gets called, that part I'm not sure about.

visr (approved):

PR #568, opened 2023-08-31
Numerical considerations

SouthEndMusic:
I started writing a docs page about numerical methods and stability/convergence considerations, for my own understanding and to formalize the discussion a bit.

PR #639, opened 2023-10-02
start using isoutofdomain

visr:
This rejects any steps that lead to a negative storage. See also:

https://docs.sciml.ai/DiffEqDocs/stable/basics/faq/#My-ODE-goes-negative-but-should-stay-positive,-what-tools-can-help?

When using adaptive solvers this will cause the dt to go down. If it is not adaptive and implicit, it will fail:

```
Warning: Newton steps could not converge and algorithm is not adaptive. Use a lower dt.
```

Both options seem desired behavior.

I managed to still get negative storages with an explicit solver with a large fixed dt, and a flux that empties a basin in well under a timestep. But adaptive is the default and should generally be used.

The `du` correction in `water_balance!` is also removed as it should normally not be needed anymore. And if it does, I'd prefer to not distort the water balance.

One alternative is to use the [`PositiveDomain`](https://docs.sciml.ai/DiffEqCallbacks/stable/step_control/#DiffEqCallbacks.PositiveDomain) callback, but this only works for `adaptive=false`, and is not easily adapted to enforce only the storage part of the state to be positive.

Also adds an unrelated test to avoid saving a t more than once when not needed.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -126,6 +126,7 @@ function BMI.initialize(T::Type{Model}, config::Config)::Model
         progress_steps = 100,
         callback,
         tstops,
+        isoutofdomain = (u, p, t) -> any(<(0), u.storage),
         config.solver.saveat,
         config.solver.adaptive,
         config.solver.dt,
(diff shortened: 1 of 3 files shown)


SouthEndMusic (approved):
Looks good to me. I feel like you should add some of your findings to https://github.com/Deltares/Ribasim/pull/568.

PR #641, opened 2023-10-02
add more solver options for timestepping

visr:
https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/

This provides options for bounding the allowed dt for adaptive timestepping, even forcing it to continue despite dtmin being too large, which obviously should be used with caution.


SouthEndMusic (approved):
Looks good, maybe change `Maybe{T}` to explicitly `Union{Nothing, T}`.

PR #643, opened 2023-10-03
update pixi lock file

visr:
I removed the old lock file, followed by `pixi run install` to install everything again.

This gets us new versions of our dependencies. Good to do every once in a while, preferable early in the release cycle in case issues with new versions of dependencies arise.


Hofer-Julian:
The CI doesn’t seem to be happy

Hofer-Julian:
I've pinned pandas and opened https://github.com/Deltares/Ribasim/issues/645

We should tackle this issue after @evetion's Ribasim Python refactor.

Hofer-Julian (approved):

Issue #199, opened 2023-05-09
Classify nodes as (non)-conserving


SouthEndMusic:
As mentioned by @visr in https://github.com/Deltares/Ribasim/issues/195#issuecomment-1539641160

visr:
Thanks for creating the issue. From the documentation it should be clear if nodes conserve water or not. We don't need it yet, but I was thinking such a classification could be useful if it is in the API, to make sure we can for instance ignore conservative nodes when creating water balance overviews and such. Though best to wait with implementing this until we have a clear use case.

PR #644, opened 2023-10-03
add model inputs and outputs to flow output

visr:
Fixes https://github.com/Deltares/Ribasim/issues/199.

The flow sparse matrix is currently only used for flows on edges between nodes. To be able to make a water balance, this adds the boundary condition flows onto the diagonal, with + for adding water to the model, and - for removing water.

For a graph, an entry onto the diagonal is a self-loop, and edge from a vertex onto itself. We don't actually have an edge_id for this, so in the [flow output](https://deltares.github.io/Ribasim/core/usage.html#edge) this is `missing`, and `from_node_id` and `to_node_id` are the same.

Still have to
- [x] clean up the PR
- [x] add tests
- [x] ask @SouthEndMusic if he accepts me removing explicit `flow = 0` statement on inactive nodes (`flow` and `du` are set to 0 at the start of `water_balance!`, so it's not needed)
- [x] check if QGIS likes this
- [x] update addnode.qmd
- [x] make an issue on the horrible way we currently check flows in the julia tests, (done using the new `Ribasim.flow_table(model)` function from this PR)
- [x] make an issue for https://github.com/Deltares/Ribasim/pull/644#issuecomment-1745584689

But welcome to hear early feedback.

We could add these self-loops to the actual `flow_graph` as well, but I feel like it would be annoying having to always filter these out of `inneighbors` and `outneighbors`, so this adds the self-loops only to the flow sparse matrix. They are only used for output.

The QGIS time series visualization just ignores this currently, which is fine:
image


visr:
One limitation of this approach is that we cannot separate the 4 basin fluxes (precipitation, evaporation, drainage, infiltration). We could make separate accommodations for those. Or perhaps these should really be separate node types with 1 flux each;

- Precipitation
- Evaporation (specify potential_evaporation and connect to Basin to get open water evaporation)
- Infiltration (combine with drainage? one node per modflow cell?)

This would also fix the timeseries duplication problem. Say you have 1000 Basins in a small area with a single precipitation timeseries. Now you have to copy this data 1000x, once per Basin. If you have a separate Precipitation node, you only have to define it once and connect it to all 1000 Basins, just like you can do already for e.g. a LevelBoundary. For visualization you'd probably want an easy way to hide these nodes and edges though.

I plan to start keeping them lumped together for now.

SouthEndMusic:
> Now you have to copy this data 1000x, once per Basin.

I thought the strategy for this was eventually going to be to have a central set of timeseries and nodes using a time series pointing to that. That is at least what came out of discussions regarding control, having conditions on variables that are not part of the model.

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/199.

SouthEndMusic (approved):
Looks good to me. This fixes I think the first issue I ever wrote ^^

visr:
> I thought the strategy for this was eventually going to be to have a central set of timeseries and nodes using a time series pointing to that.

Yes, that's a good point. I don't think that  allows us to store on the diagonal, but it also helps with avoiding copies of timeseries. Is there already an issue for that? We can link that to #661.

Issue #632, opened 2023-09-28
Allocation based on solving max flow problems


SouthEndMusic:
The PoC code in https://github.com/Deltares/Ribasim/pull/627 attempts to propagate demands from users to the source. The approach there is flawed, because it does not account for multiple demand sharing paths from users to the source. Therefore here is a new setup for this demand propagation.

For now we assume a single source from which each user gets its demand.

### Formulate a max flow problem per priority

A [max flow problem](https://en.wikipedia.org/wiki/Maximum_flow_problem) consists of a directed graph with capacities associated with the edges, and a single source and target node between which one wants to find the maximum flow as restricted by the capacities of the edges. The Ribasim model graph (which I will call the model graph) can be converted into a graph for the maximum flow problem (which I will call the MFP graph) as follows:
- These are the nodes in the MFP graph:
  - The source node given by the source in the model graph;
  - Nodes in the model graph that form junctions (e.g. nodes 2 and 6 in the model and the end of [this comment](https://github.com/Deltares/Ribasim/pull/627#issuecomment-1735899799));
  - A single (sink) node that represents all users of a given priority (see [here](https://en.wikipedia.org/wiki/Maximum_flow_problem#Multi-source_multi-sink_maximum_flow_problem) for multi-sink maximum flow problems);
- Each edge is given by a path that connects 2 nodes mentioned above in the model graph. The capacity of an edge is given by the smallest `max_flow_rate` of the pumps and outlets on this path. If there are no such nodes on the path, either the capacity is considered infinite, or if this cannot be supported, merge the MFP graph nodes on either side of this edge into one MFP graph node. Other node types that restrict flow (e.g. `TabulatedRatingCurve`) are not taken into consideration at this stage, as discussed with @gijsber.

There are a few capacity constraints that are not included into the above, which are;
- The sum total demand of the users of a given priority;
- The flow rate of the source.

To include these constraints into the MFP graph, an extra edge (and phantom node) can be added directly after the source node whose capacity is the minimum of the capacity constraints above. 

### Solve max flow problems

Obtaining the demand propagation from the users to the source then goes as follows:
- Solve the priority 1 max flow problem with an appropriate [solver](https://juliagraphs.org/GraphsFlows.jl/dev/maxflow/);
[...]

PR #646, opened 2023-10-05
Allocation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/632.

There is currently some plotting code in the Ribasim core for visualizing the max flow results, this will probably be removed before merging. 

Here is my extensive (but probably incomplete) list of what still needs to be done if we keep this approach for allocation (not meant to be solved within this PR). Feel free to comment so that it can be worked out into productive issues. These points are ordered roughly from low to high expected complexity.

To do:
- [ ] Remove plotting code from Ribasim core, how to make the plotting functionality available for users? See also allocation output file below
- [ ] Validation: subnetworks should not overlap
- [x] Add support for users having demands for multiple priorities
- [ ] Take into account that users can be inactive
- [ ] Add more residual flow distribution algorithms (see powerpoint)
- [ ] Create modeller input framework (needs design session)
- [ ] Create allocation update policy; i.e. how often are allocations re-evaluated (needs design session)
- [ ] Create output file of allocation decisions
- [ ] Take local water supply into account for allocation
- [ ] Add support for more subnetwork sources than only `FlowBoundary`, e.g. `LevelBoundary` or something on the boundary between the subnetwork and the rest of the model. Generic approach: specify an edge as a source and grab the flow over that edge after formulating the flows as flow source capacity.
- [ ] Add interaction of allocation with control; e.g. set pump and outlet flow rates to achieve required capacity
- [ ] Add support for fractional flow: splits the capacity of edges in the max flow graph, and can make constraining nodes also junction nodes which is currently not supported
- [ ] Take return flow of users into account for allocation. One thing that at least must be done to support this is to add an extra node for the return flow if it goes to the same basin as from which the user subtracts, otherwise the solver doesn't deal with return flow properly (because in some sense it sees both edges between the basin and the user as the same thing). Also, having the return flow of a user be some fixed fraction of the inflow I don't think is supported by standard max flow solvers. We might have to go to a more genereric LP solver for this.
- [ ] Add support for multiple sources in a subnetwork

Maybe:
- [ ] Take the minimum abstraction level per user into account during allocation.


SouthEndMusic:
Here is the script I'm currently using to run allocation once at t=0 for the first allocation test model:

```julia
using Ribasim
using SQLite
using GraphMakie: save
using GeoDataFrames: read

# Load test model
toml_path = "path/to/Ribasim/generated_testmodels/subnetwork/subnetwork.toml"
cfg = Ribasim.Config(toml_path)
gpkg_path = Ribasim.input_path(cfg, cfg.geopackage)
db = SQLite.DB(gpkg_path)

gdf_node = read(gpkg_path, "Node")
close(db)

# Inputs specific for this test model
subgraph_node_ids = unique(keys(p.lookup))
source_node_id = only(p.flow_boundary.node_id)

# Parsed subnetwork (contains max flow graph and additional information)
subnetwork = Ribasim.get_graph_max_flow(p, subgraph_node_ids, source_node_id)

function run_allocation_once()::Nothing
    # Plot max flow graph
    plot_folder = "path/to/your/favorite/folder"
    fig, ax = Ribasim.plot_graph_max_flow(subnetwork; gdf_node)
    save(normpath("$plot_folder/Ribasim_max_flow.png"), fig)

    # Run allocation with max flow solver,
    # plots solutions of max flow problems on the max flow graph
    t = 0.0
    source_capacity = only(p.flow_boundary.flow_rate)(t)
    Ribasim.allocate!(p.user, subnetwork, source_capacity, t; plot_folder, gdf_node)
    @show(p.user.allocated)
    return nothing
end
```

SouthEndMusic:
Closed because superseded by https://github.com/Deltares/Ribasim/pull/663.

PR #647, opened 2023-10-05
Retain the order of the schemas and generated Python code

### LINKED ISSUE #640: schema ordering reversal
When I do `pixi run codegen`, it wants to change the order of `number` and `null` here:

```json
        "look_ahead": {
            "format": "default",
            "anyOf": [
                {
                    "type": "number"
                },
                {
                    "type": "null"
                }
            ]
        }
```

This is different for others for unknown reasons, see https://github.com/Deltares/Ribasim/pull/604#discussion_r1328392984.

Minor issue, but it's a bit annoying because I need to look for signal in the noise when finding what to commit after codegen.
Would be nice if this can be sorted or otherwise standardized.

visr:
Fixes #640, by using `OrderedDict` instead of `Dict` in `gen_schema.jl`.


evetion (approved):
This is an improvement :)

PR #648, opened 2023-10-06
add optional pixi metadata

visr:
As documented here: https://prefix.dev/docs/pixi/configuration

### DIFF
--- a/core/Project.toml
+++ b/core/Project.toml
@@ -1,6 +1,6 @@
 name = "Ribasim"
 uuid = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
-authors = ["Deltares and contributors"]
+authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
 version = "0.3.0"
 
 [deps]
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,8 +1,16 @@
 [project]
 name = "Ribasim"
 version = "0.2.0"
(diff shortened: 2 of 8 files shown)


pixi.toml
```diff
@@ -1,8 +1,16 @@
 [project]
 name = "Ribasim"
 version = "0.2.0"
+description = "Water resources modeling"
+authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
```
Hofer-Julian, line 5: Nice, I had no idea that we have an email address!

core/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Ribasim"
 uuid = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
-authors = ["Deltares and contributors"]
+authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
```
Hofer-Julian, line 3: Maybe add the email address to the Python packages as well?

Hofer-Julian (approved):

core/Project.toml
visr: Like ada5bd384c754eb8f4b9c92d2256595dc8bc4fd1? This removes individual authors. Much easier to keep up to date though, it was already out of date. If we want to highlight individual contributions that would probably best be done in one place.

Hofer-Julian (approved):

PR #652, opened 2023-10-09
Test Ribasim Python against multiple versions

### LINKED ISSUE #650: test on different python versions
Our Python tests are set up to run on Python 3.9, 3.10 and 3.11.
https://github.com/Deltares/Ribasim/blob/30d468a5d4b20d8c6393d5862dbffec8e83b01d3/.github/workflows/python_tests.yml#L23-L26

I was wondering with the Python version being fixed to 3.11.5 in the pixi.lock file, how this worked. Looking at the 3 jobs it seems that all three actually test Python 3.11.5:

https://github.com/Deltares/Ribasim/actions/runs/6419269733/job/17428758052#step:5:6
https://github.com/Deltares/Ribasim/actions/runs/6419269733/job/17428758689#step:5:6
https://github.com/Deltares/Ribasim/actions/runs/6419269733/job/17428759035#step:5:6

Also, is it worth it to test every minor Python release, or is the oldest and newest we support enough?

Found while working on https://github.com/Deltares/Ribasim-NL/pull/17. There I'll only test 3.11 for now.

Hofer-Julian:
Fixes #650

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -21,9 +21,9 @@ jobs:
           - macOS-latest
           - windows-latest
         python_version:
-          - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
         arch:
           - x86
     steps:
@@ -34,7 +34,9 @@ jobs:
           pixi-version: "latest"
           cache: true
       - name: Prepare pixi
-        run: pixi run install-without-pre-commit
+        run: >
+          pixi add python=${{ matrix.python_version }}
+          pixi run install-without-pre-commit
 
       - name: Run tests
         run: pixi run test-ribasim-python-cov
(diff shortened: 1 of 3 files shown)


visr:
Read from the release notes in #651

> setup-pixi now uses --locked by default

Perhaps that is causing issues?

Also, I see you don't test Python 3.9 support. Fine by me to drop support, should we make a separate issue to update the pyproject compat and start using 3.10 goodies like https://github.com/Deltares/Ribasim/pull/351#discussion_r1235405489?

Hofer-Julian:
> Perhaps that is causing issues?

I don't think that's the issue. I am just confused by YAML (who isn't?)

> Also, I see you don't test Python 3.9 support. 

Yes, I think the latest three versions is more than enough.

> Fine by me to drop support, should we make a separate issue to update the pyproject compat and start using 3.10 goodies like https://github.com/Deltares/Ribasim/pull/351#discussion_r1235405489?

Sounds good to me. Can you take care of that?

Hofer-Julian:
Seems like `pyogrio` doesn't like Python 3.12 yet.
Let's try again in a month or so.

visr (approved):

Issue #276, opened 2023-06-05
Allow multiple priorities for user demand


SnippenE:
Extension of #76 to be able to split demand into multiple at least 2 priorities (same as RIBASIM7)

SouthEndMusic:
Should the minimum subtraction level of the user be allowed to be time independent?

visr:
I think a constant minimum abstraction level is good enough for now.

PR #655, opened 2023-10-09
Multiple demands per user

SouthEndMusic:
fixes https://github.com/Deltares/Ribasim/issues/276

### DIFF
--- a/core/src/create.jl
+++ b/core/src/create.jl
@@ -294,7 +294,8 @@ function TabulatedRatingCurve(db::DB, config::Config)::TabulatedRatingCurve
             push!(interpolations, interpolation)
             push!(active, true)
         else
-            error("TabulatedRatingCurve node #$node_id data not in any table.")
+            @error "TabulatedRatingCurve node #$node_id data not in any table."
+            errors = true
         end
         if !is_valid
             @error "A Q(h) relationship for TabulatedRatingCurve #$node_id from the $source table has repeated levels, this can not be interpolated."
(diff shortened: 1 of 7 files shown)


docs/core/usage.qmd
```diff
@@ -376,7 +376,8 @@ control_state         | String  | -            | (optional)
 
 # User
```
visr, line 377: In the table below, can you add the new sorting requirement?

core/src/create.jl
```diff
@@ -624,33 +625,98 @@ end
...
     )
+
+    for node_id in node_ids
+        first_row = nothing
+        demand = Vector{ScalarInterpolation}()
+
+        if node_id in static_node_ids
+            rows = searchsorted(static.node_id, node_id)
+            static_id = view(static, rows)
+            for p in 1:maximum_priority
```
visr, line 662: This will not go well if a modeler wan't to be sure to give the lowest priority and thus sets

```julia
julia> typemax(Int)
9223372036854775807
```

Perhaps it is better to find the `sorted(unique(priorities))`` and loop over that?

core/src/create.jl
```diff
@@ -624,33 +625,98 @@ end
...
+            push!(min_level, min_level_)
+            push!(return_factor, return_factor_)
+        end
+    end
+
+    if errors
+        error("Errors occurred when parsing User data.")
+    end
+
+    allocated = [zeros(maximum_priority) for id in node_ids]
```
visr, line 717: Same issue here for large maximum_priority. Unless we map input of [4,8,1000] to [1,2,3] first.

visr (approved):

Issue #576, opened 2023-09-07
Allow extra table columns in Python builder


evetion:
So one can keep their own type of remarks/lineage. Might be hard because we hardcode all columns at the moment.

On hold after discussion with @gijsber en @visr, we wait for a customer request for this.

gijsber:
refinement 24/10: add extra column: 
- userdefined_id
- data_origin

evetion:
## Now
Extra columns get filtered

## Request
Allow extra columns.
Externally only a specific id column is requested, but internally we also see other usages, such as an origin column.

## Solutions
Ordered roughly by complexity/restriction:

- Allow all
- Allow all by opt-in in toml
- Allow all but add prefix
- Add requested column(s) only (add to schema)
- Allow specific columns by opt-in in toml
- Allow metadata in separate (but linked) tables
- Allow metadata in separate (but linked) node/edge tables

## Objections
- Extra columns can be confusing
- Allowing extra columns (noise) is just a bad practice
- Specifying columns in the toml is 
- Metadata in separate tables (especially with foreign keys) introduces a lot of complexity for both users as developers.

## Consensus
There is no clear consensus. We need to allow for something, but ideally restrict it if possible.

## Authors opinion
Much of the discussion is *how* people should model, but it doesn't relate to any Ribasim features. However, some proposals lead to extra complexity of the Ribasim (Python/QGIS) code.
This seems unnecessary, and should not be our focus. I therefore propose:

Allow all extra columns, but prefix columns (like `nr|`) to prevent confusion.

Making things opt-in seems patronizing to me, and given our own creation of this issue and the external request, it should be opt-out if anything.

gijsber:
Yesterday, Edwin and me came up with an alternative solution for traceability how the model was constructed.

Basically you need to specify:
1. a tag to the github script version that was used to generate the model
2. a tag to the file which holds the references to the basedata from which the model was generated.
We could store these tags in the TOML file in a separate section which is neglected by our code (modelconstruction_history).
we could even include other information like the person who created the model, the creation date etc.

gijsber:
I would say any TOML-field below `modelconstruction_history` is open for expansion, so people van also add a link to a document which describes how a model is created (or manully improved)

visr:
Model metadata is also an interesting topic, but it is really a different issue. This is about node metadata. There are generally too many nodes to store the metadata in the TOML.

gijsber:
[...]

PR #658, opened 2023-10-09
Add name as column and use as visualisation in QGIS.

### LINKED ISSUE #567: add 'name' as optional string field in node type schemas
the name field can be used in communication and visualization. If the impact is low, it would be nice to log both the id (which is unique) and the name

evetion:
Fixes #567

- [x] Some warnings still need updating

Screenshot 2023-10-09 at 13 46 40


core/src/create.jl
```diff
@@ -153,7 +154,7 @@ function parse_static_and_time(
                     )
                     if !is_valid
                         errors = true
-                        @error "A $parameter_name time series for $nodetype node #$node_id has repeated times, this can not be interpolated."
+                        @error "A $parameter_name time series for $nodetype node $node_name (#$node_id) has repeated times, this can not be interpolated."
```
visr, line 157: I expect that node names will often be empty strings. Then this gives a double space `node  (#2)`. How about using `repr(node_name)` such that the quotes are written. That reads nicer for longer names with spaces as well: `node "Lake IJssel" (#2)` / `node "" (#2)`.

visr:
You meant this fixes #567, not #576 I assume? Edited your post for this.

visr (approved):

Issue #349, opened 2023-06-20
Develop PoC allocation decisions


SnippenE:
First try to address #14 and than #79 (if needed) 

Steps in the computation process
- [ ] identify main water system inlets
- [ ] route requested water abstractions (demands) from users to main water system inlets, administered by priority
- [ ] allocate water from flow boundaries and storages over main water system to inlets while accounting for their priority
- [ ] for each priority allocate water from inlet to local users and update their demands to an 'allocated abstraction'
- [ ] possibly there is a need to update control targets (flows) at main water system inlets
- [ ] run the flow simulation (with control) with users that abstract the 'allocated demand'. also register the actual realized abstraction

Assumptions: routing and allocation decisions do not take travel times/delays into account
#564  will be added as additional water demand after the PoC has shown that it can work
#565 may be added at later stage

![image](https://github.com/Deltares/Ribasim/assets/14976756/e68665ee-460b-469f-800f-21be53b705d6)

gijsber:
[allocation_test_model_1.xlsx](https://github.com/Deltares/Ribasim/files/12646449/allocation_test_model_1.xlsx) holds a basic test model to start with. It assumes fixed storage volumes. More complicated/extended versions will follow

gijsber:
As agreed while discussing with Martijn and Bart:
It is better to incorporate the discharge constraints in the demand propagation and simultaneously write an spatially explicit-output file which highlights where your supply constraints are

gijsber:
Test model with loop. to be decided whether the edge direction of the pump may need to be switched.

![Image](https://github.com/Deltares/Ribasim/assets/14976756/fdf5bd0a-4c0e-478d-8e0e-b611a7c07d78)

Issue #14, opened 2022-10-28
Metaissue: Global demand and allocation callback (nested allocation)


visr:
Aggregate the demands of the districts and allocate the available water according to defined priorities.

Define bifurcation mechanisms that don't need explicit tables for each possible split.

Define use cases (including bidirectional flow channels).

gijsber:
try to make a decision layer (on top of the physical-oriented control layer) for the prioritization of abstractions. This layer would assess what abstactions could fit given the physical situation and reduce those abstractions that have lowest priority. then those reduced abstractions are given to the physical/control layer

gijsber:
Per timestep this would result in the following processes:



![Image](https://github.com/Deltares/Ribasim/assets/14976756/06fdf6b1-d3da-43f7-8505-ed12a4bee288)

gijsber:
level control vs abstractions: how to manage priorities:


![Image](https://github.com/Deltares/Ribasim/assets/14976756/107f4c0b-3be6-4832-b444-998e63cab363)

Sldies can also be found on sharepoint: \DEL155 Oppervlaktewatermodule NHI - Documents\0. Deltares only\4_presentaties\achitecture_layer.pptx

PR #663, opened 2023-10-11
Allocation beyond max flow

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/632, fixes https://github.com/Deltares/Ribasim/issues/14, fixes https://github.com/Deltares/Ribasim/issues/349, using `JuMP.jl` and `HiGHS`.


SouthEndMusic:
Closed because superseded by https://github.com/Deltares/Ribasim/pull/663.

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+
+Both fractional flow nodes and user nodes dictate proportional relationships between flows over edges in the subnetwork. Users have a return factor $r_i, i \in U_S$.
+
+## The allocation optimization problem
+
+### The allocation graph
+
+A new graph is created from the subnetwork, which we call the allocation graph. To indicate the difference between subnetwork data and allocation graph data, the allocation graph data is denoted with a hat. The allocation graph consists of:
+
+- Nodes $\hat{V_S}$, where each basin, source and user in the subnetwork get a node in the allocation graph. Also nodes that have fractional flow outneighbors get a node in the allocation graph. The implementation makes heavy use of the node id mapping $m_S : i \mapsto \hat{i}$ to translate from subnetwork node Ds to allocation graph node IDs. Unless specified otherwise, we assume this relationship between index symbols that appear both with and without a hat.
```
visr, line 65:
```suggestion
- Nodes $\hat{V_S}$, where each basin, source and user in the subnetwork get a node in the allocation graph. Also nodes that have fractional flow outneighbors get a node in the allocation graph. The implementation makes heavy use of the node id mapping $m_S : i \mapsto \hat{i}$ to translate from subnetwork node IDs to allocation graph node IDs. Unless specified otherwise, we assume this relationship between index symbols that appear both with and without a hat.
```

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+    \begin{align}
+        \hat{V}^{\text{out}}_S(\hat{i}) = \left\{\hat{j} \in \hat{V}_S : (\hat{i},\hat{j}) \in \hat{E}_S\right\} \\
+        \hat{V}^{\text{in}}_S(\hat{j}) = \left\{\hat{i} \in \hat{V}_S : (\hat{i},\hat{j}) \in \hat{E}_S\right\}
+    \end{align}
+$$
+for the set of in-neighbors and out-neighbors of a node in the allocation graph respectively.
+
+### The allocation graph capacities
+
+The capacities of the edges of the allocation graph are collected in the sparse capacity matrix $\hat{C}_S \in \overline{\mathbb{R}}_{\ge 0}^{\hat{n}\times\hat{n}}$ where $\hat{n}$ is the number of nodes in the allocation graph. The capacities can be infinity.
```
visr, line 79:
```suggestion
The capacities of the edges of the allocation graph are collected in the sparse capacity matrix $\hat{C}_S \in \overline{\mathbb{R}}_{\ge 0}^{\hat{n}\times\hat{n}}$ where $\hat{n}$ is the number of nodes in the allocation graph. The capacities can be infinite.
```

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+
+The The capacities are determined in 3 different ways:
```
visr, line 81:
```suggestion
The capacities are determined in 3 different ways:
```

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+    F_{\hat{i}\hat{j}} = f_j \sum_{k\in \hat{V}^\text{in}_S(\hat{i})} F_{\hat{k}\hat{i}} \\
+    \forall \hat{i} \in \hat{L}_S, \\
+    \hat{j} \in \hat{V}_S^\text{out}(\hat{i}).
+\end{align}
+$$ {#eq-fractionalflowconstraint}
+
+- Flow sign: Furthermore there are the non-negativity constraints for the flows and allocations, see [The optimization variables](allocation.qmd#the-optimization-variables).
+
+:::{.callout-note}
+Note that $\Phi_{\hat{k}}$ can become negative, in which case @eq-flowconservationconstraint becomes impossible to satisfy. What to do in this case?
```
visr, line 169: How is this possible, isn't it defined to be `max(x, 0.0)`?

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+:::
+
+## Final notes on the allocation problem
+
+### Users using their own return flow
+
+If not explicitly avoided, users can use their own return flow in this allocation problem formulation. Therefore, Return flow of users is only taken into account by allocation if that return flow is downstream of the user where it comes from. That is, if there is no path in the directed allocation graph from the user outflow node back to the user.
```
visr, line 176:
```suggestion
If not explicitly avoided, users can use their own return flow in this allocation problem formulation. Therefore, return flow of users is only taken into account by allocation if that return flow is downstream of the user where it comes from. That is, if there is no path in the directed allocation graph from the user outflow node back to the user.
```

Side note, could you perhaps put sentences on separate lines, rather than a paragraph on a line? That makes it easier to edit and track diffs, and is more consistent with the other docs I think.

docs/core/allocation.qmd
```diff
@@ -0,0 +1,178 @@
...
+
+# Solving the allocation problem
```
visr, line 178: Maybe leave this out? Or add some stub text explaining what this is about.

core/src/allocation.jl
```diff
@@ -0,0 +1,503 @@
...
+An AllocationModel object, see solve.jl.
+
+"""
+function get_allocation_model(
+    p::Parameters,
+    subnetwork_node_ids::Vector{Int},
+    source_edge_ids::Vector{Int},
+    Δt_allocation::Float64,
+)::AllocationModel
+    (; connectivity, user, lookup, basin) = p
```
visr, line 55:
```suggestion
    (; connectivity, user, lookup) = p
```

Basin is not used.

core/src/allocation.jl
```diff
@@ -0,0 +1,503 @@
...
+source_edge_ids:: The IDs of the edges in the subnetwork whose flow fill be taken as
+    a source in allocation
+Δt_allocation: The timestep between successive allocation solves
+
+Outputs
+-------
+An AllocationModel object, see solve.jl.
+
+"""
+function get_allocation_model(
```
visr, line 49: You could also rename this to `AllocationModel` to signify it is a constructor function, similar to the ones in `create.jl`.

core/src/allocation.jl
```diff
@@ -0,0 +1,503 @@
...
+    return false
+end
+
+"""
+Construct the JuMP.jl model for allocation.
+
+Definitions
+-----------
+- 'subnetwork' is used to refer to the original Ribasim subnetwork;
+- 'AG' is used to refer to the allocation graph.
```
visr, line 34: I'd stick to the lowercase conventions for variable and function names: `ag`, just like we talk about IDs in docs but `ids` in code.
`ag` is a bit cryptic so it's good you define it. It sounds more like agriculture to me.
Allocation is commonly abbreviated to alloc. `allocgraph` is not too bad.

core/src/allocation.jl
```diff
@@ -0,0 +1,503 @@
...
+        AG_node_id for
+        (AG_node_id, node_type) in values(node_id_mapping) if node_type == :user
+    ]
+    AG_node_ids_user_with_returnflow = Int[]
+    for AG_node_id_user in AG_node_ids_user
+        AG_node_id_return_flow = only(outneighbors(graph_allocation, AG_node_id_user))
+        if path_exists(graph_allocation, AG_node_id_return_flow, AG_node_id_user)
+            rem_edge!(graph_allocation, AG_node_id_user, AG_node_id_return_flow)
+            # TODO: Add to logging?
+            @warn "The outflow of user #$(node_id_mapping_inverse[AG_node_id_user][1]) is upstream of this user itself and thus ignored."
```
visr, line 290: This is a valid situation, so it's more informational, since warnings tend to be for things users should change. Perhaps a debug statement is better?
Also it would be good to explain what is ignored, reading it at face value modelers could think the outflow does not happen.

core/src/allocation.jl
```diff
@@ -0,0 +1,503 @@
...
+        node_id_mapping_inverse,
+        source_edge_mapping,
+        graph_allocation,
+        capacity,
+        model,
+        Δt_allocation,
+    )
+end
+
+function allocate!(p::Parameters, allocation_model::AllocationModel, t::Float64;)::Nothing
```
visr, line 450: Could use a short explanation of what this function does in a docstring.

docs/core/allocation.qmd
SouthEndMusic: You're right, I forgot to remove this after I found a solution for it.

core/src/allocation.jl
```diff
@@ -0,0 +1,531 @@
...
+source_edge_ids:: The IDs of the edges in the subnetwork whose flow fill be taken as
+    a source in allocation
+Δt_allocation: The timestep between successive allocation solves
+
+Outputs
+-------
+An AllocationModel object.
+
+"""
+function AllocationModel(
```
Hofer-Julian, line 49: This function is really long. Ideally, every part which has its own comment block would be its own function instead.

core/test/time.jl
```diff
@@ -45,7 +45,7 @@ end
 
     t_end = Ribasim.seconds_since(cfg.endtime, cfg.starttime)
 
-    # demand[user_idx][priority](t)
+    # demand[user_idx][priority_idx](t)
```
Hofer-Julian, line 48: Why do we need this commented-out code?

core/src/allocation.jl
```diff
@@ -0,0 +1,531 @@
...
+                            allocgraph_node_id_1,
+                            allocgraph_node_id_2,
+                        )
+                        if subnetwork_outneighbor_id in user.node_id
+                            # Capacity depends on user demand at a given priority
+                            capacity[allocgraph_node_id_1, allocgraph_node_id_2] = Inf
+                        else
+                            # These direct connections cannot have capacity constraints
+                            capacity[allocgraph_node_id_1, allocgraph_node_id_2] = Inf
+                        end
```
Hofer-Julian, line 156: I am confused about this. Both branches seem to do the same thing.

core/src/allocation.jl
```diff
@@ -0,0 +1,531 @@
...
+        # No connection to allocgraph node found on this side, so edge is discarded
+        if isnothing(allocgraph_node_id_2)
+            continue
+        end
+
+        # Find capacity of this composite allocgraph edge
+        positive_flow = true
+        negative_flow = true
+        allocgraph_edge_capacity = Inf
+        for (i, subnetwork_node_id) in enumerate(allocgraph_edge_composite)
```
Hofer-Julian, line 228: I think iterating over a partition of three elements makes more sense here: https://docs.julialang.org/en/v1/base/iterators/#Base.Iterators.partition

Using raw indexes should only be seen as the last option after functions in the [Iteration utilities](https://docs.julialang.org/en/v1/base/iterators/#Iteration-utilities) have been exhausted.

Hofer-Julian (commented):
I've added a few comments :)

core/src/allocation.jl
Hofer-Julian: https://refactoring.guru/smells/long-method

core/test/time.jl
SouthEndMusic: We don't, it was there as a reminder to myself.

visr: This wasn't really commented-out code though, it was documentation explaining what the two indices represented for the lines below:

```
    @test demand[1][2](0.5 * t_end) ≈ 1.0
    @test demand[2][1](0.0) ≈ 0.0
    @test demand[2][1](t_end) ≈ 0.0
```

Which I think is useful. But no need to put it back though if you don't need it.

visr (approved):
Let's get this merged.

PR #672, opened 2023-10-17
Add dependency on testmodel for build binary tasks

Hofer-Julian:
The testmodels need to be present for the build process to work

I've also formatted, so the diff is slightly bigger than expected


visr (approved):
Right, this is because we run a test model as a precompilation step in https://github.com/Deltares/Ribasim/blob/main/build/create_binaries/precompile.jl to make sure the binaries start more quickly since more of the code is already compiled.

Issue #630, opened 2023-09-28
Implement new Python API


SnippenE:

evetion:
For this sprint, make sure that new classes integrate and the old tests work. We can then merge, and start further work on the object level API.

The idea is that tests and external people keep using the "DataFrame" API, although slight breakage might occur.

PR #673, opened 2023-10-17
support pydantic v2

### LINKED ISSUE #382: Start supporting pydantic v2
pydantic v2 is now out with a lot of nice things, most notably higher speed.

I think we will want to support v2 as fast as possible, but depending on the amount of work, it would be nice to continue supporting v1 as well.
If we don't do that, packaging Ribasim Python on Deltaforge will be hard to do in the near future because of version requirement conflicts.

The migration guide can be found here: https://docs.pydantic.dev/latest/migration/

visr:
I went a few steps further than mentioned in https://github.com/Deltares/Ribasim/issues/382#issuecomment-1765212299, but I get stuck, so it is probably better to continue after #630.

Main issues are that these nested default dicts are not appreciated by mypy: https://github.com/Deltares/Ribasim/commit/50cf27f761ebb2cc77d1bda8993d594826987c4d#diff-bdece7de80ef751956236e71b5aed3964311d59479f076c892d6745b0e30266d

The other one is that it doesn't like the optional DataFrame:

```
>       node = ribasim.Node(
            static=gpd.GeoDataFrame(
                data={
                    "type": node_type,
                    "name": [ribasim.utils.random_string() for _ in range(len(node_id))],
                },
                index=pd.Index(node_id, name="fid"),
                geometry=node_xy,
                crs="EPSG:28992",
            )
        )
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for Node
E       static
E         Value error, Cannot use DataFrame as a pydantic type as its DataFrameModel cannot be converted to a DataFrameSchema.
E       Please revisit the model to address the following errors:
E       Found missing annotations: ['model_config'] [type=value_error, input_value=                     type...vc   POINT (6.000 1.000), input_type=GeoDataFrame]
E           For further information visit https://errors.pydantic.dev/2.4/v/value_error

python\ribasim_testmodels\ribasim_testmodels\basic.py:160: ValidationError
```

Would fix #382.


evetion:
I'll take over this branch, thank you very much.

PR #674, opened 2023-10-17
Read some interpolation tables into the Model struct

Huite:


core/src/export.jl
```diff
@@ -0,0 +1,58 @@
...
+            level_exporters[name] = LevelExporter(group, node_to_basin)
+        end
+    end
+    return level_exporters
+end
+
+"""
+Compute a new water level for each external element.
+"""
+function update!(exporter::LevelExporter, basin_level)
```
Hofer-Julian, line 54:
```suggestion
function update!(exporter::LevelExporter, basin_level)::Nothing
```

core/src/export.jl
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    return LevelExporter(basin_ids, interpolations, fill(NaN, length(basin_ids)))
+end
+
+function create_level_exporters(db::DB, config::Config, basin::Basin)::Dict{String, LevelExporter}
+    node_to_basin = Dict(node_id => index for (index, node_id) in enumerate(basin.node_id))
+    tables = load_structvector(db, config, LevelExporterStaticV1)
+    level_exporters = Dict{String, LevelExporter}()
+    if length(tables) > 0
```
visr, line 42:
```suggestion
    if !isempty(tables)
```

docs/core/usage.qmd
```diff
@@ -275,6 +276,47 @@ Internally this get converted to two functions, $A(S)$ and $h(S)$, by integratin
...
+------------- | ------- | ----- | ------------------------
+name          | String  | -     |
+subgrid_id    | Int     | -     | sorted
+node_id       | Int     | -     | constant per subgrid_id
+basin_level   | Float64 | $m$   | per subgrid_id: increasing
+subgrid_level | Float64 | $m$   | per subgrid_id: increasing
+
+The name column is used to distinguish between spatially distinct water bodies that may
+exist within a single basin (e.g. perennial versus ephemeral streams). When no
+distinction is required, all rows should given the same name.
```
Hofer-Julian, line 297:
```suggestion
distinction is required, all rows should be given the same name.
```

Hofer-Julian (commented):
One small comment. I don't think I am familiar enough with the rest of the code to review it. I leave that to @visr 🙃

visr:
I've done some fixes here so we can use "Basin / subgrid_level" as you originally intended. I seem to have broken the test along the way though.

I could use more description of the design though. Main questions are:
- Why do we need the "name" column?
- Why distinguish between "subgrid level" and subgrid exporter"? Can they be named the same?

Huite:
Good question about the name. I did some thinking: this stems from my idea that Ribasim would present 1:1 levels to MODFLOW, but it turns out this is impractical for multiple reasons, because we do not necessarily want the subgrid elements to align 1:1 (e.g. because the Ribasim domain is larger spatially than the MODFLOW domain, or vice versa).

The name would still be a useful entry to have when setting up the coupling schemes, but in that sense it's no different from the x and y columns that we discussed yesterday. In that case, all the grouping is a coupler problem only, and in Ribasim everything can be flattened (which also cleans up the BMI).

I guess the idea with exporter and level is that the exporter "does the exporting"; the level is just a level.

I'll omit the name column, and see whether the exporter / level thing can be cleared up somewhat.

core/src/bmi.jl
```diff
@@ -480,6 +494,26 @@ function save_flow(u, t, integrator)
...
+        end
+    end
+end
+
+"Interpolate the levels and save them to SavedValues"
+function save_subgrid_level(u, t, integrator)
+    update_subgrid_level!(integrator)
+    return vcat(
+        [exporter.subgrid_level for exporter in values(integrator.p.subgrid_exporters)]...,
+    )
```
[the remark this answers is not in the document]
Huite: By removing the nested subgrids, this is no longer required fortunately.

visr (approved):

PR #676, opened 2023-10-18
pixi: Update test-ribasim-api

Hofer-Julian:


visr:
What does this do?

Hofer-Julian:
> What does this do?

Add the temp dir locally so its easier to inspect and generate an xml file for teamcity test discovery

visr (approved):
Interesting, thanks.

PR #678, opened 2023-10-19
update Manifests and add core Manifest

visr:
The manifests should have been re-resolved after #677, since that requires DataInterpolations 4.4, but the Manifests still used 4.0.

This just updates all packages in them also switching to a newer PackageCompiler version for instance.

This should fix TeamCity builds.


visr:
This also checks in the `core/Manifest.toml`, to avoid breakage like this in the future. With that we essentially lock all our depencencies, in pixi.lock and the several Manifest.toml files.

Hofer-Julian (approved):

Issue #10, opened 2022-10-28
Integration tests


visr:
- [ ] continuous integration on Teamcity
- [ ] runtime performance test
- [ ] variants with and without MODFLOW

PR #683, opened 2023-10-19
refine edge labels

### LINKED ISSUE #175: Labeling of edges
Edges aren't labelled, by default, and when looking at the output it becomes a little confusing:

image

4 and 3 refer to the edge FIDs here, but that isn't quite clear.

Ideally, I think, we could use `from_node_id` and `to_node_id` to label the edges, though maybe not by default:

image

This works relatively easily, as QGIS supports expressions for labels:

image

I think it's possible to configure the layer to have these labels by default, but not show them. Anyway, maybe something to discuss.

visr:
Fixes #175
Most of the work for this was already done in #658.

This changes a few things:
- Stop using brackets around the ID to reduce visual clutter: now we use `"Lake IJssel" #10` in printed errors and `Lake IJssel #10` as a QGIS label
- Add the edge ID to the label, this is the part that fixed #175
- Stop putting random strings as names in our basic test model
- Add some real names to our Dutch waterways test model

image

### DIFF
--- a/core/src/create.jl
+++ b/core/src/create.jl
@@ -154,7 +154,7 @@ function parse_static_and_time(
                     )
                     if !is_valid
                         errors = true
-                        @error "A $parameter_name time series for $nodetype node $(repr(node_name)) (#$node_id) has repeated times, this can not be interpolated."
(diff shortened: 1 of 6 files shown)


evetion (approved):
I'm gonna miss my random names.

Besides, there's an Arhnem in the screenshot, what's up with that?

PR #687, opened 2023-10-20
strip cldr artifact before build

visr:
Unfortunately #685 didn't work, since the same problematic artifact is included in older releases.
This works around the same issue by stripping the cldr artifact of all unnecessary files.


Hofer-Julian:
I've started two jobs to test this branch:

- https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_RibasimJl_BuildLibribasimWindows/2855462?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true
- https://dpcbuild.deltares.nl/buildConfiguration/BuildRibasimCliWindows/2855465?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

Hofer-Julian:
It builds now, but when actually using the `exe` it fails:
https://dpcbuild.deltares.nl/buildConfiguration/BuildRibasimCliWindows/2855465?buildTab=log&focusLine=0&linesState=534&logView=flowAware

visr:
Nice that it builds. The new error seems unrelated, possibly this is new since we updated the manifests.

I've seen it before that it downloads MKL on first usage. That seems to fail now, not sure if that is only on TeamCity or everywhere, I don't see an artifact for this build that I can download.

I've added 56eae5f73cefee598b893798352a115b128fdb86 to see if that helps. That should already download and bundle all lazy artifacts, which seems preferable anyway, granted it doesn't download many more artifacts we don't actually use.

Hofer-Julian:
> don't see an artifact for this build that I can download.

It doesn't upload artifacts if the build fails, I've changed that behavior and triggered a new build

visr:
Nice, thanks! This seems to work well, also tried the cli locally. It doesn't download MKL on first use anymore. Though now the ribasim_cli.zip grew from 250 MB to 435 MB, due to the new MKL artifact that is 640 MB unzipped.

The reason we need MKL is because we use https://github.com/SciML/LinearSolve.jl which depends on it. Not sure if it is feasible/desirable to try to avoid it.

Hofer-Julian:
> ribasim_cli.zip grew from 250 MB to 435 MB

Unpacked, they are now 1.57 GB. I feel like we reached the point that investigating these large binary sizes should become a higher priority.

Hofer-Julian (approved):
Libribasim tests are still failing but might be because of some TeamCity weirdness.
Let's give it a try!

Issue #633, opened 2023-09-29
support autodiff w.r.t. time


visr:
The algorithms `Rosenbrock23` and `Rodas5` both don't work with autodiff. So far we only support differentiation w.r.t. u, but these Rosenbrock methods need to differentiate w.r.t. time.

To reproduce, take any model and add. (Autodiff is true by default).

```toml
[solver]
algorithm="Rodas5"
autodiff=true
```

```
julia> model = Ribasim.run("generated_testmodels/non_negative/non_negative.toml")
ERROR: First call to automatic differentiation for time gradient
failed. This means that the user `f` function is not compatible
with automatic differentiation. Methods to fix this include:

1. Turn off automatic differentiation (e.g. Rosenbrock23() becomes
   Rosenbrock23(autodiff=false)). More details can be found at
   https://docs.sciml.ai/DiffEqDocs/stable/features/performance_overloads/
2. Improving the compatibility of `f` with ForwardDiff.jl automatic
   differentiation (using tools like PreallocationTools.jl). More details
   can be found at https://docs.sciml.ai/DiffEqDocs/stable/basics/faq/#Autodifferentiation-and-Dual-Numbers
3. Defining analytical Jacobians and time gradients. More details can be
   found at https://docs.sciml.ai/DiffEqDocs/stable/types/ode_types/#SciMLBase.ODEFunction

Note 1: this failure occurred inside of the time gradient function. These
time gradients are only required by Rosenbrock methods (`Rosenbrock23`,
`Rodas4`, etc.) are are done by automatic differentiation w.r.t. the
argument `t`. If your function is compatible with automatic differentiation
w.r.t. `u`, i.e. for Jacobian generation, another way to work around this
issue is to switch to a non-Rosenbrock method.

Note 2: turning off automatic differentiation tends to have a very minimal
performance impact (for this use case, because it's forward mode for a
square Jacobian. This is different from optimization gradient scenarios).
However, one should be careful as some methods are more sensitive to
accurate gradients than others. Specifically, Rodas methods like `Rodas4`
and `Rodas5P` require accurate Jacobians in order to have good convergence,
while many other methods like BDF (`QNDF`, `FBDF`), SDIRK (`KenCarp4`),
and Rosenbrock-W (`Rosenbrock23`) do not. Thus if using an algorithm which
is sensitive to autodiff and solving at a low tolerance, please change the
algorithm as well.

[...]

PR #688, opened 2023-10-20
Automatic differentiation w.r.t. time

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/633.

I don't know whether what I did to make this work is an elegant solution or an ugly hack. Also, although a simulation with `Rodas5` now runs without errors, the performance is poor so there is obviously still something wrong.


core/src/solve.jl
```diff
@@ -557,10 +591,12 @@ function formulate_basins!(
     basin::Basin,
     flow::AbstractMatrix,
     storage::AbstractVector,
+    t::Number,
 )::Nothing
     (; node_id, current_level, current_area) = basin
-    current_level = get_tmp(current_level, storage)
-    current_area = get_tmp(current_area, storage)
+    diffvar = isa(t, Dual) ? t : storage
```
visr, line 597: Could you split this line out into a function with two methods? If the types can be inferred by the compiler, that would eliminate this branch.

You wouldn't need it anymore, but the rest of the code mainly uses `x isa T` syntax rather than `isa(x, T)`.

core/src/solve.jl
```diff
@@ -4,6 +4,40 @@ const ScalarInterpolation =
...
+    rowval::Vector{Ti}      # Row indices of stored values
+    nzval::D                # Stored values, typically nonzeros
+end
+
+const SparseCache = Union{SparseMatrixCSC_DiffCache, SparseMatrixCSC_cache}
+
+SparseArrays.size(matrix::SparseCache) = (matrix.m, matrix.n)
+SparseArrays.getcolptr(matrix::SparseCache) = matrix.colptr
+SparseArrays.rowvals(matrix::SparseCache) = matrix.rowval
+SparseArrays.nonzeros(matrix::SparseCache) = matrix.nzval
```
visr, line 29: Could you explain why you need to implement your own AbstractSparseMatrixCSC types? It seems the only difference is the `eltype`. That is the first type parameter:

```
SparseMatrixCSC{Tv,Ti<:Integer} <: AbstractSparseMatrixCSC{Tv,Ti}
```

Is it also possible to define separate methods for different `Tv`, to avoid needing our own sparse matrix types?

core/src/create.jl
```diff
@@ -201,6 +201,17 @@ end
...
     Set{String}(["Basin", "LevelBoundary", "FlowBoundary", "Terminal", "User"])
 
+function get_chunk_sizes(config::Config, chunk_size::Int)::Vector{Int}
+    chunk_sizes = Int[]
+    if config.solver.autodiff
+        push!(chunk_sizes, chunk_size)
+    end
+    if config.solver.algorithm in ["Rodas5", "Rosenbrock23"]
+        push!(chunk_sizes, 1)
+    end
```
visr, line 211: Should the second `if` not be inside the first `if`? We only use autodiff when `config.solve.autodiff` right?

We might add other Rosenbrock options in the future, perhaps this check is a bit more robust:

`algorithm(config.solver) isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm`

if you do `using OrdinaryDiffEq: OrdinaryDiffEqRosenbrockAdaptiveAlgorithm`.

core/src/create.jl
```diff
@@ -225,9 +236,10 @@ function Connectivity(db::DB, config::Config, chunk_size::Int)::Connectivity
     end
     flow .= 0.0
 
-    if config.solver.autodiff
-        # FixedSizeDiffCache performs better for sparse matrix
-        flow = FixedSizeDiffCache(flow, chunk_size)
+    chunk_sizes = get_chunk_sizes(config, chunk_size)
+    if !isempty(chunk_sizes)
```
visr, line 240: If my other comment is true, you could still do this, here and below: `if config.solver.autodiff`. That is easier to read than `isempty`.

visr:
Adding a test would be good to make sure this keeps working. Perhaps it's time to set up a solver test where we structurally test all the solver variants. Could also be a separate issue however. This would need a sufficiently complex model to exercise many paths, but can be a very short simulation to keep it fast enough.

core/src/create.jl
SouthEndMusic: Ah nice, I couldn't find that more robust check.

visr: I find this function convenient for finding the right abstract type:

```
julia> supertypes(Int)
(Int64, Signed, Integer, Real, Number, Any)
```

core/src/create.jl
```diff
@@ -201,6 +201,32 @@ end
...
+        push!(chunk_sizes, 1)
+    end
+    return chunk_sizes
+end
+
+"""
+If the tuple of variables contains a Dual variable, return the first one.
+Otherwise return the last variable.
+"""
+function get_diffvar(variables::Tuple)
```
visr, line 221: I think this would be a bit more idiomatic than using a Tuple.

```
julia> f(args...) = [a for a in args]
f (generic function with 1 method)

julia> f(3, 4)
2-element Vector{Int64}:
 3
 4
```

core/src/solve.jl
```diff
@@ -4,6 +4,57 @@ const ScalarInterpolation =
...
+Previously FixedSizeDiffCache was used for the sparse matrix Connectivity.flow, but FixedSizeDiffCache
+does not support multi-level Duals.
+"""
+struct SparseMatrixCSC_cache{T <: Number, Ti <: Integer} <:
+       SparseArrays.AbstractSparseMatrixCSC{T, Ti}
+    m::Int                  # Number of rows
+    n::Int                  # Number of columns
+    colptr::Vector{Ti}      # Column j is in colptr[j]:(colptr[j+1]-1)
+    rowval::Vector{Ti}      # Row indices of stored values
+    nzval::Base.ReinterpretArray{T} # cache
```
visr, line 34: This field is not yet concretely typed, since a ReinterpretArray has many more type parameters: https://github.com/JuliaLang/julia/blob/v1.10.0-beta3/base/reinterpretarray.jl#L9

In combination with https://github.com/Deltares/Ribasim/issues/694#issuecomment-1775409127 I'm starting to wonder if we should keep using a sparse matrix for flow at all, or if we should use some other data structure like a dict.

visr:
Converting back to draft, decided with @SouthEndMusic that it's best to finish this after moving away from using a sparse array as a data structure for flow: https://github.com/Deltares/Ribasim/pull/688#discussion_r1371371192.

SouthEndMusic:
Closed in favor of https://github.com/Deltares/Ribasim/pull/850.

PR #693, opened 2023-10-23
pixi: Update registry before instantiating

Hofer-Julian:


visr:
Not sure what's wise here. Instantiate [should do this automatically](https://github.com/JuliaLang/Pkg.jl/blob/cc71837381da2569b0d8e9d2c6130a13f3f5b8f4/src/API.jl#L1653-L1654), but the reason for this PR is that we observed once that didn't happen. Perhaps we should check if we can reproduce it first? If we do this, I'd probably prefer using the keyword argument `initialize(; update_registry = true)` explicitly.

Hofer-Julian:
> initialize(; update_registry = true)

That is already the default value, so I assume if there's a bug it is deeper in the code.

I am actually fine either way, when I opened the PR I didn't know yet that this was supposed to be automatic.

visr:
Shall we just close this then? If it happens again it's relatively easy to tell since it will tell you that new versions you require don't exist.

Hofer-Julian:
Sure

PR #695, opened 2023-10-23
don't require consecutive node IDs from 1 to n

### LINKED ISSUE #694: numbering of nodes
Maybe this comment is related to the issue "stability of node and edge IDs". 

I try to change my basis model following from the lumping tool within a python script to avoid changing my model manually in Qgis each time my basis model is changing with newer versions of the lumping/volume tools.

One of my actions is to remove isolated basin nodes (not connected to other nodes). However, when you remove a node from your model.node.static geodataframe the model is not valid anymore. The validation test is based on the number of nodes and a continues numbering of nodes, which means that no numbers may be skipped. If I would rearrange my numbering of the nodes, my edges definitions are no longer correct as well. Is it possible to keep the numbering as is, also when nodes are removed?

visr:
Fixes #694

Just a draft, remaining steps are written here: https://github.com/Deltares/Ribasim/issues/694#issuecomment-1775409127
We should first decide whether we want this or not.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -215,7 +215,6 @@ class Model(BaseModel):
 
         # Check node IDs of node fields
         node_IDs_all = []
-        n_nodes = len(self.node.static)
 
         for name in self.fields():
             if name in node_names_all_snake_case:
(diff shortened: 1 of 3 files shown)


visr:
Fixed in #885

PR #696, opened 2023-10-24
Fix nodetype for Node and Edge in load_table.

### LINKED ISSUE #637: `load_structvector` fails with types `NodeV1`, `EdgeV1`
It's probably a simple fix in the function `nodetype`, which assumes that the `SchemaVersion` name always has 2 periods in it. `load_structvector` is currently used with these types.

evetion:
Fixes #637


SouthEndMusic (approved):
LGTM (y)

PR #697, opened 2023-10-25
Move from black to ruff formatter

Hofer-Julian:


ruff.toml
```diff
@@ -1,5 +1,17 @@
 select = ["D", "E", "F", "NPY", "PD", "C4", "I"]
-ignore = ["D1", "D202", "D205", "D400", "D404", "E501", "PD002", "PD901"]
+ignore = [
+    "D1",
+    "D202",
+    "D205",
+    "D206",
```
evetion, line 6: You added D206, D402 and E703? Might be good to place comments here about the why/what, or link to a page where we can immediately check these things.

evetion (approved):

ruff.toml
Hofer-Julian: I've added a link to the docs

visr (approved):

PR #698, opened 2023-10-25
use a TimeZones.jl fork for now

visr:
Alternative to #687, which wasn't working well with PackageCompiler due to long paths still, leading to broken builds.

This switches to a fork. A PR to merge this upstream has already been submitted: https://github.com/JuliaTime/TimeZones.jl/pull/450

Clearly this is temporary, but since we use Manifests this should work ok.


Hofer-Julian (approved):

Hofer-Julian:
It seems like this didn't solve the long path problem :(
https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_RibasimJl_BuildLibribasimWindows/2869733?buildTab=log&focusLine=0&linesState=593&logView=flowAware

visr:
Actually I think both #687 and this PR solved the long path problem, but old artifacts with long paths from previous builds are causing issues. Maybe it's time for clean checkout...

Hofer-Julian:
Will try that.

visr:
Looks like 72faf454c76c6eed735dcab892d10f3496cb6740 worked, right? Should we enable that for all Ribasim jobs?

Hofer-Julian:
Yeah, I'll enable it for everything now.

Issue #656, opened 2023-10-09
Modeller input for allocation


SouthEndMusic:
What should the input look like from the modeler to specify allocation? E.g. how are subnetworks, sources etc. specified?

SnippenE:
Subnetworks have an ID and 1 or more sources where source is an edge. Nodes have an optional reference to a subnetwork ID. We reserve groups for different purposes. 

We call this allocation sub-network

SouthEndMusic:
Priority of sources is at the level of a sub-network.

gijsber:
add optional column to node table called: allocationnetwork_id (long integer)

add optional column to edge table: allocationnetwork_id (long integer)

Issue #659, opened 2023-10-09
Generate allocation output file 


SouthEndMusic:
For each time the allocation procedure is called, add a row to the output file `allocation.arrow` with the following columns:

- `time`
- `subnetwork_id`
- `node_id` (of user)
- `priority`
- `demand` (of user at given time and priority)
- `allocated` (to user at given time and priority)

For full insight it is nice to optionally write the full sparse flow matrix per allocation solve in say `allocation_flow.arrow` with the following columns:

- `time`
- `subnetwork_id`
- `from_id`
- `to_id`
- `flow`

Here `from_id` and `to_id` are the node_ids whose equivalent in the max flow graph have an edge between them.

PR #699, opened 2023-10-25
Modeler allocation input

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/656.


SouthEndMusic:
For some reason I now get an `SQLiteException("out of memory")`, M'W'E:

```julia
using Ribasim
using SQLite

toml_path = "path/to/generated_testmodels/subnetwork/ribasim.toml"
cfg = Ribasim.Config(toml_path)
gpkg_path = Ribasim.input_path(cfg, cfg.database)
db = SQLite.DB(gpkg_path)
close(db)

Ribasim.generate_allocation_models!(p, db, cfg)
```

SouthEndMusic:
@reviewer: I suggest adding more tests after https://github.com/Deltares/Ribasim/issues/659.

core/src/bmi.jl
```diff
@@ -223,6 +229,13 @@ function create_callbacks(
     tabulated_rating_curve_cb = PresetTimeCallback(tstops, update_tabulated_rating_curve!)
     push!(callbacks, tabulated_rating_curve_cb)
 
+    if config.allocation.use_allocation
+        t_end = seconds_since(config.endtime, config.starttime)
+        tstops = 1:(config.allocation.timestep):t_end
```
visr, line 234: Do you intentionally start after 1 second rather than at the beginning?

core/src/bmi.jl
```diff
@@ -223,6 +229,13 @@ function create_callbacks(
...
+        allocation_cb = PresetTimeCallback(tstops, update_allocation!)
```
visr, line 235: For a fixed interval the PeriodicCallback is a better match: https://docs.sciml.ai/DiffEqCallbacks/stable/timed_callbacks/

visr (commented):
Nice work! Left a few comments.

core/src/bmi.jl
SouthEndMusic: No, this is probably due to me not being completely used to 1-based indexing (although I know this doesn't actually have anything to do with that) ^^

PR #701, opened 2023-10-26
Add osx-arm64 to pixi.toml

evetion:
Make pixi useable on osx-arm64, by ignoring quarto.


visr (approved):

visr:
I think for now we want to keep the `environment.yml` up to date as well? Could you do the same change there?

PR #713, opened 2023-10-31
Update to python 3.12, pandas 2.1.2

### LINKED ISSUE #645: Start supporting pandas >=2.1
At the moment, Ribasim Python fails with pandas >= 2.1.
We should adapt our code, so it starts working, and remove the version requirement.

Original issue description: https://github.com/Deltares/Ribasim/pull/586#issuecomment-1714490399
I assumed 2.1.1 would fix it for us: https://github.com/Deltares/Ribasim/pull/643#issuecomment-1745000090

### LINKED ISSUE #654: test python 3.12
It only came out a week or two ago, so our dependencies aren't there yet. pyogrio should have a 3.12 build in the next release after 0.6.0, I'm watching that for releases, then we can try again. We can also directly update our pixi.lock so people use it.

Came up in https://github.com/Deltares/Ribasim/pull/652#issuecomment-1752704778

visr:
Part of #654
Fixes #645

I ran `pixi add pandas=2.1.2 python=3.12 pyogrio=0.7.2` using pixi v0.5.0 so as not to update the lockfile format.

The new pyogrio release is needed to be able to use Python 3.12. We only update the Python version in the Pixi lockfile and on CI, this does not change our minimum supported Python version.

The pandas update to 2.1.2 fixes a regression present in 2.1.0 and 2.1.1, no further changes were needed in our code.

@Hofer-Julian I simply removed the `pandas < 2.1` restriction in the pyproject.toml, assuming that is fine since with the release of pandas 2.1.2 most people won't get 2.1.0 or 2.1.1 anymore, though I guess we could exclude multiple patch releases.


pixi.toml
```diff
@@ -87,18 +87,18 @@ geopandas = "*"
...
-pandas = "<2.1"
+pandas = "2.1.2.*"
 pandas-stubs = "*"
 pandera = "*"
 pip = "*"
 pre-commit = "*"
 pyarrow = "*"
 pydantic = "~=1.0"
-pyogrio = "*"
+pyogrio = "0.7.2.*"
```
Hofer-Julian, line 97: Why do we need to pin pyogrio?
It should just take the newest, no?

Hofer-Julian (commented):
Looks mostly good to me.
But tests are still failing

pixi.toml
visr: I just restored this. `pixi add` writes this by default, but good to keep this open and only write down real version requirements.

Hofer-Julian: Yeah, not a big fan of the default behaviour of `pixi add`.
It makes kind of sense for `cargo add`, but no too much in the conda ecosystem IMO.
They disagree with me though when I opened an issue for that :P

visr:
Locally tests pass for me, but that is using the pixi environment. On CI we don't use pixi yet but pip, leading to `ERROR: Failed building wheel for pyarrow`.

Hofer-Julian:
> Locally tests pass for me, but that is using the pixi environment. On CI we don't use pixi yet but pip, leading to `ERROR: Failed building wheel for pyarrow`.

pyarrow doesn't provide python 3.12 wheels yet.
Seems like we have to wait for that.

visr:
Ok support for that is merged upstream, we just need to wait for a pyarrow 14 release on PyPI.
https://pypi.org/project/pyarrow/#history

Hofer-Julian:
> Ok support for that is merged upstream, we just need to wait for a pyarrow 14 release on PyPI. https://pypi.org/project/pyarrow/#history

I'd be fine with only merging the pixi part and start testing 3.12 for Ribasim Python as soon as the pyarrow wheels for 3.12 are out.

visr:
Sounds good, done. I kept pre-commit CI on 3.12 as well since that is also working fine.

Hofer-Julian (approved):

PR #716, opened 2023-10-31
allow control of *Resistance and FractionalFlow

### LINKED ISSUE #718: Allow control of ManningResistance and LinearResistance
The error comes from our validation code working as intended. We need to allow control of ManningResistance in validation and add a test case to make sure it is and keeps working.

### LINKED ISSUE #722: Allow control of FractionalFlow
Added on Samenwerkingsdag 8 (31-10-2023) to secure fall back option to reproduce DM verdeelsleutels. Martijn has made a draft pull request for Hydrologic. They will make a simple testmodel that can be used in testbank Ribasim

visr:
Fix #718
Fix #722

Not sure if this works yet or if the validation is just too strict. Before merging this should be covered by a test model.


visr:
Copying commit message a52d4905be375e24f2112e1955d4a846cf7bbe27 for visibility:

> Fix FractionalFlow sum to 1 validation with control states

> The initial fractions come from whatever happens to be the last control state per node, and therefore can trigger false positive validation errors.

> Now we validate this per control state. To make that work I had to add the missing control state to the `control_mapping`. For simplicity I coalesce a missing `control_state` to the empty string. I wonder if that is the right approach though, since for e.g. name we default to the empty string rather than support missing. We should probably settle on either for all string columns.

@evetion what do you recommend for missing / default string columns; `missing` or `""`?

SouthEndMusic (approved):
Looks good (:

evetion:
> @evetion what do you recommend for missing / default string columns; `missing` or `""`?

For naming things, so which not influence the numerics, I'm fine with "". Otherwise I like the explicit handling of missing with coalesce.

Issue #252, opened 2023-05-30
Ribasim Python: stop exposing `Ribasim.Node` and node IDs to users


Hofer-Julian:
Thanks to @SouthEndMusic's work, users will soon have fewer chances to mess up their model input when using Ribasim Python.
https://github.com/Deltares/Ribasim/issues/245

What would be even better, if users wouldn't have to think about node IDs at all. They would create specific node types with the corresponding geometries attached. When new nodes are created that way, they would access a singleton allocator which returns new and unique node IDs.

For creating edges, the interfaces might change like this:

```diff
- from_id = np.array(
-    [1, 2], dtype=np.int64
-)
- to_id = np.array(
-    [2, 3], dtype=np.int64
-)

+ from_id = [basin.nodes[0], linear_resistance.nodes[0]]
+ to_id = [linear_resisitance.nodes[0], basin.nodes[1]]
```

visr:
#446 introduces a `get_node_ids_and_types` function that addresses a part of this. This probably needs to be removed/refined later. Right now it is guessing available tables for node types, `["static", "time", "condition"]`.

https://github.com/Deltares/Ribasim/pull/446/files#diff-1781035826cbdb5bd206a7c07398ee9d29ad5c34fa4312b6a0b07db7e154d3ab

visr:
Fixing this would probably also fix #212, since you would not pass the Node table anymore so there is no way it cannot match.

Issue #653, opened 2023-10-09
drop python 3.9 support


visr:
As of #652 we no longer test on Python 3.9, but our pyproject.toml still say it is supported. We should update this to require 3.10, and we can then directly make use of new language features from Python 3.10, like the new Optional syntax from https://github.com/Deltares/Ribasim/pull/351#discussion_r1235405489.

visr:
Also update to `datamodel-codegen --target-python-version=3.10`.

https://github.com/Deltares/Ribasim/pull/673/commits/50cf27f761ebb2cc77d1bda8993d594826987c4d

PR #731, opened 2023-11-02
Overhaul of Python architecture

### LINKED ISSUE #512: Refactor Python Model

evetion:
Fixes #512
Groundwork for #630 & #252

The code has become much simpler in some places (e.g. write_toml).

```python
from ribasim import Model

m = Model(filepath="generated_testmodels/basic/ribasim.toml")
m.database.node.df # Node table
m.basin.static.df # BasinStatc table
m.write("test")
```

### Some notes:
- The config.py file cannot be autogenerated anymore. The schemas still can, but I disabled it for now to be sure (some imports error).

### Changes:
I created new (parent) classes:
- BaseModel, from Pydantic, with our own config
- FileModel, like hydrolib-core (but now Pydantic v2), which can take a single filepath for initilization. Models who inherit require defining _load and _save, dealing with that filepath.
- NodeModel, a class for nodes (where `add` will be).
- TableModel, a class to read/write individual tables from/to gpkg/arrow.
- SpatialTableModel, inherits TableModel, but reads/writes spatial stuff.

I changed:
- the Model class to be a carbon copy of Config (which has been deleted), so it mirrors the toml.
- in turn this created a `database` NodeModel class (reflecting the field in the toml), with only Node and Edge underneath.
- the NodeModel classes Basin from their node_type version to the one in Config, and set the type of the underlying table with a TypeVar like so:
```python
class Terminal(NodeModel):
    static: TableModel[TerminalStaticSchema]
```

### Yet to do:
- [x] Update tests
- [x] Fix sort! rules
- [x] Delete node_types folder
- [x] Link schemas to their Pydantic class (TableModel[TerminalStaticSchema] => TerminalStatic)


visr:
> The config cannot be autogenerated without overwriting this code.

What do you mean by this?

evetion:
> > The config cannot be autogenerated without overwriting this code.
> 
> What do you mean by this?

We currently autogenerate all code in config.py, but I changed that code manually and use it extensively. Unless I write a new template for the codegenerator, I don't see autogeneration happening in the near future.

pixi.toml
```diff
@@ -101,12 +101,12 @@ pandera = "*"
...
 pre-commit = "*"
 pyarrow = "*"
-pydantic = "~=1.0"
+pydantic = ">=2"
 pyogrio = "*"
 pytest = "*"
 pytest-cov = "*"
 pytest-xdist = "*"
-python = ">=3.9"
+python = ">=3.9, <3.12"
```
Hofer-Julian, line 109: Do we still have a blocker for 3.12?

If yes, we should also adapt the `pyproject.toml`

python/ribasim/ribasim/config.py
```diff
@@ -3,17 +3,51 @@
...
 
 class Allocation(BaseModel):
     timestep: Optional[float] = None
     use_allocation: bool = False
 
 
+class Results(BaseModel):
+    basin: str = "results/basin.arrow"
+    flow: str = "results/flow.arrow"
+    control: str = "results/control.arrow"
```
Hofer-Julian, line 45: Could those be of type `pathlib.Path` instead?

python/ribasim/ribasim/config.py
```diff
@@ -3,17 +3,51 @@
...
+    outstate: Optional[str] = None
+    compression: str = "zstd"
```
Hofer-Julian, line 47: Could this be a `str enum` instead?

python/ribasim/ribasim/input_base.py
```diff
@@ -30,31 +55,144 @@ def exists(connection: Connection, name: str) -> bool:
     return result is not None
 
 
-class TableModel(BaseModel):
-    class Config:
-        validate_assignment = True
+PathOrStr = Union[Path, str]
```
Hofer-Julian, line 58:
```suggestion
PathOrStr = Path | str
```

Tbh, with the new syntax, I don't think there's a need for an alias anymore

python/ribasim/ribasim/input_base.py
```diff
@@ -76,19 +214,31 @@ def __repr__(self) -> str:
...
+            node_IDs.update(self.df["node_id"])
 
         return node_IDs
 
     @classmethod
-    def _layername(cls, field) -> str:
-        return f"{cls.get_input_type()}{delimiter}{field}"
+    def _load(cls, filepath: Path) -> Dict[str, Any]:
+        db = context_file_loading.get().get("database")
+        if filepath is not None and Path(filepath).exists():
```
Hofer-Julian, line 225:
```suggestion
        if filepath is not None and filepath.exists():
```

python/ribasim/ribasim/model.py
```diff
@@ -1,43 +1,84 @@
...
+    Solver,
+    TabulatedRatingCurve,
+    Terminal,
+    User,
+)
+from ribasim.geometry.edge import Edge, EdgeStaticSchema
+from ribasim.geometry.node import Node, NodeStaticSchema
 
 # Do not import from ribasim namespace: will create import errors.
 # E.g. not: from ribasim import Basin
```
Hofer-Julian, line 38: I guess you have a circular import there.
Might be good to solve that.

python/ribasim/ribasim/model.py
```diff
@@ -80,42 +121,39 @@ class Model(BaseModel):
...
-    @classmethod
-    def fields(cls):
-        """Return the names of the fields contained in the Model."""
-        return cls.__fields__.keys()
+    update_timestep: float = 86400
+    relative_dir: str = "."
+    input_dir: str = "."
+    results_dir: str = "."
+
+    database: Database = Database()
```
Hofer-Julian, line 138:
```suggestion
    database: Database = Database()
```

Why is `= Database()` necessary? Isn't that the default behavior?
Or did that change with Pydantic v2?

python/ribasim/ribasim/model.py
```diff
@@ -80,42 +121,39 @@ class Model(BaseModel):
...
+    allocation: Optional[Allocation] = Allocation()
+    results: Optional[Results] = Results()
+    solver: Optional[Solver] = Solver()
+    logging: Optional[Logging] = Logging()
```
Hofer-Julian, line 142:
```suggestion
    allocation: Allocation | None = Allocation()
    results: Results | None = Results()
    solver: Solver | None = Solver()
    logging: Logging | None = Logging()
```

I am not sure how the `None` case would be triggered here

python/ribasim/ribasim/model.py
```diff
@@ -1,43 +1,84 @@
...
+    FileModel,
+    NodeModel,
+    TableModel,
+    context_file_loading,
+)
 from ribasim.types import FilePath
 
 
-class Model(BaseModel):
+class Database(NodeModel, FileModel):
```
Hofer-Julian, line 48: I couldn't find anything whether inheriting from multiple pydantic models is supported.
Might be good to double check: https://docs.pydantic.dev/latest/help_with_pydantic/

python/ribasim/ribasim/config.py
```diff
@@ -3,17 +3,51 @@
 
 from __future__ import annotations
```
Hofer-Julian, line 4: I assume this import isn't needed anymore?

Hofer-Julian (commented):
Looks good @evetion.

To later review the usage, tests or even better the example jupyter notebook would be helpful.

pixi.toml
visr: For context, on main our pixi.lock gives us 3.12 and that seems to be working fine: https://github.com/Deltares/Ribasim/pull/713

python/ribasim/ribasim/model.py
evetion: It was never default behaviour, but this is the way to set the default, if none is provided. If it isn't set (only var: type), it is required to be provided on init. While this is the case if we open a toml, we also need to be able to start from scratch.

Hofer-Julian: I see, this is for the interactive use case.
If I understood you correctly the following makes it optional

```python
database: Database | None = Database()
```

But what is the difference between the following?

```python
database: Database
```

and 

```python
database: Database = Database()
```

evetion: So the `= something` provides a default, which is used if nothing is provided on init.

So `thing()` will fail in the first case with database not provided. In the second case, `thing()` will work, and the database field is set with the default. In your optional case, `thing(database=None)` would be a valid option as well.

python/ribasim/ribasim/config.py
evetion: I'd have to check, might be for `-> "SomeClass":` underneath the SomeClass itself.

Hofer-Julian:
I've added https://github.com/Deltares/Ribasim/pull/731/commits/507d925362b9acc1451db7a331b0881c98f6d5dd to your branch so similar type hint lints are caught by ruff directly.

python/ribasim/ribasim/config.py
evetion: I made it to an Enum (I did read about that Literal is faster, but that doesn't matter here).

evetion:
> I've added [507d925](https://github.com/Deltares/Ribasim/commit/507d925362b9acc1451db7a331b0881c98f6d5dd) to your branch so similar type hint lints are caught by ruff directly.

I had to revert it for now, as the autogenerated schemas still use the old notation. :/ Will see if I can fix that at a later stage.

visr:
> the autogenerated schemas still use the old notation

Is that fixed by `target-python-version`? `datamodel-codegen --target-python-version=3.10`
Mentioned in https://github.com/Deltares/Ribasim/issues/653#issuecomment-1766130686.

evetion:
> > the autogenerated schemas still use the old notation
> 
> Is that fixed by `target-python-version`? `datamodel-codegen --target-python-version=3.10` Mentioned in [#653 (comment)](https://github.com/Deltares/Ribasim/issues/653#issuecomment-1766130686).

Nope, but it was `--use-union-operator` 🎉

pixi.toml
```diff
@@ -77,9 +77,7 @@ tests = { depends_on = ["lint", "test-ribasim-python", "test-ribasim-core"] }
 generate-schema = { cmd = "julia --project=docs docs/gen_schema.jl", depends_on = [
     "instantiate-julia",
 ] }
-generate-python = """\
-    datamodel-codegen --use-title-as-name --use-double-quotes --disable-timestamp --use-default --strict-nullable --input-file-type=jsonschema --input docs/schema/root.schema.json --output python/ribasim/ribasim/models.py && \
-    datamodel-codegen --use-title-as-name --use-double-quotes --disable-timestamp --use-default --strict-nullable --input-file-type=jsonschema --input docs/schema/Config.schema.json --output python/ribasim/ribasim/config.py"""
+generate-python = "" # "datamodel-codegen --use-union-operator --use-title-as-name --use-double-quotes --disable-timestamp --use-default --strict-nullable --input-file-type=jsonschema --input docs/schema/root.schema.json --output python/ribasim/ribasim/models.py"
```
Hofer-Julian, line 80: Is that # a typo?
If not, can you please explain it?

evetion: So the generation of the config.py is removed (I substantially modified these classes). I disabled the generation of the models.py, because using it fails with an odd error message (`E   TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'`) when running ribasim-python. This comes from the use of `--use-union-operator`, resulting in `Something | None = None`, which doesn't seem to work. `Optional[Something] = None` does seem to work, so I'm falling back on that one.

Hofer-Julian:
I've added https://github.com/Deltares/Ribasim/pull/731/commits/be766f9797516b05fee86fa579ce854cf447121e as an example how to simplify the code

python/ribasim/ribasim/input_base.py
```diff
@@ -30,65 +60,181 @@ def exists(connection: Connection, name: str) -> bool:
...
+    #                 colnames = textwrap.indent(
+    #                     textwrap.fill(colnames, width=50), prefix="    "
+    #                 )
+    #                 entry = f"{field}: DataFrame(rows={len(attr)})\n{colnames}"
+    #             else:
+    #                 entry = f"{field}: DataFrame(rows={len(attr)}) {colnames}"
+    #         else:
+    #             entry = f"{field}: {attr}"
+    #         content.append(textwrap.indent(entry, prefix="   "))
+    #     return "\n".join(content)
```
Hofer-Julian, line 197: I assume this can be deleted?

Hofer-Julian (approved):
The changeset is naturally massive so I didn't check every detail.
But overall it looks good

python/ribasim/ribasim/input_base.py
evetion: Yeah. I'm not sure how we want our models to be represented, but it seemed this was some effort, so I commented it out for later adaption.

python/ribasim/tests/test_model.py
```diff
@@ -44,42 +45,48 @@ def test_invalid_node_id(basic):
     model = basic
 
     # Add entry with invalid node ID
-    model.pump.static = model.pump.static._append(
+    df = model.pump.static.df._append(
         {"flow_rate": 1, "node_id": -1, "remarks": "", "active": True},
         ignore_index=True,
     )
+    # Currently can't handle mixed NaN and None in a DataFrame
```
visr, line 52: Is this a pandas limitation? We could benefit from a clear guide on how we deal with / represent missing data for different data types on both Julia, Python, Arrow and GeoPackage sides.

See also [this comment](https://github.com/Deltares/Ribasim/pull/716#issuecomment-1792749731): @evetion what do you recommend for missing / default string columns; missing or ""?

python/ribasim_testmodels/ribasim_testmodels/trivial.py
```diff
@@ -97,8 +97,10 @@ def trivial_model() -> ribasim.Model:
     )
 
     model = ribasim.Model(
-        node=node,
-        edge=edge,
+        database=ribasim.Database(
+            node=node,
+            edge=edge,
+        ),
```
visr, line 103: I like that these are now together in a single class, and understand you named it `Database` since these tables always have to be in the database. But wouldn't a name like `Network` better represent what it really is? If all tables are stored in the database, users may expect to find all tables in `ribasim.Database`.

visr (approved):
Looks good to me! Left some comments, but they could also be moved out to separate issues or follow up PRs if you prefer.

python/ribasim_testmodels/ribasim_testmodels/trivial.py
evetion: The field is database because it is named so in the toml. I thought about naming the class Topology, but Network is a good alternative. This can be renamed in the follow up PR, when we actually start using said class more extensively.

python/ribasim/tests/test_model.py
evetion: Pandas is fine, it's just that with adding new columns, mixing None (object) and NaN (float) doesn't play nice with pandera. I might've fixed that in the last commit. I will write a guide in the subsequent PR on how to handle missing values in Python. (I think it's pretty clear in Julia & geopackage).

PR #734, opened 2023-11-03
Work around quarto in a cross-platform way

Hofer-Julian:
See https://quarto.org/docs/computations/python.html#installation


SouthEndMusic (approved):
Looks good (as far as I can tell)

PR #736, opened 2023-11-03
set FractionalFlow inflow equal to outflow

### LINKED ISSUE #733: Flow discontinuity over FractionalFlow
Right now the inflow and outflow of a FractionalFlow node is not the same. Everything is working as intended, but for instance for water quality coupling, it would be better to have both the same, so the water balance adds up.

image

It's a bit hacky, but we could just update the inflow after calculating the fractional outflow from it.

https://github.com/Deltares/Ribasim/blob/v2023.10.0/core/src/solve.jl#L990

visr:
Fixes #733


SouthEndMusic (approved):
Looks good to me

PR #738, opened 2023-11-04
Switch to Julia 1.10-rc1

visr:
Release notes here: https://github.com/JuliaLang/julia/blob/v1.10.0-rc1/NEWS.md

Since at this point we don't care about supporting multiple julia versions, I updated everything to 1.10.

Marking as a draft since:
- [x] Julia 1.10-rc1 needs to be installed on TeamCity
- [x] `%env.JULIA_1_10%` needs to be added on TeamCity to point to that install (this PR already attempts to use `JULIA_1_10`)

Main reasons I'm excited about 1.10 for Ribasim; mostly for developer experience:
- Less latency
- Better syntax errors
- Shorter stacktraces

If people want we can wait for the final release, though its good to start kicking the 1.10 tires at least.


visr:
Thanks! We can wait for now, and put this PR on hold.
I reverted the first commit, which updated all Manifests: 546b636fa0fc77da3106b135c571d270dbe8372b
Julia 1.10 also accepts 1.9 manifests, just with a warning, and that commit gathers conflicts easily.

visr:
@Hofer-Julian it looks like TeamCity is currently using Julia 1.9.0 instead of the latest 1.9.3: https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_RibasimJl_BuildLibribasimWindows/2928560?buildTab=log&focusLine=1294&linesState=600&logView=flowAware

We are now running in #740 into issues that are fixed in later patch releases, blocking progress here. Could we prepare the latest stable release and ask to trigger an Ansible run manually?

Hofer-Julian:
Willem is looking into running ansible to get the following on TeamCity
-pixi 0.6.0
- julia 1.9.3
- julia 1.10.rc

PR #739, opened 2023-11-04
Add compat entries for stdlibs

visr:
https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958

We aren't registered in General yet, but might as well do this already.


Hofer-Julian (approved):
Great to see that stdlibs are now decoupled from Julia releases

Issue #621, opened 2023-09-25
renew Julia API docs


visr:
#470 added the Julia core API reference to the docs, which is hosted at https://deltares.github.io/Ribasim/build/.

This uses the [DocumenterMarkdown](https://github.com/JuliaDocs/DocumenterMarkdown.jl) backend which is not actively maintained, meaning we are stuck for now on the old Documenter 0.27.

That isn't a big issue right now since we pick up the rendered markdown with Quarto anyway. But it's good to keep an eye out for alternatives.

A [Discourse thread](https://discourse.julialang.org/t/using-quarto-to-create-package-documentation/102698) mentions two alternatives:
- https://github.com/medyan-dev/QMDDocTools.jl
- https://github.com/vituri/Quartomenter.jl

Or we could try to update DocumenterMarkdown, not sure how much work that would be since Documenter changed a lot, if I understand correctly it is now using https://github.com/JuliaDocs/MarkdownAST.jl as an internal representation.

Perhaps some https://github.com/kdheepak/Pandoc.jl related solution could work, since Quarto builds on pandoc.

visr:
I suggest to remove the Julia API docs altogether from the online docs. They aren't very useful, and look bad.

PR #740, opened 2023-11-05
Share manifests

visr:
One frequent reason that `libribasim` and `ribasim_cli` builds break is because we add a dependency to the core, and we forget to resolve the Manifests of `libribasim` and `ribasim_cli`. (e.g. #484).

This adds an entry like `manifest = "../Manifest.toml"` to the `core`, `libribasim` and `ribasim_cli` Project.toml, which all point to the new `root` Manifest. This option exists to make it easier to work with multiple related projects in one monorepo.

As a follow-up we could consider also doing this for the remaining two projects, `docs` and `create_binaries`. `create_binaries` doesn't depend on Ribasim, so there is not as much benefit. `docs` does, currently not via Project.toml, but via adding it to the LOAD_PATH. It would be cleaner to also use a shared Manifest there. One thing to check if the unmaintained docs dependencies don't hold back Ribasim dependencies, see #621.

This is on top of #738, but could be separated out if needed.

Also simplifies the docs, and has a not-directly-related commit 16ce0cf3702964251d1824c58f297dd5a8f21dbc that runs CompatHelper for all projects, not just core.


visr:
Blocked since this seems to be hitting an issue on Julia 1.9.0 that is fixed in the latest 1.9 release, see https://github.com/Deltares/Ribasim/pull/738#issuecomment-1798548518.

Hofer-Julian (approved):

PR #742, opened 2023-11-06
Fix test-ribasim-core by adding `using Pkg;`

deltamarnix:
Otherwise you would receive an error during `pixi run test-ribasim-core`


Hofer-Julian (approved):
Nice find!

Issue #726, opened 2023-11-02
Fairness of allocation in times of scarcity


SouthEndMusic:
Now that allocation is back to an LP solve per priority (https://github.com/Deltares/Ribasim/pull/699), fairness in allocation has to be rethought. 

On way to do it is to introduce a new variable $M$ to maximize, so the objective function becomes simply

$$
\max M
$$

with either constraints

$$
M \le F_{\hat{i}\hat{j}} \quad \forall (\hat{i},\hat{j}) \in \hat{E}_S : \hat{j} \in \hat{U}_S,
$$

so the objective is to maximize the minimum flow to a user, or

$$
M \le \frac{F_{\hat{i}\hat{j}}}{d^p_j(t)} \quad \forall (\hat{i},\hat{j}) \in \hat{E}_S : \hat{j} \in \hat{U}_S,
$$

so the objective is to maximize the minimum demand fraction allocated to the users.

A problem with this is that this method doesn't 'care' about water that cannot be distributed fairly, so not all available water is distributed.

_This isssue has been scheduled to be discussed with members of the RTC-Tools team._

SouthEndMusic:
This issue has been discussed with the RTC-Tools team, concluding that the objective for allocation flows for a given priority $p$ can have the following forms.

1. All deviations from the demand flow have the same weight:

$$
\min \sum_{(\hat{i},\hat{j}) \in \hat{E}\_S : \hat{j} \in \hat{U}_S} \left|F\_{\hat{i}\hat{j}} - d^p_j(t) \right|
$$

2. The same as 1 but quadratic:

$$
\min \sum_{(\hat{i},\hat{j}) \in \hat{E}\_S : \hat{j} \in \hat{U}_S} \left(F\_{\hat{i}\hat{j}} - d^p_j(t) \right)^2
$$

3. The same as 1 but demand fraction deviations are minimized:

$$
\min \sum_{(\hat{i},\hat{j}) \in \hat{E}\_S : \hat{j} \in \hat{U}_S} \left|1 - \frac{F\_{\hat{i}\hat{j}}}{d^p_j(t)} \right|
$$

4. The same as 2 but scaled by the demands:

$$
\min \sum_{(\hat{i},\hat{j}) \in \hat{E}\_S : \hat{j} \in \hat{U}_S} \frac{\left(F\_{\hat{i}\hat{j}} - d^p_j(t) \right)^2}{d^p_j(t)}
$$

If desired we could support custom weights for these terms as input from the modeller.

PR #743, opened 2023-11-06
Allocation output file

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/659.

notes:
- Allocation not done  `t = 0` as no flows for sources are available yet.
- The saved abstraction is now the last abstraction before the allocation update, should be the average abstraction since the last allocation solve #754


SouthEndMusic:
Here is an example of the behavior of allocation:

simple_subnetwork
allocation_result

The 2 users are completely equivalent, only the demand on one of them is increasing over time. Currently the goal is to maximize the sum of the flows to the users, so the optimization does not care about the composition of that sum. Fair distribution is next (https://github.com/Deltares/Ribasim/issues/726).

SouthEndMusic:
@visr we have to decide what users should do when they are not in a subnetwork or allocation is not activated in the TOML. Should they default to demand = allocated?

visr:
> Should they default to demand = allocated?

Yes I think that makes sense here.

SouthEndMusic:
Now I'm actually done with this PR 😋.

core/test/allocation.jl
```diff
@@ -22,7 +22,54 @@ using PreallocationTools: get_tmp
...
+    @test allocated[2] ≈ [4.0, 0.0]
     @test allocated[3] ≈ [0.0, 0.0]
 end
+
+@testset "Simulation with allocation" begin
+    toml_path =
+        normpath(@__DIR__, "../../generated_testmodels/simple_subnetwork/ribasim.toml")
+    @test ispath(toml_path)
+    model = Ribasim.run(toml_path)
+    record = model.integrator.p.user.record
```
visr, line 35: Since this is a NamedTuple of vectors, this could be converted to a DataFrame directly in the tests. This could make the tests easier to understand.

core/src/solve.jl
```diff
@@ -449,6 +454,27 @@ struct User <: AbstractParameterNode
...
+        Tuple{
+            Vector{Float64},
+            Vector{Int},
+            Vector{Int},
+            Vector{Int},
+            Vector{Float64},
+            Vector{Float64},
+            Vector{Float64},
+        },
+    }
```
visr, line 477: I think here the NamedTuple macro may lead to more intuitive syntax: https://docs.julialang.org/en/v1/base/base/#Base.@NamedTuple

core/src/create.jl
```diff
@@ -780,6 +781,18 @@ function User(db::DB, config::Config)::User
...
     allocated = [zeros(length(priorities)) for id in node_ids]
 
+    record = (
+        time = Vector{Float64}(),
+        allocation_network_id = Vector{Int}(),
+        user_node_id = Vector{Int}(),
+        priority = Vector{Int}(),
+        demand = Vector{Float64}(),
+        allocated = Vector{Float64}(),
+        abstracted = Vector{Float64}(),
```
visr, line 791:
```suggestion
        time = Float64[],
        allocation_network_id = Int[],
        user_node_id = Int[],
        priority = Int[],
        demand = Float64[],
        allocated = Float64[],
        abstracted = Float64[],
```

core/src/solve.jl
```diff
@@ -787,12 +814,23 @@ function formulate_flow!(
...
         end
 
-        # For now allocated = demand
-        for priority in eachindex(allocated[i])
-            allocated[i][priority] = demand[i][priority](t)
-        end
+        q = 0.0
 
-        q = sum(allocated[i])
+        if allocation_optimized[i]
```
visr, line 819: I wonder if introducing `allocation_optimized` is worth it. Would it also be sufficient to initialize `allocated` to `Inf`, and then take the minimum of that and the demand in either case?

core/src/config.jl
```diff
@@ -112,6 +112,7 @@ end
     basin::String = "results/basin.arrow"
     flow::String = "results/flow.arrow"
     control::String = "results/control.arrow"
+    allocation::String = "results/allocation.arrow"
```
visr, line 115: Looks like this changes the schema, so needs a `pixi run codegen`.

core/src/allocation.jl
```diff
@@ -778,6 +809,7 @@ and flows, solve the allocation problem and assign the results to the users.
 function allocate!(p::Parameters, allocation_model::AllocationModel, t::Float64)::Nothing
     (; user) = p
     (; problem) = allocation_model
+    (; priorities, record) = user
```
visr, line 812: Looks like `record` is not used.

core/test/allocation.jl
```diff
@@ -29,47 +30,56 @@ end
...
+            collect(range(1.0e-3, 2.0e-3, n_datapoints))[2:end];
             rtol = 0.01,
         ),
     )
+
+    allocation_output_path = normpath(
+        @__DIR__,
+        "../../generated_testmodels/minimal_subnetwork/results/allocation.arrow",
+    )
+    @test ispath(allocation_output_path)
```
visr, line 84:
```suggestion
    @test isfile(allocation_output_path)
```
`ispath` is for either files or folders, so this is a bit more specific.

visr (commented):
Thanks for the update, the code looks great. One thing I missed earlier, I think we should document this on https://deltares.github.io/Ribasim/core/usage.html. Perhaps under the User section, just like you did for DiscreteControl results.

visr (approved):

PR #763, opened 2023-11-09
update to ruff 1.5, add notebook settings

visr:
This follows the suggestions from https://astral.sh/blog/ruff-v0.1.5

Also moves `settings_template.json` to `settings.json` as suggested by @deltamarnix. The only part I'm not sure about if it is wise to remove the file from `.gitignore`. If local changes keep appearing there, we can revert that part.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -146,9 +146,6 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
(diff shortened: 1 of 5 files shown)


Hofer-Julian (approved):
Looks good!

PR #770, opened 2023-11-10
C4 diagrams

deltamarnix:
Add diagrams for C4 model showing the different components for ribasim, the preprocessing API, and the QGIS plugin.

We could go even further and make a detailed diagram of the different modules that are connected to each other. But these diagrams already give an overview of the components that are involved in the system.

### DIFF
--- a/docs/assets/c4_system.mmd
+++ b/docs/assets/c4_system.mmd
@@ -1,4 +1,3 @@
-```{mermaid}
 flowchart TB
 modeler([Modeler]):::user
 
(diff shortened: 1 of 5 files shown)


docs/assets/c4_system.mmd
```diff
@@ -0,0 +1,32 @@
...
+    QGIS-->qgisPlugin
+end
+modeler-->|prepare model|qgisBoundary
+
+model[("input model data\n[toml + geopackage + arrow]")]:::system
+qgisPlugin-->|read/write|model
+api-->|read/write|model
+ribasim-->|simulate|model
+
+output[("simulation output\n[arrow]")]:::system
```
visr, line 22:
```suggestion
output[("simulation results\n[arrow]")]:::system
```

visr (approved):

PR #772, opened 2023-11-10
Fix getindex of local `truth_values` with global `condition_idx`

visr:
This fixes a
```
BoundsError: attempt to access 2-element Vector{String} at index [4]
```

A test model that doesn't work without this is here (I also have code to generate this):

[2Basins_2Pumps_1_Boezem_TargetLevelBasin1.zip](https://github.com/Deltares/Ribasim/files/13321623/2Basins_2Pumps_1_Boezem_TargetLevelBasin1.zip)

Though perhaps it's better to see if we can easily update one of our own test models to trigger this path.

- [x] add test

Reported by @harm-nomden-sweco.


visr:
Added to #716.

PR #773, opened 2023-11-10
lower reltol from 1e-3 to 1e-5

visr:
@harm-nomden-sweco shared a case where a pump is emptying a basin with an inflow smaller than the pump capacity. We know this case leads to small timesteps, but it appeared this also led to large errors in the water balance. This lowering of the reltol is enough to fix that issue. Since such a case can be common, it is good to have a safe default.

This also adds `using Test` to all test files. And `test/time.jl` needed some refactoring since there were a few sub-millisecond timesteps, whereas we round our time to milliseconds.


SouthEndMusic (approved):
LGTM

PR #776, opened 2023-11-13
Ignore errors when cleaning generated testmodels

deltamarnix:
Otherwise it would give an error halfway if you still have a terminal open.


visr (approved):
Thanks!

PR #780, opened 2023-11-13
add navbar to Home

visr:
This goes from

image

to

image

The Ribasim button to the left of "Home" also acts as a home button, but perhaps this is not obvious to users. The default template at least includes this "Home" button, so I think it is best to have it as well.


Hofer-Julian:
I think it's really common that the logo leads to the home page. Did you encounter somebody who struggled with that?

visr:
Not really, no. I did this mainly since it follows the Quarto template. But happy to close, it looks less good anyway.

PR #781, opened 2023-11-14
Support toml in QGIS plugin

### LINKED ISSUE #629: Load TOML, not GeoPackage, in QGIS plugin
Right now QGIS loads a model from a GeoPackage, but really we should load the TOML and find the GeoPackage from there. When we do this, we can directly read the TOML, check if the output exists, and if it does, associate the node and edge output automatically on loading the model. This saves quite some clicks when loading a model for visualization.

Originally mentioned in, and needed for, #318.

deltamarnix:
Fixes part of #629 (the results are not automatically associated yet).

The Open button now supports toml instead of gpkg. The toml file checks for the "database" entry and loads that file as it used to do.
The New button now also creates a toml file with the database entry pointing to the gpkg that is also created like before. And the start and end dates are 01-01-2020 until 01-01-2030.

An option for another PR would be to add the start and end dates as fields in the main widget so that they can be adjusted. But that would mean that we need to keep the open toml file in memory and set the start and end date, and flush the file after those changes.

Also, error handling is not fancy yet. If the "database" entry is missing from the toml file, you will get a python KeyError

The folder `qgis` was renamed to `ribasim_qgis` for pylance to handle the imports better.


visr:
From CI: `ModuleNotFoundError: No module named 'toml'`. We probably need to check a bit in which cases (which QGIS version, setup method) the `toml` package is included with the QGIS bundled Python.

Hofer-Julian:
> From CI: `ModuleNotFoundError: No module named 'toml'`. We probably need to check a bit in which cases (which QGIS version, setup method) the `toml` package is included with the QGIS bundled Python.

`toml` shouldn't be used in general.
Starting from Python 3.11 you can use `tomllib` from the stdlib, for older python versions we have to depend on `tomli`.

visr:
Yeah but we're bound to what QGIS' Python comes with here. Worst case we vendor something.

Hofer-Julian:
> Even the latest release QGIS 3.34 that just came out has Python 3.9 still, so waiting for Python 3.11 will take to long. Since we just need to read the top level `database` key we think it is easier to do our own regex parsing rather than vendoring a TOML package (even though tomllib is only 1000 lines).

Or we vendor the relevant part of tomllib?
https://github.com/python/cpython/tree/3.12/Lib/tomllib

visr:
That is an option, but since we really only require a single top level key a regex was deemed easier.

visr:
(until we can switch to tomllib or ribasim-python at least)

visr:
Oh wait as @deltamarnix just mentioned we also will need to parse the results section, so vendoring might be better after all. @Hofer-Julian that's just copying the files you linked into a tomllib folder in the plugin right? Or could that break other plugins, so that we should name it `ribasim_tomllib` or something like that?

Hofer-Julian:
> Oh wait as @deltamarnix just mentioned we also will need to parse the results section, so vendoring might be better after all. @Hofer-Julian that's just copying the files you linked into a tomllib folder in the plugin right? Or could that break other plugins, so that we should name it `ribasim_tomllib` or something like that?

In my opinion, copying the folder. has the lowest risk of breakage and the lowest maintenance costs.

Hofer-Julian:
I don't see how it would break other plugins. So no renaming should be necessary
I assume every plugin is in its own folder?

visr:
Yes, ok good to know.

visr:
One thing to look out for is if tomllib uses any python features for 3.10 or later. If it does perhaps we should vendor tomli instead, which is Python 3.7+: https://github.com/hukkin/tomli/blob/master/pyproject.toml

PR #786, opened 2023-11-14
rename ribasim.Database to ribasim.Network

visr:
Follow-up as mentioned in https://github.com/Deltares/Ribasim/pull/731#discussion_r1392641169

Not passing the tests yet, as I don't understand how to have a different TOML key `geopackage` and field `network`. Writing to TOML is 'fixed' by bbb2982e33449b4506801894593319a091e5de5f, but tests are still failing. @evetion could you help out?


evetion:
I'm not sure if I'm in favour of this change, as the attributes of the model *are* (not just mimic) the toml file. I understand renaming only the class (which I thought you wanted) is confusing (database=Network()), so I'm not sure about it either.

Btw, this is probably also why the tests fail, as the database is loaded by the identical toml field, and there's no network key provided.

FWIW with the planned API, users should not touch model.database/network at all.

visr:
Thanks for 47234d174722c2b722f473a25a042800327b0bc0, that seems like a good way of dealing with this.

> the attributes of the model are (not just mimic) the toml file

Yeah I understand the hesitation, though overall I'm still in favor of this change. I guess the issue here is that the TOML file and the data model don't really match. In the data model the network is a required part of the database. In the TOML `database` is just the path to the database. So the TOML file doesn't need the concept of network at all, since it is part of the database. I feel that it would be confusing to refer to the database as either the GeoPackage or the network, hence this change.

> FWIW with the planned API, users should not touch model.database/network at all.

Yes I see that for the "add API" for model building, but not in general. For many operations you only need the network, so I imagine other API that can build on this. For instance converting the network to UGrid, networkx, or so.

Issue #784, opened 2023-11-14
Combine flow and control graphs using `MetaGraphsNext.jl`


SouthEndMusic:
Introduce a single graph `graph` for flow and control where the edge type is specified in the edge metadata. Here node and edge ID structs can be introduced. node type (as a replacement of lookup) and edge type is specified with `EnumX.jl`.

SnippenE:
Obsolete

PR #789, opened 2023-11-15
Use a single `MetaGraph` that combines `graph_flow` and `graph_control`

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/784.


SouthEndMusic:
Locallly a test from `Expand logic_mapping` fails, but not when run in the REPL 🤔

core/src/allocation.jl
```diff
@@ -1,30 +1,47 @@
+"""
+Temporary solution before 'edge_ids_flow_inv' is no longer needed.
```
SouthEndMusic, line 2: This is a temporary fix, will be removed after refactoring allocation with the metagraph

core/src/utils.jl
```diff
@@ -12,25 +12,73 @@ function valid_edge_types(db::DB)::Bool
...
+            error("Invalid edge type $edge_type.")
         end
+        graph[NodeID(from_node_id), NodeID(to_node_id)] =
+            EdgeMetadata(EdgeID(fid), edge_type)
     end
-    return graph, edge_ids, edge_connection_types
+    return graph
+end
+
+function inneighbor_labels_type(
```
SouthEndMusic, line 45: This and the following 2 functions might be badly named. Also, they might be allocating and thus slow.

core/src/utils.jl
```diff
@@ -77,8 +125,8 @@ function get_storage_from_level(basin::Basin, state_idx::Int, level::Float64)::F
     bottom = first(level_discrete)
 
     if level < bottom
-        node_id = basin.node_id[state_idx]
-        @error "The level $level of basin #$node_id is lower than the bottom of this basin $bottom."
+        node_id = basin.node_id[NodeID(state_idx)]
```
SouthEndMusic, line 128: This seems like an ugly fix

visr:
Ha, I messed up here, and reviewed #807 instead. I was wondering why it was so big haha. Though given that, maybe we should just merge #807 when the comments are addressed and close this.

visr:
Closing in favor of #807.

PR #794, opened 2023-11-16
Prefix extra columns with meta_.

### LINKED ISSUE #576: Allow extra table columns in Python builder

evetion:
Fixes #576

Seemingly required, but seemingly unrelated fixes:
- The unused, but required RootModel is generated automatically, but had the wrong casing in fieldnames, leading to errors.
- Utils required a Node import, leading to circular import errors when using utils elsewhere. Moved the functions to the Node class where they belong.

### DIFF
--- a/python/ribasim/ribasim/models.py
+++ b/python/ribasim/ribasim/models.py
@@ -5,7 +5,9 @@ from __future__ import annotations
 
 from datetime import datetime
 
-from pydantic import BaseModel, Field
+from pydantic import Field
+
+from ribasim.input_base import BaseModel
 
 
 class BasinProfile(BaseModel):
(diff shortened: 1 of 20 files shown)


python/ribasim/ribasim/models.py
```diff
@@ -218,27 +220,27 @@ class UserTime(BaseModel):
...
-    OutletStatic: OutletStatic | None
-    PidControlStatic: PidControlStatic | None
-    PidControlTime: PidControlTime | None
-    PumpStatic: PumpStatic | None
-    TabulatedRatingCurveStatic: TabulatedRatingCurveStatic | None
-    TabulatedRatingCurveTime: TabulatedRatingCurveTime | None
-    TerminalStatic: TerminalStatic | None
-    UserStatic: UserStatic | None
-    UserTime: UserTime | None
+    basinprofile: BasinProfile | None = None
```
visr, line 223: Shouldn't this be `basin_profile`?

visr:
Does the prefix now get added for all dataframes?

python/ribasim/ribasim/models.py
evetion: It's unused, the whole Root is there just to generate all child models from this single root definition.json.

evetion:
> Does the prefix now get added for all dataframes?

Validation always runs, also on assignment, so yes. The only exception are the node and edge tables, which have no pydantic record (cause geometry).

visr (approved):

PR #798, opened 2023-11-16
Support arrow input.

### LINKED ISSUE #170: Support Arrow input tables
It should work already in the core, but since there are no tests let's assume it's not working. Only the output tables are written to Arrow.

The example models in `conftest.py` currently write all input tables to the GeoPackage file. This is what `model.write(path)` does. Usually Arrow input tables are mainly interesting for large (time-dependent) tables, but we need to make sure it is working in CI.

I see the `class InputMixin` has `_write_arrow` and `_write_geopackage` that are both not used. Instead `write` loops over the underlying tables and writes them one-by-one to GeoPackage. Perhaps we need something like an attribute in all tables (e.g. classes that inherit from `pa.SchemaModel`, that determines if it should be written to GeoPackage (the default) or Arrow.

A minimum first test is added here, which contains a utility function to convert GeoPackage tables to Arrow.
https://github.com/Deltares/Ribasim/pull/187/files#diff-dd42657be5ede608d919dd58408933bac74a4661098f89d8b48dc53adf0478d7

evetion:
Fixes #170

What do we support in terms of input/relative dir as put in the settings? Currently, this ignores the setting completely.

### DIFF
--- a/core/test/docs.toml
+++ b/core/test/docs.toml
@@ -6,7 +6,7 @@ endtime = 2021-01-01   # required
 # input files
 database = "database.gpkg" # required
 input_dir = "."            # optional, default "."
-results_dir = "."          # optional, default "."
+results_dir = "results"    # optional, default "."
 
 # Specific tables can also go into Arrow files rather than the database.
 # For large tables this can benefit from better compressed file sizes.
--- a/core/test/docs.toml
+++ b/core/test/docs.toml
@@ -5,6 +5,8 @@ endtime = 2021-01-01   # required
 
 # input files
 database = "database.gpkg" # required
+input_dir = "."            # optional, default "."
+results_dir = "."          # optional, default "."
 
 # Specific tables can also go into Arrow files rather than the database.
 # For large tables this can benefit from better compressed file sizes.
@@ -37,9 +39,9 @@ timing = false     # optional, whether to log debug timing statements
 
 [results]
 # These results files are always written
-flow = "results/flow.arrow"       # optional, default "results/flow.arrow"
-basin = "results/basin.arrow"     # optional, default "results/basin.arrow"
-control = "results/control.arrow" # optional, default "results/control.arrow"
+flow = "results/flow.arrow"             # optional, default "results/flow.arrow"
+basin = "results/basin.arrow"           # optional, default "results/basin.arrow"
+control = "results/control.arrow"       # optional, default "results/control.arrow"
 allocation = "results/allocation.arrow" # optional, default "results/allocation.arrow"
-compression = "zstd"             # optional, default "zstd", also supports "lz4"
-compression_level = 6            # optional, default 6
+compression = "zstd"                    # optional, default "zstd", also supports "lz4"
+compression_level = 6                   # optional, default 6
--- a/docs/python/examples.ipynb
+++ b/docs/python/examples.ipynb
@@ -471,7 +471,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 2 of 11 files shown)


visr:
Ah, great, surprised that this just works on the Julia side. Indeed since this is input I'd say the path has to be relative to both the TOML directory and the optional `input_dir`.

Actually, I see `input_dir` and `results_dir` are missing from `docs.toml`, if you feel like addressing that as well.

Hofer-Julian (approved):

evetion:
> Ah, great, surprised that this just works on the Julia side. Indeed since this is input I'd say the path has to be relative to both the TOML directory and the optional `input_dir`.
> 
> Actually, I see `input_dir` and `results_dir` are missing from `docs.toml`, if you feel like addressing that as well.

I can add the input_dir, and leave the relative_dir for now, as I think it's unused and should probably be removed. Other naming things should probably be refined and changed in another PR.

visr:
Great, thanks. I thought `relative_dir` isn't really a TOML key, it is just a way on the Julia side to interpret all paths relative to the TOML.

evetion:
Ok, I've now added using input_dir in both reading and writing. I've also added a test in Python to check we can read arrow files back in again (from the relative input_path).

docs/python/examples.ipynb
```diff
@@ -471,7 +471,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "model = ribasim.Model.from_toml(datadir / \"basic/ribasim.toml\")"
+    "model = ribasim.Model(filepath=datadir / \"basic/ribasim.toml\")"
```
Hofer-Julian, line 474: I liked having separate constructors here. Is there anything blocking us from being explicit here?
Also, what is the behavior if I now specify `filepath` as well as other parameters? Will the TOML or the manually set parameters take precedence?

evetion: Hmm, I found it confusing. Would `read` also be acceptable (as there's a `write`)?

Manually set should override, I've fixed that now.

Hofer-Julian: `read` sounds good. There is no alternative to our toml.

evetion:
Thanks for the review, all addressed now.

core/test/docs.toml
```diff
@@ -37,9 +39,9 @@ timing = false     # optional, whether to log debug timing statements
 
 [results]
 # These results files are always written
-flow = "results/flow.arrow"       # optional, default "results/flow.arrow"
-basin = "results/basin.arrow"     # optional, default "results/basin.arrow"
-control = "results/control.arrow" # optional, default "results/control.arrow"
+flow = "results/flow.arrow"             # optional, default "results/flow.arrow"
```
Hofer-Julian, line 42: Why do you still have to specify "results" here? :)

Shouldn't this just be flow.arrow now?

evetion: Yeah probably, but it might be good to have that discussion first before changing it. Results are out of scope for this PR?

Hofer-Julian: But you introduced results_dir.

Shouldn't this part be adapted accordingly?

Or am I misunderstanding what results_dir does?

evetion: This PR does not introduce results_dir functionality. I added it to the docs.toml because it was missing there on request of @visr, and autoformatting did a bit more to that file. The Python code does not (yet) read results files.

Hofer-Julian: I see.
If this is unrelated to your PR then let's merge it.
Let's remember to discuss that result_dir thing on Monday :)

PR #803, opened 2023-11-18
Run Julia tests in parallel

visr:
This runs our tests on 8 workers. I haven't played around with the number yet, but as-is this is over 2x faster locally.

It uses https://github.com/JuliaTesting/ReTestItems.jl/. This runs each testitem in a separate module. Before we were using SafeTestSets to run each test file in a separate module. Therefore some changes were needed, to make each item self-contained. Possibly we can make things a bit nicer with the `testsetup` macro as well.

This also collects the output, so overall seems a lot more like our pytest setup. It also makes it easier to run a part of the tests with a regex or tag.

I removed the TeamCity and TimerOutput tests, since they were just a PoC and not actually used. I we bring the XML report back we should probably use the [ReTestItems report keyword](https://github.com/JuliaTesting/ReTestItems.jl/blob/7bdf3bb79c756555bb9a55e2902a21b413fae8d3/src/ReTestItems.jl#L157).


Hofer-Julian (approved):
Nice, that change is much appreciated.

Apart from speeding up runs, selecting a subset with regex is something I missed a lot from pytest

visr:
> This runs our tests on 8 workers. I haven't played around with the number yet

Ran a quick grid search locally:
```
nworkers =  1: 3m02.4s
nworkers =  2: 2m17.1s
nworkers =  4: 1m48.1s
nworkers =  8: 1m42.9s
nworkers = 12: 2m12.9s
nworkers = 16: 3m08.7s
nworkers = 20: 3m30.2s
```

So 8 was a good guess. I changed it to 4 since it is almost as fast, and CI and some other devs have less CPU resources.

core/test/runtests.jl
```diff
@@ -1,21 +1,3 @@
...
-    @safetestset "Utility functions" include("utils.jl")
-    @safetestset "Control" include("control.jl")
-    @safetestset "Allocation" include("allocation.jl")
-    @safetestset "Time" include("time.jl")
-    @safetestset "Docs" include("docs.jl")
-    @safetestset "Command Line Interface" include("cli.jl")
-    @safetestset "libribasim" include("libribasim.jl")
-    Aqua.test_all(Ribasim; ambiguities = false, persistent_tasks = false)
-end
+runtests(Ribasim; nworkers = 4)
```
evetion, line 3:
```suggestion
runtests(Ribasim; nworkers = min(4, Sys.CPU_THREADS))
```

Or even just the CPU_threads, because scaling depends on number of threads available. Number of cores do matter, but are not known without using Hwloc.jl or CpuId.jl.

PR #805, opened 2023-11-20
Develop QGIS plugins via VS Code

deltamarnix:
Install QGIS via pixi, ensures that vs code can find QGIS.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -17,11 +17,13 @@ repository = "https://github.com/Deltares/Ribasim"
 install-ribasim-python = "pip install --no-deps --editable python/ribasim"
 install-ribasim-api = "pip install --no-deps --editable python/ribasim_api"
 install-ribasim-testmodels = "pip install --no-deps --editable python/ribasim_testmodels"
+install-ribasim-qgis = "pip install --no-deps --editable ribasim_qgis"
 install-pre-commit = "pre-commit install"
 install-without-pre-commit = { depends_on = [
     "install-ribasim-python",
     "install-ribasim-api",
     "install-ribasim-testmodels",
+    "install-ribasim-qgis",
 ] }
 install = { depends_on = ["install-without-pre-commit", "install-pre-commit"] }
 # Instantiate
--- a/pixi.toml
+++ b/pixi.toml
@@ -121,6 +121,7 @@ twine = "*"
 typing-extensions = ">=4.6"
 xarray = "*"
 xmipy = "*"
+qgis = "3.34.0.*"
 
 [target.win-64.dependencies]
 quarto = "*"
--- a/pixi.toml
+++ b/pixi.toml
@@ -73,6 +73,8 @@ test-ribasim-core = { cmd = "julia --project=core --eval 'using Pkg; Pkg.test()'
     "instantiate-julia",
     "generate-testmodels",
 ] }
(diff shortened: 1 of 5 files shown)


ribasim_qgis/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[build-system]
+requires = ["setuptools>=61"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ribasim_qgis"
+description = "Ribasim QGIS Plugin"
+readme = "README.md"
```
visr, line 8: This file doesn't exist yet, but would be good to create, just like https://github.com/Deltares/Ribasim/blob/main/python/ribasim_api/README.md for instance.

ribasim_qgis/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+authors = [{ name = "Deltares and contributors", email = "ribasim.info@deltares.nl" }]
+license = { text = "MIT" }
+classifiers = [
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering :: Hydrology",
+]
+requires-python = ">=3.10"
```
visr, line 15:
```suggestion
requires-python = ">=3.9"
```

QGIS default installer still comes with Python 3.9, see also https://github.com/Deltares/Ribasim/pull/795.

ribasim_qgis/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+dynamic = ["version"]
+
+[tool.setuptools]
+zip-safe = true
+
+[tool.setuptools.dynamic]
+version.attr = "ribasim_testmodels.__version__"
```
visr, line 22:
```suggestion
version.attr = "ribasim_qgis.__version__"
```

ribasim_qgis/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[tool.setuptools.packages.find]
+include = ["*"]
+
+[tool.setuptools.package-data]
+"ribasim_testmodels" = ["py.typed"]
```
visr, line 28:
```suggestion
"ribasim_qgis" = ["py.typed"]
```

I think you also need to add a py.typed file like the other Python packages.

pixi.toml
```diff
@@ -121,6 +125,7 @@ twine = "*"
 typing-extensions = ">=4.6"
 xarray = "*"
 xmipy = "*"
+qgis = "3.34.0.*"
```
visr, line 128:
```suggestion
qgis = ">=3.28"
```

We generally use * here unless we know some other bound, like we do for QGIS.

pixi.toml
```diff
@@ -71,6 +73,8 @@ test-ribasim-core = { cmd = "julia --project=core --eval 'using Pkg; Pkg.test()'
     "instantiate-julia",
     "generate-testmodels",
 ] }
+start-docker-qgis = { cmd = "sh ./start.sh", cwd = ".docker" }
+test-ribasim-qgis = { cmd = "sh ./test.sh; sh ./stop.sh", cwd = ".docker", depends_on = ["start-docker-qgis"] }
```
visr, line 77: This requires a local Docker install right? Good to add to dev docs.

pixi.toml
deltamarnix: I changed it to 3.28.*, since that is the LTS release

Issue #785, opened 2023-11-14
Refactor `allocation.jl` with `MetaGraphsNext.jl`


SouthEndMusic:
Should be done after https://github.com/Deltares/Ribasim/issues/784.

Currently the allocation code is a bit messy because there is a lot of translation going on between flow graph and allocation graph identifiers. With the introduction of `MetaGraphsNext.jl` this can be resolved by storing data about allocation graphs in the metadata of the main `graph`.

SouthEndMusic:
The edge capacities for allocation could maybe be stored in the graph weights.

PR #807, opened 2023-11-20
Refactor allocation with `MetaGraph`

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/785.


core/src/validation.jl
```diff
@@ -406,59 +406,82 @@ function sorted_table!(
...
+
+function Base.setindex!(
+    M::AbstractArray,
+    value::T,
+    id_row::NodeID,
+    id_col::NodeID,
+)::Nothing where {T}
+    M[id_row.value, id_col.value] = value
+    return nothing
+end
```
visr, line 432: These are only to support `flow::SparseMatrix` handling right, and therefore can be removed once we stop storing flow as a SparseMatrix? Perhaps good to write this in a comment above.

core/src/validation.jl
```diff
@@ -406,59 +406,82 @@ function sorted_table!(
...
+    for e in edges(graph)
+        id_src = label_for(graph, e.src)
+        id_dst = label_for(graph, e.dst)
+        type_src = graph[id_src].type
+        type_dst = graph[id_dst].type
+
+        if !(type_dst in neighbortypes(type_src))
+            errors = true
+            edge_id = graph[id_src, id_dst].id.value
+            @error "Cannot connect a $type_src to a $type_dst (edge #$edge_id from node $id_src to $id_dst)."
```
visr, line 453: If you would define a `show` method for EdgeID as well, you wouldn't have to take the `.value` out here right?

Perhaps it's worth it to create an `abstract type ID`, and make both NodeID and EdgeID an ID, and then you can directly define these methods for both.

core/test/utils_test.jl
```diff
@@ -1,9 +1,9 @@
 @testitem "id_index" begin
     using Dictionaries: Indices
 
-    ids = Indices([2, 4, 6])
-    @test Ribasim.id_index(ids, 4) === (true, 2)
-    @test Ribasim.id_index(ids, 5) === (false, 0)
+    ids = Indices([Ribasim.NodeID(2), Ribasim.NodeID(4), Ribasim.NodeID(6)])
```
visr, line 4:
```suggestion
    ids = Indices(Ribasim.NodeID[2, 4, 6])
```

This works with the extra `Base.convert` method suggested.

core/test/utils_test.jl
```diff
@@ -23,7 +23,7 @@ end
     darea = zeros(2)
     storage = Ribasim.profile_storage.(level, area)
     basin = Ribasim.Basin(
-        Indices([5, 7]),
+        Indices([Ribasim.NodeID(5), Ribasim.NodeID(7)]),
```
visr, line 26:
```suggestion
        Indices(Ribasim.NodeID[5, 7]),
```

core/test/utils_test.jl
```diff
@@ -116,19 +131,19 @@ end
...
+    logic_mapping[(Ribasim.NodeID(1), "*T*")] = "foo"
+    logic_mapping[(Ribasim.NodeID(2), "FF")] = "bar"
     logic_mapping_expanded = Ribasim.expand_logic_mapping(logic_mapping)
 
-    @test logic_mapping_expanded[(1, "TTT")] == "foo"
-    @test logic_mapping_expanded[(1, "FTT")] == "foo"
-    @test logic_mapping_expanded[(1, "TTF")] == "foo"
-    @test logic_mapping_expanded[(1, "FTF")] == "foo"
-    @test logic_mapping_expanded[(2, "FF")] == "bar"
+    @test logic_mapping_expanded[(Ribasim.NodeID(1), "TTT")] == "foo"
```
visr, line 139: If you get tired of typing `Ribasim.NodeID`, you can also add `using Ribasim: NodeID` at the top of the testitem, so you can just use `NodeID` directly.

Note that: By default, Test and the package being tested will be imported into the testitem automatically.

core/test/validation_test.jl
```diff
@@ -49,30 +49,45 @@ end
...
+    graph[Ribasim.NodeID(3), Ribasim.NodeID(1)] =
+        Ribasim.EdgeMetadata(Ribasim.EdgeID(2), Ribasim.EdgeType.flow)
+    graph[Ribasim.NodeID(6), Ribasim.NodeID(2)] =
+        Ribasim.EdgeMetadata(Ribasim.EdgeID(3), Ribasim.EdgeType.flow)
+    graph[Ribasim.NodeID(5), Ribasim.NodeID(6)] =
+        Ribasim.EdgeMetadata(Ribasim.EdgeID(4), Ribasim.EdgeType.control)
 
     pump = Ribasim.Pump(
-        [1, 6],
+        [Ribasim.NodeID(1), Ribasim.NodeID(6)],
```
visr, line 79:
```suggestion
        Ribasim.NodeID[1, 6],
```

core/test/validation_test.jl
```diff
@@ -49,30 +49,45 @@ end
...
+
+    for i in 1:6
+        type = i in [1, 6] ? :pump : :other
+        graph[Ribasim.NodeID(i)] = Ribasim.NodeMetadata(type, 9)
+    end
 
-    graph_control = DiGraph(6)
-    add_edge!(graph_control, 5, 6)
+    graph[Ribasim.NodeID(2), Ribasim.NodeID(1)] =
+        Ribasim.EdgeMetadata(Ribasim.EdgeID(1), Ribasim.EdgeType.flow)
```
visr, line 70: It probably makes sense to add `using Ribasim: ...` for these ID and metadata types at the top of this testitem.

core/test/validation_test.jl
```diff
@@ -186,13 +224,13 @@ end
...
           "Fractional flow nodes must have non-negative fractions."
-    @test logger.logs[2].kwargs[:node_id] == 3
+    @test logger.logs[2].kwargs[:node_id].value == 3
     @test logger.logs[2].kwargs[:fraction] ≈ -0.1
     @test logger.logs[2].kwargs[:control_state] == ""
     @test logger.logs[3].level == Error
     @test logger.logs[3].message ==
           "The sum of fractional flow fractions leaving a node must be ≈1."
-    @test logger.logs[3].kwargs[:node_id] == 7
+    @test logger.logs[3].kwargs[:node_id].value == 7
```
visr, line 233: A better test would be to remove the `.value` and `== NodeID(...)`, since that explicitly tests if it is a NodeID.

core/src/create.jl
```diff
@@ -227,17 +233,7 @@ function Connectivity(db::DB, config::Config, chunk_size::Int)::Connectivity
...
-        graph_control,
-        flow,
-        edge_ids_flow,
-        edge_ids_flow_inv,
-        edge_ids_control,
-        edge_connection_types_flow,
-        edge_connection_types_control,
-        allocation_models,
-    )
+    return Connectivity(graph, flow, allocation_models)
```
visr, line 236: Love it.

core/src/solve.jl
```diff
@@ -34,10 +34,33 @@ struct AllocationModel
     Δt_allocation::Float64
 end
 
+@enumx EdgeType flow control
+
+"""
+Type for storing metadata of nodes in the graph.
+"""
+struct NodeMetadata
+    type::Symbol
```
visr, line 43: Do you think it's worthwhile to create a NodeType enum as well? Would be for another PR though, this is big enough as is :)

core/src/solve.jl
```diff
@@ -764,9 +749,9 @@ function continuous_control!(
 
         # When the controlled pump flows out into fractional flow nodes
         if controls_pump
-            for id in outneighbors(graph_flow, controlled_node_id)
+            for id in outneighbor_labels_type(graph, controlled_node_id, EdgeType.flow)
```
visr, line 752: Slightly less generic, but how about creating a `outflow_ids` function for this, since we use it so much? Together with `inflow_ids`. We can even make `<in/out>flow_id` that do `only(x)` for us.

core/src/utils.jl
```diff
@@ -12,25 +12,97 @@ function valid_edge_types(db::DB)::Bool
...
-    edge_type_::String,
-)::Tuple{DiGraph, Dictionary{Tuple{Int, Int}, Int}, Dictionary{Int, Tuple{Symbol, Symbol}}}
+"""
+Return a directed metagraph with data of nodes (NodeMetadata):
+- Node type (snake case)
+- Allocation network ID
+
+and data of edges (EdgeMetadata):
+- Edge ID (EdgeID)
+- type (flow/control)
```
visr, line 22: The content of NodeMetadata and EdgeMetadata are also in those docstrings. Rather than copying, it's better to link to those, using

```
[`NodeMetadata`](@ref)
```

core/src/utils.jl
```diff
@@ -12,25 +12,97 @@ function valid_edge_types(db::DB)::Bool
...
-        if edge_type == edge_type_
-            add_edge!(graph, from_node_id, to_node_id)
-            insert!(edge_ids, (from_node_id, to_node_id), fid)
-            insert!(edge_connection_types, fid, (nodes[from_node_id], nodes[to_node_id]))
+    graph = MetaGraph(
+        DiGraph();
+        label_type = NodeID,
+        vertex_data_type = NodeMetadata,
+        edge_data_type = EdgeMetadata,
+        graph_data = nothing, # In theory all remaining connectivity data could be put here
```
visr, line 32: If you think this would be an improvement, it'd be good to make a follow up issue/task.

visr (commented):
Great job here! I left a lot of small comments, though nothing structurally different than what is here.

core/src/solve.jl
SouthEndMusic: Might be, but that requires more maintenance when adding new node types. I haven't found a use for it yet.

visr (approved):

Issue #811, opened 2023-11-21
Pydantic model_dump(exclude_unset=True, ...) causes model.write to ignore mutation


Huite:
Here's an MWE:

```python
import ribasim_testmodels

model = ribasim_testmodels.trivial.trivial_model()
model.solver.saveat = 86400.0
model.write("here")
```

If you now check the TOML, I'd expect to see a non-default saveat entry, but instead:

```toml
starttime = 2020-01-01 00:00:00
endtime = 2021-01-01 00:00:00
database = "database.gpkg"
```

I believe this is due to the (expected) behavior of exlude_unset, see also:
https://github.com/pydantic/pydantic/discussions/5749

Creating a new object does work, probably because the solver entry is detected as being set by pydantic?

```python
model = ribasim_testmodels.trivial.trivial_model()
model.solver = ribasim.Solver(saveat=86400.0)
model.write("here")
```

TOML:

```toml
starttime = 2020-01-01 00:00:00
endtime = 2021-01-01 00:00:00
database = "database.gpkg"

[solver]
saveat = 86400.0
```

Given that this seems expected behavior by pydantic, I'm not sure it's easy to fix (without dirty hacks).

A pragmatic solution might be something like this:

```python
model.solver = ribasim.solver.update(saveat=86400.0)
```

The update method just returns a new object.

You can argue this is consistent with how you're supposed to work with pandas and xarray, where the `inplace` operations are mostly discouraged or removed; although you might as well make the object immutable in that case to avoid confusion?

Huite:
Another thoughts: this can be avoided entirely by not using `exclude_unset`. This then writes all the defaults to the TOML, but that's fine. I mostly prefer having the settings in that way, because it explicitly communicates what e.g. the solver settings are.

evetion:
I'll try a fix by making a relationship between Model and its children.

PR #815, opened 2023-11-21
Use hardcoded filenames.

### LINKED ISSUE #799: Default results design
Currently the entire results section is optional, with the following defaults:

```toml
[results]
# These results files are always written
flow = "results/flow.arrow"       # optional, default "results/flow.arrow"
basin = "results/basin.arrow"     # optional, default "results/basin.arrow"
control = "results/control.arrow" # optional, default "results/control.arrow"
allocation = "results/allocation.arrow" # optional, default "results/allocation.arrow"
compression = "zstd"             # optional, default "zstd", also supports "lz4"
compression_level = 6            # optional, default 6
```

In https://github.com/Deltares/Ribasim/issues/638#issuecomment-1746283362 I suggested to stop using default paths, so it is always explicit from the TOML file which files belong to the model, without having to know the defaults. I still like having this explicit, though it does mean that the minimal TOML becomes bigger (not really a problem), and that in Ribasim Python you need to add `results = ribasim.Results(flow = "results/flow.arrow", basin = "results/basin.arrow", ...)`

One related thing is that we now always write all 4 files. If e.g. control and allocation are not part of the model, Arrow files with 0 rows are written to disk. If we decide to require explicit paths for the results, we have a way to turn off output that we are not interested in. That would be nice to support in the future as we may add some more output that is not always interesting or cheap to write.

I'm not sure anymore what's the best design here and would like some input from others. If we go for explicit paths that's a breaking change that would be best to make soon.

evetion:
Fixes #799

### DIFF
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -139,7 +139,7 @@ end
 
 struct Config
     toml::Toml
-    relative_dir::String
+    dir::String
 end
 
 Config(toml::Toml) = Config(toml, ".")
@@ -152,27 +152,22 @@ keys from a subsection, e.g. `dt` from the `solver` section, use underscores: `s
 """
 function Config(config_path::AbstractString; kwargs...)::Config
     toml = from_toml(Toml, config_path; kwargs...)
-    relative_dir = dirname(normpath(config_path))
-    Config(toml, relative_dir)
+    dir = dirname(normpath(config_path))
+    Config(toml, dir)
 end
 
-Base.getproperty(config::Config, sym::Symbol) = getfield(getfield(config, :toml), sym)
 Base.getproperty(config::Config, sym::Symbol) = getproperty(getfield(config, :toml), sym)
-"""
-    relative_dir(config::Config)
 
-Get the relative directory of the configuration
-"""
-relative_dir(config::Config) = getfield(config, :relative_dir)
+Base.dirname(config::Config) = getfield(config, :dir)
 
 "Construct a path relative to both the TOML directory and the optional `input_dir`"
 function input_path(config::Config, path::String)
-    return normpath(relative_dir(config), config.input_dir, path)
+    return normpath(dirname(config), config.input_dir, path)
 end
 
 "Construct a path relative to both the TOML directory and the optional `results_dir`"
 function results_path(config::Config, path::String)
-    return normpath(relative_dir(config), config.results_dir, path)
+    return normpath(dirname(config), config.results_dir, path)
 end
 
 function Configurations.from_dict(::Type{Logging}, ::Type{LogLevel}, level::AbstractString)
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -157,7 +157,7 @@ function Config(config_path::AbstractString; kwargs...)::Config
 end
 
 Base.getproperty(config::Config, sym::Symbol) = getfield(getfield(config, :toml), sym)
-
+Base.getproperty(config::Config, sym::Symbol) = getproperty(getfield(config, :toml), sym)
 """
     relative_dir(config::Config)
 
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -15,7 +15,7 @@ using Logging: LogLevel, Debug, Info, Warn, Error
 using ..Ribasim: Ribasim, isnode, nodetype
 using OrdinaryDiffEq
 
-export Config, Solver, Results, Logging
+export Config, Solver, Results, Logging, Toml
 export algorithm, snake_case, zstd, lz4
 
 const schemas =
@@ -109,10 +109,6 @@ end
 
 # Separate struct, as basin clashes with nodetype
 @option struct Results <: TableOption
-    basin::String = "basin.arrow"
-    flow::String = "flow.arrow"
-    control::String = "control.arrow"
-    allocation::String = "allocation.arrow"
     outstate::Union{String, Nothing} = nothing
     compression::Compression = "zstd"
     compression_level::Int = 6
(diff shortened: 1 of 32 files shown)


core/src/config.jl
```diff
@@ -109,10 +109,10 @@ end
...
 # Separate struct, as basin clashes with nodetype
 @option struct Results <: TableOption
-    basin::String = "results/basin.arrow"
-    flow::String = "results/flow.arrow"
-    control::String = "results/control.arrow"
-    allocation::String = "results/allocation.arrow"
+    basin::String = "basin.arrow"
+    flow::String = "flow.arrow"
+    control::String = "control.arrow"
+    allocation::String = "allocation.arrow"
```
Hofer-Julian, line 115: To fit https://github.com/Deltares/Ribasim/issues/799#issuecomment-1818799618

Those fields need to be removed as far as I can see

python/ribasim/ribasim/config.py
```diff
@@ -44,9 +44,10 @@ class Compression(str, Enum):
 
 
 class Results(BaseModel):
-    basin: Path = Path("results/basin.arrow")
-    flow: Path = Path("results/flow.arrow")
-    control: Path = Path("results/control.arrow")
+    basin: Path = Field(default=Path("basin.arrow"), exclude=True, repr=False)
+    flow: Path = Field(default=Path("flow.arrow"), exclude=True, repr=False)
+    control: Path = Field(default=Path("control.arrow"), exclude=True, repr=False)
+    allocation: Path = Field(default=Path("allocation.arrow"), exclude=True, repr=False)
```
Hofer-Julian, line 50: These need to be removed too

core/src/config.jl
```diff
@@ -134,12 +134,12 @@ end
     endtime::DateTime
 
     # optional, when Config is created from a TOML file, this is its directory
-    relative_dir::String = "."  # ignored(!)
+    relative_dir::String = "."
     input_dir::String = "."
-    results_dir::String = "."
+    results_dir::String = "results"
```
Hofer-Julian, line 139: Here you need to remove the default values, right?

core/src/config.jl
evetion: These fields need to be hardcodes somewhere for core to write them.

Hofer-Julian: Sure, but if you hardcode them here, it will be possible to override them with the TOML right?

python/ribasim/ribasim/config.py
evetion: I've excluded them from any exports. They are here for when we need to parse the output files.

core/src/config.jl
evetion: This is the same as the results fields. Let's see if I can fix them, as in not set by toml.

python/ribasim/ribasim/config.py
Huite: Just an FYI: I just found out that TOML hates these Path objects, and complains that it cannot serialize them.

However, due to the exclude_unset logic, they aren't included unless you set them yourself. For extra confusion, if you set the path (in current main) as a string, pydantic will coerce it to a Path. Then pydantic detects it has been set, and then TOML complains that it cannot serialize it.

To be honest, together with issue #811, I'm getting really radicalized against exclude_unset.

evetion: That's why there's 

    @field_serializer("input_dir", "results_dir") 
      def serialize_path(self, path: Path) -> str: 
          return str(path)


Right? Which fields are you trying to set?

I don't like the unset behaviour either, but that should be a separate issue/discussion.

core/src/config.jl
evetion: I now always override them.

core/src/config.jl
```diff
@@ -149,6 +144,13 @@ end
     results::Results = Results()
 end
 
+struct Config
+    toml::Toml
+    relative_dir::String
+end
```
visr, line 150: I like separating the TOML from other things. I don't like having to always type e.g. `config.toml.starttime`. Would it work for you if we forward `Base.getproperty` on `Config` to the `toml` field? To maintain the old API. To access the `relative_dir` field, which perhaps should just be `dir` or `directory`, we could add `Base.dirname(config::Config) = getfield(config, :relative_dir)`.

visr (changes requested):

core/src/consts.jl
```diff
@@ -0,0 +1,4 @@
+BASIN_NAME = "basin.arrow"
+FLOW_NAME = "flow.arrow"
+CONTROL_NAME = "control.arrow"
+ALLOCATION_NAME = "allocation.arrow"
```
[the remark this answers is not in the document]
Hofer-Julian: > perhaps this isn't needed?

I am fine either way. @evetion had the idea to "make a stub package with all defaults written out".
I've added the consts for that.

core/src/config.jl
Hofer-Julian: > forward Base.getproperty on Config to the toml field

Nice idea. It adds a bit complexity, but I think it's worth it.

> relative_dir field, which perhaps should just be dir or directory

I think relative_dir is more descriptive

> we could add Base.dirname(config::Config) = getfield(config, :relative_dir)

The [docs](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.dirname) Base.dirname talk about paths.
Since config isn't a path, I don't think it is applicable here. I've created a function `relative_dir` instead.

> The Base.show method in this file probably also need updating.

It seems to work fine for me. It's what is called if I evaluate a `Config` instance in the REPL, right?

visr: Thanks! Not sure what you think, but the name `relative_dir` has always been a bit confusing to me. It is not necessarily relative, it is just a directory, it's other paths in the TOML that are relative to it.

Regarding `Base.dirname`, good that you checked the docstring for it. The line "Get the directory part of a path." really seems to describe the particular `::String` method though. If it is a true function docstring it should describe what is does ("get the directory"), not from where. And what it does seems to match fine. Wflow also has this method on their `Config`.

Indeed Base.show seems to work fine, I mistakenly assumed it would be broken by this PR.

visr (approved):
Left some final minor comments.

Hofer-Julian:
@visr I've renamed `relative_dir` to `dir`, used `getproperty` instead of `getfield` where applicable and started using `Base.dirname`

PR #818, opened 2023-11-22
Update issue templates

evetion:


Hofer-Julian:
Bildschirmfoto vom 2023-11-22 09-21-20

Is this the intended render?
I assume you either want those things you've emphasized to be `h2` instead, or add a colon afterward.

evetion:
This is the default styling as proposed by GitHub. I only changed the text. I'm happy to add colons if you like it.

Hofer-Julian:
> This is the default styling as proposed by GitHub. I only changed the text. I'm happy to add colons if you like it.

I am fine either way. Just wanted to make sure that this is indeed what you wanted :)

Hofer-Julian (approved):

evetion:
This is how it now looks like:

Screenshot 2023-11-22 at 17 38 02

Hofer-Julian:
Hah, I assume the Github Diff render doesn't know yet that Github changed the behavior of newlines :P

PR #823, opened 2023-11-23
fixing mypy issues for ribasim_qgis plugin

### LINKED ISSUE #819: Fix mypy issues in QGIS plugin
There are several mypy issues to be resolved in the QGIS plugin.

- [ ] Mypy is different when running during development in vs code than when running it in via `pixi run mypy ribasim_qgis`
- [ ] There are several mypy issues to be resolved for typing.

deltamarnix:
Fixes #819 

Add mypy correctly to vs code and pixi. Required to add the mypy.ini file to the command line in vs code, also required to specify which mypy to use and which paths to look for.

The mypy.ini required some excludes for the qgis modules.

Also fixes a bug in the nodes widget. I broke it during #826. See nodes_widget.py:62.


mypy.ini
```diff
@@ -10,3 +10,11 @@ strict_concatenate = True
 disallow_subclassing_any = True
 disallow_untyped_decorators = True
 disallow_any_generics = True
+exclude=/tomllib/
+mypy_path=.pixi/env/Library/python
```
Hofer-Julian, line 14: Probably good for @evetion to check if that path also works for unix-like systems

evetion: For me its under `.pixi/env/share/qgis/python` (assuming you expect a `qgis` folder in there with Python files).

visr: Ha, didn't know they were so different. I guess we could just list both? `mypy_path=.pixi/env/Library/python:.pixi/env/share/qgis/python`

ribasim_qgis/core/topology.py
```diff
@@ -16,9 +22,11 @@ def explode_lines(edge: QgsVectorLayer) -> None:
         # Avoid infinite recursion and stackoverflow
         edge.blockSignals(True)
         provider = edge.dataProvider()
+        assert provider is not None
 
         with edit(edge):
-            provider.deleteFeatures([f.id() for f in edge.getFeatures()])
+            edge_iterator = cast(Iterable[QgsFeature], edge.getFeatures())
```
deltamarnix, line 28: `getFeatures()` actually returns a `QgsFeatureIterator`, but somehow mypy doesn't understand that it has an `__iter__()` and `__next()__` function

deltamarnix: I figured mypy doesn't understand, because the `__iter__()` function of qgis returns an `Optional[QgsFeatureIterator`

```py
    def __iter__(self) -> typing.Optional['QgsFeatureIterator']: ...
```

ribasim_qgis/core/topology.py
```diff
@@ -29,7 +37,9 @@ def explode_lines(edge: QgsVectorLayer) -> None:
     return
 
 
-def derive_connectivity(node_index, node_xy, edge_xy) -> tuple[np.ndarray, np.ndarray]:
+def derive_connectivity(
+    node_index, node_xy, edge_xy
+) -> tuple[np.ndarray[Any, Any], np.ndarray[Any, Any]]:
```
deltamarnix, line 42: I would like to get rid of these `Any`s. Do we know the types of these objects?

ribasim_qgis/widgets/dataset_widget.py
```diff
@@ -93,8 +99,9 @@ def remove_geopackage_layers(self) -> None:
             return
 
         # Start deleting
-        elements = {item.element for item in selection}
+        elements = {item.element for item in selection}  # type: ignore
```
deltamarnix, line 102: This needs to be fixed. `item.element` is dynamically added and is not part of the variables in `item`. Therefore we should change this. Maybe just with a simple dictionary in `self` to keep track of the elements per item.

ribasim_qgis/widgets/dataset_widget.py
```diff
@@ -124,10 +131,13 @@ def remove_geopackage_layers(self) -> None:
 
 
 class DatasetWidget(QWidget):
-    def __init__(self, parent):
+    def __init__(self, parent: QWidget):
+        from ribasim_qgis.widgets.ribasim_widget import RibasimWidget
```
deltamarnix, line 135: preventing circular references

ribasim_qgis/core/topology.py
visr: These are both 1D integer arrays.

Hofer-Julian: Also `numpy.typing.NDArray` is more useful here, it's seldom useful to specify the shape.
Here that would be `NDArray[np.int_]`

ribasim_qgis/core/nodes.py
```diff
@@ -47,102 +49,123 @@
 class Input(abc.ABC):
     """Abstract base class for Ribasim input layers."""
 
-    input_type = ""
-
     def __init__(self, path: str):
-        self.name = self.input_type
-        self.path = path
-        self.layer = None
+        self.name = self.input_type()
```
deltamarnix, line 53: I find this `self.name` a weird concept and I think we can remove it. Since `input_type()` describes exactly this on a class level. Meaning that we could even check up front if we should create the instance or not (line 80).

And for the rest we could simply call input_type() where needed.

.vscode/settings.json
```diff
@@ -16,5 +16,9 @@
     },
     "python.analysis.extraPaths": [
         ".pixi/env/Library/python"
+    ],
+    "mypy-type-checker.importStrategy": "fromEnvironment",
+    "mypy-type-checker.args": [
+        "--config-file=${workspaceFolder}/mypy.ini"
```
Hofer-Julian, line 22: Did you double check that this is really necessary?

ribasim_qgis/core/topology.py
Hofer-Julian: Maybe “assert is not None” before would help then?

Hofer-Julian (commented):
Looks like a big improvement to me. Great job!

i had a few comments. After they are discussed and/or addressed, we can merge this from my side

ribasim_qgis/core/topology.py
deltamarnix: Yes the alternative would be:

```py
edge_iterator = edge.getFeatures().__iter__()
assert edge_iterator is not None
```

But I'm not sure if that really makes a difference here.

ribasim_qgis/widgets/dataset_widget.py
deltamarnix: I've added TODOs

ribasim_qgis/core/topology.py
Hofer-Julian: I leave the decision to you

.vscode/settings.json
deltamarnix: You are right, no longer needed

visr (approved):

PR #826, opened 2023-11-24
Fix QGIS plugin now that database entry is not in TOML

Huite:
This code gives an error in the QGIS plugin:

```python
    def _get_database_path_from_model_file(self) -> str:
        with open(self.path, "rb") as f:
            model_filename = tomllib.load(f)["database"]
            return str(Path(self.path).parent.joinpath(model_filename))
```

Because the database entry is no longer present in the TOML because of #815.

### DIFF
--- a/ribasim_qgis/widgets/dataset_widget.py
+++ b/ribasim_qgis/widgets/dataset_widget.py
@@ -260,8 +260,11 @@ class DatasetWidget(QWidget):
 
     def _get_database_path_from_model_file(self) -> str:
         with open(self.path, "rb") as f:
-            input_dir = tomllib.load(f)["input_dir"]
-        return str((Path(self.path).parent / input_dir / "database.gpkg").resolve())
+            input_dir = Path(tomllib.load(f)["input_dir"])
+        if input_dir.is_absolute():
+            return str(input_dir / "database.gpkg")
+        else:
+            return str((Path(self.path).parent / input_dir / "database.gpkg").resolve())
 
     def new_model(self) -> None:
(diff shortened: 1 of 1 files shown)


ribasim_qgis/widgets/dataset_widget.py
```diff
@@ -260,8 +260,8 @@ def load_geopackage(self) -> None:
 
     def _get_database_path_from_model_file(self) -> str:
         with open(self.path, "rb") as f:
-            model_filename = tomllib.load(f)["database"]
-            return str(Path(self.path).parent.joinpath(model_filename))
+            input_dir = tomllib.load(f)["input_dir"]
+        return str((Path(self.path).parent / input_dir / "database.gpkg").resolve())
```
deltamarnix, line 264: Is `input_dir` always relative? Perhaps perform some more checks before you just append the paths. @visr what are the agreements on `input_dir`?

deltamarnix (changes requested):

ribasim_qgis/widgets/dataset_widget.py
visr: It can be absolute as well, but `/` handles that case as desired, removing the preceding part.

Huite: I tested it manually to be sure, I learned something today

Hofer-Julian: Oh nice

deltamarnix (approved):

Issue #814, opened 2023-11-21
`Connectivity` refactor


SouthEndMusic:
With the introduction of the `MetaGraph`, the `Connectivity` struct has been stripped down a lot. In fact, I think the struct is obsolete and can be replaced by the `MetaGraph` altogether. For the remaining fields:

- [ ] The `flow` data can be distributed over the `EdgeMetadata` instances. The problem with that however is that control edges do not have a flow associated with them, so this field would be redundant for control edges. Making separate metadata types for flow and control edges is probably also not nice since then `MetaGraph.label_type` is not fully typed. Perhaps the flow can be made a length 0 or 1 vector (that probably also works together more nicely with `PreAllocationTools.jl` and also makes them mutable). Or maybe the flow is a dense array in the `graph_data` and the edges contain only a reference to it. Thoughts @visr?
- [ ] The `AllocationModel` vector can be moved to the `graph_data`.

SouthEndMusic:
Suggestion: distribute flow over edge metadata. Make `EdgeMetadata` mutable struct with everything exept `flow` constant (https://docs.julialang.org/en/v1/manual/types/#Mutable-Composite-Types). To make it work with `ForwardDiff.jl`, the flow field type must be a type parameter of the `EdgeMetadata` struct.

PR #828, opened 2023-11-24
connectivity refactor

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/814.

### DIFF
--- a/core/src/io.jl
+++ b/core/src/io.jl
@@ -170,29 +170,31 @@ end
 function flow_table(model::Model)::NamedTuple
     (; config, saved_flow, integrator) = model
     (; t, saveval) = saved_flow
-    (; connectivity) = integrator.p
-    (; graph) = connectivity
+    (; graph) = integrator.p
 
-    I, J, _ = findnz(get_tmp(connectivity.flow, 0))
     # self-loops have no edge ID
-    # unique_edge_ids = [get(connectivity.edge_ids_flow, ij, missing) for ij in zip(I, J)]
+    from_node_id = Int[]
+    to_node_id = Int[]
     unique_edge_ids_flow = Union{Int, Missing}[]
-    for (i, j) in zip(I, J)
-        if i == j
+    for e in edges(graph)
+        edge_metadata = metadata_from_edge(graph, e)
+        id = edge_metadata.id
+        push!(from_node_id, edge_metadata.from_id.value)
+        push!(to_node_id, edge_metadata.to_id.value)
+        if id == 0
             push!(unique_edge_ids_flow, missing)
         else
-            edge_metadata = metadata_from_edge(graph, Edge(i, j))
-            push!(unique_edge_ids_flow, edge_metadata.id)
+            push!(unique_edge_ids_flow, id)
         end
     end
 
-    nflow = length(I)
+    nflow = length(unique_edge_ids_flow)
     ntsteps = length(t)
 
     time = repeat(datetime_since.(t, config.starttime); inner = nflow)
     edge_id = repeat(unique_edge_ids_flow; outer = ntsteps)
-    from_node_id = repeat(I; outer = ntsteps)
-    to_node_id = repeat(J; outer = ntsteps)
+    from_node_id = repeat(from_node_id; outer = ntsteps)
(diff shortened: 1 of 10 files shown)


core/src/io.jl
```diff
@@ -181,29 +181,37 @@ function flow_table(
...
-    for (i, j) in zip(I, J)
-        if i == j
-            push!(unique_edge_ids_flow, missing)
-        else
-            edge_metadata = metadata_from_edge(graph, Edge(i, j))
-            push!(unique_edge_ids_flow, edge_metadata.id)
-        end
+    from_node_id = Int[]
+    to_node_id = Int[]
+    unique_edge_ids_flow = Vector{Union{Int, Missing}}()
```
Hofer-Julian, line 190:
```suggestion
    unique_edge_ids_flow = Union{Int, Missing}[]
```

core/src/io.jl
```diff
@@ -181,29 +181,37 @@ function flow_table(
...
+    flow_vertical_dict_inverse = Dict(value => key for (key, value) in flow_vertical_dict)
+    flow_dict_inverse = Dict(value => key for (key, value) in flow_dict)
+
+    for i in 1:length(flow_vertical_dict_inverse)
+        id = flow_vertical_dict_inverse[i]
+        push!(from_node_id, id.value)
+        push!(to_node_id, id.value)
+        push!(unique_edge_ids_flow, missing)
     end
```
Hofer-Julian, line 199:
```suggestion
    for id in flow_vertical_dict_inverse
        push!(from_node_id, id.value)
        push!(to_node_id, id.value)
        push!(unique_edge_ids_flow, missing)
    end
```

core/src/io.jl
```diff
@@ -181,29 +181,37 @@ function flow_table(
...
 
-    nflow = length(I)
+    for i in 1:length(flow_dict_inverse)
+        from_id, to_id = flow_dict_inverse[i]
+        push!(from_node_id, from_id.value)
+        push!(to_node_id, to_id.value)
+        push!(unique_edge_ids_flow, graph[from_id, to_id].id)
+    end
```
Hofer-Julian, line 206:
```suggestion
    for (from_id, to_id) in flow_dict_inverse
        push!(from_node_id, from_id.value)
        push!(to_node_id, to_id.value)
        push!(unique_edge_ids_flow, graph[from_id, to_id].id)
    end
```

core/src/utils.jl
```diff
@@ -63,14 +71,33 @@ function create_graph(db::DB)::MetaGraph
...
         end
-        graph[id_src, id_dst] = edge_metadata
     end
+
+    flow = zeros(flow_counter)
+    flow_vertical = zeros(flow_vertical_counter)
+    if config.solver.autodiff
+        flow = DiffCache(flow, chunk_size)
+        flow_vertical = DiffCache(flow_vertical, chunk_size)
+    end
```
Hofer-Julian, line 90:
```suggestion
    flow, flow_vertical = if config.solver.autodiff
        DiffCache(flow, chunk_size), DiffCache(flow_vertical, chunk_size)
    else
        zeros(flow_counter), zeros(flow_vertical_counter)
    end
```
It's clearer and saves doesn't allocate needlessly

Hofer-Julian (changes requested):
I had a few comments.
Apart from them, it looks good to me.

core/src/io.jl
Hofer-Julian: So you verified that iterating through it and indexing one-by-one doesn't behave the same?

core/src/utils.jl
SouthEndMusic: The reason it is written this way is because allocation of `flow` and `flow_vertical` is always needed, and in the case of autodiff `DiffCache` allocates a bit more for AD computations. No needless allocation happens

Hofer-Julian: Your solution always allocates the zeros.
Mine only allocates them if `config.solver.autodiff` is false.

Hofer-Julian: I've overlooked that flow is an input parameter of `DiffCache`

Hofer-Julian (approved):

PR #830, opened 2023-11-26
Use custom iterators to speed up `inflow_ids` and friends

visr:
@Huite mentioned the performance of medium-sized models regressed recently, so I looked into it a bit, doing some profiling. I saw a lot of time was spent on finding the in- and outneighbors of flow. This changed quite a bit in #807, and whilst the API is a lot nicer, unfortunately there was also a performance regression. The profile showed allocations. This PR removes those allocations, leading to a 2x speedup for a test model, from 16s to 8s. Pre-#807 this ran in 3s however, so there is still a considerable gap to close. I don't really want to revert #807, though perhaps for flow connections we need another data structure, or some other way to speed up.

This PR changes from using a filtered array comprehension, which creates a `Vector{NodeID}`, to a custom iterator `InNeighbors` which iterates over the filtered `inneighbors`. This avoids allocations but otherwise returns the same result.

### DIFF
--- a/core/src/allocation.jl
+++ b/core/src/allocation.jl
@@ -13,8 +13,8 @@ function allocation_graph_used_nodes!(p::Parameters, allocation_network_id::Int)
         node_type = graph[node_id].type
         if node_type in [:user, :basin]
             push!(used_nodes, node_id)
-
-        elseif length(inoutflow_ids(graph, node_id)) > 2
+        elseif count(x -> true, inoutflow_ids(graph, node_id)) > 2
+            # use count since the length of the iterator is unknown
             push!(used_nodes, node_id)
         end
     end
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -49,7 +49,13 @@ using Legolas: Legolas, @schema, @version, validate, SchemaVersion, declared
 using Logging: current_logger, min_enabled_level, with_logger
 using LoggingExtras: EarlyFilteredLogger, LevelOverrideLogger
 using MetaGraphsNext:
-    MetaGraphsNext, MetaGraph, label_for, labels, outneighbor_labels, inneighbor_labels
+    MetaGraphsNext,
+    MetaGraph,
+    label_for,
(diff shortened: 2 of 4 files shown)


SouthEndMusic:
@visr nice that you picked this up. Regarding the gap to close: this might be due to the dictionary to map from an edge to an index in the dense `flow` vector. The `SparseMatrixCSC` we used before does not use a dictionary internally, so I suspect we can learn something from that implementation.

SouthEndMusic:
Also, `get_tmp` is called much more often now but I thought that is basically free

core/src/allocation.jl
```diff
@@ -13,8 +13,8 @@ function allocation_graph_used_nodes!(p::Parameters, allocation_network_id::Int)
         node_type = graph[node_id].type
         if node_type in [:user, :basin]
             push!(used_nodes, node_id)
-
-        elseif length(inoutflow_ids(graph, node_id)) > 2
+        elseif count(x -> true, inoutflow_ids(graph, node_id)) > 2
```
Hofer-Julian, line 16: Why is that preferable? I've assumed that `length` simply consumes unsized iterators, just like your code using `count`.

Hofer-Julian: I see :)

visr:
> this might be due to the dictionary to map from an edge to an index in the dense flow vector.

This didn't show up in the profile so I think that's fast enough. Though the dict lookup to check if an edge is a flow edge did show up. This was for a model with only flow edges. `get_tmp` also didn't show up.

@SouthEndMusic is this ok to merge for you? I think this fixes the issue enough to make a release, we can look after into more speedups.

SouthEndMusic:
@visr disregard my comments above, those are about https://github.com/Deltares/Ribasim/pull/828 which I am working on at the moment. 

If this works and indeed gives the required speedup it is fine by me, I would like you to walk me trough how it works this afternoon if you have time

SouthEndMusic (commented):
LGTM

PR #832, opened 2023-11-27
Add plugin reloader to QGIS installation

deltamarnix:
Make it easier to debug the QGIS plugin by simply reloading it with CTRL+F5


visr (approved):
Does this automatically configure the plugin reloader to reload the Ribasim QGIS plugin on Ctrl-F5?

image

If that's easy to do, it might be a nice addition: https://github.com/borysiasty/plugin_reloader/blob/36ddcb2a71fc117ba754ad5ccd9d86b463b28986/reloader_plugin.py#L36-L40

PR #840, opened 2023-11-27
Declare numpy dependency of ribasim-python

visr:
Found in https://github.com/conda-forge/ribasim-feedstock/pull/4

It is already mentioned as a dependency in ribasim_testmodels and ribasim_api.


Hofer-Julian (approved):
Nice find!

PR #842, opened 2023-11-28
Demo branch

visr:
I noticed while preparing a demo that allocated water was subtracted from capacities, but not from the source. This mattered for my model where the edges have infinite capacities. It would allocate more water than came in when using multiple priorities.

Putting this as a draft to clean up later. I also want d12430cce7592128d223213e9b48a4ca005e3be4 locally, which should not be merged but allows editing parameters for the demo model at least. See #841 for context.

PR #845, opened 2023-11-28
edge validation fix

visr:
Separated out of #842. Testing this path will be covered by the test model from https://github.com/Deltares/Ribasim/pull/842#issuecomment-1830045720 that we will add.


Hofer-Julian (approved):

visr:
Closing in favor of #848 which contains this fix and another one.

PR #853, opened 2023-11-30
Plot fix

SouthEndMusic:
Fixes the plot of the PID example in the docs as the order of the data in a data frame changed: https://deltares.github.io/Ribasim/python/examples.html#model-with-pid-control.


visr (approved):

evetion:
@SouthEndMusic check Teams.

PR #857, opened 2023-11-30
Fix plotting PID listen edges

visr:
This was broken in the Python overhaul, `self.network.node.static.df` needed to be updated to `self.network.node.df`.
I added a test to make sure the plotting code at least doesn't error, using a model that has both discrete as well as PID control.

The rest of the diff is to fix printing PID listen edges if there is only time dependent PID control, like in https://deltares.github.io/Ribasim/python/examples.html#model-with-pid-control, since we only checked if `if self.pid_control.static.df is not None`.

This now looks good, `discrete_control_of_pid_control`:
image


SouthEndMusic (approved):
LGTM

PR #859, opened 2023-12-01
Unit tests for qgis/core/model.py

### LINKED ISSUE #813: Create local QGIS unit tests
The tests can currently only run in the QGIS environment over docker. It would be easier in the workflow if we could create unit tests that can simply be run and debugged in vs code.

This might already work, but we have no tests proving it.

deltamarnix:
Can be run by calling `pixi run code.` to ensure the environment is set correctly, then simply run it by pressing F5 to debug the local file. The tests also run over `pixi run test-ribasim-qgis` in docker.

The tests cannot be found by VS code, since we are mixing `unittests` and `pytest` and VS Code can only handle one type. Also, it seems to try to activate a conda environment that is not present.

Fixes #813

### DIFF
--- a/ribasim_qgis/tests/core/test_model.py
+++ b/ribasim_qgis/tests/core/test_model.py
@@ -13,9 +13,7 @@ class TestModel(unittest.TestCase):
     data_folder_path = tests_folder_path / "data"
 
     def test_get_directory_path_from_model_file(self):
-        """_summary
-        Tests that get_directory_path_from_model_file() can resolve paths from a toml file.
-        """
+        """Tests that get_directory_path_from_model_file() can resolve paths from a toml file."""
         for test_case in [("input_dir", "."), ("results_dir", "results")]:
             with self.subTest(property=test_case[0], value=test_case[1]):
                 path = get_directory_path_from_model_file(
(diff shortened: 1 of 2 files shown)


ribasim_qgis/tests/core/test_model.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class TestModel(unittest.TestCase):
+    tests_folder_path = Path(__file__).parent.parent.resolve()
+    data_folder_path = tests_folder_path / "data"
+
+    def test_get_directory_path_from_model_file(self):
+        """_summary
+        Tests that get_directory_path_from_model_file() can resolve paths from a toml file.
+        """
```
Hofer-Julian, line 18:
```suggestion
        """Tests that get_directory_path_from_model_file() can resolve paths from a toml file."""
```
We try to follow numpydoc style: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

Without the `summary` its compliant I would say

ribasim_qgis/tests/core/test_model.py
```diff
@@ -0,0 +1,48 @@
...
+                self.assertTrue(
+                    path.is_relative_to(self.tests_folder_path),
+                    msg=f"Path '{path}' is not relative to {self.tests_folder_path}",
+                )
+                self.assertEqual(path, self.data_folder_path / test_case[1])
+
+    def test_get_database_path_from_model_file(self):
+        """_summary
+        Tests that get_database_path_from_model_file() can find the input directory and appends the database.gpkg to it.
+        """
```
Hofer-Julian, line 35: Also remove the `_summary` here. For some reason, GitHub doesn't let me add a suggestion here.

Hofer-Julian (commented):
Two small comments, apart from that it looks good.

PR #873, opened 2023-12-06
Make QGIS tables editable

### LINKED ISSUE #841: QGIS support editing attribute tables
**What**
In QGIS, only the records in the attribute tables of Node and Edge can be edited or deleted, but not any of the others.

**Why**
Without this it becomes nearly impossible to create a model, since you can create records but not edit or delete them. It is impossible to change parameters in an existing model.

**How**
QGIS requires a primary key to enable these features (https://github.com/qgis/QGIS/issues/43664). We have this only for the Node and Edge tables. I confirmed this by setting the primary key for only the "Basin / state" table with the following patch of `input_base.py`:

```python
        if table == "Basin / state":
            dtype = {"node_id": "INTEGER PRIMARY KEY"}
        else:
            dtype = {}

        assert self.df is not None
        with closing(connect(temp_path)) as connection:
            self.df.to_sql(
                table,
                connection,
                index=False,
                if_exists="replace",
                dtype=dtype,
            )
```

I suppose we could define a primary key for each data table. For some it would be just `node_id`, for others you need a combination of columns for it to be unique, like (`node_id`, `control_state`), or (`subgrid_id`, `node_id`).

It should be noted that Arrow tables cannot be edited at all in QGIS, but that is less of an issue since they are more used for long timeseries, which you probably don't want to edit by hand anyway.

**Additional context**
Here you can see the desired behavior on "Basin / state" due to the above patch:
image
And here you can see the current behavior, where you can only edit cells for records that you are newly creating:
image

deltamarnix:
Add an fid column to all tables and make them a PRIMARY KEY AUTOINCREMENT.
From now on, QGIS understands that they are editable.

Fixes #841

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -4,7 +4,7 @@ from collections.abc import Callable, Generator
 from contextlib import closing
 from contextvars import ContextVar
 from pathlib import Path
-from sqlite3 import Connection, Cursor, connect
+from sqlite3 import Connection, connect
 from typing import (
     Any,
     Generic,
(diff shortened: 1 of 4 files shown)


visr (approved):
This looks good, great to have this feature.

python/ribasim/ribasim/input_base.py
```diff
@@ -241,8 +241,20 @@ def _write_table(self, temp_path: Path) -> None:
         """
         table = self.tablename()
         assert self.df is not None
+
+        # Add `fid` to all tables as primary key
+        # Enables editing values manually in QGIS
+        df = self.df.copy()
```
evetion, line 247: Why can't we add the column in place?

deltamarnix: Because you don't want it to be part of the model. If a user wants to do other operations on the dataframe, you don't want to bother them with extra columns.

PR #874, opened 2023-12-07
Don't prefix meta to fid columns.

evetion:
Fix on top of #873

### DIFF
--- a/python/ribasim/tests/test_io.py
+++ b/python/ribasim/tests/test_io.py
@@ -8,7 +8,7 @@ from pydantic import ValidationError
 from ribasim import Pump
 
 
-def __assert_equal(a: DataFrame, b: DataFrame, ignore_metas: bool = True) -> None:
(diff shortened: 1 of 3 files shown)


python/ribasim/tests/test_io.py
```diff
@@ -27,6 +27,11 @@ def __assert_equal(a: DataFrame, b: DataFrame, ignore_metas: bool = True) -> Non
         a["time"] = a.time.astype("datetime64[ns]")
         b["time"] = b.time.astype("datetime64[ns]")
 
+    if "fid" in a:
```
deltamarnix, line 30: Should we get rid of the `ignore_metas` that I added specifically to fix the tests? It might no longer be required.

evetion: Feel free to takeover this branch to try so, I didn't dare to touch that code ;)

deltamarnix (approved):

PR #885, opened 2023-12-11
Don't require consecutive Node IDs

### LINKED ISSUE #694: numbering of nodes

visr:
Fixes #694

It basically was already fixed in the metagraphs refactor, I just had to remove the validation check from Python.

This also adds support for 0, but not negative IDs, as I feel those could be confusing.

I used the trivial model to add some ID variations, and directly added some tests of the Arrow file results that we didn't really have yet.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -248,7 +248,7 @@ function create_callbacks(
             save_subgrid_level,
             saved_subgrid_level;
             saveat,
-            save_start = false,
+            save_start = true,
         )
         push!(callbacks, export_cb)
     end
--- a/core/test/run_models_test.jl
+++ b/core/test/run_models_test.jl
@@ -1,15 +1,89 @@
 @testitem "trivial model" begin
     using SciMLBase: successful_retcode
+    using Tables: Tables
+    using Tables.DataAPI: nrow
+    using Dates: DateTime
+    import Arrow
+    using Ribasim: timesteps
 
     toml_path = normpath(@__DIR__, "../../generated_testmodels/trivial/ribasim.toml")
     @test ispath(toml_path)
     model = Ribasim.run(toml_path)
     @test model isa Ribasim.Model
     @test successful_retcode(model)
+    (; p) = model.integrator
 
-    # The exporter interpolates 1:1 for three subgrid elements, but shifted by 1.0 meter.
-    subgrid = model.integrator.p.subgrid
-    @test all(diff(subgrid.level) .≈ 1.0)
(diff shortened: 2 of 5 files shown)


core/src/bmi.jl
```diff
@@ -248,7 +248,7 @@ function create_callbacks(
             save_subgrid_level,
             saved_subgrid_level;
             saveat,
-            save_start = false,
+            save_start = true,
```
visr, line 251: @Huite I'm now saving initial subgrid levels as well, just like the regular basin.arrow. I assume you copied this from flow, which is undefined at t0 so we don't save it.

core/test/run_models_test.jl
```diff
@@ -1,15 +1,89 @@
...
+        @test flow.to_node_id[1:4] == [6, typemax(Int), typemax(Int), 0]
+
+        @test basin.storage[1] == 1.0
+        @test basin.level[1] ≈ 0.044711584
+
+        # The exporter interpolates 1:1 for three subgrid elements, but shifted by 1.0 meter.
+        @test length(p.subgrid.level) == 3
+        @test diff(p.subgrid.level) ≈ [-1.0, 2.0]
+        # TODO The original subgrid IDs are lost and mapped to 1, 2, 3
+        @test subgrid.subgrid_id[1:3] == [11, 22, 33] broken = true
```
visr, line 83: @Huite I assume you want to keep the original subgrid IDs right? If so let's create an issue to fix this. We'd have to store them in the `struct Subgrid`.

Hofer-Julian (approved):

core/src/bmi.jl
Huite: Yes, this sounds fine.

core/test/run_models_test.jl
Huite: Yes, I think we should keep them around.

visr: Ok, I opened #895

PR #886, opened 2023-12-12
Add flow cost

SouthEndMusic:
Adding flow cost terms to linear objective functions. See also the added documentation.

### DIFF
--- a/core/src/allocation.jl
+++ b/core/src/allocation.jl
@@ -661,7 +661,7 @@ function set_objective_priority!(
         ex = sum(problem[:F_abs])
     end
 
-    demand_max = 0
+    demand_max = 0.0
 
     for edge_id in edge_ids
         node_id_user = edge_id[2]
@@ -718,7 +718,7 @@ function set_objective_priority!(
             JuMP.add_to_expression!(ex, cost_per_flow * flow)
         end
     elseif objective_type == :linear_relative
-        if demand_max > 0
+        if demand_max > 0.0
             cost_per_flow = 0.5 / (demand_max * length(F))
             for flow in F
                 JuMP.add_to_expression!(ex, cost_per_flow * flow)
--- a/core/src/allocation.jl
+++ b/core/src/allocation.jl
@@ -661,6 +661,8 @@ function set_objective_priority!(
         ex = sum(problem[:F_abs])
     end
 
+    demand_max = 0
+
     for edge_id in edge_ids
         node_id_user = edge_id[2]
         if graph[node_id_user].type != :user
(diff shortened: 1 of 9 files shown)


core/test/docs.toml
```diff
@@ -16,7 +16,7 @@ time = "basin/time.arrow"
 [allocation]
 timestep = 86400
 use_allocation = true
-objective_type = "quadratic_relative"
+objective_type = "linear_absolute"
```
visr, line 19:
```suggestion
objective_type = "linear_absolute"  # optional, default "linear_absolute"
```

And could you add such a comment also to `use_allocation` and `timestep`? The latter is optional, but required if `use_allocation`.

python/ribasim/ribasim/models.py
visr, line 1: Thanks for filing #887. To not get hold up on that, shall we change our pixi files to use datamodel-code-generator 0.24.2?

core/src/allocation.jl
```diff
@@ -661,6 +661,8 @@ function set_objective_priority!(
         ex = sum(problem[:F_abs])
     end
 
+    demand_max = 0
```
visr, line 664:
```suggestion
    demand_max = 0.0
```

Later you do `demand_max = max(demand_max, d)`, which means that `demand_max` becomes either an Int or Float64.

python/ribasim/ribasim/models.py
Hofer-Julian: If that fixes it, yes.
Ideally within that PR already.

core/test/docs.toml
```diff
@@ -14,9 +14,9 @@ results_dir = "results" # required
 time = "basin/time.arrow"
 
 [allocation]
-timestep = 86400
-use_allocation = true
-objective_type = "quadratic_relative"
+timestep = 86400                   # optional (required if use_allocation = true), default 86400
+use_allocation = true              # optional, default true
```
visr, line 18:
```suggestion
use_allocation = false             # optional, default true
```

Default is false. I think requiring modelers to opt-in still makes sense, so they can focus on getting a correct model without allocation first.

core/test/allocation_test.jl
```diff
@@ -75,7 +78,14 @@ end
     @test objective isa JuMP.AffExpr # Affine expression
     @test :F_abs in keys(problem.obj_dict)
     F_abs = problem[:F_abs]
-    @test objective == F_abs[NodeID(5)] + F_abs[NodeID(6)]
+    @test string(objective) == string(
```
visr, line 81: Both these string representation tests are still failing. What is the reason you moved away from testing the objective function object directly?

SouthEndMusic: Before the objective function only had trivial coefficients (namely 1). With non-trivial float coefficients I could not get equality tests to pass on the object itself, I'll have another look at it.

SouthEndMusic: I think the tests fail because the order of the terms in the expression is not constant. I'll test it in a different way

visr (approved):

PR #891, opened 2023-12-13
Test Ribasim Python with Python 3.12

### LINKED ISSUE #654: test python 3.12

Hofer-Julian:
Fixes #654

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -23,6 +23,7 @@ jobs:
         python-version:
           - "3.10"
           - "3.11"
+          - "3.12"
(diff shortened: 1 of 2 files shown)


deltamarnix (approved):
Seems legit. The tests are green, the packages get installed. What more do you want?

Issue #852, opened 2023-11-30
Add validation rules to documentation


SouthEndMusic:
**What**
The Ribasim core (and Ribasim Python as well to a lesser extent) enforce validation rules for model schematization. These rules are not yet present in the documentation and thus should be added.

**Why**
So that modelers can keep these rules into account when creating a schematization, instead of going trough a possibly lengthy trial and error process before Ribasim accepts the model.

**How**
It might be best to add a validation page to the 'Julia core' section of the docs. Many rules can be written down concisely in a tabular format, e.g. a square table with binary data of allowed outflow neighbors (maybe one for flow and one for control), or a table with the bounds of allowed number of in-neighbors and out-neighbors.

PR #896, opened 2023-12-14
Validation rules in docs

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/852.


Hofer-Julian:
The docs workflow does not seem to be happy with your changes

docs/core/equations_files/validation_tables/connectivity.jl
visr, line 1: https://deltares.github.io/Ribasim/core/validation.html looks good!

Does this empty .jl file serve any purpose or can it be removed?

Hofer-Julian: I just pinged @SouthEndMusic for that as well :)

PR #900, opened 2023-12-15
Update PackageCompiler to 2.1.6

visr:
This may help with compilation time, let's try it.

https://github.com/JuliaLang/PackageCompiler.jl/releases/tag/v2.1.16


visr:
Nice, the ribasim_cli build now takes as long as the libribasim build, both ~1h. Still too long, but progress.

Hofer-Julian (approved):
One step after the other :P

Issue #816, opened 2023-11-21
Add allocation example in `Examples.ipynb`


SouthEndMusic:
Currently there is no example of allocation in `Examples.ipynb` yet. It probably suffices to copy a model from the tests, properly document it and generate some plots.

visr:
I propose we use the allocation demo model from https://github.com/Deltares/Ribasim/pull/842 for this.

PR #901, opened 2023-12-15
Allocation example

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/816


visr (approved):

visr:
Looks good to me! Shall we also run this test model during core tests?
There is a merge conflict that needs to be resolved, but otherwise good to go.

Issue #601, opened 2023-09-14
replace (time, node_id) order by (node_id, time)


visr:
I thought I fixed the sorting issue by updating the sorting functions as described in https://github.com/Deltares/Ribasim/pull/597#issuecomment-1719792565.

But this fixed sorting actually exposed as issue fixed in 8bc58952ec5dfe0e4f5efa6d9060c529959be660. That now filters the time tables to create contiguous vectors with the timeseries per node. For this access pattern it would be faster to have it first sorted on node_id, and then on time, instead of the other way around.

I initially went for time first since we were reading in time series at runtime, and then loading them in for all nodes. But now we create an interpolation object per node, that takes a vector of the entire timeseries. So it's better to sort by node_id first like all other tables.

- [ ] update documented sort order in usage.qmd
- [ ] update sort methods in python
- [ ] update sort methods in julia
- [ ] address TODO in 8bc58952ec5dfe0e4f5efa6d9060c529959be660 and use views like below
- [ ] use interpolation objects instead of `update_basin`, which still relies on time being sorted first
- [ ] use interpolation objects (which already seem to exist) instead of `update_tabulated_rating_curve!`, which still relies on time being sorted first

```julia
rows = searchsorted(time.node_id, node_id)
time_id = view(time, rows)
```

PR #903, opened 2023-12-15
Fix and change table sorting in Python

visr:
Fixes the first two items of #601.

Also adds a test.


docs/core/usage.qmd
visr, line 1: I reordered the columns such that the first entry is the top-level sort, and lower level sorts are followed, which I hope will make it easier to interpret the sorting from these tables.

python/ribasim/ribasim/input_base.py
```diff
@@ -426,7 +430,7 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath, **kwargs):
             getattr(self, field)._save(
                 directory,
                 input_dir,
-                sort_keys=self._sort_keys.get("field", ["node_id"]),
```
visr, line 429: This was broken, since it got the `"field"` instead of `field`, so it always only sorted by `node_id`. I decided to get rid of the default, since the default is usually not right, and makes it easier to forget to add the sort keys.

python/ribasim/ribasim/config.py
```diff
@@ -185,20 +201,29 @@ class DiscreteControl(NodeModel):
         default_factory=TableModel[DiscreteControlLogicSchema]
     )
 
+    # TODO complete the ordering, and make sure it aligns with truth_state
+    _sort_keys: dict[str, list[str]] = {"static": ["node_id"], "logic": ["node_id"]}
```
deltamarnix, line 205: I tried to understand the code from a Julia point of view. DiscreteControlLogic seems to be coupled to a `StructVector`, which calls `sorted_table!`, which will `sort_by_function`. This function has multiple implementations. `DiscreteControlLogicV1` does not have a specific implementation. Therefore Julia will `sort_by_time_id`. So that's not the same as what is stated here. Correct?

deltamarnix: Ah, found something else. `sort_by_function(table::StructVector{<:Legolas.AbstractRecord}) = sort_by_id`
So, indeed they are sorted by node_id, just as mentioned in validation.jl

python/ribasim/ribasim/config.py
```diff
@@ -82,6 +82,8 @@ class Terminal(NodeModel):
         default_factory=TableModel[TerminalStaticSchema]
     )
 
+    _sort_keys: dict[str, list[str]] = {"static": ["node_id"]}
```
deltamarnix, line 85: I don't agree with adding this protected variable in multiple places. There are other ways to do this. For example, add an `__init__` function everywhere to add the required keys to the dict.

But I am trying to look a bit further. It would be better if we receive this information from the julia code. For example, by adding it in the schemas and reading the schemas in python. Would that be something that is possible with pydantic?

visr: `_sort_keys` was already used, falling back to a wrong default if it wasn't present. This PR removes the wrong default and requires explicitly stating the correct sorting order.

If possible I agree it would be nice to share this or run this through pydantic. I'd like to keep this PR focused on a bugfix and not add a refactor on top.

deltamarnix: @visr @evetion take a look at feat/sort-keys-in-tablemodel. I don't have it working yet. But to me it seems like we need to put the sort keys inside of the table model as some class var which depends per initialization. Or maybe in another way. Just a Field?

deltamarnix: I have been playing a bit more with this, and it seems to me that the best option would be to make the sort_keys an instance Field

```py
sort_keys: list[str] = Field(init_var=True, exclude=True, repr=False)
```

I set it to init_var, so it can never be changed afterwards, and I set it to exclude, because the data is always set in python.

The only problem that I still had was the moment that you set a pandas DataFrame, because then it tells me that:

> pydantic_core._pydantic_core.ValidationError: 1 validation error for Terminal
static.sort_keys
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]        
    For further information visit https://errors.pydantic.dev/2.5/v/list_type
    
I believe we could solve this by using a different scheme for custom data types, instead of the `@model_validator` as described here: https://docs.pydantic.dev/latest/concepts/types/#custom-types

evetion: Is init_var=True needed? Because if you set a DataFrame, the TableModel already exists, including its specific sort_keys. Ideally we get that from somewhere.

> But I am trying to look a bit further. It would be better if we receive this information from the julia code. For example, by adding it in the schemas and reading the schemas in python. Would that be something that is possible with pydantic?

Fully agree, but I couldn't find a way of doing so. It requires some custom info passed to the JSON Schema, the https://github.com/koxudaxi/datamodel-code-generator/, to read it and pass it into some class as a private field, or Config option, that we here can pick up on.

deltamarnix: I created #956 for the first effort on getting the sort_keys in pydantic. You could decide to enhance the ticket, or otherwise just make a new ticket to get the info from Julia.

deltamarnix (approved):
I have not been able to test this. Please change the PR information before merging and test yourself.

PR #911, opened 2023-12-18
Logger writes to both Terminal and File

### LINKED ISSUE #502: FR: Log to file
- Config should have an option to log to a file
- Option to disable logging to stdout (i.e. a quiet modus?)
- Option to filter which Modules log

### LINKED ISSUE #747: Automate tasks with pixi
- [x] pixi run generate-testmodels # e.g. `pixi run generate-testmodels basic` should generate the basic model
- [x] pixi run ribasim-model # `pixi run ribasim /path/to/ribasim.toml`, implementation might be similar to `using Ribasim; Ribasim.run(ARGS)` (in progress as part of #502)
- [x] pixi run resolve-manifests # run pkg resolve on the checked in manifests (not needed after #740)

Follow up to https://github.com/Deltares/Ribasim/issues/598

deltamarnix:
Fixes #502 
Fixes #747

Using `LoggingExtras` as a basis to write logs to two places at the same time.
The verbosity only applies to the file logger.
The terminal logger will always be on Progress level (-1).

The settings.json was adjusted to exclude some folders that do not help in our development.


Project.toml
```diff
@@ -3,7 +3,10 @@ authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
 description = "Meta-project used to share the Manifest of Ribasim and its dependents"
 
 [deps]
+ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 Ribasim = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
+TestEnv = "1e6cf692-eddd-4d53-88a5-2d735e33781b"
```
deltamarnix, line 9: I've added these to the main Ribasim project to make it easier to develop.

deltamarnix: This Project.toml is not the core/Project.toml. Therefore these packages are only loaded when you start the right environment. Nothing changed for the build step. It's to ensure that every developer has the same packages installed. @visr what do you think? Should I remove these lines?

evetion: Ah, sorry, then this is the dev Project.toml, that should be fine.

.vscode/settings.json
```diff
@@ -22,5 +22,23 @@
...
+        "**/.hg": true,
+        "**/CVS": true,
+        "**/.DS_Store": true,
+        "**/Thumbs.db": true,
+        ".pixi": true,
+        ".mypy_cache": true,
+        ".pytest_cache": true,
+        ".ruff_cache": true,
+        "juliaup": true
+    },
```
visr, line 38: Would this also work? Then we don't need to keep maintaining two lists. `files.exclude` points to this setting.


```suggestion
    "explorer.excludeGitIgnore": true,
```

core/src/Ribasim.jl
```diff
@@ -46,8 +47,7 @@ using Graphs:
     rem_edge!
 
 using Legolas: Legolas, @schema, @version, validate, SchemaVersion, declared
-using Logging: current_logger, min_enabled_level, with_logger
-using LoggingExtras: EarlyFilteredLogger, LevelOverrideLogger
+using Logging: with_logger, LogLevel, AbstractLogger, SimpleLogger
```
visr, line 50:
```suggestion
using Logging: with_logger, LogLevel, AbstractLogger
```

SimpleLogger isn't used.

.vscode/settings.json
Hofer-Julian: As discussed with @deltamarnix, I would miss those folders.
Maybe the setting that @visr suggested should just be added globally by people who prefer that?

core/src/bmi.jl
```diff
@@ -660,37 +660,9 @@ BMI.get_end_time(model::Model) = seconds_since(model.config.endtime, model.confi
 BMI.get_time_units(model::Model) = "s"
 BMI.get_time_step(model::Model) = get_proposed_dt(model.integrator)
```
visr, line 662: Shall we add this back? Then we can also keep using `Ribasim.run(toml_path)` in the tests.

```suggestion

"""
    run(config_file::AbstractString)::Model
    run(config::Config)::Model
Run a [`Model`](@ref), given a path to a TOML configuration file, or a Config object.
Running a model includes initialization, solving to the end with `[`solve!`](@ref)` and writing results with [`BMI.finalize`](@ref).
"""
run(config_file::AbstractString)::Model = run(Config(config_file))

```

core/src/main.jl
```diff
@@ -0,0 +1,48 @@
+function help(x::AbstractString)::Cint
+    println(x)
+    println("Usage: ribasim path/to/model/ribasim.toml")
+    return 1
+end
+
+function main(ARGS::Vector{String})::Cint
```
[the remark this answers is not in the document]
visr: It would also be nice to have a `main(toml_path::AbstractString) = main([toml_path])`. Most users will use a CLI, but for testing branches ease of use for the interactive API is worth facilitating. Then we also don't need the square brackets in `examples.ipynb`.

visr (changes requested):

.vscode/settings.json
deltamarnix: Reverted

deltamarnix:
@evetion @Hofer-Julian @visr I reverted the .vssettings and reverted the removal of run(string), so that is now available again.

.vscode/settings.json
```diff
@@ -22,5 +22,10 @@
         "."
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "julia.lint.disabledDirs": [
+        ".pixi",
+        "juliaup"
```
Hofer-Julian, line 28: That would be now "utils/juliaup"

visr (approved):

visr:
> Moved the main entrypoint

This we did together last year, it is related since before we only did logger setup in the ribasim_cli module, and that was now moved over to the core. Using `Ribasim.main(ARGS)` is convenient since from Julia 1.11 onwards that will be the standardized entrypoint.

PR #913, opened 2023-12-20
Move export QUARTO_PYTHON to env_setup

deltamarnix:
@Hofer-Julian Not sure how you used to test this. Was something broken with the normal python that was used by Quarto?

### DIFF
--- a/env_setup.bat
+++ b/env_setup.bat
@@ -1,1 +1,2 @@
 set JULIAUP_DEPOT_PATH=%~dp0
+set QUARTO_PYTHON=python
--- a/env_setup.sh
+++ b/env_setup.sh
@@ -1,3 +1,5 @@
 #!/bin/bash
 JULIAUP_DEPOT_PATH=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 export JULIAUP_DEPOT_PATH
+QUARTO_PYTHON=python
+export QUARTO_PYTHON
--- a/pixi.toml
+++ b/pixi.toml
@@ -33,11 +33,11 @@ build-julia-docs = { cmd = "julia --project=docs docs/make.jl", depends_on = [
     "instantiate-julia",
 ] }
 quartodoc-build = { cmd = "quartodoc build && rm objects.json", cwd = "docs" }
-quarto-preview = { cmd = "export QUARTO_PYTHON=python && quarto preview docs", depends_on = [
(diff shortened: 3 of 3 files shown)


Hofer-Julian:
> @Hofer-Julian Not sure how you used to test this. Was something broken with the normal python that was used by Quarto?

It detected the wrong Python on windows. Unfortunately quarto uses pylauncher on Windows instead of simply using the first python in Path. Feel free to experiment with removing it, maybe it's fixed nowadays.

Hofer-Julian:
This also means that you only have to set the env var on windows. You can leave UNIX alone

deltamarnix:
@Hofer-Julian How do I see that it 'detects the wrong python'? What was going wrong?

Hofer-Julian:
> @Hofer-Julian How do I see that it 'detects the wrong python'? What was going wrong?

Windows adds a dummy python on every machine, which does nothing but redirecting you to the windows apps store. On my machine, it detected that.

deltamarnix:
So you're saying it doesn't detect python at all and therefore doesn't start the preview.

Hofer-Julian:
I've tried it again, and without `QUARTO_PYTHON` it still breaks for me:

```
  File "C:\checkouts\github.com\Deltares\Ribasim\.pixi\env\Library\share\quarto\jupyter\jupyter.py", line 21,
in <module>
    from notebook import notebook_execute, RestartKernel
  File "C:\checkouts\github.com\Deltares\Ribasim\.pixi\env\Library\share\quarto\jupyter\notebook.py", line 17,
 in <module>
    import nbformat
ModuleNotFoundError: No module named 'nbformat'
Python 3 installation:
  Version: 3.10.10 (Conda)
  Path: C:/Users/hofer_jn/AppData/Local/mambaforge/python.exe
  Jupyter: (None)
```

deltamarnix:
@Hofer-Julian Can you tell me what is "it"? What are you running?

Hofer-Julian:
> @Hofer-Julian Can you tell me what is "it"? What are you running?

Fair enough. I ran `pixi run quarto-render`. During the run it picked up the wrong Python.

deltamarnix:
@Hofer-Julian This doesn't happen on my machine, because I don't have another python installed. Could you give it another go? I have changed it so that `QUARTO_PYTHON` is set during the activation of pixi.

Hofer-Julian (approved):
The current changes work fine on my machine.
It is also a bit cleaner now. Thanks, @deltamarnix!

Issue #912, opened 2023-12-19
Ability to merge two Ribasim models in Python


evetion:
@SouthEndMusic 

This would be helpful in writing larger testmodels that don't duplicate code, but can merge smaller test models.

Could be generically implemented as follows:
- *Merge method on the [TableModel](https://github.com/Deltares/Ribasim/blob/836dc7178bf06abd0e60d61029f1be787c04cc3a/python/ribasim/ribasim/input_base.py#L159) class*, that merges the existing table with another TableModel (e.g. PumpStatic) using pandas, taking care of unique ids, after checking the schemas are the same.
- *Merge method on the [NodeModel](https://github.com/Deltares/Ribasim/blob/836dc7178bf06abd0e60d61029f1be787c04cc3a/python/ribasim/ribasim/input_base.py#L384) class*, that takes another NodeModel (e.g. Pump) class and merges with it, by calling the merge method (see above) on all its children that are tablemodels.
- *Merge method on the [Model](https://github.com/Deltares/Ribasim/blob/836dc7178bf06abd0e60d61029f1be787c04cc3a/python/ribasim/ribasim/model.py#L100) class* that merges with another Model. It will call the above merge method for all its children that are Nodes.

The difficulty lies in making sure the renamed node_ids are either known at the Model merge level, and used in other methods, or the renaming of the node_ids follows a fixed schema. So when the Nodes and Edges are renamed, they still are linked to things like Pumps.

visr:
This is probably easier to implement now after #1110. Node IDs only have to be made unique per node type.

SouthEndMusic:
Closed in favor of https://github.com/Deltares/Ribasim/issues/1297.

PR #851, opened 2023-11-29
Main allocation network

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/768.

PR #914, opened 2023-12-20
Merging models

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/912.

The general design is as in the above mentioned issue. 2 things that are kept in mind when merging model B into model A:
- B itself should not be modified, so a copy of it is made and this copy is modified. This modified B is then passed on to merge methods on node and table level;
- Only the merge method on the model level makes modifications, not the ones on node and table level. Only the table merge method does asserts for node ID conflicts (so allocation network IDs are not checked, as these are not unique. Also node/edge geometries are not checked).


SouthEndMusic:
@evetion mypy still complains about a few things, I don't know what the best way would be to fix them.

python/ribasim/tests/test_model.py
```diff
@@ -143,3 +144,19 @@ def test_write_adds_fid_in_tables(basic, tmp_path):
...
+
+
+def test_model_merging(basic, subnetwork, tmp_path):
+    model = deepcopy(basic)
+    model_added = deepcopy(subnetwork)
+    model.merge_model(model_added)
+    model.merge_model(model_added)
+    assert (model.network.node.df.index == range(1, 44)).all()
+    assert model.max_allocation_network_id() == 2
+    for node_type, node_added in model_added.nodes().items():
```
SouthEndMusic, line 156: Regarding this loop: I tried comparing models directly, e.g.
```python
assert model_added == subnetwork
```

but I kept running into 'maximum recursion depth reached' errors, and I could not find out why; there is no loop in the model object as far as I could find.

python/ribasim_testmodels/ribasim_testmodels/allocation.py
```diff
@@ -1065,3 +1071,123 @@ def allocation_example_model():
     )
 
     return model
+
+
+def main_network_with_subnetworks_model():
```
SouthEndMusic, line 1076: I'll pick up this model in https://github.com/Deltares/Ribasim/pull/851 for testing.

python/ribasim/ribasim/geometry/node.py
```diff
@@ -59,6 +63,34 @@ def node_ids_and_types(*nodes):
 
         return node_id, node_type
 
+    def translate_spacially(
```
evetion, line 66: Both functions here are identical to the one in edge.py, try to generalize these.

python/ribasim/ribasim/input_base.py
```diff
@@ -202,6 +216,44 @@ def node_ids(self) -> set[int]:
 
         return node_ids
 
+    def offset_node_ids(self, offset_node_id: int) -> "TableModel[TableT]":
```
evetion, line 219: This does also create a copy, which could be reflected in the name. Maybe we want to have a mutable version as well?

python/ribasim/ribasim/input_base.py
```diff
@@ -202,6 +216,44 @@ def node_ids(self) -> set[int]:
...
+                "node_id",
+                "from_node_id",
+                "to_node_id",
+                "listen_feature_id",
+            ]:
+                if hasattr(df, name_column):
+                    df[name_column] += offset_node_id
+        return copy
+
+    def merge_table(
```
evetion, line 235: I'd expected this function to also fix any index/id clashes by calling some form of offset?

python/ribasim/ribasim/input_base.py
```diff
@@ -421,6 +473,61 @@ def node_ids_and_types(self) -> tuple[list[int], list[str]]:
...
+        """Add the same offset to all node IDs in all underlying tables."""
+        node_copy = deepcopy(self)
+        for field in node_copy.fields():
+            attr = getattr(node_copy, field)
+            if isinstance(attr, TableModel):
+                table = attr
+                setattr(
+                    node_copy,
+                    field,
+                    table.offset_node_ids(offset_node_id),
```
evetion, line 486: This will again make a copy of the already copied table (by deepcopy)!

python/ribasim/ribasim/model.py
```diff
@@ -58,6 +61,75 @@ def n_nodes(self):
 
         return n
 
+    def translate_spacially(
```
evetion, line 64: Time to do a search-replace on spacially ;)

python/ribasim/ribasim/model.py
```diff
@@ -376,6 +448,68 @@ def reset_contextvar(self) -> "Model":
...
+    def max_node_id(self) -> int:
+        return self.network.node.df.index.max()
+
+    def max_allocation_network_id(self) -> int:
+        m = self.network.node.df.allocation_network_id.max()
+        if pd.isna(m):
+            m = 0
+        return m
+
+    def merge_model(
```
evetion, line 460: nit: Your function names are sometimes a bit verbose. It's not wrong, but you might want to think about it. This would result in `modela.merge_model(modelb)`. Purely `merge` would thus also work.

python/ribasim/ribasim/model.py
```diff
@@ -376,6 +448,68 @@ def reset_contextvar(self) -> "Model":
...
+        self,
+        model_added: "Model",
+        offset_node_id: int | None = None,
+        offset_allocation_network_id: int | None = None,
+        offset_spacial: tuple[float, float] = (0.0, 0.0),
+        inplace: bool = True,
+    ):
+        """
+        Merge copies of the nodes and edges of an added model into this model.
+        The added model is not modified, but the following modificadions are made to the added data:
```
evetion, line 470:
```suggestion
        The added model is not modified, but the following modifications are made to the added data:
```

A spellcheck in your IDE might help you 😄

python/ribasim/ribasim/model.py
```diff
@@ -376,6 +448,68 @@ def reset_contextvar(self) -> "Model":
...
+        )
+        min_offset_node_id = model.max_node_id()
+        min_offset_allocation_network_id = model.max_allocation_network_id()
+
+        if offset_allocation_network_id is None:
+            offset_allocation_network_id = min_offset_allocation_network_id
+        else:
+            assert (
+                offset_allocation_network_id >= min_offset_allocation_network_id
+            ), f"The allocation network ID offset must be at least the maximum allocation network ID of the main model ({min_offset_allocation_network_id}) to avoid conflicts."
```
evetion, line 492: If you always check this, you might as well set the default argument to 0 instead of None.

python/ribasim/ribasim/model.py
```diff
@@ -443,11 +577,22 @@ def plot(self, ax=None) -> Any:
         if ax is None:
             _, ax = plt.subplots()
             ax.axis("off")
+
         self.network.edge.plot(ax=ax, zorder=2)
         self.plot_control_listen(ax)
         self.network.node.plot(ax=ax, zorder=3)
 
-        ax.legend(loc="lower left", bbox_to_anchor=(1, 0.5))
+        handles, labels = ax.get_legend_handles_labels()
```
evetion, line 585: This should ideally not be part of the PR (but will ignore it here)?

python/ribasim/tests/test_model.py
evetion: It might be that a comparison somewhere (especially given that you implement `__eq__` yourself) is calling itself again.

python/ribasim/tests/test_model.py
```diff
@@ -143,3 +145,20 @@ def test_write_adds_fid_in_tables(basic, tmp_path):
         assert "fid" in df.columns
         fids = df.get("fid")
         assert fids.equals(pd.Series(range(1, len(fids) + 1)))
+
+
+def test_model_merging(basic, subnetwork, tmp_path):
+    model = deepcopy(basic)
+    model_added = deepcopy(subnetwork)
+    model.merge_model(model_added)
+    model.merge_model(model_added)
```
evetion, line 154: Why twice?

evetion (changes requested):

python/ribasim/tests/test_model.py
SouthEndMusic: The added model is a subnetwork with `allocation_network_id = 1` (which will probably be changed in the future because that is reserved for the main network), so I check with `assert model.max_allocation_network_id() == 2` whether a shift was added to the `allocation_network_id` of the second one. I added some clarification in the code. I'll also test the coordinate translation while I'm at it.

python/ribasim/ribasim/model.py
SouthEndMusic: You mean the whole subnetwork plotting thing? I guess it would have bean cleaner if that was done in a separate PR, but it did help with debugging of the model merging

python/ribasim/tests/test_model.py
SouthEndMusic: https://github.com/Deltares/Ribasim/issues/928

SouthEndMusic:
To do:
- [ ] Process comments by Maarten above
- [ ] Remove `Model.merge` options: `offset_node_id` and `offset_allocation_network_id` are always computed
- [ ] `offset_spatial`: make functions in tests
- [ ] `inplace`: don't make copy of entire model, only loop over nodes and modify those
- [ ] https://github.com/Deltares/Ribasim-NL/pull/49

python/ribasim/ribasim/input_base.py
SouthEndMusic: I've added a check that the tables have no node IDs in common.

evetion:
- [ ] Remove special code to merge models (default states, after its removed by another PR)
- [ ] Make merge methods on lower levels (node, tables) private
- [ ] Make use of model_copy(deep=True) and check whether it is needed everywhere
- [ ] Make inplace merge private (only merge (which copies) on model level)

visr:
This is paused, after #1110 we can decide how to proceed with this.

visr:
Closing this PR, since so many things changed that it probably needs a new start. The PR can still be used as inspiration for #912.

PR #919, opened 2024-01-08
Remove QGIS checkboxes in dataset widget

### LINKED ISSUE #916: QGIS: Remove checkboxes from dataset widget
**What?**
The checkboxes in the dataset widget in QGIS don't seem to have a function. When we look in code, there seems to be no connection to these checkboxes.

**Why?**
The interface is cluttered and the user may be confused on what these checkboxes might do. Especially in combination with the buttons underneath (Add to QGIS, Remove from Dataset). These buttons work with the selection instead.

**How?**
Make them simple list view items without a checkbox.

deltamarnix:
Fixes #916


Huite (approved):
lgtm

evetion:
Not sure why the CI is failing here, should investigate before merging though.

Hofer-Julian:
Seems to be fine now. I guess a re-run was enough?

Issue #921, opened 2024-01-08
Allocation main network initialization


SouthEndMusic:
**What**
There is already initialization of the optimization problems for subnetworks. This is also needed for the main network, which is slightly different.

**How**
- [x] Make sure that the model for the main allocation network is initialized first (just do them in ascending order by allocation network ID)
- [x] When initializing the main allocation model, collect edges to subnetworks (`allocation_network_id != 1`) in `Allocation.main_network_connections`.
- [ ] Each edge to a subnetwork is treated as a user in the objective function and constraints of the optimization problem for the main network.
*To clarify: yes, this means that subnetwork can be represented by multiple 'users' in the optimization problem for the main network if there are multiple edges connecting the main network to the subnetwork. This is because the subnetwork will have different demands for each such connecting edge determined by the subnetwork structure. The choice of how to treat the distribution of demand from a subnetwork over these multiple connecting edges has a significant effect on the behavior of the solver, I think especially for non-linear objective functions.*
- [ ] In the optimization problems for the subnetworks, add constraints for the supply/capacity coming from the main network.

As discussed with @visr, users within the main network are allowed.

Issue #922, opened 2024-01-08
Allocation optimization with main network


SouthEndMusic:
**What**
This is issue is for completing the main structure of allocation optimization.

**How**
There are the following steps:
- [x] I think it is quite straightforward to support a model that does not have a main network, where allocation is only solved in one subnetwork. This can be handled with a simple `if` statement in `update_allocation!`. To check whether the model has a subnetwork, use the following function:

```julia
function has_main_network(allocation::Allocation)::Bool
   return first(allocation.allocation_network_ids) == 1
end
```

This should always work given the ordering of the allocation network IDs and the validation introduced in https://github.com/Deltares/Ribasim/issues/710.

- [x] Add a kwarg `collect_demands::Bool = true` to the function `allocate!`. If `collect_demands == false` the function should behave as it currently does, otherwise:
- In `adjust_source_flows!`, make sure the capacities of the sources that are connections from the main network to the subnetwork are set to $\infty$ (or deactivate the source constraints entirely if possible);
- In `assign_allocations!`, make sure the allocations are written to `allocation.subnetwork_demands` instead of `user.allocated`. 

- [x] At the beginning of `update_allocation!`, loop over the allocation models of the subnetworks to compute the subnetwork demands, i.e. call `allocate!` on the subnetworks with `collect_demands = true`.

- [x] In `set_objective_priority!`, make sure that for the main network allocation the subnetworks are also reflected in the objective function, with the demands coming from `allocation.subnetwork_demands`.

- [x] In `adjust_source_flows!` for a subnetwork when `collect_demands = false`, for sources that come from the main network, make sure that the source capacity is set to what is allocated to that edge in the main network problem.

- [ ] Add tests:
- Test whether the optimization problem for the main network is correct;
- Test whether not more water is allocated within the subnetwork than is allocated to the subnetwork from the main network;.
- running the `main_network_with_subnetworks` test model to completion with `succesful_retcode`.

PR #924, opened 2024-01-08
Allocation main network initialization

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/921.


SouthEndMusic:
After doing a manual depth-first traversal of the call tree sprouting from `function AllocationModel`, I found the need for these modifications to support the main network:
- [ ]  In `function allocation_graph_used_nodes!`: Make sure that for each edge that connects the main network to a subnetwork, both nodes connected to this edge are used in the main network problem (which indeed means that the optimization problem for the main network considers some nodes which are not in the main network)
- [ ] In `function add_variables_absolute_value!`: Use the outgoing node in the main-sub connection edges as a user for the main network
- [ ] In `function add_constraints_absolute_value!`: Add constraints also for the nodes mentioned above.

Other functions in `allocation.jl` also need to be checked for whether they need updates to support the main network, but this is left for https://github.com/Deltares/Ribasim/issues/922.

core/src/allocation.jl
```diff
@@ -251,24 +287,23 @@ function valid_sources(p::Parameters, allocation_network_id::Int)::Bool
...
+                from_main_network = graph[id_source].allocation_network_id == 1
 
-            ids_allocation_out = [
-                label for label in outneighbor_labels(graph, id_source) if
-                graph[id_source, label].allocation_flow
-            ]
-            if length(ids_allocation_out) !== 1
-                errors = true
-                @error "Source edge ($id_source, $id_dst) is not the only allocation edge coming from $id_source"
+                if !from_source_node & !from_main_network
```
visr, line 303:
```suggestion
                if !from_source_node && !from_main_network
```

core/test/allocation_test.jl
```diff
@@ -165,3 +165,40 @@ end
...
+    allocation_edges_main_network = graph[].edge_ids[1]
+    @test Tuple{NodeID, NodeID}[(2, 11), (6, 24), (10, 38)] ⊆ allocation_edges_main_network
+
+    allocation_model_main_network = Ribasim.get_allocation_model(p, 1)
+    problem = allocation_model_main_network.problem
+    @test problem[:F_abs].axes[1] == NodeID[11, 24, 38]
+    @test problem[:abs_positive].axes[1] == NodeID[11, 24, 38]
+    @test problem[:abs_negative].axes[1] == NodeID[11, 24, 38]
+
+    @test Ribasim.get_allocation_model(p, 2).problem[:source].axes[1] ==
```
visr, line 198: Could use a comment

visr (approved):

PR #933, opened 2024-01-10
replace setuptools with hatchling

### LINKED ISSUE #797: Pydantic support VS Code integration
If I copy this example in VS Code I get similar autocompletion; VS Code and the language server know the arguments and types.

But if I try e.g.

```python
import ribasim
ribasim.Solver(
```

I don't see any help from the language server. Do others see the same?
There is documentation here: https://docs.pydantic.dev/latest/integrations/visual_studio_code/
Though as far as I can tell this should just work.

### LINKED ISSUE #925: Use hatchling backend for Python packages
We currently use the setuptools backend for our Python backends.

There is a new standard emerging called hatchling that has advantages over setuptools: https://hatch.pypa.io/latest/why/

Most of them, I consider nice-to-haves. However, this one is enough in my opinion to justify the switch:

> Editable installations: The default behavior of Hatchling allows for proper static analysis by external tools such as IDEs. With setuptools, you must provide [additional configuration](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior) which means that by default, for example, you would not get autocompletion in Visual Studio Code. This is marked as a legacy feature and may in fact be removed in future versions of setuptools.

Packages that should be adapted are:

- [ ] python/ribasim
- [ ] python/ribasim_api
- [ ] python/ribasim_testmodels

deltamarnix:
Fixes #925 
Fixes #797 

Use hatchling as build backend to create wheels.

Makes editable installs better accessible for development. Can be tested by looking into e.g. `backwater.py`, and see that `import ribasim` is colored and that `ribasim.Model` contains documentation when you hover over it in VS Code.

### DIFF
--- a/python/ribasim/pyproject.toml
+++ b/python/ribasim/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools>=61"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
 
 [project]
 name = "ribasim"
@@ -31,18 +31,9 @@ dynamic = ["version"]
 [project.optional-dependencies]
 tests = ["pytest", "pytest-xdist", "pytest-cov", "ribasim_testmodels"]
 
-[tool.setuptools]
-zip-safe = true
-
-[tool.setuptools.dynamic]
-version = { attr = "ribasim.__version__" }
-
-[tool.setuptools.packages.find]
-include = ["ribasim", "ribasim.*"]
-
-[tool.setuptools.package-data]
-"ribasim" = ["py.typed"]
-
 [project.urls]
 Documentation = "https://deltares.github.io/Ribasim"
 Source = "https://github.com/Deltares/Ribasim"
+
+[tool.hatch.version]
+path = "ribasim/__init__.py"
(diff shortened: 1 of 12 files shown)


python/ribasim/pyproject.toml
```diff
@@ -31,18 +31,9 @@ dynamic = ["version"]
...
-zip-safe = true
-
-[tool.setuptools.dynamic]
-version = { attr = "ribasim.__version__" }
-
-[tool.setuptools.packages.find]
-include = ["ribasim", "ribasim.*"]
-
-[tool.setuptools.package-data]
-"ribasim" = ["py.typed"]
```
deltamarnix, line 44: I'm not sure if this is required. The whl and sdist folders were the same between setuptools and hatchling

Hofer-Julian: So the `py.typed` file is still included?
In that case it should be fine

Hofer-Julian:
Seems like that also made mypy work across internals packages, and it discovers a bunch of new errors.

Hofer-Julian (commented):
I had to remove the `.pixi` folder, run `pixi run install` and then restart the Python language server. After that, the language server started working for `ribasim`.

After the mypy errors are fixed, this PR is ready to be merged from my side.

python/ribasim/pyproject.toml
deltamarnix: yes it is!

Hofer-Julian (approved):
Looks good!

PR #936, opened 2024-01-11
Test models doc

### LINKED ISSUE #915: Add a catalog of test models in the docs
**What & why**
I keep a folder on my PC with plots of all the test models in `ribasim_testmodels` for reference (e.g. to find an existing test model to test a new feature). As discussed with @visr, this would be a nice feature to add to the docs.

**How**
Make a new documentation quarto markdown page (`*.qmd`), I suppose under **Python tooling**, with a short introduction and then a Python code block that produces plots of all the test models. For reference, here is the script I use to populate the folder of plots mentioned above:

```python
import ribasim_testmodels
import matplotlib.pyplot as plt

for model_name, model_constructor in ribasim_testmodels.constructors.items():
    print(f"Plotting {model_name}")
    model = model_constructor()
    model.plot()

    fig = plt.gcf()
    fig.tight_layout()
    fig.savefig(r"path\to\Testmodel_graphs\\" + model_name + ".png")
```

I suggest adding the name of the model as a title to the plot.

**Additional context**
See [here](https://github.com/Deltares/Ribasim/blob/836dc7178bf06abd0e60d61029f1be787c04cc3a/docs/core/equations.qmd#L108) for an example of a Python codeblock used in the documentation.

Jingru923:
Fixes #915

### DIFF
--- a/docs/python/test-models.qmd
+++ b/docs/python/test-models.qmd
@@ -12,10 +12,12 @@ import matplotlib.pyplot as plt
 for model_name, model_constructor in ribasim_testmodels.constructors.items():
     if model_name.startswith("invalid"):
         continue
+
     model = model_constructor()
     fig, ax = plt.subplots()
     model.plot(ax)
-    ax.set_title(label=model_name,loc="left")
+    ax.set_title(label=model_name, loc="left")
+    fig.text(0,1,model_constructor.__doc__)
     fig.tight_layout()
     plt.show()
     plt.close(fig)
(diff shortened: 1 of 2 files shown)


docs/python/test-models.qmd
```diff
@@ -0,0 +1,24 @@
...
+
+for model_name, model_constructor in ribasim_testmodels.constructors.items():
+    if model_name.startswith("invalid"):
+        continue
+
+    model = model_constructor()
+    fig, ax = plt.subplots()
+    model.plot(ax)
+    ax.set_title(label=model_name, loc="left")
+    fig.text(0,1,model_constructor.__doc__)
```
evetion, line 20: Don't we run an autoformatter for these snippets? (missing spaces)

Jingru923: We don't do it for quarto markdown

SouthEndMusic (approved):
LGTM (Looks good to me) (y)

PR #938, opened 2024-01-11
[fix] Apply meta renaming to spatial tables.

### LINKED ISSUE #934: Loading existing models in ribasim Python with extra columns
**What**
https://github.com/Deltares/Ribasim-NL/pull/49#issuecomment-1883140814 point 1, see also https://github.com/Deltares/Ribasim-NL/pull/49#issuecomment-1884948242. 

If I understand correctly, we have to make a decision what we do when loading a model in ribasim Python with extra columns that are not prefixed by `meta_`. We either have to add the prefix when loading the model or raise an error.

evetion:
Fixes #934

We now always apply the meta prefix to unknown columns in the Node and Edge spatial tables as well.


SouthEndMusic (approved):
LTGM (:

PR #939, opened 2024-01-11
[ci] Fix intermittent QGIS test failures.

evetion:
Long story short, we did a `sleep 10` to wait for the container to be ready. I removed it, and now start a displayserver (`xvfb-run -a`, separate from the container one) as part of the test, which seems to be much more reliable.


visr:
Same issue, do you need an older patch release?

evetion:
No this worked locally, whereas the old version failed. Might not be related to the version at all.

deltamarnix:
A new version was pushed ~6 hours ago and that version also works on my machine now. So it might be a race condition.

visr (approved):

visr:
Same test failure on master after merging this: https://github.com/Deltares/Ribasim/actions/runs/7530011861/job/20495487468

PR #940, opened 2024-01-12
No more warning in Ribasim Python when plotting model without edges

### LINKED ISSUE #937: Warning in Ribasim Python when plotting model without edges
In the wiki page of Test models under Python tooling, there is a warning:
```
[C:\Project\Ribasim\python\ribasim\ribasim\geometry\edge.py:76](file:///C:/Project/Ribasim/python/ribasim/ribasim/geometry/edge.py:76): UserWarning: The GeoDataFrame you are attempting to plot is empty. Nothing has been displayed.
  self.df[where_flow].plot(**kwargs_flow)
``` 
This warning is printed because one of our test models, "bucket", has no edges. It is not a desired type of model for us. But the warning is not necessary. 

As a result, the Ribasim python needs to be fixed to adapt the model that without edges so the warning like this will not be shown.

Jingru923:
Fixes #937

### DIFF
--- a/python/ribasim/ribasim/geometry/edge.py
+++ b/python/ribasim/ribasim/geometry/edge.py
@@ -74,7 +74,8 @@ class Edge(SpatialTableModel[EdgeSchema]):
         where_flow = self.get_where_edge_type("flow")
         where_control = self.get_where_edge_type("control")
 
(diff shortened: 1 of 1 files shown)


SouthEndMusic (approved):
Looks good

PR #943, opened 2024-01-12
Remove docker from QGIS tests

deltamarnix:
The tests are now run via the qgis_testrunner script that is a clone from the original QGIS runner: https://github.com/qgis/QGIS/blob/4891c6bda8922ff5c6a70021f097ce43c486cddf/.docker/qgis_resources/test_runner/qgis_testrunner.py. I removed the if check for being inside or outside of QGIS, because not every module was available in Windows.

The QGIS profile information and the plugins themselves are now installed inside of the .pixi folder, so that cleanup is easier. And there is no need anymore for looking into the roaming folder.

The QGIS tests that really require the user interface are called via `test-ribasim-qgis-ui`. This task is not automated.
The QGIS tests that do not require the user interface are called via `test-ribasim-qgis` and are automated via github actions.


Hofer-Julian (approved):
It's a bit sad that we cannot test GUI functionality anymore.
On the flip side, this setup is much simpler, and we now test on macOS and Windows as well.

evetion (changes requested):
This is akin to removing tests because they fail, instead of fixing them. Especially the first removed test (plugin load) seems essential for guaranteeing the QGIS plugin will work. Note that the remaining plugins purely tests internal consistency of code (syntax mostly), not whether it is a valid plugin at all.

visr:
I guess this can be closed in favor of #939. Still would be nice to be able to drop Docker from our test stack, but preferably not at the cost of reduced testing capability.

Hofer-Julian:
> I guess this can be closed in favor of #939. Still would be nice to be able to drop Docker from our test stack, given we have QGIS in our Pixi environment as well, but preferably not at the cost of reduced testing capability.

Since the latest attempt also didn't fix it, maybe we merge this and add a high priority issue to add something equivalent with pixi? Or does anyone want to give docker another try?

evetion:
> > I guess this can be closed in favor of #939. Still would be nice to be able to drop Docker from our test stack, given we have QGIS in our Pixi environment as well, but preferably not at the cost of reduced testing capability.
> 
> Since the latest attempt also didn't fix it, maybe we merge this and add a high priority issue to add something equivalent with pixi? Or does anyone want to give docker another try?

Yeah, I have given up on the Docker path (although I don't expect it is the root cause), but I can try the pixi path in this branch.

visr:
Make sure to merge main, which has a more recent QGIS that is not broken.

ribasim_qgis/scripts/qgis_testrunner.py
```diff
@@ -0,0 +1,132 @@
...
+    Email                : apasotti at boundlessgeo dot com
+***************************************************************************
+*                                                                         *
+*   This program is free software; you can redistribute it and/or modify  *
+*   it under the terms of the GNU General Public License as published by  *
+*   the Free Software Foundation; either version 2 of the License, or     *
+*   (at your option) any later version.                                   *
+*                                                                         *
+***************************************************************************
+"""
```
deltamarnix, line 42: I need to add information here that I changed the file so that it works on Windows and Linux. Add the commit hash.

Also, what about this specific license. Am I allowed to commit it like this?

ribasim_qgis/scripts/run_qgis_tests.py
```diff
@@ -0,0 +1,22 @@
...
+        "ribasim_qgis.tests",
+    ],
+    stdout=subprocess.PIPE,
+    stderr=subprocess.STDOUT,
+    text=True,
+)
+
+print(qgis_process.stdout)
+qgis_process.check_returncode()
+if any(s in qgis_process.stdout for s in ["QGIS died on signal", "FAILED"]):
```
deltamarnix, line 21: Add why we need this check.

Because qgis always returns 0 when all tests have run, even when the tests failed.

ribasim_qgis/scripts/qgis_testrunner.py
Hofer-Julian: We don't distribute this file, we only use it to run tests. It is therefore fine to use it.

If I remember correctly, your current code barely resembles that original testrunner and only uses a few general concepts like monkey patching QGIS. IMO, you can just start a new file without mentioning any copyright information then.

evetion:
I can't run the tests locally because of https://github.com/qgis/QGIS/issues/52987 on mac, but it seems to be setup ok. 👍🏻

PR #953, opened 2024-01-15
Launch debug julia

Huite:
Add an entry to launch.json to debug Julia

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,6 +1,15 @@
 {
     "version": "0.2.0",
     "configurations": [
+         {
+            "type": "julia",
+            "request": "launch",
+            "name": "Julia: current file",
+            "program": "${file}",
+            "stopOnEntry": false,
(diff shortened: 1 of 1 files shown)


visr (approved):

.vscode/launch.json
```diff
@@ -1,6 +1,15 @@
 {
     "version": "0.2.0",
     "configurations": [
+         {
+            "type": "julia",
+            "request": "launch",
+            "name": "Julia: current file",
+            "program": "${file}",
+            "stopOnEntry": false,
+            "cwd": "${workspaceFolder}",
```
deltamarnix, line 10: Will this automatically activate `--project=core`? Do we want to add that, so that all the packages are correct?

visr: Yeah I think the current `"juliaEnv": "${command:activeJuliaEnvironment}"` is fine since it allows users to control it, and still has a good default setting of the `root` environment.

PR #958, opened 2024-01-15
Pixi: use latest QGIS release

visr:
I tried `pixi run qgis`, which up to now was set to the LTS release, and it was broken on loading the plugin:

```
Windows fatal exception: access violation

Current thread 0x00006700 (most recent call first):
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\ribasim_widget.py", line 47 in __init__
    self.tabwidget.addTab(self.__dataset_widget, "Model")
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\ribasim_qgis.py", line 48 in toggle_ribasim
    widget = RibasimWidget(self.ribasim_widget, self.iface)
```

I saw this happens on 3.28.12, which is the last LTS in conda-forge, but locally my OSGeo4w qgis-ltr release is 3.28.14, which doesn't have this issue. I propose we switch to the latest QGIS release, which fixes this. In pixi we set QGIS to the LTS since that is the oldest we support, but this is the only dependency for which we set it to the oldest version. And also it holds a lot of other dependencies in our environment back as can be seen from the size of the diff. If we want to check the oldest supported QGIS release we can perhaps use https://pixi.sh/design_proposals/multi_environment_proposal/ in the future for this, same as for Python versions.


Hofer-Julian (approved):
Tbh I forgot why we pinned qgis to begin with.

Having the newest release in development sounds like a good idea to me in general.

PR #962, opened 2024-01-16
Add pixi task to update registry

visr:
With our current workflow, it can happen that dev A updates the packages in Manifest.toml, and dev B gets the error:

```
ERROR: Unsatisfiable requirements detected for package X
```

The problem is that dev B's local registry is outdated, so it is not aware of the new versions of X that are being used. This adds a task to update the registry, and runs it by default when building or installing. That should make this error more rare, and perhaps easier to resolve.

Note that this is an issue that may get resolved in Pkg in the future; https://github.com/JuliaLang/Pkg.jl/issues/3713

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -21,16 +21,22 @@ install-ribasim-testmodels = "pip install --no-deps --editable python/ribasim_te
 install-pre-commit = "pre-commit install"
 install-without-pre-commit = { depends_on = [
     "install-julia",
+    "update-registry-julia",
     "install-ribasim-python",
     "install-ribasim-api",
     "install-ribasim-testmodels",
 ] }
 install = { depends_on = ["install-without-pre-commit", "install-pre-commit"] }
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -21,16 +21,22 @@ install-ribasim-testmodels = "pip install --no-deps --editable python/ribasim_te
...
 install-without-pre-commit = { depends_on = [
     "install-julia",
+    "update-registry-julia",
     "install-ribasim-python",
     "install-ribasim-api",
     "install-ribasim-testmodels",
 ] }
 install = { depends_on = ["install-without-pre-commit", "install-pre-commit"] }
 # Instantiate
+update-registry-julia = "julia --project --eval='using Pkg; Registry.update()'"
```
Hofer-Julian, line 31: Is `--project` needed here?

visr: Yes, good point.

PR #964, opened 2024-01-17
Complete allocation algorithm with main network and subnetworks

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/922.


visr:
The AllocationModels are created once, with Parameter initialization.

SouthEndMusic:
To do:
- [ ] Add convenience function `get_main_network_connections` analogous to `get_allocation_model`;
- [ ] Subnetworks picking up what is allocated to them from the main network doesn't work yet. In `function assign_allocations`, allocations from the main network to a subnetwork should be written to a new field `Allocated.subnetwork_demands::Dict{Tuple{NodeID, NodeID}, Vector{Float64}}`. Currently the allocated value is taken as the optimized flow over the connecting edge in the main network problem, but this does not work at all since this is only the allocated value for the last priority.
- [ ] Complete tests

core/src/allocation.jl
```diff
@@ -952,17 +1037,22 @@ function adjust_edge_capacities!(
...
+
     for edge_id in edge_ids
         c = capacity[edge_id...]
 
-        # Edges with infinite capacity have no capacity constraints
-        if isinf(c)
+        # These edges have no capacity constraints:
+        # - With infinite capacity
+        # - Being a source from the main network to a subnetwork
+        if isinf(c) || edge_id ∈ main_network_source_edges
```
visr, line 1055: Seems like `prevfloat` would not work here.

```
julia> isinf(prevfloat(Inf))
false
```

core/src/allocation.jl
```diff
@@ -984,18 +1074,35 @@ end
...
-    (; problem) = allocation_model
+function allocate!(
+    p::Parameters,
+    allocation_model::AllocationModel,
+    t::Float64;
+    collect_demands::Bool = false,
+)::Nothing
+    (; user, allocation) = p
+    (; problem, allocation_network_id) = allocation_model
+    (; allocation_network_ids, main_network_connections) = allocation
```
visr, line 1085:
```suggestion
```

visr (approved):

core/src/allocation.jl
SouthEndMusic: There's the sparse capacity matrix `AllocationModel.capacity` which is determined at initialization which was already allowed to have inifinities, which is related to but not exactly the same as the capacity values in the capacity constraints in the optimization problem. As it turns out, having infinite capacities in these constraints works fine.

Issue #692, opened 2023-10-23
Testmodel `looped_subnetwork_model` is not working


Hofer-Julian:
`looped_subnetwork_model` fails when run with `ribasim.exe`

PR #965, opened 2024-01-17
Fix looped subnetwork

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/692.

It now also runs fine with the cli (which I think I ran for the first time ever ^^).


SouthEndMusic:
@gijsber this fixes the allocation test model which you designed a while back:
looped_subnetwork

The allocation runs without errors. Maybe we should have a look at the actual allocation behavior.

PR #970, opened 2024-01-19
Refactor `sort_keys` implementation

### LINKED ISSUE #956: Sort keys in Ribasim python should should be moved to TableModel
**What**
Every different model that contains a TableModel now has a `_sort_keys` attribute. These attributes can now be changed by other programmers dynamically. It would be better if this information was stuck inside of every table model.

**Why**
The `_sort_keys` per model are overwriting the existing variable in the base class. If the base class would ever initialize other than an empty list, it would mean that these values get lost.

And as mentioned, it is now possible to change the values dynamically at run time. This should be avoided.

**How**
Make the `sort_keys` a pydantic `Field()` in order to set the correct values via the constructor. Also ensure that the value should no longer change after the initialization.

**Additional context**
Some try-out code can be found in branch feat/sort-keys-in-tablemodel.
The idea was to get `TableModel`: `sort_keys: list[str] = Field(init_var=True, exclude=True, repr=False)`

evetion:
Fixes #956

While ideally we set the sort_keys on the Schema of a TableModel, or on the TableModel themselves, this is not directly possible AFAIK:
- Schemas are autogenerated, and sort_keys is not something we can serialize or parse in a JSONSchema
- The TableModels are generic (TableModel[T]), so specific sort_keys can't be hardcoded to a class.

The previous approach was to hold the sort keys in a `Dict[table, sort_keys]` on a NodeModel (which holds multiple tables), and passing them on when save/sort was called from the Node. This PR improves on that approach by introducing a private _sort_keys on the TableModel, which is always set (from an empty list by default) on initialization from the NodeModel. 

Instead of keeping a separate Dict with sort_keys on a NodeModel, we now set them in the `Field` attribute `json_schema_extra` (the only customizable attribute of a Field) for each field. Each field of a NodeModel is now validated by default (also on assignment), and if a sort_keys is present for a fieldtype, it is set on the TableModel.

Based on the initial exploration by @deltamarnix in https://github.com/Deltares/Ribasim/tree/feat/sort-keys-in-tablemodel

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -20,6 +20,7 @@ from pydantic import (
     ConfigDict,
     DirectoryPath,
     Field,
+    PrivateAttr,
     ValidationInfo,
     field_validator,
     model_serializer,
@@ -159,7 +160,7 @@ class FileModel(BaseModel, ABC):
 
 class TableModel(FileModel, Generic[TableT]):
     df: DataFrame[TableT] | None = Field(default=None, exclude=True, repr=False)
-    sort_keys: list[str] = Field(default=[], exclude=True, repr=False)
+    _sort_keys: list[str] = PrivateAttr(default=[])
 
     @field_validator("df")
     @classmethod
@@ -296,7 +297,7 @@ class TableModel(FileModel, Generic[TableT]):
         Sorting is done automatically before writing the table.
         """
         if self.df is not None:
-            self.df.sort_values(self.sort_keys, ignore_index=True, inplace=True)
+            self.df.sort_values(self._sort_keys, ignore_index=True, inplace=True)
 
     @classmethod
     def tableschema(cls) -> TableT:
@@ -408,7 +409,7 @@ class NodeModel(ChildModel):
             field = cls.model_fields[getattr(info, "field_name")]
             extra = field.json_schema_extra
             if extra is not None and isinstance(extra, dict):
-                v.sort_keys = extra.get("sort_keys", [])  # type: ignore
+                v._sort_keys = extra.get("sort_keys", [])  # type: ignore
         return v
 
     @classmethod
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -20,6 +20,7 @@ from pydantic import (
     ConfigDict,
     DirectoryPath,
     Field,
+    ValidationInfo,
     field_validator,
     model_serializer,
     model_validator,
@@ -158,6 +159,7 @@ class FileModel(BaseModel, ABC):
 
 class TableModel(FileModel, Generic[TableT]):
     df: DataFrame[TableT] | None = Field(default=None, exclude=True, repr=False)
+    sort_keys: list[str] = Field(default=[], exclude=True, repr=False)
 
     @field_validator("df")
     @classmethod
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/input_base.py
```diff
@@ -392,15 +393,24 @@ def check_parent(self) -> "ChildModel":
...
 
+    @field_validator("*")
+    @classmethod
+    def set_sort_keys(cls, v: Any, info: ValidationInfo) -> Any:
+        """Set sort keys for all TableModels if present in FieldInfo."""
+        if isinstance(v, (TableModel,)):
+            field = cls.model_fields[getattr(info, "field_name")]
+            extra = field.json_schema_extra
+            if extra is not None and isinstance(extra, dict):
+                v.sort_keys = extra.get("sort_keys", [])  # type: ignore
```
deltamarnix: Could you add the reason why you added a type ignore? Is it something we want to fix?

python/ribasim/ribasim/input_base.py
```diff
@@ -392,15 +393,24 @@ def check_parent(self) -> "ChildModel":
...
     @model_serializer(mode="wrap")
     def set_modeld(
         self, serializer: Callable[[type["NodeModel"]], dict[str, Any]]
     ) -> dict[str, Any]:
         content = serializer(self)
         return dict(filter(lambda x: x[1], content.items()))
 
+    @field_validator("*")
+    @classmethod
+    def set_sort_keys(cls, v: Any, info: ValidationInfo) -> Any:
```
deltamarnix, line 406: Could you add some more information in the pull request on how and why you solved the issue for future reference?

python/ribasim/ribasim/input_base.py
```diff
@@ -392,15 +393,24 @@ def check_parent(self) -> "ChildModel":
...
+        """Set sort keys for all TableModels if present in FieldInfo."""
+        if isinstance(v, (TableModel,)):
+            field = cls.model_fields[getattr(info, "field_name")]
+            extra = field.json_schema_extra
+            if extra is not None and isinstance(extra, dict):
```
deltamarnix, line 411: Is there a chance that extra is not a `dict`? Should we throw an assertion error in this case, because some programmer needs to do more work?

deltamarnix (changes requested):
The code itself looks good, I would like to see some more error handling and documentation on the decisions that were made so that it will be in the git history

python/ribasim/ribasim/input_base.py
evetion: I updated the code, and provided an explanation. Not something we want to fix imho.

python/ribasim/ribasim/input_base.py
evetion: It's to make mypy happy, as json_schema_extra can be a callable as well according to the type. We always set it to a dict though.

visr (approved):

python/ribasim/ribasim/input_base.py
```diff
@@ -392,15 +395,26 @@ def check_parent(self) -> "ChildModel":
...
         self, serializer: Callable[[type["NodeModel"]], dict[str, Any]]
     ) -> dict[str, Any]:
         content = serializer(self)
         return dict(filter(lambda x: x[1], content.items()))
 
+    @field_validator("*")
+    @classmethod
+    def set_sort_keys(cls, v: Any, info: ValidationInfo) -> Any:
+        """Set sort keys for all TableModels if present in FieldInfo."""
+        if isinstance(v, (TableModel,)):
```
deltamarnix, line 409: Because of the nesting I would prefer some early returns here instead.

deltamarnix (approved):
I like the solution. Making use of the schema to set the information. I would love to eventually see this coming from the generated code instead.

PR #972, opened 2024-01-19
Validate allocation input

Jingru923:
Fixes #


SouthEndMusic (commented):
I added some suggestions. I was also thinking about how to organize the test, because triggering different validation errors with the same model might be complicated due to the structure of the initialization code. Here are my thoughts per task of the issue:

- For this one it is probably best to create an invalid test model in Ribasim Python as you already started making.
- For this one I was a bit naughty, as I wrote the mentioned source validation function without testing it. Maybe you can skip this one for now, as I changed this validation function in the branch I am working on.
- For this one it might be educational to create an instance of a `User` object within the test, just as is done for instance [here](https://github.com/Deltares/Ribasim/blob/4d226ddb32f42e953a35fa1cbf27e72f9dcf102a/core/test/utils_test.jl#L38) for `Basin`. Also, you said you created an `if` statement in the inner constructor, but `function User` is not an inner constructor. Doing validation in an inner constructor looks something like this:

```julia
struct Circle
   r::Float64
   function Circle(r::Float64)
      if r < 0
         error("Circle radius must be non-negative, got $r.")
      end
      return new(r)
   end
end
```
We the same in our own code for instance [here](https://github.com/Deltares/Ribasim/blob/4d226ddb32f42e953a35fa1cbf27e72f9dcf102a/core/src/solve.jl#L255).
- The last one is probably also best done with a model made in Ribasim python. Creating a subnetwork is quite easy: you only have to add a column to the `Node` table called `allocation_network_id`, see also [here](https://deltares.github.io/Ribasim/core/usage.html#node) in the docs. 

Also a small nitpick:
- In the description of the PR you put the number of this PR instead of the number of the issue (710).

SouthEndMusic:
I also saw that this docstring is not correct, so you can fix my mistake (:

https://github.com/Deltares/Ribasim/blob/306d29a69245577bdd0f8935f03a08ea77235032/core/src/create.jl#L644

It should be something like `# All provided priorities`.

On a related note, we use a VSCode extension called Git Blame, which tells you at the bottom of your screen who wrote a line of code you don't like if you put your cursor on it 😄

core/src/create.jl
```diff
@@ -691,7 +704,7 @@ function User(db::DB, config::Config)::User
             first_row = time[first_row_idx]
             is_active = true
         else
-            @error "User node #$node_id data not in any table."
+            @error "User node $node_id data not in any table."
```
SouthEndMusic, line 707: I had another look at this. `node_id` comes from `node_ids` as the third output of the function `static_and_time_node_ids`, which is a vector with element type `Int` which you can see from the function signature:

```julia
function static_and_time_node_ids(
    db::DB,
    static::StructVector,
    time::StructVector,
    node_type::String,
)::Tuple{Set{Int}, Set{Int}, Vector{Int}, Vector{String}, Bool}
```

Thus adding the `#` here is needed for consistency. My mistake! This also indicates that there is no unit test for this error message.

core/src/utils.jl
```diff
@@ -88,6 +88,10 @@ function create_graph(db::DB, config::Config, chunk_sizes::Vector{Int})::MetaGra
         end
     end
 
+    if incomplete_subnetwork(graph, node_ids)
+        error("Incomplete connection in network")
```
SouthEndMusic, line 92:
```suggestion
        error("Incomplete connectivity in network")
```

core/src/validation.jl
```diff
@@ -660,3 +660,39 @@ function valid_subgrid(
...
+        Vector{LinearInterpolation{Vector{Float64}, Vector{Float64}, true, Float64}},
+    },
+    priorities::Vector{Int},
+)::Bool
+    errors = false
+
+    for (col, id) in zip(demand, node_id)
+        for (demand_p_itp, p_itp) in zip(col, priorities)
+            if any(demand_p_itp.u .< 0.0)
+                @error "Demand of user node $id with priority #$p_itp should be non-negative"
```
SouthEndMusic, line 676:
```suggestion
                @error "Demand of user node $id with priority $p_itp should be non-negative"
```
We only prefix node IDs with a `#`.

core/test/validation_test.jl
```diff
@@ -401,3 +401,27 @@ end
...
+            [0.9],
+            [1],
+            [],
+        )
+    end
+
+    @test length(logger.logs) == 1
+    @test logger.logs[1].level == Error
+    @test logger.logs[1].message ==
+          "Demand of user node #1 with priority #1 should be non-negative"
```
SouthEndMusic, line 426:
```suggestion
          "Demand of user node #1 with priority 1 should be non-negative"
```

core/test/create_test.jl
```diff
@@ -0,0 +1,46 @@
+@testitem "Non-positive allocation network ID" begin end
+
+@testitem "Incomplete subnetwork" begin
```
SouthEndMusic, line 3: Regarding this test: It is good practice to create your own test model to do this, but I thought of an easier way: load the parameters of an existing model, say

```julia
toml_path = normpath(@__DIR__, "../../generated_testmodels/allocation_example/ribasim.toml")
p = Ribasim.Model(toml_path).integrator.p
```

[remove an edge](https://juliagraphs.org/MetaGraphsNext.jl/dev/api/#Base.delete!-Tuple{MetaGraph,%20Any,%20Any}) to make it invalid and run your validation test on that.

Jingru923: This morning I tried to load an existing model and change one of the allocation id to negative number. It gave error saying the immutable struct of type cannot be changed. I remembered you mentioned this yesterday, it probably can't be changed or maybe I didn't do it the correct way. Anyhow, in my new code I created a graph that has the required feature and used it as the test model.

Hofer-Julian:
Looks to me like merging main into your branch went wrong.
This needs to be fixed up before we can review this :)

core/src/create.jl
```diff
@@ -199,6 +199,12 @@ const nonconservative_nodetypes =
 function generate_allocation_models!(p::Parameters, config::Config)::Nothing
     (; graph, allocation_models) = p
 
+    errors = non_positive_id(graph)
```
SouthEndMusic, line 202:
```suggestion
    errors = non_positive_allocation_network_id(graph)
```

core/test/create_test.jl
```diff
@@ -0,0 +1,100 @@
...
+
+    node_ids = Dict{Int, Set{Ribasim.NodeID}}()
+    node_ids[1] = Set{Ribasim.NodeID}()
+    push!(node_ids[1], Ribasim.NodeID(1))
+    push!(node_ids[1], Ribasim.NodeID(2))
+    push!(node_ids[1], Ribasim.NodeID(3))
+    node_ids[2] = Set{Ribasim.NodeID}()
+    push!(node_ids[2], Ribasim.NodeID(4))
+    push!(node_ids[2], Ribasim.NodeID(5))
+    push!(node_ids[2], Ribasim.NodeID(5))
```
SouthEndMusic, line 76: Duplicate line?

SouthEndMusic (approved):
Looks good, I added a few small comments.

PR #977, opened 2024-01-22
Always require 'Basin / state' as an initial condition

visr:
Before we put a volume of 1 m3 in if no initial condition was given. But since this will not always work well, and there is no other number that makes a good default, it is best to remove this feature and request an explicit initial condition.

### DIFF
--- a/core/src/utils.jl
+++ b/core/src/utils.jl
@@ -374,16 +374,28 @@ function get_storage_from_level(basin::Basin, state_idx::Int, level::Float64)::F
 end
 
 """Compute the storages of the basins based on the water level of the basins."""
-function get_storages_from_levels(
-    basin::Basin,
-    levels::Vector,
-)::Tuple{Vector{Float64}, Bool}
-    storages = Float64[]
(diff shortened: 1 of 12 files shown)


core/src/validation.jl
visr, line 1: Unrelated change, but this was dead code that is unlikely to be used in such form.

core/src/utils.jl
```diff
@@ -374,16 +374,28 @@ function get_storage_from_level(basin::Basin, state_idx::Int, level::Float64)::F
...
-    basin::Basin,
-    levels::Vector,
-)::Tuple{Vector{Float64}, Bool}
-    storages = Float64[]
+function get_storages_from_levels(basin::Basin, levels::Vector)::Vector{Float64}
+    errors = false
+    state_length = length(levels)
+    basin_length = length(basin.level)
+    if state_length != basin_length
+        @error "Unexpected 'Basin / state' length." state_length basin_length
```
SouthEndMusic, line 382: Had a look at what adding the variables at the end does, very nice

SouthEndMusic (approved):
Looks good to me (y)

PR #978, opened 2024-01-23
Add cron job to auto-update pixi lock file

Hofer-Julian:
This way there will be a PR that updates the lock file every month

### DIFF
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -1,7 +1,7 @@
 name: Julia Tests
 on:
   push:
-    branches: [main]
+    branches: [main, update/pixi-lock]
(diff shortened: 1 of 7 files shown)


visr (approved):
Looks good to me! So we can start this job manually as well right? Would be good to give it a shot right away.

Would be nice to have something similar for updating the manifest, e.g. `Pkg.update()` on the root environment.

Hofer-Julian:
> Looks good to me! So we can start this job manually as well right? Would be good to give it a shot right away.

Yes, we can. Happy to test it manually, I am pretty sure it works though. Just added the same for the imod docs :)
https://github.com/Deltares/iMOD-Documentation/blob/main/.github/workflows/pixi_auto_update.yml

PR #981, opened 2024-01-23
Julia manifest update the root environment

visr:
Instead to the global one.
Now it works, I manually triggered the action from this branch in #982.


Hofer-Julian (approved):
Thanks, I already wondered why the job didn't trigger any updates 😅

PR #986, opened 2024-01-25
Add update_subgrid_level to BMI functions

### LINKED ISSUE #985: Add update_subgrid_level to BMI functions
The MODFLOW coupling currently always requires a subgrid interpolation to be active.
However, the interpolation does not occur until the SavingCallback is triggered (defined by the solver.saveat entry).

To guarantee that the subgrid levels are to update, this function should be called in the imod_coupler before using the levels.

Huite:
Fixes #985

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -633,6 +633,11 @@ function BMI.update_until(model::Model, time)::Model
     return model
 end
 
+function BMI.update_subgrid_level(model::Model)::Model
+    update_subgrid_level!(model.integrator)
+    return model
+end
+
 function BMI.get_value_ptr(model::Model, name::AbstractString)
     if name == "volume"
         model.integrator.u.storage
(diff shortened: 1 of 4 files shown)


core/src/bmi.jl
```diff
@@ -633,6 +633,11 @@ function BMI.update_until(model::Model, time)::Model
     return model
 end
 
+function BMI.update_subgrid_level(model::Model)::Model
+    update_subgrid_level!(model.integrator)
+    return model
+end
```
visr, line 639:
```suggestion
function update_subgrid_level(model::Model)::Model
    update_subgrid_level!(model.integrator)
    return model
end
```

I don't think creating a function in an external package is the right approach here, since this is not a BMI function. You also need to add a C callable version in https://github.com/Deltares/Ribasim/blob/main/build/libribasim/src/libribasim.jl to get it in libribasim.

Huite: You're obviously right, not really awake yet I guess...

Hofer-Julian (approved):

PR #992, opened 2024-01-26
Use flow_rate rather than discharge in TabulatedRatingCurve

### LINKED ISSUE #983: flow_rate versus discharge
Almost everywhere (pumps/outlets/flowboundary/etc.) the term "flow_rate" is used but for TabulatedRatingCurve the term "discharge" is used. For consistency it would be better to use "flow_rate" also for the TabulatedRatingCurve.

visr:
A breaking change, but this is more consistent. Fixes #983.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -607,9 +607,9 @@ function update_tabulated_rating_curve!(integrator)::Nothing
         # update the existing LinearInterpolation
         id = first(group).node_id
         level = [row.level for row in group]
-        discharge = [row.discharge for row in group]
+        flow_rate = [row.flow_rate for row in group]
         i = searchsortedfirst(node_id, NodeID(id))
-        tables[i] = LinearInterpolation(discharge, level; extrapolate = true)
+        tables[i] = LinearInterpolation(flow_rate, level; extrapolate = true)
     end
     return nothing
 end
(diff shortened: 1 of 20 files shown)


Huite:
Talking semantics...:
Is there a reason to prefer flow_rate over discharge? You've currently replaced all discharge -> flow_rate, but why not flow_rate -> discharge?

discharge has the benefit of not having an underscore. Otherwise they seem interchangeable.

visr:
This use of discharge as a column name predates the use flow_rate in the other tables. I went for discharge since it is used in e.g. https://en.wikipedia.org/wiki/Rating_curve.

I think discharge is more often used in the sense that some flow comes out of some confinement (like an upstream reservoir for a weir), so it seems to me flow_rate is a bit more general. Though I feel both would be fine.

Huite:
I had do some thinking about it: it feels more natural to speak of negative flow_rates rather than negative discharges, which I guess relates to your comment "out of confinement" which has a high to low pressure kind of assocation (and hence an implied positive direction).

I think this is the strongest argument: since discharge is "outwardly" coded, it's confusing when an external boundary would have a **positive** discharge which flows **into** the model; I think relatively many people would expect it to act as an outflow instead.
Flow rate "feels" more abstracted and it's easier to apply the consciously apply the rule "positive is inflow, negative is outflow" (at least I'm hoping we're following that rule consistently...).

Huite (approved):
lgtm

Although I haven't searched for every instance of discharge myself...

PR #995, opened 2024-01-26
Always write fid index of Node and Edge tables

visr:
We are using `GeoDataFrame.to_file` to write these tables to GeoPackage. By default GeoPandas only writes the index if it has a name. We don't want to rely on users naming indices, so this sets the name and ensure it is written.

I ran into this since I noticed the `model.network.edge.df.index` (by default 0-based) did not correspond to the 1-based fid column that GeoPandas writes.

https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_file.html

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -374,8 +374,9 @@ class SpatialTableModel(TableModel[TableT], Generic[TableT]):
 
         gdf = gpd.GeoDataFrame(data=self.df)
         gdf = gdf.set_geometry("geometry")
+        gdf.index.name = "fid"
 
-        gdf.to_file(path, layer=self.tablename(), driver="GPKG")
+        gdf.to_file(path, layer=self.tablename(), driver="GPKG", index=True)
 
     def sort(self):
         self.df.sort_index(inplace=True)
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/input_base.py
```diff
@@ -374,8 +374,9 @@ def _write_table(self, path: FilePath) -> None:
 
         gdf = gpd.GeoDataFrame(data=self.df)
         gdf = gdf.set_geometry("geometry")
+        gdf.index.name = "fid"
 
-        gdf.to_file(path, layer=self.tablename(), driver="GPKG")
+        gdf.to_file(path, layer=self.tablename(), driver="GPKG", index=True)
```
Huite, line 379: Mostly out of curiosty: What happens if index=True but no name is provided, does it error?

Huite (approved):

python/ribasim/ribasim/input_base.py
visr: Oef, then it writes the index to an `index` column, and still adds a 1-based autoincrement fid.
This is the trivial model, which has edge_id 9 and 11 only.
image

PR #997, opened 2024-01-29
Allow compiled binaries to get version

### LINKED ISSUE #364: `get_version` doesn't work in iMOD Coupler
`get_version` is an API function of `ribasim_api` and we test that it works on TeamCity: https://github.com/Deltares/Ribasim/blob/b1c9b32609a0cfd1eb295cabb10e0eb91bbe1cad/python/ribasim_api/tests/test_bmi.py#L116

However, it when calling it with iMOD Coupler it returns the string "nothing".
My guess is that this somehow breaks when relocating the DLLs.

Hofer-Julian:
We started using the function that Julia provides, but it doesn't seem to work with compiled binaries

Fixes #364

### DIFF
--- a/core/src/utils.jl
+++ b/core/src/utils.jl
@@ -1,3 +1,11 @@
+"Get the package version of a given module"
+function pkgversion(m::Module)::VersionNumber
+    rootmodule = Base.moduleroot(m)
+    pkg = Base.PkgId(rootmodule)
+    pkgorigin = get(Base.pkgorigins, pkg, nothing)
+    return pkgorigin.version
+end
+
 "Check that only supported edge types are declared."
 function valid_edge_types(db::DB)::Bool
     edge_rows = execute(
--- a/build/libribasim/src/libribasim.jl
+++ b/build/libribasim/src/libribasim.jl
@@ -175,7 +175,7 @@ end
 
 Base.@ccallable function get_version(version::Cstring)::Cint
     @try_c_uninitialized begin
-        ribasim_version = pkgversion(Ribasim)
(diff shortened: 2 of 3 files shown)


deltamarnix:
I did a test on the Main branch and I checked main_test.jl.

I changed the test to

```jl
@testitem "version" begin
    using IOCapture: capture

    (; value, output) = capture() do
        Ribasim.main(["--version"])
    end
    @test value == 0
    @test output != "nothing"
    @test VersionNumber(output) > v"0.0.0"
end
```

It works locally, but as described in the issue, the binary .exe returns "nothing".
The solution provided in this PR breaks this adjusted test in code mode, but will make the binary.exe return correct values.

I would suggest to make a bug report that shows these differences with an example: https://github.com/JuliaLang/julia/issues

core/src/utils.jl
```diff
@@ -1,9 +1,15 @@
 "Get the package version of a given module"
 function pkgversion(m::Module)::VersionNumber
-    rootmodule = Base.moduleroot(m)
-    pkg = Base.PkgId(rootmodule)
-    pkgorigin = get(Base.pkgorigins, pkg, nothing)
-    return pkgorigin.version
+    version = Base.pkgversion(Ribasim)
+    if !isnothing(version)
+        return version
+    else
```
deltamarnix, line 6: I don't know how the Ribasim team is on this, but to have less nesting you could write it as: `!isnothing(version) && return version`

Hofer-Julian: Fine by me

visr (approved):

Issue #968, opened 2024-01-18
Multiple define of variable "S"


Jingru923:
**What**
In the documentation Equation and Allocation under Julia core, the variable $S$ and its lower-case are defined more than once. $S$ denotes source in Equation and denotes subnetwork in Allocation. @SouthEndMusic 

**Why**
To avoid confusion, we should find other letter to denote one or more of the variable. Meanwhile we should also check if other variables are defined twice.

PR #1003, opened 2024-01-29
Remove using S for storage (use u)

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/968.

### DIFF
--- a/docs/core/equations.qmd
+++ b/docs/core/equations.qmd
@@ -42,7 +42,7 @@ $\mathbf{u}(t)$ is given by all the states of the model, which are (currently) t
 Given a single basin with node ID $i \in B$, the equation that dictates the change of its storage over time is given by
 
 $$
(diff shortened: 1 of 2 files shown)


Jingru923:
I see the new letter you are using is also used somewhere. For example, in equation section, equation (1) has also 'u'. In numerical considerations, equation (1), there is also 'u'. 
For me, I think the doc is less confusing than before, it's improving. However, how do you think about the overlapping of 'u'? @SouthEndMusic

SouthEndMusic:
@Jingru923 these all refer to the same $u$, which is the state vector of the model.

Jingru923:
So the storages are states?

SouthEndMusic:
> So the storages are states?

Indeed, the system of differential equations which is solved by the Ribasim core is in terms of the storages of the basins (and the integral term of the PID controller).

Jingru923 (approved):

PR #1021, opened 2024-01-31
Keep only the root and core project

visr:
This attempts to simplifiy our Julia project and build setup.

Where possible, existing projects are moved into the root project like in #1014. All core dependencies are now also direct root dependencies to make using the root project more convenient for development.

While working on this I realized that we can remove the `libribasim` and `ribasim_cli` projects entirely. `libribasim` is now a module of the Ribasim package, and `ribasim_cli` already was just a tiny wrapper around `Ribasim.main`. The content of the README.md files from those two folders now reside in the `build_lib` and `build_app` docstring.

Since that left only the `create_binaries` folder under `/build`, I collapsed that level out. `create_binaries` is no longer a package, but the code is still used, but now simply directly included in `build.jl`.


Hofer-Julian (approved):
If the tests work, it can be merged from my side.
Nice yak shave

Project.toml
```diff
@@ -3,37 +3,62 @@ authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
...
+LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
 MarkdownTables = "1862ce21-31c7-451e-824c-f20fa3f90fa2"
 MetaGraphsNext = "fa8bd995-216d-47f1-8a91-f3b68fbeb377"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+PackageCompiler = "9b87118b-4619-50d2-8e1e-99f35a4d4d9d"
+PreallocationTools = "d236fae5-4411-538c-8e31-a6e3d9e00b46"
 ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 Ribasim = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
```
deltamarnix, line 46: I see that Ribasim itself is still in this list. Does that make sense?

Project.toml
```diff
@@ -3,37 +3,62 @@ authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
...
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
 MarkdownTables = "1862ce21-31c7-451e-824c-f20fa3f90fa2"
 MetaGraphsNext = "fa8bd995-216d-47f1-8a91-f3b68fbeb377"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+PackageCompiler = "9b87118b-4619-50d2-8e1e-99f35a4d4d9d"
+PreallocationTools = "d236fae5-4411-538c-8e31-a6e3d9e00b46"
 ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
deltamarnix, line 45: The split made it possible to add debug items that would not be needed to package in the binary. How do we know if we don't pack too many dependencies now? Should we remove some developer only packages and set it up in a different way?

deltamarnix (commented):
I have some questions about the root Project.toml

Project.toml
visr: Yes. The root project is the superset of the core + all other dev/build/other deps we need. Including Ribasim here means we can use Ribasim from the root environment.

Project.toml
visr: The root project is used to build the binaries, but it is only the core project that is being built. So there is no risk to adding more dev dependencies to the root project.

Before this change we did also have `libribasim` and `ribasim_cli` projects. So we do lose the ability to add dependencies for our binaries that are not depdendencies of the core. But there are no such dependencies right now.

Issue #1027, opened 2024-02-01
NaN entries are basin.time are not succesfully filtered


Huite:
I'm running a model with drainage and infiltration set to NaN (ideally so they are skipped in a coupled run). As I understand, this should be supported.

However, it results in the following error:

```
julia> Ribasim.run("ribasim.toml")
ERROR: MethodError: Cannot `convert` an object of type Missing to an object of type Float64

Closest candidates are:
  convert(::Type{T}, ::VectorizationBase.AbstractSIMD) where T<:Union{Bool, Float16, Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, SIMDTypes.Bit}
   @ VectorizationBase C:\Users\bootsma\.julia\packages\VectorizationBase\xE5Tx\src\base_defs.jl:201
  convert(::Type{T}, ::VectorizationBase.LazyMulAdd{M, O, I}) where {M, O, I, T<:Number}
   @ VectorizationBase C:\Users\bootsma\.julia\packages\VectorizationBase\xE5Tx\src\lazymul.jl:25
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:84
  ...

Stacktrace:
  [1] macro expansion
    @ C:\Users\bootsma\.julia\packages\StructArrays\irgIp\src\utils.jl:184 [inlined]
  [2] bypass_constructor(::Type{Ribasim.BasinTimeV1}, args::Tuple{Int64, Dates.DateTime, Missing, Float64, Missing, Float64, Float64})
    @ StructArrays C:\Users\bootsma\.julia\packages\StructArrays\irgIp\src\utils.jl:184
  [3] createinstance(::Type{Ribasim.BasinTimeV1}, ::Int64, ::Vararg{Any})
    @ StructArrays C:\Users\bootsma\.julia\packages\StructArrays\irgIp\src\interface.jl:49
  [4] _getindex
    @ C:\Users\bootsma\.julia\packages\StructArrays\irgIp\src\structarray.jl:353 [inlined]
  [5] getindex
    @ C:\Users\bootsma\.julia\packages\StructArrays\irgIp\src\structarray.jl:348 [inlined]
  [6] _issorted(v::StructArrays.StructVector{Ribasim.BasinTimeV1, @NamedTuple{…}, Int64}, lo::Int64, hi::Int64, o::Base.Order.By{typeof(Ribasim.sort_by_time_id), Base.Order.ForwardOrdering})
[...]

PR #1028, opened 2024-02-01
Allow missing 'Basin / time' data

Huite:
This is a very opportunistic fix to #1027

I'm not sure the `!isnan(value)` check currently does anything meaningful to be honest?

It's still a bit tedious to setup: to get it working I need to explicitly set the first date to zero, otherwise it triggers the following error:             `error("Missing initial data for the Basin variable drainage")`

One way of dealing with this would be to initialize to zeros, and allow "unset" values for optionals. Not sure about it.

Needs a decision, and then we could add some simple tests.


Huite:
From the imod coupler perspective, it's a bit onerous. Having to set all entries to NaN, except the first feels very specific.

Ideally, I think the following would work as preparation for the coupling step:

```python
df = ribasim_model.basin.df
df.loc[df["node_id"].isin(coupled_node_ids), ["drainage", "infiltration"]] = np.nan
```

Requires altering this:

```
    drainage = fill(NaN, length(node_id))
    infiltration = fill(NaN, length(node_id))
```

using `zeros` instead and ommitting the check.

Having zero as values if they are not defined feels reasonable?

visr (approved):
I initialized with zeros and allowed NaN for the other columns as well.
Also added a simple test model.
Does this work well for you?

Huite:
Judging from the test, this does exactly the right thing. I'll try and run this branch on one of my local models to be sure.

PR #1063, opened 2024-02-02
Stop running CompatHelper on root project again

visr:
In #1021 I enabled to run CompatHelper on the root project, but seeing the flood of PRs this morning and thinking about it some more, I think this was a mistake. We should close all those PRs.

Compat is needed in core for all dependencies. For other dependencies, in principle we shouldn't declare compat, since we use a Manifest anyway. This is similar to the way we allow any version in pixi.toml, and the monthly pixi.lock update also brings in possibly breaking changes that we can then evaluate.

We could have compat entries in the root project. Those makes sense if we are not compatible with newer breaking releases to hold it back. This way the manifest update will not bring in this incompatible release. We should not forget about these, so ideally we have an issue labeled `tech-debt` for those.


Hofer-Julian (approved):
Sounds good to me

PR #1080, opened 2024-02-05
Testmodels: run one year instead of ten

### LINKED ISSUE #1031: Minimize running time of backwater and two basin tests models
I've been adding some additional tests to imod_coupler. That means running more models, which means the tests take more time to run.

I myself am at least the worst offender, since both the backwater model and the two basin model run for 10 years. Looking at what's tested, this guarantees they're approximated a steady-state condition, but 10 years is quite excessive.
The other models seem to run for a month to a year, which is more reasonable.

Huite:
No adverse effects: results have already achieved steady-state after one year.
This obviously reduces running times, especially when a MODFLOW6 model is run as well.

Fixes #1031

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ report.xml
 
 # Designated working dir for working on Ribasim models
 models/
+
+# Result of build steps.
+build/
--- a/.gitignore
+++ b/.gitignore
@@ -154,6 +154,3 @@ report.xml
 
 # Designated working dir for working on Ribasim models
 models/
-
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -154,3 +154,6 @@ report.xml
 
 # Designated working dir for working on Ribasim models
 models/
+
+# Result of build steps.
+build/
```
visr, line 159:
```suggestion
```

visr (approved):

Issue #893, opened 2023-12-13
Extend and exposing user demands for coupling with MetaSWAP


HendrikKok:
For the unsaturated zone we will couple Ribasim to MetaSWAP. For irrigation from surface water, MetaSWAP uses the soil moister content to trigger the irrigation demand. Via the coupling we want to connect this irrigation demand to a user-demand in Ribasim. After a timestep we want to return a realized volume back to MetaSWAP.  

To be able to do this we need a user demand + realized memory address to de exposed in the BMI. Most likely we also need the node id array to be exposed since we can’t assume that user one is always index one of the exposed arrays?

A possible problem is that now, the user demand is a Vector{ScalarInterpolation}. For demand from input this is probably right, from the coupling perspective we want something like a Vector{Float64}. Also I'm not sure is the realized volume is now part of the user type.

SouthEndMusic:
For the demand I propose:
- A `demand::Vector{Vector{Float64}}` field with the actual demands per user per priority of the current timestep, either coming from coupling or from the existing time series input. This can then be exposed via the BMI. I think indeed also the `node_id` array should be exposed (although that may be a bit tricky since that array does not contain `Int64` directly but `Int64` wrapped in a `NodeID` type. The node IDs are not consecutive, but they are sorted). Maybe also the `priorities` vector, because those do not have to be consecutive either. 
- Rename the current `demand::Vector{Vector{ScalarInterpolation}}` field to something like `demand_time`.

Some practical questions:
- Does this irrigation demand always have the same priority?
- How should the user node know where the demand comes from?
- Should user nodes be able to have both types of demand or only either from coupling or from a time series?

visr:
> - Does this irrigation demand always have the same priority?

I'd say so, yes.

> - How should the user node know where the demand comes from?
> - Should user nodes be able to have both types of demand or only either from coupling or from a time series?

I think that we should only support coupling static user demands, where the static only serves as an initial demand, which is then set by Metaswap through BMI.

In general I don't think we should let BMI conventions of flat arrays dictate our internal datastructures. I don't like that BMI seems to have these expectations. I'd prefer if Metaswap could just tell Ribasim, "set the demand for User ID X to Y", but BMI doesn't seem to offer this.

SouthEndMusic:
> I think that we should only support coupling static user demands, where the static only serves as an initial demand, which is then set by Metaswap through BMI.

Ok, but currently we handle static demands by converting them to trivial interpolation objects

[...]

PR #1081, opened 2024-02-05
Introduce flat demand vector and expose via BMI

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/893.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -681,6 +681,8 @@ function BMI.get_value_ptr(model::Model, name::AbstractString)
         model.integrator.p.basin.drainage
     elseif name == "subgrid_level"
         model.integrator.p.subgrid.level
+    elseif name == "demand"
+        model.integrator.p.demand
     else
         error("Unknown variable $name")
     end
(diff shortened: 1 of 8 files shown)


core/src/utils.jl
```diff
@@ -1282,3 +1282,23 @@ end
...
+    n_priorities = length(user.priorities)
+    return demand[(user_idx - 1) * n_priorities + priority_idx]
+end
+
+function set_user_demand!(
+    user::User,
+    node_id::NodeID,
+    priority_idx::Int,
+    value::Float64,
+)::Nothing
```
visr, line 1298: Could use tests.

visr (approved):

core/src/bmi.jl
```diff
@@ -681,6 +681,8 @@ function BMI.get_value_ptr(model::Model, name::AbstractString)
         model.integrator.p.basin.drainage
     elseif name == "subgrid_level"
         model.integrator.p.subgrid.level
+    elseif name == "demand"
+        model.integrator.p.demand
```
Hofer-Julian, line 685: Would be good to add a test on the `ribasim_api` side.

PR #1082, opened 2024-02-06
Water supply/demand by basins in allocation (new node type `AllocationTarget`)

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/564.


gijsber:
Hi @SouthEndMusic, just a small check. When using the vertical fluxes from the simulation, I assume you take the average or accumulation and not the instantaneous values

SouthEndMusic:
@gijsber, good point, I currently take instantaneous values. I am somewhat aware of efforts to obtain integrated values (also requested for coupling: https://github.com/Deltares/Ribasim/issues/1084). As far as I am aware the ball for this one is in @visr's court.

SouthEndMusic:
To do for adding flows into basins to the objective function:
- [ ] Define absolute value variables for the flows into the basins
- [ ] Define constraints on these variables so that they act as the absolute value of the right expression
- [ ] Finish `add_basin_term!` by passing the proper constraints to `add_objective_term!`
- [ ] Properly call `add_basin_term!` add the end of `set_objective_priority!`

SouthEndMusic:
@gijsber in this PR I will keep using instantaneous values, I made a new issue for using averaged values: https://github.com/Deltares/Ribasim/issues/1121.

SouthEndMusic:
I added some tests with this model:

allocation_level_control

The idea is that basin 2 starts with its level below the minimum level, and so user 3 will only abstract after the basin has been filled. This does work:

allocation_target_level_time

There are 3 stages:
1. Before the first allocation solve, where the user abstracts fully (identical to when allocation is not used);
2. Between the first allocation solve and when the basin reaches its maximum level, where the user does not abstract;
3. After the basin has reached its maximum level, where the user abstracts fully again.

All this is qualitatively as expected, but when I writing the tests I found some weird but structured errors. Here are the errors per stage (expected - simulated):

allocation_target_level_time
allocation_target_level_time
allocation_target_level_time

See also the code for the tests in `allocation_test.jl`.

core/src/allocation_init.jl
```diff
@@ -684,6 +779,26 @@ function add_constraints_fractional_flow!(
...
+"""
+Add the basin flow constraints to the allocation problem.
+The constraint indices are the basin node IDs.
+NOTE: negative basin flow means flow out of the basin.
+
+Constraint:
+flow basin >= - basin capacity
+"""
+function add_constraints_basin_flow!(problem::JuMP.Model)::Nothing
+    F_basin_in = problem[:F_basin_in]
```
visr, line 791: Unused
```suggestion
```

core/src/allocation_init.jl
```diff
@@ -684,6 +779,26 @@ function add_constraints_fractional_flow!(
     return nothing
 end
 
+"""
+Add the basin flow constraints to the allocation problem.
+The constraint indices are the basin node IDs.
+NOTE: negative basin flow means flow out of the basin.
+
+Constraint:
+flow basin >= - basin capacity
```
visr, line 788: Can you explain this constraint a bit more? How is basin flow and capacity defined? What does a negative capacity mean? And how does it relate to the 0.0 in the code below?

core/src/allocation_optim.jl
visr, line 1: I like that you split the file into `allocation_init.jl` and `allocation_optim.jl`, but it makes it difficult to review the changes to the optim code, since the diff says that optim.jl is all new. So best to split these kind of non-functional refactors into separate PRs.

core/src/allocation_optim.jl
```diff
@@ -0,0 +1,592 @@
...
+- The index of the basin
+"""
+function get_basin_data(
+    allocation_model::AllocationModel,
+    p::Parameters,
+    u::ComponentVector,
+    node_id::NodeID,
+)
+    (; graph, basin, allocation_target) = p
+    (; Δt_allocation) = allocation_model
```
visr, line 374:
```suggestion
    (; Δt_allocation) = allocation_model
    @assert node_id.type == NodeType.Basin
```

core/src/allocation_optim.jl
```diff
@@ -0,0 +1,592 @@
...
+Storages are converted to flows by dividing by the allocation timestep.
+"""
+function get_basin_capacity(
+    allocation_model::AllocationModel,
+    u::ComponentVector,
+    p::Parameters,
+    t::Float64,
+    node_id::NodeID,
+)::Float64
+    (; allocation_target) = p
```
visr, line 403:
```suggestion
    (; allocation_target) = p
    @assert node_id.type == NodeType.Basin
```

core/src/allocation_optim.jl
```diff
@@ -0,0 +1,592 @@
...
+Storages are converted to flows by dividing by the allocation timestep.
+"""
+function get_basin_demand(
+    allocation_model::AllocationModel,
+    u::ComponentVector,
+    p::Parameters,
+    t::Float64,
+    node_id::NodeID,
+)::Float64
+    (; allocation_target) = p
```
visr, line 428:
```suggestion
    (; allocation_target) = p
    @assert node_id.type == NodeType.Basin
```

visr (commented):
Left some comments, but looks good overall. I can rename to `TargetLevel` in a separate PR.

core/src/allocation_init.jl
SouthEndMusic: I updated the description. This was from when I still used a single variable for flow in and out of the basin, now I have separate variables for that. The $0.0$ below does not have a specific meaning, it's a dummy value to initialize the constraint.

docs/core/usage.qmd
```diff
@@ -486,29 +458,37 @@ Note that a `node_id` can be either in this table or in the static one, but not
...
+This table is the transient form of the `AllocationTarget` table, in which time-dependent minimum and maximum levels can be supplied.
+Similar to the static version, only a single priority per `AllocationTarget` node can be provided.
+
+column        | type     | unit         | restriction
+------------- | -------  | ------------ | -----------
+node_id       | Int      | -            | sorted
+time          | DateTime | -            | sorted per node id
+min_level     | Float64  | $m$          | -
+max_level     | Float64  | $m$          | -
+priority      | Int      | -            | positive
```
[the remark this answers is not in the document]
SouthEndMusic: This isn't true for this table because there's only one priority per node.

core/src/allocation_init.jl
```diff
@@ -489,6 +522,30 @@ function outflow_ids_allocation(graph::MetaGraph, node_id::NodeID)
...
+    else
+        0.0
+    end
+end
+
+function get_basin_outflow(
+    problem::JuMP.Model,
+    node_id::NodeID,
+)::Union{JuMP.VariableRef, Float64}
+    F_basin_out = problem[:F_basin_out]
```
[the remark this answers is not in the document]
SouthEndMusic: Here I want to allow that the node ID is not from a basin, in that case the function returns 0.0.

visr (approved):

PR #1083, opened 2024-02-06
change leq to eq constrain in flow conservation

### LINKED ISSUE #954: Use equality constraints for flow conservation
**What**
After discussion with the RTC-Tools team, we came to the conclusion that equality constraints are fine for flow conservation in the optimization problems for allocation. 

**Why**
The advantage of equality constraints is that the flow cost term introduced in https://github.com/Deltares/Ribasim/pull/886 is probably no longer needed.

**How**
- [x] See `function add_constraints_flow_conservation!`;
- [ ] Also update this in the documentation [here](https://deltares.github.io/Ribasim/core/allocation.html#the-optimization-variable-constraints);
- [ ] Check whether it's now fine to remove the flow cost and if so, also remove it in the documentation [here](https://deltares.github.io/Ribasim/core/allocation.html#the-optimization-objective).

Jingru923:
Fixes #954

I change the less than or equal constraints to equal constraints. I also delete the flow cost term in the `linear_absolute` and `linear_relative` objective functions.

This result in a change in the optimization solutions.


core/src/allocation.jl
```diff
@@ -883,20 +883,20 @@ function set_objective_priority!(
...
+    #         JuMP.add_to_expression!(ex, cost_per_flow * flow) #
+    #     end
+    # elseif objective_type == :linear_relative
+    #     if demand_max > 0.0
+    #         cost_per_flow = 0.5 / (demand_max * length(F))
+    #         for flow in F
+    #             JuMP.add_to_expression!(ex, cost_per_flow * flow) #
+    #         end
+    #     end
+    # end
```
SouthEndMusic, line 899: Don't forget to remove commented out code before merging. Also, note that `demand_max` is no longer needed without the flow costs so you can remove the initialization and computation of that variable. 

Our collegues are quite strict with removing unneeded code. If a variable shows up grey in the code, that usually means that it is not used and thus that you can remove it.

Jingru923: Yes! I left it here for other purpose. I will remove it before merging

core/src/allocation.jl
```diff
@@ -1128,7 +1113,9 @@ function allocate!(
         @debug JuMP.solution_summary(problem)
         if JuMP.termination_status(problem) !== JuMP.OPTIMAL
             (; allocation_network_id) = allocation_model
-            priority = priorities[priority_index]
+            priority = priorities[priority_idx]
+            println(JuMP.solution_summary(problem))
+            println(problem)
             error(
```
SouthEndMusic, line 1119: We don't want these print statements on the main branch. The line starting with `@debug` above adds the solution summary to the logging data if Ribasim is run in debug mode. I don't know how to access the logging data, but it shouldn't be too hard.

core/test/allocation_test.jl
```diff
@@ -254,7 +246,8 @@ end
     end
```
SouthEndMusic, line 246: You don't have to mark any test in this testitem as broken. Your changes altered the behaviour of the allocation algorithm a bit, but mostly for the better since a larger flows are allocated. You can simply change the reference values to the new values.

SouthEndMusic (changes requested):
It mainly looks good, I've added some comments.

Jingru923 (commented):
All comments are now resolved

SouthEndMusic (approved):
Nice, work, looks good to me! One thing you could do for clarity is to add to the titles of the new testitems your created that those names indicate objective function types. If you decide to change this, you can merge without asking for a new review.

PR #1086, opened 2024-02-06
Switch components over to the same calver version number

### LINKED ISSUE #1076: Improve versions of Ribasim Suite components
At the moment, versioning is a bit confusing.
We use CalVer for the whole release and SemVer for the separate packages.
Additionally, it isn't always obvious to users which versions they currently use.

## Proposal 

@visr and I propose the following

- Use the same CalVer version for Ribasim Suite, Ribasim Core, Ribasim QGIS, Ribasim Python and Ribasim API
- Log version when Ribasim CLI is run

## Regressions

- Users cannot rely on semantic versions anymore when it comes to Ribasim Python and Ribasim API

visr:
Fixes #1076

The new logging looks like this:

```
│   ribasim_version = v"2024.2.0"
│   starttime = 2020-01-01T00:00:00
└   endtime = 2021-01-01T00:00:00
Simulating 100%|███████████████████████████████████████| Time: 0:00:00
[ Info: The model finished successfully
```

### DIFF
--- a/core/Project.toml
+++ b/core/Project.toml
@@ -2,7 +2,7 @@ name = "Ribasim"
 uuid = "aac5e3d9-0b8f-4d4f-8241-b1a7a9632635"
 authors = ["Deltares and contributors <ribasim.info@deltares.nl>"]
 manifest = "../Manifest.toml"
-version = "0.7.0"
+version = "2024.02.0"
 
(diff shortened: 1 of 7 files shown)


Hofer-Julian (approved):
Looks good to me

PR #1088, opened 2024-02-06
Correct precipitation equation

visr:
Before we used the current wetted area to convert the precipitation $m/s$ input to a flow rate $m^3/s$. It was deemed simpler and more stable to use the maximum profile area instead, otherwise empty Basins with 0 wetted area could never fill back up due to rain. This updates the equations to reflect that.

### DIFF
--- a/docs/core/equations.qmd
+++ b/docs/core/equations.qmd
@@ -149,19 +149,21 @@ plt.show()
 The precipitation term is given by
 
 $$
-    Q_P = P \cdot A(u).
+    Q_P = P \cdot A
 $$ {#eq-precip}
 
-Here $P = P(t)$ is the precipitation rate and $A$ is the wetted area. $A$ is a
-function of the storage $u = u(t)$: as the volume of water changes, the area of the free water
-surface may change as well, depending on the slopes of the surface waters.
+Here $P = P(t)$ is the precipitation rate and $A$ is the maximum area given in the `Basin / profile` table.
+Precipitation in the Basin area is assumed to be directly added to the Basin storage.
+The modeler needs to ensure all precipitation enters the model, and there is no overlap in the maximum profile areas, else extra water is created.
+If a part of the catchment is not in any Basin profile, the modeler has to verify that water source is not forgotten.
+It can for instance be converted to a flow rate and added to a Basin as a FlowBoundary.
 
 ## Evaporation
 
 The evaporation term is given by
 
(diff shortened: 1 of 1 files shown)


docs/core/equations.qmd
```diff
@@ -149,19 +149,21 @@ plt.show()
...
+If a part of the catchment is not in any Basin profile, the modeler has to verify that water source is not forgotten.
+It can for instance be converted to a flow rate and added to a Basin as a FlowBoundary.
 
 ## Evaporation
 
 The evaporation term is given by
 
 $$
-    Q_E = E_\text{pot} \cdot A(u) \cdot \phi(d;0.1).
+    Q_E = E_\text{pot} \cdot A(u) \cdot \phi(d;0.1)
```
SouthEndMusic, line 166: In a scientific writing course I was thought that equations in this context must be treated as part of the sentence, and so there should be a period at the end of the equation.

docs/core/equations.qmd
```diff
@@ -149,19 +149,21 @@ plt.show()
...
 
 $$
-    Q_P = P \cdot A(u).
+    Q_P = P \cdot A
 $$ {#eq-precip}
 
-Here $P = P(t)$ is the precipitation rate and $A$ is the wetted area. $A$ is a
-function of the storage $u = u(t)$: as the volume of water changes, the area of the free water
-surface may change as well, depending on the slopes of the surface waters.
+Here $P = P(t)$ is the precipitation rate and $A$ is the maximum area given in the `Basin / profile` table.
```
SouthEndMusic, line 155: Is it really the maximum area that is used? Considering 

```julia
fixed_area = basin.area[i][end]
```
it depends on the ordering of these vectors.

SouthEndMusic (commented):
I have some questions

docs/core/equations.qmd
visr: Interesting, I didn't know. I added them back. There do seems to be quite some equations without dots though.

docs/core/equations.qmd
visr: Indeed it depends on the ordering, but that is enforced to be increasing. We rely on these orderings in many places in the code, so I don't think an explicit max is needed.

https://deltares.github.io/Ribasim/core/usage.html#basin-profile

SouthEndMusic (approved):

PR #1095, opened 2024-02-07
Validate the equality of timestamps in a test

deltamarnix:
Subsecond precision is not working when loading it back in from a SQLite database. The validation in the test did not check this.

The question is if we want to have such high precision for a datetime.


visr:
> The question is if we want to have such high precision for a datetime.

The Julia DateTime we use has millisecond precision, and the solver uses a Float64 in seconds since simulation start. Do you know if millisecond precision passes the test?

deltamarnix:
I have adjusted the test to prove that it was never working, that's why I decided to remove the precision from the test model.

visr:
Yeah my question is if any subsecond precision is lost, or just submillisecond precision.

deltamarnix:
I tried to change the code to `00:00:00.001`, but that was also invalid. I see that the function is also fine with a format that we pass to the reader: https://pandas.pydata.org/pandas-docs/version/1.2.4/reference/api/pandas.read_sql_query.html

I will try to see if I can change the `parse_dates` parameter in read_sql_query. Now it's only a list `parse_dates=["time"]`, but we could go for a dictionary with a format: `parse_dates={"time": "<cool_format_here>"}`

deltamarnix:
@visr I was able to fix it by digging deeper into the internals of pandas.

PR #1096, opened 2024-02-07
Generate python code for schemas directly from Julia

### LINKED ISSUE #887: Support datamodel-code-generator 0.25+
**What**
With #897 we use datamodel-code-generator 0.24, but if we use 0.25 we get the following issue when running
```
pixi run codegen
pixi run generate-testmodels
```
the first line introduces problems that become apparent when running the second line. Error when trying to generate `basic_transient` model:
```
Value error, Error while coercing 'BasinTimeSchema' to type <class 'ribasim.models.BasinTime'>: Could not coerce <class 'pandas.core.frame.DataFrame'> data_container into type <class 'ribasim.models.BasinTime'>
      index                                failure_case
0         0  {'time': Timestamp('2020-01-01 00:00:00')}
1         1  {'time': Timestamp('2020-01-01 00:00:00')}
2         2  {'time': Timestamp('2020-01-01 00:00:00')}
3         3  {'time': Timestamp('2020-01-01 00:00:00')}
4         4  {'time': Timestamp('2020-01-02 00:00:00')}
...     ...                                         ...
1463   1463  {'time': Timestamp('2020-12-31 00:00:00')}
1464   1464  {'time': Timestamp('2021-01-01 00:00:00')}
1465   1465  {'time': Timestamp('2021-01-01 00:00:00')}
1466   1466  {'time': Timestamp('2021-01-01 00:00:00')}
1467   1467  {'time': Timestamp('2021-01-01 00:00:00')}

[1468 rows x 2 columns] [type=value_error, input_value=      node_id       time ...
[1468 rows x 7 columns], input_type=DataFrame]
    For further information visit https://errors.pydantic.dev/2.5/v/value_error
```

It is probably due to the fact that in `python/ribasim/ribasim/models.py`, `datetime.datetime` gets replaced by `pydantic.AwareDatetime`.

**Why**
Generating test models should just work.

**How**
Not sure yet. Options are:

- We switch to properly using and testing timezone aware datetimes everywhere, so we can use `pydantic.AwareDatetime`.
- We (automatically) replace `pydantic.AwareDatetime` with `pydantic.NaiveDatetime`
- Contribute such an option upstream if there is demand

https://github.com/koxudaxi/datamodel-code-generator/pull/1735
https://docs.pydantic.dev/latest/api/types/#pydantic.types.AwareDatetime

### LINKED ISSUE #904: Better schema validation errors in Ribasim Python
**What**

When adding the node_id, time, precipitation and potential_evaporation for basins in transient, the validation error is shown. 
Makes sense, it needs also the columns 'drainage' etc. However, the validation error was not clear. Which columns or which types were needed? Especially, if you don't need them for your current model simulation.

**Why and how**

Making the error more specific. The user will, most likely, understand the error and what needs to be adjusted. In this case adding empty 'drainage', 'urban_runoff', and 'infiltration' columns, even though they are not used. Also, the value should then be 0.0 not 0, since it needs float type. 

**Additional context**

```
model.basin.time.df = forcing

  Cell In[115], line 1
    model.basin.time.df = forcing

  File C:\Ribasim9\.pixi\env\Lib\site-packages\pydantic\main.py:786 in __setattr__
    self.__pydantic_validator__.validate_assignment(self, name, value)

ValidationError: 1 validation error for TableModel[BasinTimeSchema]
df
  Value error, Error while coercing 'BasinTimeSchema' to type <class 'ribasim.models.BasinTime'>: Could not coerce <class 'pandas.core.frame.DataFrame'> data_container into type <class 'ribasim.models.BasinTime'>
             column  index         failure_case
0              time      0  2016-01-01 00:00:00
1              time      1  2016-01-02 00:00:00
2              time      2  2016-01-03 00:00:00
3              time      3  2016-01-04 00:00:00
4              time      4  2016-01-05 00:00:00
            ...    ...                  ...
10229  infiltration   1457                  0.0
10230  infiltration   1458                  0.0
10231  infiltration   1459                  0.0
10232  infiltration   1460                  0.0
10233  infiltration   1461                  0.0

[10234 rows x 3 columns] [type=value_error, input_value=           time  node_id ...
[1462 rows x 7 columns], input_type=DataFrame]
    For further information visit https://errors.pydantic.dev/2.5/v/value_error
```

### LINKED ISSUE #1077: Re-evaluate code-generation in Ribasim Python
**What**
We currently generate code from Julia core for Ribasim Python.

**Why**
The code-generation in Ribasim Python is convenient for developers, but complicated to understand.
We also see problems that we can trace back to it.

**How**
We should take a closer look and see if we can improve/simplify it.
There are two angles to it:
- generate pandera models directly instead of going via pydantic
- remove code generation

deltamarnix:
Remove intermittent json generation and dependency on the data model code generator

Fixes #904
Fixes #1077 
Fixes #887


python/ribasim/ribasim/schemas.py
```diff
@@ -1,37 +1,299 @@
...
+
+
+class BasinProfileSchema(pa.DataFrameModel):
+    node_id: Series[int] = pa.Field(nullable=False)
+    area: Series[float] = pa.Field(nullable=False)
+    level: Series[float] = pa.Field(nullable=False)
+
+    class Config:
+        add_missing_columns = True
+        coerce = True
```
deltamarnix, line 17: Perhaps we could simplify all these configs into one by creating our own base class.

python/ribasim_testmodels/ribasim_testmodels/basic.py
```diff
@@ -336,8 +336,7 @@ def tabulated_rating_curve_model() -> ribasim.Model:
             data={
                 "node_id": [3, 3, 3, 3, 3, 3],
                 "time": [
-                    # test subsecond precision
-                    pd.Timestamp("2020-01-01 00:00:00.000001"),
+                    pd.Timestamp("2020-01-01 00:00:01"),
```
deltamarnix, line 339: Unfortunately subsecond precision wasn't working at all. See #1095 for a separate PR.

python/ribasim/ribasim/input_base.py
```diff
@@ -311,24 +311,11 @@ def tableschema(cls) -> TableT:
         T: TableT = fieldtype.__args__[0]
```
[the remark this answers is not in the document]
deltamarnix: This is the right way, python doesn't know anything about the generics on `cls` time.

utils/templates/model.py.jinja
```diff
@@ -0,0 +1,18 @@
+# Automatically generated file.
+# DO NOT MODIFY.
```
visr, line 2:
```suggestion
# Automatically generated file, do not modify.
```

No need to scream ;)

python/ribasim/ribasim/schemas.py
```diff
@@ -1,37 +1,204 @@
-"""
-Generate Pandera Schemas from all autogenerated Pydantic Models
-
-These classes have Schema as a postfix, so Ribasim.models.PumpStatic
-becomes Ribasim.schemas.PumpStaticSchema.
-"""
-import inspect
-import sys
+# Automatically generated file.
+# DO NOT MODIFY.
```
visr, line 2:
```suggestion
# Automatically generated file, do not modify.
```

utils/gen_python.jl
```diff
@@ -0,0 +1,57 @@
...
+using InteractiveUtils
+using Legolas
+using OteraEngine
+using Ribasim
+
+pythontype(::Type{<:AbstractString}) = "Series[str]"
+pythontype(::Type{<:Integer}) = "Series[int]"
+pythontype(::Type{<:AbstractFloat}) = "Series[float]"
+pythontype(::Type{<:Number}) = "Series[float]"
+pythontype(::Type{<:Bool}) = "Series[pa.BOOL]" # pa.BOOL is a nullable boolean type, bool is not
```
Hofer-Julian, line 12:
```suggestion
pythontype(::Type{<:Bool}) = "Series[pa.BOOL]" # pa.BOOL is a nullable boolean type, bool is not nullable
```

I assume that's what you mean? Had to read the sentence a few times to grasp it :)

python/ribasim/tests/test_io.py
```diff
@@ -62,7 +62,7 @@ def test_basic_transient(basic_transient, tmp_path):
     time = model_loaded.basin.time
     assert model_orig.basin.time.df.time[0] == time.df.time[0]
     __assert_equal(model_orig.basin.time.df, time.df)
-    assert time.df.shape == (1468, 9)
+    assert time.df.shape == (1468, 8)
```
visr, line 65: Is this because the "remarks" string column (needed as a hack for pandera) is gone? There are a few other occurances of "remarks" in the test code that would be good to update.

deltamarnix: thanks, done

visr (approved):
This looks like a good improvement, thanks!
The error message from #904 has gotten a lot better, pointing at the actual problem:

```
>>> import ribasim
>>> import pandas as pd
>>> ribasim.Terminal(static=pd.DataFrame(data={"no_id":[1]}))
  File "<stdin>", line 1, in <module>
  File "D:\Ribasim\.pixi\env\Lib\site-packages\pydantic\main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Terminal
static.df
  Value error, column 'node_id' in DataFrameSchema {'node_id': <Schema Column(name=node_id, type=DataType(int64))>} requires a default value when non-nullable add_missing_columns is enabled [type=value_error, input_value=   no_id
0      1, input_type=DataFrame]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error
```

Still I wonder how many users will understand "requires a default value when non-nullable add_missing_columns is enabled", instead of something simply stating that a required column is missing.

Hofer-Julian (approved):

PR #1099, opened 2024-02-08
Remove `update_timestep` in Ribasim Python

### LINKED ISSUE #1093: `update_timestep` in Ribasim Python's `Model` is not used
That config has been moved by @SouthEndMusic with https://github.com/Deltares/Ribasim/commit/c3c69a90d3444e11c95a9801f0d26c7015e79df8 on the Julia sized, but copied on the Python side.

We should remove `Model.update_timestep`.

Hofer-Julian:
Fixes #1093

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -12,7 +12,6 @@ from pydantic import (
     DirectoryPath,
     Field,
     field_serializer,
-    field_validator,
     model_serializer,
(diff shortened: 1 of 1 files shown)


visr (approved):
This is now `Solver(saveat=3600.0)`, which saves every hour. We could support fancy `datetime.timedelta` stuff for that just like we used to for `update_timestep`, but as-is is fine as well.

PR #1100, opened 2024-02-08
Add max_flow_rate to LinearResistance

visr:
This is needed to be able to model structures that exchange flow based on a water level difference, but only up to a certain capacity.

Ref https://github.com/Deltares/Ribasim-NL/issues/62

The updated test model now looks like this:

image

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -315,17 +315,16 @@ function formulate_flow!(
     t::Number,
 )::Nothing
     (; graph) = p
-    (; node_id, active, resistance) = linear_resistance
+    (; node_id, active, resistance, max_flow_rate) = linear_resistance
     for (i, id) in enumerate(node_id)
         basin_a_id = inflow_id(graph, id)
         basin_b_id = outflow_id(graph, id)
 
         if active[i]
-            q =
-                (
-                    get_level(p, basin_a_id, t; storage) -
-                    get_level(p, basin_b_id, t; storage)
-                ) / resistance[i]
+            h_a = get_level(p, basin_a_id, t; storage)
+            h_b = get_level(p, basin_b_id, t; storage)
+            q_unlimited = (h_a - h_b) / resistance[i]
+            q = clamp(q_unlimited, -max_flow_rate[i], max_flow_rate[i])
 
             # add reduction_factor on highest level
             if q > 0
--- a/core/test/equations_test.jl
+++ b/core/test/equations_test.jl
@@ -23,19 +23,24 @@
     @test ispath(toml_path)
     model = Ribasim.run(toml_path)
     @test successful_retcode(model)
-    p = model.integrator.p
+    (; p) = model.integrator
 
     t = Ribasim.timesteps(model)
     storage = Ribasim.get_storages_and_levels(model).storage[1, :]
-    basin_area = p.basin.area[1][2] # Considered constant
-    # The storage of the basin when it has the same level as the level boundary
-    limit_storage = 450.0
-    decay_rate = -1 / (basin_area * p.linear_resistance.resistance[1])
-    # TODO adapt for max_flow_rate of 6e-5
-    storage_analytic =
-        @. limit_storage + (storage[1] - limit_storage) * exp.(decay_rate * t)
+    A = p.basin.area[1][2]  # needs to be constant
+    u0 = A * 10.0
+    L = p.level_boundary.level[1].u[1]
+    R = p.linear_resistance.resistance[1]
+    Q_max = p.linear_resistance.max_flow_rate[1]
 
-    @test all(isapprox.(storage, storage_analytic; rtol = 0.005)) # Fails with '≈'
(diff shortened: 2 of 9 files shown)


core/test/equations_test.jl
```diff
@@ -31,6 +31,7 @@
     # The storage of the basin when it has the same level as the level boundary
     limit_storage = 450.0
     decay_rate = -1 / (basin_area * p.linear_resistance.resistance[1])
+    # TODO adapt for max_flow_rate of 6e-5
     storage_analytic =
         @. limit_storage + (storage[1] - limit_storage) * exp.(decay_rate * t)
```
visr, line 36: @SouthEndMusic it would be great if you could help updating this test value.

Huite: Solve for Q(level) == max_flow_rate

Find out at which times it switches from max_flow_rate to dynamic flow rate

Compute from there

SouthEndMusic:
DE for $u(t)$:

$$
u'(t) = \text{clamp}\left(\frac{L - \frac{u(t)}{A}}{R}, -q_{\max}, q_{\max}\right), \quad R, A, q_{\max} > 0.
$$

For simplicity we assume a constant area here, which is not the case in the current linear resistance test model but is now supported.

Initial condition:

$$
u(0) = u_0 > 0.
$$

We assume $u_0$ is big enough such that

$$
\frac{L - \frac{u_0}{A}}{R} < -q_{\max}.
$$

Hence for $0 < t < t_\text{shift}$ we have that

$$
u(t) = u_0 - q_{\max} t.
$$

$t_\text{shift}$ is given by

$$
\frac{L - \frac{u(t_\text{shift})}{A}}{R} = -q_{\max}.
$$

Substituting the above expression for $u(t)$ yields

$$
t_\text{shift} = \frac{u_0 - A(L + Rq_{\max})}{ q_{\max}},
$$

and so

$$
u(t_\text{shift}) = A(L + Rq_{\max}).
$$

Now we solve from here with

$$
u'(t) = \frac{L - \frac{u(t)}{A}}{R},
$$

which yields

$$
AL = ARu'(t) + u(t) = AR \exp\left(-\frac{t}{AR}\right)\left[u(t)\exp\left(\frac{t}{AR}\right)\right]'
$$

and so

$$
u(t)\exp\left(\frac{t}{AR}\right) = \frac{L}{R}\int \exp\left(\frac{t}{AR}\right) \text{d} t = AL  \exp\left(\frac{t}{AR}\right) + C,
$$

which gives

$$
u(t) = AL + C \exp\left(-\frac{t}{AR}\right).
$$

This makes sense; $u(t)$ will converge to $AL$, the volume in the basin at the level of the level boundary.

Filling in our initial condition:

$$
u(t_\text{shift}) = AL + C \exp\left(-\frac{t_\text{shift}}{AR}\right) = A(L + Rq_{\max})
$$

yields

$$
C = ARq_{\max} \exp\left(\frac{t_\text{shift}}{AR}\right).
$$

We conclude:

$$
u(t) = 
\begin{cases}
 u_0 - q_{\max} t \quad \text{if} \quad 0 \leq t \leq t_\text{shift}, \\ \\
AL + ARq_{\max} \exp\left(-\frac{t - t_\text{shift}}{AR}\right) \quad \text{if} \quad t > t_\text{shift}
\end{cases}.
$$

core/src/solve.jl
```diff
@@ -182,13 +182,15 @@ Requirements:
 
 node_id: node ID of the LinearResistance node
 active: whether this node is active and thus contributes flows
-resistance: the resistance to flow; Q = Δh/resistance
+resistance: the resistance to flow; Q = min(Δh/resistance, max_flow_rate)
```
SouthEndMusic, line 185: Is this just about flow magnitude and not direction?

docs/core/equations.qmd
```diff
@@ -269,15 +269,15 @@ of outlets, or even reversal of flows for high precipitation events.
...
 
 $$
-    Q = \frac{h_a - h_b}{R}
+    Q = \mathrm{clamp}(\frac{h_a - h_b}{R}, -Q_{\max}, Q_{\max})
 $$ {#eq-basinflow}
 
-Here $h_a$ is the water level in the first basin, $h_b$ is the water level in
-the second basin, and $R$ is the resistance of the link. A `LinearResistance` makes
-no assumptions about the direction of the flow: water flows from high to low.
+Here $h_a$ is the water level in the first basin, $h_b$ is the water level in the second basin.
```
SouthEndMusic, line 278:
```suggestion
Here $h_a$ is the water level in the first basin and $h_b$ is the water level in the second basin.
```

SouthEndMusic (approved):
Looking good, I trust you can update the analytical test with the maximum flow 👍

core/src/solve.jl
visr: No also direction, fixed. I forgot at first to handle negative Q.

visr:
image

The storage in black nicely goes from your u pre-shift in green to u post-shift in blue.

Issue #649, opened 2023-10-06
PoC: Create initial offline exchange with Delwaq


gijsber:
[hyd file_commented_EM.pdf](https://github.com/Deltares/Ribasim/files/12832297/hyd.file_commented_EM.pdf)

Attached file describes the content of a HYD file which holds delwaq input. the HYD file itself holds references to other files
Or use an alternative method for the coupling with Delwaq (together with Hidde)

Q4: create plan together with WES how to connect to Delwaq
If the plan is easy to realize (e.g. sequential file exchange, e.g. in HYD format) than we migt try to implement in Q4

gijsber:
LWKM needs a sequential coupling. All agree that online BMI coupling would be useful for salinity an other substances

Implementation 1:
ribasim.exe-> arrow->converter->hyd-->delwaq

Implementation 2 (towards BMI):
ribasim->BMI->hyd->dwq

Online coupling implementation:
ribasim->BMI->dwq, still needs topological info via UGrid

Arrow file needs to keep track of individual fluxes: Precip/Evap/Inf/Drainage
Flows should be average over previous timestep while storage should be actual at end of timestep.

Conversion needed from geopackage->ugrid (filter only basins and exchanges (connections)) 
TODO WES: investigate how they could make better use of Ugrid holding all data (pointer info, all dynamic data etc)

proposal:
hydromt_ribasim reader for geopackage and arrow (action by ribasim team)
hydromt_delwaq writes hyd file including pointer-file (action by ribasim team)

evetion:
Talked with arthur, it would make sense to have Delwaq `hyd` support in hydrolib-core. From there on, an updated HydroMT package could be made. From these building blocks a connection should be possible.

gijsber:
Please consider in your design also how your want to bring waste loads into the delwaq model (e.g. salinity injections)

PR #1111, opened 2024-02-12
Implement external timeseries for Delwaq coupling

### LINKED ISSUE #520: External data (for DiscreteControl)
Currently conditions are possible on basin levels and (transient) flow boundary flow rates. We also want to support conditions on variables which are not part of the model at all, but are only given by a time series (first mentioned in https://github.com/Deltares/Ribasim/issues/331). This should also support look ahead (https://github.com/Deltares/Ribasim/pull/516).

This functionality is also usefull for control based on external salinity conditions

evetion:
Fixes #520 
Required for reading in results from (and possibly input to) Delwaq (#649), which can be used in coupling.

This adds a generic External, node_id less Node, with only a time table, with fields `time` and `external`. It can be used as a global condition (climate forcing?). Furthermore, this adds a *single* `concentration` column to the Basin static/time tables. This can be used as both input for Delwaq coupling, as output from the Delwaq model.

The code for both inputs is very similar, hence the combined PR. Both inputs are now hardcoded single names, as a generic dynamic solution seems much more complex. Eventually we want to have several concentration/input fields, for setting multiple tracers.


evetion:
Closed in favour of #1137 (which show *how* these fields can be used).

PR #1113, opened 2024-02-12
Make node ID unique to the node type

visr:
Needed for #1110.

Before node IDs were globally unique. If you had the ID, you could look up the type in the Node table. When designing #1110 we came to the conclusion that this was not the right choice. It requires you to be aware of the IDs that are used throughout the model, whereas with this you only need to make sure that `Pump #5` doesn't already exist.

Most of the updates in this PR were to correct the tests. Some of the code and error messages became easier to read. If we talk about a node then we always know what the type is, there is no need to look it up first.

Most tables stay the same, e.g. if you have a `Terminal / static` table with a `node_id` Int column, you know that this refers to a Terminal NodeID. Only when connecting to other nodes do we need to specify the type next to the ID. So the `Edge` table now gets `from_node_type` next to `from_node_id`, the `PidControl / static` gets `listen_node_type` next to `listen_node_id`, etc. These extra columns are currently automatically filled in by Ribasim-Python on model save, hence they don't require changing the test models.

In terms of implementation, this basically adds the `type` field to `NodeID` and fixes the resulting errors.

```julia
struct NodeID
    type::NodeType.T
    value::Int
end
```

It does not yet test if models with the same node IDs (`Pump #1` and `Basin #1`) work, but this is hard to do right now with Ribasim Python, so best left for a later moment.

### DIFF
--- a/core/src/schema.jl
+++ b/core/src/schema.jl
@@ -1,7 +1,5 @@
 # These schemas define the name of database tables and the configuration file structure
 # The identifier is parsed as ribasim.nodetype.kind, no capitals or underscores are allowed.
-@schema "ribasim.node" Node
-@schema "ribasim.edge" Edge
 @schema "ribasim.discretecontrol.condition" DiscreteControlCondition
 @schema "ribasim.discretecontrol.logic" DiscreteControlLogic
 @schema "ribasim.basin.static" BasinStatic
@@ -56,22 +54,6 @@ function nodetype(
     end
 
     return Symbol(node[begin:length(n)]), k
-end
-
-@version NodeV1 begin
-    fid::Int
-    name::String = isnothing(s) ? "" : String(s)
-    type::String = in(Symbol(type), nodetypes) ? type : error("Unknown node type $type")
-    allocation_network_id::Union{Missing, Int}
-end
-
-@version EdgeV1 begin
-    fid::Int
-    name::String = isnothing(s) ? "" : String(s)
-    from_node_id::Int
-    to_node_id::Int
-    edge_type::String
-    allocation_network_id::Union{Missing, Int}
 end
 
 @version PumpStaticV1 begin
@@ -199,6 +181,7 @@ end
 
 @version DiscreteControlConditionV1 begin
     node_id::Int
+    listen_feature_type::String
     listen_feature_id::Int
     variable::String
     greater_than::Float64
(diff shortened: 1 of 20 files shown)


core/src/schema.jl
```diff
@@ -58,22 +56,6 @@ function nodetype(
...
-
-@version EdgeV1 begin
-    fid::Int
-    name::String = isnothing(s) ? "" : String(s)
-    from_node_id::Int
-    to_node_id::Int
-    edge_type::String
-    allocation_network_id::Union{Missing, Int}
-end
-
```
visr, line 76: These schemas were not used in the core, and their translations in Python were also not used, they are handwritten elsewhere. So best to just remove.

python/ribasim/ribasim/model.py
visr, line 1: This should be removed when Ribasim-Python is ready for it.

Hofer-Julian:
Julia Tests are still red :)

core/src/util.jl
```diff
@@ -383,17 +378,15 @@ function get_level(
     storage::Union{AbstractArray, Number} = 0,
 )::Union{Real, Nothing}
     (; basin, level_boundary) = p
-    hasindex, i = id_index(basin.node_id, node_id)
-    current_level = get_tmp(basin.current_level, storage)
-    return if hasindex
+    return if node_id.type == NodeType.Basin
```
Hofer-Julian, line 381:
```suggestion
   if node_id.type == NodeType.Basin
```

I am fine with having "return" in general if @Huite insists, but here it really doesn't make sense :P

python/ribasim/ribasim/model.py
```diff
@@ -309,6 +309,28 @@ def validate_model(self):
...
+        self._add_node_type(self.pid_control.static.df, id_col, type_col)
+        self._add_node_type(self.pid_control.time.df, id_col, type_col)
+        self._add_node_type(
+            self.discrete_control.condition.df,
+            "listen_feature_id",
+            "listen_feature_type",
+        )
+
     @classmethod
     def read(cls, filepath: FilePath) -> "Model":
```
Hofer-Julian, line 335:
```suggestion
    def read(cls, filepath: FilePath) -> Self:
```

Hofer-Julian (approved):
A few small nitpicks, but apart from that it looks good

PR #1116, opened 2024-02-13
Set mypypath dynamically

deltamarnix:
Dynamically try to get the relative path for the current working environment to set `MYPYPATH`

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -10,7 +10,6 @@ strict_concatenate = True
 disallow_subclassing_any = True
 disallow_untyped_decorators = True
 disallow_any_generics = True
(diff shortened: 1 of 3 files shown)


deltamarnix:
Tested with pixi 0.13.0 windows and pixi 0.11.0 linux

Hofer-Julian:
As discussed with @deltamarnix there's no need to support older pixi versions.
Let's not add additional logic to support that :)

Hofer-Julian (approved):
@deltamarnix just told me that we need this PR if we move to multi-environments
I am trusting him on that, so I've reopened this PR and approve

Issue #1119, opened 2024-02-13
Stop retrieving node type from node metadata 


SouthEndMusic:
**What**
After https://github.com/Deltares/Ribasim/pull/1113 it is no longer needed to retrieve the type of a node by its node ID from the node metadata, since the node type is now part of the `NodeID` struct.

**Why**
Simplify the code, fewer lookups.

PR #1120, opened 2024-02-13
Avoid lookup of node type from Node metadata

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1119.

At first I wanted to completely remove the field `NodeMetadata.type::Symbol`. After trying this, it turns out, this field is quite useful to obtain a snake case symbol version of the node type, for use in `getfield` and `neighbortypes`.

### DIFF
--- a/core/src/allocation.jl
+++ b/core/src/allocation.jl
@@ -494,14 +494,14 @@ Add the flow conservation constraints to the allocation problem.
 The constraint indices are user node IDs.
 
 Constraint:
-sum(flows out of node node) <= flows into node + flow from storage and vertical fluxes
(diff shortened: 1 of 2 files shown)


core/src/allocation.jl
```diff
@@ -1109,7 +1108,7 @@ function allocate!(
         @debug JuMP.solution_summary(problem)
         if JuMP.termination_status(problem) !== JuMP.OPTIMAL
             (; allocation_network_id) = allocation_model
-            priority = priorities[priority_idx]
+            priority = priorities[priority_index]
```
visr, line 1111:
```suggestion
            priority = priorities[priority_idx]
```

visr (approved):
1 suggestion

PR #1130, opened 2024-02-15
Introduce consistent naming of subnetworks and allocation networks

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/709.

The idea is that the modeler inputs a _subnetwork_, and what is derived from that by allocation initialization is called an _allocation network_.

The breaking change is that in the `Node` and `Edge` input tables, as well as the `allocation.arrow` output table, `allocation_network_id` has been replaced by `subnetwork_id`.


visr:
Can you describe the breaking change and provide some code in the top post to update models, just like the breaking PRs in https://github.com/Deltares/Ribasim/releases/tag/v2024.01.1?

visr (approved):

visr:
Updated the top post with code that can be used to upgrade. This is quite useful when updating old models.

PR #1135, opened 2024-02-15
Don't add `Basin / static` if there are no flows

visr:
One thing I didn't realize about #1079, is that we can make most example models shorter!


Huite:
Ah, nice catch!

Hofer-Julian (approved):

PR #1136, opened 2024-02-16
Use `Self` annotation when returning own type in methods

Hofer-Julian:


Hofer-Julian:
Seems like `Self` is only available starting from 3.11.
Let's close this for now

PR #1137, opened 2024-02-16
Offline coupling with Delwaq

### LINKED ISSUE #649: PoC: Create initial offline exchange with Delwaq

evetion:
Fixes #649

This PR adds a folder `coupling/delwaq` for offline coupling with Delwaq, with a more extensive description provided in `README.md` in the same folder.

A test is provided in `test.py`, that is also run on Teamcity on Windows, using the weekly DIMR releases. The `basic` testmodel (the only with built-in substance concentration values) is used for this. In order to speed-up the CI one can now do `pixi run ribasim-core-testmodels basic` to only run the `basic` model (but any or multiple model names can be used).


evetion:
So this is up to date with main again (I think), and not working, given the non consecutive boundaries ids. Next up is to remove non-basin nodes, it was nice to copy the Ribasim network 1-1 to Delwaq, but it seems to introduce some error in the results. The continituity tracer gives the worst mass balance at non-basin nodes (expected, as we set some non-zero storage), so let's remove them.

However, pruning the network is not *that* straightforward on a pure topology basis. While it's easy to remove a single node between two basins, there are some nodes that require recalculating flows based on flow directions/rules. See our basic model for example, where I crossed the easy ones, but removing the circled tabulated rating curve `#4` requires more work.
Screenshot 2024-04-20 at 21 38 53
I would propose to introduce a dependency on networkx to handle the graph related logic to prune the network. This is also a good intermediate issue that can be tackled by others.

SouthEndMusic:
Ah, you're starting to see why `FractionalFlow` is my nemesis. 

If I understand correctly, you want a routine that takes in the Ribasim model and spits out a new graph with edges only between basins and boundary nodes. You also want to know the flows over these edges. This might work (looking only at flow edges of course):

1. Start with the original graph and make each edge its own representative edge;
2. Loop over all nodes in the network. If it is a connector node, remove it and its connected edges, and create new edges from the upstream node to all downstream nodes. The representative edge of the new edges is the edge between the connector node and the downstream node;
3. Now you should be left with the graph you want. It can happen that there are multiple edges between the same pair of nodes (so the procedure above should allow this), so in that case the flow is given by the sum of the flows over their representative edges.

Using this representative edge system, I don't think any additional calculations or checks are needed.

evetion:
Ha, yes I understand your struggles.

What you propose is indeed what I want, with the addition that the nodes need to be renumbered so they're consecutive. Same for the boundary nodes, but then negatively numbered.

Is representative edge a name you came up with? You mean as an object to reason about (like done in the  networkx package)? 

This approach requires directed edges, and I wonder whether it always holds. I can imagine that if the fractionalflows here would be nodes that allow bidirectional (negative) flow, a solution would be morw complex. Not sure if such configurations are possible with our current nodetypes.

SouthEndMusic:
Yes representative edge is a term I came up with, I haven't looked into the networkx package yet. What I mean is that the edges you have in your delwaq graph are generally not in the original graph, so the flow over the edges in the delwaq graph is given by the sum of the flows over the representative edges in the original graph.

I'm not sure why the solution would be more complex with bidirectional edges. Do you want to convert to edges with only positive flow, so that you need edges in both directions?

I wrote this function which I think does what you want (the renumbering can be added in):

```julia
function create_graph_delwaq(graph::MetaGraph)::MetaGraph
    graph_delwaq = MetaGraph(
        DiGraph();
        label_type = NodeID,
        vertex_data_type = Nothing,
        edge_data_type = @NamedTuple{
            type::EdgeType.T,
            representative_edges::Vector{Tuple{NodeID, NodeID}},
        },
        graph_data = nothing,
    )

    # Add all flow edges to the delwaq graph
    for edge_metadata in values(graph.edge_data)
        (; type, from_id, to_id) = edge_metadata
        if type == EdgeType.flow
            graph_delwaq[from_id] = nothing
            graph_delwaq[to_id] = nothing

            edge = (from_id, to_id)
            graph_delwaq[edge...] = (type = EdgeType.flow, representative_edges = [edge])
        end
    end

    # Remove connector nodes from the delwaq graph
    for node_id in collect(values(graph_delwaq.vertex_labels))
        if node_id.type ∉ [
            NodeType.Basin,
            NodeType.Terminal,
            NodeType.LevelBoundary,
            NodeType.FlowBoundary,
            NodeType.UserDemand,
        ]
            inneighbor_id = inflow_id(graph_delwaq, node_id)
            outneighbor_ids = collect(outflow_ids(graph_delwaq, node_id))

            # From MetagraphsNext
            rem_vertex!(graph_delwaq, code_for(graph_delwaq, node_id))

            for outneighbor_id in outneighbor_ids
                edge = (inneighbor_id, outneighbor_id)
                representative_edge = (node_id, outneighbor_id)
                # If the new edge already exists, add a representative edge,
                # otherwise create the new edge
                if haskey(graph_delwaq, edge...)
                    push!(graph_delwaq[edge...].representative_edges, representative_edge)
                else
                    graph_delwaq[edge...] =
                        (type = EdgeType.flow, representative_edges = [representative_edge])
                end
            end
        end
    end

    return graph_delwaq
end
```

For testmodel `basic` this gives
```
julia> graph_delwaq.edge_data
Dict{Tuple{Ribasim.NodeID, Ribasim.NodeID}, @NamedTuple{type::Ribasim.EdgeType.T, representative_edges::Vector{Tuple{Ribasim.NodeID, Ribasim.NodeID}}}} with 9 entries:
  (Basin #6, Basin #9)          => (type = flow, representative_edges = [(Pump #7, Basin #9)])
  (Basin #3, Basin #6)          => (type = flow, representative_edges = [(TabulatedRatingCurve #4, Basin #6)])
  (FlowBoundary #16, Basin #1)  => (type = flow, representative_edges = [(FlowBoundary #16, Basin #1)])
  (Basin #3, Basin #9)          => (type = flow, representative_edges = [(TabulatedRatingCurve #4, Basin #9)])
  (FlowBoundary #15, Basin #6)  => (type = flow, representative_edges = [(FlowBoundary #15, Basin #6)])
  (Basin #1, Basin #3)          => (type = flow, representative_edges = [(ManningResistance #2, Basin #3)])
  (LevelBoundary #11, Basin #3) => (type = flow, representative_edges = [(LinearResistance #12, Basin #3)])
  (Basin #9, LevelBoundary #17) => (type = flow, representative_edges = [(LinearResistance #10, LevelBoundary #17)])
  (Basin #3, Terminal #14)      => (type = flow, representative_edges = [(TabulatedRatingCurve #4, Terminal #14)])
```

and a simple example where there are multiple paths between 2 basins, testmodel `tabulated_rating_curve`:

```
julia> graph_delwaq.edge_data
Dict{Tuple{Ribasim.NodeID, Ribasim.NodeID}, @NamedTuple{type::Ribasim.EdgeType.T, representative_edges::Vector{Tuple{Ribasim.NodeID, Ribasim.NodeID}}}} with 1 entry:
  (Basin #1, Basin #4) => (type = flow, representative_edges = [(TabulatedRatingCurve #3, Basin #4), (TabulatedRatingCurve #2, Basin #4)])

**Comment by evetion** [comment]
I've made a Python translation, which seems to work fine. 👍🏻 FWIW, I use the original edge_id instead of the "representative edge", but the idea is the same. Still needs work on applying the new mapping to the nodes and edges (volumes, flows), and adding some flows if needed.

image

File: coupling/delwaq/README.md [hunk 1]
```diff
@@ -0,0 +1,37 @@
+# Ribasim Delwaq coupling
+
+
+## Steps
+Choose a testmodel and run it. For example, running the basic model with ribasim_cli:
+```bash
+./build/ribasim_cli/bin/ribasim generated_testmodels/basic/ribasim.toml
```
visr, line 7:
```suggestion
ribasim generated_testmodels/basic/ribasim.toml
```

The `./build/ribasim_cli/bin/ribasim` path is outdated.

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
+# Ribasim Delwaq coupling
+
+
+## Steps
+Choose a testmodel and run it. For example, running the basic model with ribasim_cli:
```
visr, line 5:
```suggestion
Choose a model and run it. For example, running the basic model with the Ribasim CLI:
```

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
...
+```bash
+./build/ribasim_cli/bin/ribasim generated_testmodels/basic/ribasim.toml
+```
+
+Run `gen_delwaq.py` to couple the models. Open a terminal in the directory of Ribasim
```
visr, line 10:
```suggestion
Run `generate.py` to generate a Delwaq model. Open a terminal in the Ribasim repository root and run:
```

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
...
+```python
+python ./coupling/delwaq/gen_delwaq.py
+```
+The script can:
+- Open an existing Ribasim model (hardcoded to `generated_testmodels/basic/ribasim.toml` for now)
+- Convert the metadata, topology and substances to Delwaq format and put it into the `model` folder
+
+You now should manually run Delwaq. You can do so with the Docker image.
```
visr, line 18: Good to make it clear from the start that this only works for Deltares employees.

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
...
+docker login containers.deltares.nl  # use your deltares email + token
+```
+
+You can now run the Delwaq model from this directory.
+```bash
+docker run --mount type=bind,source="$(pwd)/model",target=/mnt/myModel \
+  --workdir /mnt/myModel containers.deltares.nl/delft3d/delft3dfm run_dimr.sh
+```
+
+If everything worked out, there's now a new netcdf output. We'll use this to update the Ribasim model.
```
visr, line 32:
```suggestion
If everything worked out, there's now a new netCDF output. We'll use this to update the Ribasim model.
```

coupling/delwaq/generate.py
```diff
@@ -0,0 +1,359 @@
...
+delwaq_dir = Path(__file__).parent
+
+env = Environment(autoescape=True, loader=FileSystemLoader(delwaq_dir / "template"))
+
+# Add evaporation edges, so mass balance is correct
+# To simulate salt increase due to evaporation, set to False
+USE_EVAP = True
+
+
+def generate(modelfn: Path) -> tuple[nx.DiGraph, set[str]]:
```
visr, line 30: For consistency with Ribasim Python it would perhaps be better to use `toml_path` rather than `modelfn` everywhere.

coupling/delwaq/generate.py
```diff
@@ -0,0 +1,359 @@
...
+    basins = pd.read_feather(modelfn.parent / "results" / "basin.arrow")
+    flows = pd.read_feather(modelfn.parent / "results" / "flow.arrow")
+
+    output_folder = delwaq_dir / "model"
+    output_folder.mkdir(exist_ok=True)
+
+    # Simplify network, only keep Basins and Boundaries
+    G = nx.DiGraph()
+    for row in model.node_table().df.itertuples():
+        if "Control" not in row.node_type:
```
visr, line 44: Would using something like this be more robust?

https://github.com/Deltares/Ribasim/blob/333e2acc23a7789b3f35f469ff3920a3a226dddd/python/ribasim/ribasim/geometry/edge.py#L60-L62

That would also include LevelDemand.

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Steps
+Choose a testmodel and run it. For example, running the basic model with ribasim_cli:
+```bash
+./build/ribasim_cli/bin/ribasim generated_testmodels/basic/ribasim.toml
+```
+
+Run `gen_delwaq.py` to couple the models. Open a terminal in the directory of Ribasim
+```python
+python ./coupling/delwaq/gen_delwaq.py
```
visr, line 12:
```suggestion
```bash
python ./coupling/delwaq/generate.py
```

coupling/delwaq/parse.py
```diff
@@ -0,0 +1,51 @@
+"""Read a Delwaq model generated from a Ribasim model and inject it back to Ribasim."""
```
visr, line 1:
```suggestion
"""Read a Delwaq model generated from a Ribasim model and inject the results back to Ribasim."""
```

coupling/delwaq/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+You can now run the Delwaq model from this directory.
+```bash
+docker run --mount type=bind,source="$(pwd)/model",target=/mnt/myModel \
+  --workdir /mnt/myModel containers.deltares.nl/delft3d/delft3dfm run_dimr.sh
+```
+
+If everything worked out, there's now a new netcdf output. We'll use this to update the Ribasim model.
+Run `parse_delwaq.py` again to update the Ribasim model with the new concentrations.
```
visr, line 33:
```suggestion
Run `parse.py` again to update the Ribasim model with the new concentrations.
```

Why again? This looks like it is the only time.

coupling/delwaq/util.py
```diff
@@ -0,0 +1,169 @@
...
+            f.write(struct.pack("<i", int(time)))
+            f.write(group.flow_rate.to_numpy().astype("float32").tobytes())
+
+        # Delwaq needs an extra timestep after the end
+        f.write(struct.pack("<i", int(time + timestep.total_seconds())))
+        f.write(group.flow_rate.to_numpy().astype("float32").tobytes())
+
+
+def ugrid(G):
+    # TODO Deduplicate with ribasim.utils
```
visr, line 102:
```suggestion
    # TODO Deduplicate with ribasim.Model.to_xugrid
```

coupling/delwaq/parse.py
```diff
@@ -0,0 +1,51 @@
...
+        df.node_id += 1
+        df.node_id = df.node_id.map(mapping)
+
+        dfs.append(df)
+
+    df = pd.concat(dfs).reset_index(drop=True)
+    df.sort_values(["time", "node_id"], inplace=True)
+
+    model.basin.concentrationexternal = df
+    df.to_feather(modelfn.parent / "results" / "basinconcentrationexternal.arrow")
```
visr, line 49:
```suggestion
    df.to_feather(modelfn.parent / "results" / "basin-concentration-external.arrow")
```

coupling/delwaq/test.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_offline_delwaq_coupling():
+    repo_dir = delwaq_dir.parents[1]
+    modelfn = repo_dir / "generated_testmodels/basic/ribasim.toml"
+
+    graph, substances = generate(modelfn)
+    run_delwaq()
+    model = parse(modelfn, graph, substances)
+    df = model.basin.concentrationexternal
```
visr, line 17: Shouldn't this be `concentration_external`, like here?

https://github.com/Deltares/Ribasim/pull/1267/files#diff-bdece7de80ef751956236e71b5aed3964311d59479f076c892d6745b0e30266dR298

coupling/delwaq/util.py
```diff
@@ -0,0 +1,169 @@
...
+import pandas as pd
+from ribasim.utils import MissingOptionalModule
+
+try:
+    import xugrid
+except ImportError:
+    xugrid = MissingOptionalModule("xugrid")
+
+
+def strfdelta(tdelta):
```
visr, line 20: Most functions here have type hints on the arguments but not return type annotations. Would be a bit nicer for readability.

pixi.toml
```diff
@@ -173,6 +171,8 @@ tomli-w = "*"
 xmipy = "*"
 xugrid = "*"
 xarray = "*"
+ipykernel = ">=6.29.3,<6.30"
```
visr, line 174: Can be removed?

pixi.toml
```diff
@@ -173,6 +171,8 @@ tomli-w = "*"
...
+networkx = ">=3.3,<4"
```
visr, line 175:
```suggestion
networkx = ">=3.3"
```

I'd say let's leave the upper bound off for now. It's not there for other dependencies, and we have no automated tooling to notify us of new breaking releases like dependabot set up for pixi.toml. And when updating the pixi.lock we rely on our tests.

python/ribasim/ribasim/model.py
```diff
@@ -204,7 +204,7 @@ def _apply_crs_function(self, function_name: str, crs: str) -> None:
 
     def node_table(self) -> NodeTable:
         """Compute the full NodeTable from all node types."""
-        df_chunks = [node.node.df.set_crs(self.crs) for node in self._nodes()]  # type: ignore
+        df_chunks = [node.node.df.set_crs(self.crs) for node in self._nodes()]
```
visr, line 207: CI still seems to disapprove, though I understood locally you saw the opposite? Not sure how to satisfy both.

utils/testmodelrun.jl
```diff
@@ -4,6 +4,9 @@ include("utils.jl")
 
 function main(ARGS)
     toml_paths = get_testmodels()
+    if length(ARGS) > 0
+        toml_paths = filter(x -> basename(dirname(x)) in ARGS, toml_paths)
+    end
```
visr, line 9: Can you add this usage to a docstring?

visr (approved):
Left a bunch of comments, mostly small. If they are addressed to the degree that they are applicable, it is good to merge from my side.

visr:
Quality work!

evetion:
Thanks, took me way longer that I would've liked, but very happy now that I've got it to work, including the CI.

Issue #1103, opened 2024-02-09
Quadratic absolute object function can't be solved by the solver


Jingru923:
**What**
After using equal constraint for flow conservation (PR: https://github.com/Deltares/Ribasim/pull/1083), a bug appeared in `allocation_test.jl`. When using quadratic absolute as object function, minimal subnetwork as test model, the solver give error of 
```
Termination status : OTHER_ERROR
  Message from the solver:
  "There was an error calling optimize!"
```
The reason of the error is unknown. When I print out the optimization problem, I notice the redundant constraints. If I simplify the constraints, I can solve it with help of some online tools. This means that this optimization problem is feasible, but somehow the solver HiGHS can't deal with it.

**How**

Some possible ways to solve this issue and will be :

- [x] Try problem with other solvers
- [x] Try problem with HiGHS but redundant constraints removed
- [ ] Contact developer of HiGHS with problem (make issue)

Jingru923:
The issue #1191 make this one no longer applicable. So this issue is closed.

Issue #708, opened 2023-10-27
Take terminal nodes into account as sinks in allocation


SouthEndMusic:
In allocation networks some water is lost trough flow into terminal nodes, which is currently not taken into account in allocation solves. The flow to the terminal nodes can be used as the sink flow magnitude, but it is not immediately clear to me how this should be taken into account in the allocation problem.

gijsber:
Hi @SouthEndMusic , to 'force' water on a trajectory without being consumed we should introduce some 'instream demand'. From an allocation algorithm perspective could see this as a user that abstracts the instreaf demand and has a 100% returnflow, but we need to discuss with @visr how we envision that being modelled in the ribasim network.

SnippenE:
Discuss with RTC tool team

SouthEndMusic:
For now it is fine to model the terminal node as a user with a single demand with higher priority than all the user demands, and the demand is given by the flow over the edge to the terminal node in the physical layer at the starting time of the allocation optimization.

SouthEndMusic:
Closed because of https://github.com/Deltares/Ribasim/issues/1208#issuecomment-1988404773, @gijsber argued that we do not need the terminal nodes in allocation at all.

PR #1143, opened 2024-02-19
Allocation fixes

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1103.


SouthEndMusic:
Some Rabbit holes I went down:
- I thought maybe the problem was that the coefficients in the quadratic absolute objective function differ too much in magnitude, since there is a coefficient of 1 and also a coefficient of d^2 for every demand d. To test this I removed the d^2 terms (since adding a constant to the objective function should not have any effect) but this didn't solve the problem. If we want to look into rescaling to tackle this issue, we would have to come up with a robust way to do this based on the order of magnitude of demands and flows in a subnetwork.  
- I also looked into using the Ipopt optimizer instead of HiGHS. There were some difficulties with this involving being allowed to modify the constraints and retrieving values of variables after the model became 'dirty' (i.e. modified after optimization). I believe the first problem was solved by called `JuMP.set_optimizer` which initializes the optimizer in a different mode, and the second by storing the latest flow values in `AllocationModel.flows::flows::Dict{Tuple{NodeID, NodeID}, Float64}`. 
- When I got the Ipopt optimizer to work, I noticed some differences in optimization results. Most of them were just small deviations from the HiGHS results, but there were some notable exceptions, see below. Both of these were due to unconstrained variables in optimizations for collecting subnetwork demands, so the fact that these behaved desirably before was just a lucky coincidence (especially after the flow cost was removed). It would be nice if we could formalize or at least test that there are no unconstrained variables in the problem.
  - `Terminal` nodes: Currently these are already part of the allocation networks, but the allocation flow to them was not yet constrained. Adding their demands is part of #708, but I added constraints to the flow to the terminals so that they do not receive unnecessary flow when collecting subnetwork demands.
  - `FractionalFlow` nodes: Unlike the flow conservation constraints, the fractional flow constraints still have '$\le$' constraints instead of equality constraints, as I feel equality would be too restrictive. To still restrict the flow more (as the priority 1 flow towards the fractional flow node in the middle subnetwork in the `main_network_with_subnetwork` testmodels was some arbitrary high number), I made sure that there is also a flow conservation constraint for nodes that have downstream `FractionalFlow` nodes. The result is now though that less demand is collected from this subnetwork.

There's another rabbit hole I came across which I didn't go into yet, which is all the options JuMP has for making an optimization problem suitable to be solved by a specific solver (which was also briefly discussed by Ailbhe during our latest meeting with the RTC-tools team). In general my feeling is that we've come pretty far with trial and error, but maybe organising a meeting with an expert in JuMP would go a long way at this point (if possible).

SouthEndMusic:
I also noticed a silent error where in the new testmodel `target_level_model` the column `allocation_network_id` wasn't properly renamed to `subnetwork_id`, I'll fix it in my upcoming PR.

SouthEndMusic:
This PR can be closed because:
- With the current choice of only the linear absolute objective function we can keep the HiGHS optimizer;
- The problems with the terminal and fractional flow nodes are mentioned in https://github.com/Deltares/Ribasim/issues/1208.

PR #1145, opened 2024-02-19
Add `ribasim_version` metadata to Arrow results

visr:
The Arrow IPC format supports adding metadata both at the table and column level. The table metadata is a `Dict{String, String}`. This adds `ribasim_version` = `2024.2.0` to each output Arrow table. That way we can always check what Ribasim version was used to simulate a particular model, helping reproducability.

We don't test any results with Ribasim Python, but if you want to read the `ribasim_version` with Python, you can use:

```python
from pyarrow import feather
table = feather.read_table("results/flow.arrow")
table.schema.metadata[b"ribasim_version"]
table.schema.metadata
# {b'ribasim_version': b'2024.2.0'}
```

### DIFF
--- a/core/test/io_test.jl
+++ b/core/test/io_test.jl
@@ -110,3 +110,18 @@ end
     # reversed_arrow_table throws an AssertionError
     @test_throws "not sorted as required" Ribasim.sorted_table!(reversed_arrow_table)
 end
+
+@testitem "results" begin
+    using SciMLBase: successful_retcode
+
+    toml_path = normpath(@__DIR__, "../../generated_testmodels/bucket/ribasim.toml")
+    @test ispath(toml_path)
+    config = Ribasim.Config(toml_path)
+    model = Ribasim.run(config)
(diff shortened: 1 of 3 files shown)


core/test/io_test.jl
```diff
@@ -110,3 +110,18 @@ end
...
+    toml_path = normpath(@__DIR__, "../../generated_testmodels/bucket/ribasim.toml")
+    @test ispath(toml_path)
+    config = Ribasim.Config(toml_path)
+    model = Ribasim.run(config)
+    @test successful_retcode(model)
+
+    path = Ribasim.results_path(config, Ribasim.RESULTS_FILENAME.basin)
+    bytes = read(path)
+    tbl = Arrow.Table(bytes)
+    @test Arrow.getmetadata(tbl) === Base.ImmutableDict("ribasim_version" => "2024.2.0")
```
Hofer-Julian, line 126: Maybe extract the version rather than hardcode it.

PR #1147, opened 2024-02-19
Make results.compression a Boolean

visr:
I wanted to add an option for no compression in addition to the current options of zstd and lz4.
No compression is good to have because then that file can be directly memory mapped to the IPC format.

When doing this I wondered if there is a real need to keep lz4 as an option. We looked at write and read speed here: https://github.com/Deltares/Ribasim/issues/419#issuecomment-1648428575.

# Upgrade information

The results.compression key is now a Boolean instead of a string. Because before compression was always on, modelers can simply delete the key and rely on the default of compression being on. Since the compression is rarely set explicitly, not many models will break.

```toml
[results]
compression = "zstd"  # or "lz4"
```

```python
import tomli
import tomli_w


def compression_bool(toml_path):
    """Remove results.compression key if it exists and is a string."""
    with open(toml_path, "rb") as f:
        config = tomli.load(f)

    results = config.get("results")
    if results is not None:
        val = results.pop("compression", None)
        if isinstance(val, str):
            with open(toml_path, "wb") as f:
                tomli_w.dump(config, f)
```


Hofer-Julian (approved):
Makes sense to me

Issue #1122, opened 2024-02-14
Add supply/demands for basins in allocation to `allocation.arrow` output


SouthEndMusic:
**What**
The supply/demand of basins is computed (since https://github.com/Deltares/Ribasim/pull/1082) but never stored. I propose to store this information in the `allocation.arrow` output file. It fits in there quite well, apart from that the column name `user_node_id` must be generalized to something that encompasses all nodes that have a demand.

For testing, expand on  `@testitem "Allocation level control"`. Also add a test for stage 5 of the graph in https://github.com/Deltares/Ribasim/pull/1082#issuecomment-1943469944.

visr:
> It fits in there quite well, apart from that the column name user_node_id must be generalized to something that encompasses all nodes that have a demand.

Sounds like the combination of `node_type` and `node_id` will do, in light of #1113.

PR #1148, opened 2024-02-19
Basin allocation output

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1122.

# Upgrade information

The flow rate in `flow.arrow` is now in a column named `flow_rate` rather than `flow` to be consistent with the input.


core/src/allocation_optim.jl
```diff
@@ -475,7 +459,76 @@ function adjust_basin_capacities!(
...
+                if priority_idx == basin_priority_idx
+                    # Basin demand
+                    _, basin_idx = id_index(basin.node_id, node_id)
+                    demand += basin.demand[basin_idx]
+                end
+                allocated =
+                    JuMP.value(F_basin_in[node_id]) - JuMP.value(F_basin_out[node_id])
+                # TODO: Abstracted for a basin is not so clear, maybe it should be Δstorage/Δt
+                # over the last allocation interval?
+                abstracted = 0.0
```
visr, line 508: I think `Δstorage/Δt` would make sense. Could you create an issue so we can track it?

Also, `abstracted` made a lot of sense when we were only talking about Users. Now we added water level control and soon we should add instream demand, both of which don't **abstract** water. Shall we already rename the column to `realized`? Or maybe @gijsber has other suggestions, perhaps `flow_rate`?

Actually if you don't mind, I realize just now we consistently use `flow_rate` in input tables, but `flow` columns in `flow.arrow` and `allocation_flow.arrow`. Mind updating those as well?

core/src/allocation_optim.jl
```diff
@@ -475,7 +459,76 @@ function adjust_basin_capacities!(
...
+            end
+        end
+
+        if has_demand
+            # Save allocations and demands to record
+            push!(record_demand.time, t)
+            push!(record_demand.subnetwork_id, allocation_network_id)
+            push!(record_demand.node_type, String(Symbol(node_id.type)))
```
visr, line 516:
```suggestion
            push!(record_demand.node_type, string(node_id.type))
```

visr (commented):
Good stuff. Left some comments.

core/src/write.jl
```diff
@@ -205,19 +207,19 @@ function allocation_flow_table(
...
-        record.edge_id,
-        record.from_node_id,
-        record.to_node_id,
-        record.subnetwork_id,
-        record.priority,
-        record.flow,
-        record.collect_demands,
+        record_flow.edge_id,
+        record_flow.from_node_id,
+        record_flow.to_node_id,
```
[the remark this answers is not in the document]
SouthEndMusic: I think you mean `to_node_type` for the second suggestion?

core/src/allocation_optim.jl
SouthEndMusic: Well, if we're breaking stuff anyway, might as well do a lot xd

core/src/write.jl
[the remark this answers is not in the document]
visr: Yes, sorry.

visr (approved):

PR #1149, opened 2024-02-19
Add `node_id` column to `Node` table

Hofer-Julian:
Breaking ⚠️

- Document changes
- Make sure that existing scripts still work
- Adapt tests
- Add update script for
   - Adding `node_id` to `Node` table if it doesn't exist already
   - Renaming `type` to `node_type` in `Node` table

```python
import geopandas as gpd

path = "path/to/database.gpkg"

# Load the GeoPackage as a GeoDataFrame
gdf = gpd.read_file(path, layer="Node")

# Rename the column "type" to "node_type"
gdf.rename(columns={"type": "node_type"}, inplace=True)

# Add a new column "node_id" with the values of the index
gdf["node_id"] = gdf.index

# Save it back to a GeoPackage
gdf.to_file(path, layer="Node", driver="GPKG")
```


docs/core/usage.qmd
```diff
@@ -162,6 +162,7 @@ between nodes, and never have storage of their own.
 column        | type     | restriction
 --------------| -------- | -----------
 fid           | Int      | unique, sorted
+node_id       | Int      | -
 type          | String   | known node type
```
visr, line 166:
```suggestion
node_id       | Int      | sorted
node_type          | String   | known node type
```

The `fid` is needed by GeoPackage, but not really part of our datamodel, right? And GeoPandas adds it for us. So perhaps better to leave out?

Since you added `node_id`, not `id` to the Node table, perhaps we should rename `type` to `node_type` for consistency. It is slightly redundant since the table is also called Node, but it is better to always use the same name. That way it is also easier to join columns between different tables.

python/ribasim/ribasim/geometry/node.py
```diff
@@ -27,11 +29,20 @@ class NodeSchema(pa.SchemaModel):
...
 class Node(SpatialTableModel[NodeSchema]):
     """The Ribasim nodes as Point geometries."""
 
+    # TODO: Remove as soon as add api has been merged
+    @field_validator("df", mode="before")
+    @classmethod
+    def add_node_id_column(cls, df: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+        if "node_id" not in df.columns:
+            df.insert(0, "node_id", df.index)
+        return df
```
visr, line 44: Doesn't this mean it is not breaking? If it is breaking it would be good to have a script to update existing models.

docs/core/usage.qmd
Hofer-Julian: Fine by me!

python/ribasim/ribasim/geometry/node.py
Hofer-Julian: Depending on how so see it.
I would say that it's still a breaking change.
Breaking Ribasim Core API is a breaking change, even though Ribasim Python users are spared from the consequences.

visr (approved):
This is good to merge from my side, if you can still add upgrade information with a script, similar to https://github.com/Deltares/Ribasim/pull/992.

visr:
Yeah, using Ribasim Python if possible or sqlite3 if not. I know it is a bit of a hassle, but otherwise it becomes impossible to debug user models.

Hofer-Julian:
> Yeah, using Ribasim Python if possible or sqlite3 if not. I know it is a bit of a hassle, but otherwise it becomes impossible to debug user models.

I can take a look, but would be curious about the motivation.
I assumed you can always just take an older Ribasim Python version.

visr:
You can take an older Ribasim Python to load the model, but to debug possible issues we need to be able to run the core on the main branch.

Ideally it is always fully reproducible and we can just regenerate models with scripts, but that is often not possible. Only for our test models it is that simple.

I think at some point we should start making an upgrade script that automatically updates models to the current version. For that to work we should start adding `ribasim_version` to the TOML file as well.

Issue #935, opened 2024-01-11
Change flow results from instantaneous to average over past timestep


visr:
**What**
Change flow results from instantaneous to average over past timestep.
Right now the flow results at time $t_i$ represent the instantaneous flow sampled at time $t_i$.
This should be changed to the mean flow from $t_{i-1}$ to $t_i$.

**Why**
To calculate the water balance, sampled flows are not very useful. Mean flows allow us to calculate total volumes over time periods. For coupling with Delwaq our flows need to be mean flows.

**How**
The `flow` we store in our metagraph and use extensively in our solver code need to stay instantaneous. The `saved_flow` for our results need to become mean flows over the period since the last save. To calculate the mean flow, we add a `flow_volume` and `flow_prev` vector in our graph next to the `flow`. At every timestop, we add the contribution of the past timestep to the `flow_volume`. We initially calculate this simply by taking the `@. flow_volume += (flow + flow_prev) * 0.5 * (t - tprev)`, after which we `copyto!(flow_prev, flow)`.

Then, at each `saveat`, we calculate the mean flow by dividing `flow_volume` by the time since the last saveat. After this we `fill!(flow_volume, 0.0)`.

This should be plenty for one PR. Follow ups:

- In `BMI.update_until`, check that the `saveat` is equal to the time since last BMI stop. This way we can use the same memory for both saving and BMI, at the cost of less flexibility.
- Add mean flow columns to the non-conservative node type structs
- Implement a more precise interpolation using Gaussian quadrature like described in #642

**Additional context**
I can't think of good reasons to keep the sampled flow output as well, so let's just switch the existing results.

evetion:
Note that Delwaq needs *both* the instantaneous flow, as the averaged flow over the past timestep.

Huite:
Are you sure? Instantaneous flow might literally be a flow that persists for a single second (or even a millisecond!) in the adapative timestepping.

evetion:
Ah, sorry, mixed up my notes, we need the instantaneous volume, and indeed, averaged flow.

PR #1159, opened 2024-02-21
Mean output flows

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/935.


SouthEndMusic:
To do:
- incorporate https://github.com/Deltares/Ribasim/pull/1155
- At test of mean flow output

core/src/parameter.jl
```diff
@@ -558,8 +558,13 @@ struct Parameters{T, C1, C2}
             edges_source::Dict{Int, Set{EdgeMetadata}},
             flow_dict::Dict{Tuple{NodeID, NodeID}, Int},
             flow::T,
+            flow_prev::Vector{Float64},
+            flow_integrated::Vector{Float64},
             flow_vertical_dict::Dict{NodeID, Int},
             flow_vertical::T,
+            flow_vertical_prev::Vector{Float64},
+            flow_vertical_integrated::Vector{Float64},
```
visr, line 566: I think it's fine to leave vertical flows out of this PR. I think we should remove them  altogether. For instance for a Basin, we'll need to integrate the separate vertical fluxes and put those in the Basin struct. Or do you rely on them in allocation?

SouthEndMusic: I think it's equally fine if allocation obtains the mean vertical fluxes from the basin struct.

visr: Ok yeah, I realize it wouldn't make sense to have both instantaneous and mean flows in flows.arrow, but I figured it would be ok if we follow up quickly. I'm fine either way!

core/src/callback.jl
```diff
@@ -379,10 +405,17 @@ end
 
 "Copy the current flow to the SavedValues"
 function save_flow(u, t, integrator)
-    vcat(
-        get_tmp(integrator.p.graph[].flow_vertical, 0.0),
-        get_tmp(integrator.p.graph[].flow, 0.0),
-    )
+    (; graph) = integrator.p
+    (; flow_integrated, flow_vertical_integrated, tprev_flow_save) = graph[]
+    Δt = t - tprev_flow_save[1]
```
SouthEndMusic, line 410: I don't like the length 1 vector `tprev_flow_save` very much, but we still need something mutable like this because if all timesteps are saved the interval between saves is not constant. Or is there a way to obtain this time interval in a different way?

core/test/time_test.jl
```diff
@@ -29,5 +29,5 @@
     t = Ribasim.seconds_since.(flow_1_to_2.time, model.config.starttime)
     flow_expected = @. 1 + sin(0.5 * π * (t - t[1]) / (t[end] - t[1]))^2
     # some difference is expected since the modeled flow is for the period up to t
-    @test isapprox(flow_added_1, flow_expected, rtol = 0.005)
+    @test isapprox(flow_added_1, flow_expected, rtol = 0.01)
```
SouthEndMusic, line 32: Since the saved flows are no longer instantaneous, they should not be used for comparing against exact solutions any further

core/src/parameter.jl
```diff
@@ -558,8 +558,13 @@ struct Parameters{T, C1, C2}
             edges_source::Dict{Int, Set{EdgeMetadata}},
             flow_dict::Dict{Tuple{NodeID, NodeID}, Int},
             flow::T,
+            flow_prev::Vector{Float64},
+            flow_integrated::Vector{Float64},
```
SouthEndMusic, line 562: I don't think the integrated flows have to be differentiated

core/test/time_test.jl
visr: Yeah that's a bit unfortunate. I guess as long as `saveat` is short enough and the tolerance large enough we can keep using it?

core/src/callback.jl
visr: Perhaps if we save all timesteps we use `integrator.tprev`, and otherwise the dt is constant.

core/test/time_test.jl
SouthEndMusic: Turns out we can, I wrote this before I noticed that initially my averaged flows were all wrong (:

SouthEndMusic:
To do:
- [ ] Remove horizontal flows from flow output
- [ ] Update docstrings and documentation

core/src/callback.jl
```diff
@@ -379,10 +425,26 @@ end
 
 "Copy the current flow to the SavedValues"
 function save_flow(u, t, integrator)
-    vcat(
-        get_tmp(integrator.p.graph[].flow_vertical, 0.0),
-        get_tmp(integrator.p.graph[].flow, 0.0),
-    )
+    (; tprev, p) = integrator
```
visr, line 428: There is actually also `dt` in `integrator`, does that help the edge cases a bit?

core/test/run_models_test.jl
```diff
@@ -476,3 +476,28 @@ end
...
+        model = Ribasim.run(config)
+        df = DataFrame(Ribasim.flow_table(model))
+        return filter(
+            [:from_node_id, :to_node_id] => (from, to) -> from === 3 && to === 2,
+            df,
+        ).flow_rate
+    end
+    @test all(get_flow(true, 86400.0) .≈ 1.0)
+    @test all(get_flow(true, 0.0) .≈ 1.0)
+    @test all(get_flow(true, Inf) .≈ 1.0)
```
visr, line 499: Would be good to also test the length of `tsaves` here, to make sure they are in fact different. And also test `saveat` values that don't fit n times in the tspan, to trigger the remainder path.

core/src/callback.jl
```diff
@@ -377,12 +422,34 @@ function set_control_params!(p::Parameters, node_id::NodeID, control_state::Stri
...
+            iszero(rem) ? saveat : rem
+        end
+    end
+
+    mean_flow_vertical = flow_vertical_integrated / Δt
+    mean_flow = flow_integrated / Δt
+
+    fill!(flow_vertical_integrated, 0.0)
+    fill!(flow_integrated, 0.0)
+    return vcat(mean_flow_vertical, mean_flow)
```
visr, line 452: Can you vcat the integrated flows first and then divide inplace? That way we are sure to not have more then one array allocation here.

core/src/callback.jl
```diff
@@ -87,6 +92,46 @@ function create_callbacks(
...
+            flow_vertical
+        else
+            0.5 * (flow_vertical + flow_vertical_prev)
+        end
+
+    @. flow_integrated += flow_effective * dt
+    @. flow_vertical_integrated += flow_vertical_effective * dt
+
+    copyto!(flow_prev, flow)
+    copyto!(flow_vertical_prev, flow_vertical)
```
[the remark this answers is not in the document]
SouthEndMusic: Nice suggestion, the result also looks cleaner

core/src/config.jl
```diff
@@ -233,19 +233,26 @@ end
...
 function convert_saveat(saveat::Float64, t_end::Float64)::Union{Float64, Vector{Float64}}
+    errors = false
     if iszero(saveat)
         # every step
-        Float64[]
+        saveat = Float64[]
     elseif saveat == Inf
         # only the start and end
-        [0.0, t_end]
+        saveat = [0.0, t_end]
```
visr, line 242: Would it not be simpler to return `t_end` (or perhaps `[t_end]`) here instead of filtering the 0.0 out for the SavingCallback?

core/test/run_models_test.jl
```diff
@@ -476,3 +476,70 @@ end
...
+    @test ispath(toml_path)
+    function get_flow(solver_dt::Union{Float64, Nothing}, solver_saveat::Float64)
+        config = Ribasim.Config(toml_path; solver_dt, solver_saveat)
+        model = Ribasim.run(config)
+        df = DataFrame(Ribasim.flow_table(model))
+        return filter(
+            [:from_node_id, :to_node_id] => (from, to) -> from === 3 && to === 2,
+            df,
+        ).flow_rate,
+        Ribasim.tsaves(model)
```
visr, line 494:
```suggestion
        flow = filter(
            [:from_node_id, :to_node_id] => (from, to) -> from === 3 && to === 2,
            df,
        ).flow_rate,
        return flow, Ribasim.tsaves(model)
```

visr (approved):

core/src/config.jl
SouthEndMusic: I don't know whether that's desirable since `saveat` is also used for the save subgrid level callback.

visr: Yeah I was thinking for that we can use save start = true

PR #1170, opened 2024-02-22
Remove `adaptive` solver config option

visr:
Instead, the presence of `dt` will determine if it is adaptive or not.

The only feature we lose with this is that we cannot set the initial timestep for an adaptive run, which to my knowledge wasn't used.

# Upgrade information

In the TOML, remove the `adaptive` key in the `[solver]` section. Provide a `dt` key only if you want a fixed timestep.


core/test/equations_test.jl
```diff
@@ -168,6 +168,7 @@ end
 @testitem "MiscellaneousNodes" begin
     using PreallocationTools: get_tmp
     using SciMLBase: successful_retcode
+    using Ribasim: timesteps, get_storages_and_levels
```
Jingru923, line 171:
```suggestion
    using Ribasim: timesteps, get_storages_and_levels
```

Jingru923 (approved):
Get `dt` from config instead of typing it out

PR #1175, opened 2024-02-23
Rename User to UserDemand

visr:
This was a pain to rename. I manually edited quite a bit.

I prefer to always refer to the node types by their proper PascalCase name in comments and docs, so Basin and not basin.

A part of the allocation docs were also outdated, referring to only User nodes, whereas it should refer to all Demand nodes.

Rename suggested by @gijsber and on top of https://github.com/Deltares/Ribasim/pull/1172.

PR is ready for review, I'll still add upgrade instructions.


Hofer-Julian (approved):
LGTM (after the merge conflicts are resolved)

Issue #319, opened 2023-06-15
Investigate Ribasim binary size


Hofer-Julian:
At the moment, the unzipped libribasim binaries take ~800 MB.
For comparison, libmf6 takes ~30 MB.

Two links by @visr that might be useful:

- https://discourse.julialang.org/t/packagecompiler-reduce-system-image/97952
- https://julialang.github.io/PackageCompiler.jl/dev/apps.html#Filtering-stdlibs

Hofer-Julian:
At last JuliaCon Local, we've got a couple of suggestions in order to reduce that:

- filter_stdlibs
- strip binary (with common tools)
- strip metadata (not advisable, since that would lead to less readable stack-traces according to @visr)
- strip-ir

PR #1176, opened 2024-02-24
Reduce binary size by leaving out artifacts

### LINKED ISSUE #1153: installing ribasim (long paths)
installing ribasim gives this complain:

image

not everybody can install in the root, maybe something can be done about this...?
regards!

visr:
Takes a good chunk out of #319.
Fixes #1153.
Fixes https://github.com/Deltares/Ribasim/pull/687#issuecomment-1772635668

Our `ribasim_cli_windows.zip` is 535 MB on main, this reduces it to 227 MB.
Unpacked, it goes from 1.9 GB to 1.0 GB.
Libribasim shrinks by the same amount.

This is because we no longer include two 400 MB artifacts, MinGW and MKL. Three PackageCompiler keyword arguments come into play here:

- set to true: `include_preferences::Bool`: If true, store all preferences visible by the project in package_dir in the app bundle. Defaults to true.
- set to false: `include_lazy_artifacts::Bool`: if lazy artifacts should be included in the bundled artifacts, defaults to false.
- set to false: `include_transitive_dependencies::Bool`: If true, explicitly put all transitive dependencies into the sysimage. This only makes a difference if some packages do not load all their dependencies when themselves are loaded. Defaults to true.

I added an explicit `include_preferences=true`. This is mostly to highlight that we are making use of Preferences to tell LinearSolve that we don't want to load MKL_jll. This would be enough outside of PackageCompiler, however more was needed. The `include_lazy_artifacts=false` is restoring the default behavior. We had set it to true to stop it from downloading MKL_jll the first time the CLI was used. When I added this option, it would still do that, so more was needed. The `include_transitive_dependencies=false` is deviating from the default, but this causes several issues. I'm not sure, but for MKL_jll perhaps the problem was that the code was already fixed to the default load MKL_jll behavior in the sysimage.

One more note about choosing to opt out of using MKL_jll, [by default we use KLUFactorization](https://github.com/SciML/LinearSolve.jl/blob/v2.24.0/src/default.jl#L114), since in `[solver]` we set `sparse = true` by default. It could be that `sparse = false` will be a bit slower. I haven't taken the time to measure, but I doubt it is worth the binary size, since `sparse = false` only makes sense for small models.

A nice side effect is that this combination of keys also took MinGW out of the sysimage. The fact that we got it might be a PackageCompiler bug, though I could not repoduce this in small test executables.

Some references for those interested:
https://julialang.github.io/PackageCompiler.jl/stable/refs.html#PackageCompiler.create_app
https://juliapackaging.github.io/Preferences.jl/dev/
https://github.com/SciML/LinearSolve.jl/pull/425/files

I snuck in some extra preferences. When precompiling, OrdinaryDiffEq always takes very long since it precompiles a lot of combinations of solvers and specializations. I had a look at the docs and precompilation code linked below, and concluded that we probably don't gain too much from these precompile workloads. Our default solver, QNDF, is not precompiled, and for snappy CLI startup we have our own precompilation script that runs Ribasim, which is probably more effective anyway. That means developers using Julia have to spend less time waiting for OrdinaryDiffEq to precompile, possibly at the cost of waiting slightly longer to compile a first Ribasim run from Julia. I haven't measured. Another benefit is that we will create less code; the `lib/julia/sys.dll` shrinks by 50 MB to 672 MB due to less unused OrdinaryDiffEq code.

https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep/
https://github.com/SciML/OrdinaryDiffEq.jl/blob/v6.71.0/src/OrdinaryDiffEq.jl#L236-L334
https://github.com/trixi-framework/libtrixi/blob/v0.1.4/LibTrixi.jl/Project.toml

The builds from this branch do run the test models just fine, don't download anything on start, and start more quickly.


Hofer-Julian (approved):
I can confirm that the binary size is reduced significantly, the tests are green as well.

Great work, Martijn!

gijsber:
nice work !

PR #1177, opened 2024-02-26
Error quadratic absolute

### LINKED ISSUE #954: Use equality constraints for flow conservation

### LINKED ISSUE #1103: Quadratic absolute object function can't be solved by the solver

Jingru923:
Fixes #1103 

As some updated happened in other branch, this error no longer exists. So I made this pull request to recover the skipped test for quadratic absolute optimization.


visr:
Closing since we'll only support linear_absolute going forward.

PR #1180, opened 2024-02-26
Update codegen pixi task

Hofer-Julian:
- Running codegen again leads to a different result
- Linting is not needed
- Initializing Julia is needed

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -93,7 +93,7 @@ generate-testmodels = "python utils/generate-testmodels.py"
 tests = { depends_on = ["lint", "test-ribasim-python", "test-ribasim-core"] }
 # Codegen
 generate-python = { cmd = "julia --project utils/gen_python.jl && ruff format python/ribasim/ribasim/schemas.py"}
-codegen = { depends_on = ["generate-python", "lint"] }
+codegen = { depends_on = ["initialize-julia", "generate-python"] }
 # Publish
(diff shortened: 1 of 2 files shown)


visr (approved):
Thanks. Yes I didn't run codegen after renaming nodes last week.

pixi.toml
```diff
@@ -93,7 +93,7 @@ generate-testmodels = "python utils/generate-testmodels.py"
 tests = { depends_on = ["lint", "test-ribasim-python", "test-ribasim-core"] }
 # Codegen
 generate-python = { cmd = "julia --project utils/gen_python.jl && ruff format python/ribasim/ribasim/schemas.py"}
-codegen = { depends_on = ["generate-python", "lint"] }
+codegen = { depends_on = ["initialize-julia", "generate-python"] }
```
deltamarnix, line 96: Linting was needed, because it solves the whitespacing issues. pre-commit didn't let me pass it.

Hofer-Julian: Mmmh good to keep in mind.

Pre-commit runs anyway before, well, a commit, so I think we should still be fine :)

deltamarnix: Correct, but that would mean you can never commit at once.

Hofer-Julian: > Correct, but that would mean you can never commit at once.

That would be annoying. I can at least tell that I was able to commit at once.
If that was an outlier and most of the time pre-commit complains, then let's add the task again.

PR #1181, opened 2024-02-26
Write `ribasim_version` to TOML and warn in core if it is different

### LINKED ISSUE #1178: Add `ribasim_version` to TOML
**Why**
Ribasim models are created for a specific Ribasim version. If you receive a model, you don't know what Ribasim version it targets. If you have results then you could take it out of the Arrow metadata using https://github.com/Deltares/Ribasim/pull/1145, but if you only have input there is no way of knowing this.

In the future, we could possibly start using this information to help migrating across breaking changes, but implementing that is out of scope for this issue. Similarly, we could start throwing warnings or even errors in the core if the version of the CLI does not match this `ribasim_version`.

**How**
If a model is written with Ribasim Python we can easily add this information automatically. We could do this as a top level key `ribasim_version = "2024.2.0"`. One open question is whether to make this a required key, making it a breaking change. I'm leaning toward yes, because not knowing this information makes hard to give clear error messages in some cases.

visr:
Fixes #1178

At first I tried to add `ribasim_version` to the Python data model, but struggled to get it to write to the TOML even if it had its default values. But then I thought, when we `model.write`, with Ribasim Python version X, we should always set that key to X. Therefore there is no point in having it be part of the data model.

This does some hardcoded current version numbers in our test TOML files. I guess we could script a way around that, but updating the version number is a documented find and replace action already anyway, so it doesn't matter much.

The warning from the core looks like this:

```
┌ Warning: The Ribasim version in the TOML config file does not match the used Ribasim CLI version.
│   config.ribasim_version = "2024.2.1"
│   ribasim_version = "2024.2.0"
└ @ Ribasim d:\Ribasim\core\src\main.jl:63
```

### DIFF
--- a/core/src/main.jl
+++ b/core/src/main.jl
@@ -57,19 +57,19 @@ function main(ARGS::Vector{String})::Cint
             logger =
                 Ribasim.setup_logger(; verbosity = config.logging.verbosity, stream = io)
             with_logger(logger) do
-                ribasim_version = string(pkgversion(Ribasim))
+                cli = (; ribasim_version = string(pkgversion(Ribasim)))
                 (; starttime, endtime) = config
-                if string(ribasim_version) != config.ribasim_version
-                    @warn "The Ribasim version in the TOML config file does not match the used Ribasim CLI version." config.ribasim_version ribasim_version
+                if config.ribasim_version != cli.ribasim_version
+                    @warn "The Ribasim version in the TOML config file does not match the used Ribasim CLI version." config.ribasim_version cli.ribasim_version
                 end
-                @info "Starting a Ribasim simulation." ribasim_version starttime endtime
+                @info "Starting a Ribasim simulation." cli.ribasim_version starttime endtime
                 model = Ribasim.run(config)
                 if successful_retcode(model)
                     @info "The model finished successfully"
                     return 0
                 end
 
-                t = Ribasim.datetime_since(model.integrator.t, model.config.starttime)
+                t = Ribasim.datetime_since(model.integrator.t, starttime)
                 retcode = model.integrator.sol.retcode
                 @error "The model exited at model time $t with return code $retcode.\nSee https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#retcodes"
                 return 1
--- a/core/src/main.jl
+++ b/core/src/main.jl
@@ -57,8 +57,11 @@ function main(ARGS::Vector{String})::Cint
             logger =
                 Ribasim.setup_logger(; verbosity = config.logging.verbosity, stream = io)
             with_logger(logger) do
-                ribasim_version = pkgversion(Ribasim)
+                ribasim_version = string(pkgversion(Ribasim))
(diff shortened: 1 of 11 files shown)


core/src/main.jl
```diff
@@ -57,8 +57,11 @@ function main(ARGS::Vector{String})::Cint
             logger =
                 Ribasim.setup_logger(; verbosity = config.logging.verbosity, stream = io)
             with_logger(logger) do
-                ribasim_version = pkgversion(Ribasim)
+                ribasim_version = string(pkgversion(Ribasim))
                 (; starttime, endtime) = config
+                if string(ribasim_version) != config.ribasim_version
```
Hofer-Julian, line 62: Why do you convert to a string twice?

core/src/main.jl
```diff
@@ -57,8 +57,11 @@ function main(ARGS::Vector{String})::Cint
...
+                    @warn "The Ribasim version in the TOML config file does not match the used Ribasim CLI version." config.ribasim_version ribasim_version
+                end
```
Hofer-Julian, line 64: Maybe make it a bit easier to see which one is the version in the config and which one in the binary?

docs/contribute/release.qmd
```diff
@@ -25,13 +25,8 @@ Determine the new version number like `2023.1.0`, filling in the current year, a
...
 Update the version numbers in the repository to the new version number.
-A single find and replace can update all 5 locations:
-
-- `pixi.toml`
-- `core/Project.toml`
-- `python/ribasim/ribasim/__init__.py`
-- `python/ribasim_api/ribasim_api/__init__.py`
-- `ribasim_qgis/metadata.txt`
+A single find and replace can update all locations.
+The Ribasim version number is not in `pixi.lock`, that file does not need updating.
```
Hofer-Julian, line 29: Why mention that specifically?

Hofer-Julian (commented):
A few small comments, apart from that ready to go from my side

core/src/main.jl
visr: Is this better? I added the `cli.` In general I like this type of logging that splits the values from the message.

```
┌ Warning: The Ribasim version in the TOML config file does not match the used Ribasim CLI version.
│   config.ribasim_version = "a_different_version"
│   cli.ribasim_version = "2024.2.0"
└ @ Ribasim d:\Ribasim\core\src\main.jl:63
```

docs/contribute/release.qmd
visr: Because the string 2024.2.1 occurs there as well as e.g. the current dask-core version we use. We don't want to edit that.

core/src/main.jl
visr: Mistake, fixed, thanks.

docs/contribute/release.qmd
Hofer-Julian: Oh no, that is indeed really easy to mess up by accident.

Hofer-Julian (approved):

PR #1186, opened 2024-02-27
Don't auto add `meta_` prefix, error.

### LINKED ISSUE #1156: Error on unknown columns that don't start with `meta_`
**What**
DataFrames with unknown columns are automatically prefixed with `meta_`. We should error instead. Unknown columns that start with `meta_` can stay.

**Why**
If a user mistypes an optional column name, it gets turned into a meta_mistake; a wrong model. Also the automatic prefixing might be a bit too magical and confuse users. There is no problem with requiring users to opt in to this by prefixing the columns they want to add as metadata themselves.

visr:
Fixes #1156

### DIFF
--- a/python/ribasim/ribasim/__init__.py
+++ b/python/ribasim/ribasim/__init__.py
@@ -51,6 +51,5 @@ __all__ = [
     "TabulatedRatingCurve",
     "Terminal",
     "UserDemand",
-    "utils",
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
CI is still red

Hofer-Julian (approved):

PR #1187, opened 2024-02-27
Fix broken hyperlinks

visr:
With `quarto run docs` I couldn't reproduce #1138, but I'm fairly sure this will fix #1138 since these are the only links that have backslashes. 

With backslashes apparently these links keep their .qmd extenstion rather than .html: 

https://deltares.github.io/Ribasim/core/usage.qmd#sec-tabulated-rating-curve


Hofer-Julian (approved):
After merging this, let's make sure this actually also works on Linux

visr:
Links are working.

Issue #78, opened 2023-03-01
Add `FlowDemand` node for instream demand


gijsber:
This node connects to a flow node which is part of an allocation edge, and prescribes a flow demand there. This flow node is then treated as a user with 100% returnflow. 

What came out of discussion with the RTC-Tools team:
- return flow from `UserDemand` and `FlowDemand` should act as a source for later priorities. This means that a source must be introduced for every return flow to which capacity is added when the node has an inflow.

Note that this also changes the behavior of `UserDemand` nodes, and makes https://github.com/Deltares/Ribasim/issues/899 obsolete.

SnippenE:
Discuss how we can take a minimum flow into account in the allocation optimisation step (incl the model specification)

gijsber:
For allocation:
introduce node TargetFlow which is associated with a connector node.
Specify a miniflow_flow_rate and priority.
Include in allocation problem as a flow_rate to preserve

SouthEndMusic:
It is relatively easy to do a naïve implementation of this, that is: add a term to the objective function for the flow trough the specified node. However:
- The way the allocation optimization problems are set up now, the optimizer would only be able to satisfy (part of) the flow demand if the flow can end up somewhere afterwards. Currently that is only at user nodes (and in the future that could also be at terminal nodes), but that would only happen if these users have sufficient demand.
-  Given the problem above, would it be useful to let the flow demand persist in priorities later than its own? Especially given that unlike users, the flow demand does not use up the flow.

gijsber:
Hi @SouthEndMusic,
Than that Terminal node implementation becomes more important
I think there is no objection when we have a FlowDemand with priority 2 that the same FlowDemand is repeated at priority 3

PR #1188, opened 2024-02-28
Add FlowDemand node type

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/78.


SouthEndMusic:
Could there be conflicting/changing code formatting? Since my latest commit contains formatting changes to lines of code I didn't touch.

SouthEndMusic:
# Concept of flow demand within an optimization problem

1. Flow conservation constraint of a node with a flow demand at priority other than the flow demand priority ($p \ne p_\text{fd}$):

$$
F_\text{flow out} = F_\text{flow in} + F_\text{buffer out}
$$

where the value of $F_\text{buffer out}$ is bounded from above by the capacity of the buffer, which is set to $0.0$ before the optimization for the first priority.

2. Flow conservation constraint of a node with a flow demand at the flow demand priority ($p = p_\text{fd}$):

$$
F_\text{buffer in} = F_\text{flow in}
$$

where $F_\text{buffer in}$ is set as the capacity of the buffer after optimization for this priority.  $|F_\text{buffer in} - d|$ is added as a term to the objective function. Here $d = 0$ when $p \ne p_\text{fd}$ and when $d=d_\text{fd}$ then $d$ is what is left of the flow demand after subtracting the flows trough the node with the flow demand at at all $d < d_\text{fd}$.
 
3. The above can be combined into:

$$
F_\text{flow out} + F_\text{buffer in}  = F_\text{flow in} + F_\text{buffer out}
$$

This constraint can exist unmodified as a combination of 1. and 2. because:

- For $p \ne p_\text{fd}$ we have that $F_\text{buffer in} = 0$ because $d = 0$ in the objective function;
- For $p = p_\text{fd}$ we have that $F_\text{buffer out} = 0$ because the buffer capacity at that point is still $0.0$ and $F_\text{flow out} = 0$ because we can constrain it that way.

It must be noted that the choice to do it in this way makes it potentially hard in the future to allow flow demands for multiple priorities for the same node.

# Concept of the user demand within an optimization problem

This is slightly different from the flow demand, because the user does not act as a 'conductor' of flow in the absence of demand. Therefore we can just add a buffer capacity constraint on the user outflow edge variable and add to this buffer the return fraction of the abstraction of the user.

Note that with this concept we are not all the way there when it comes to avoiding circular flows:
- We _**do**_ prevent that the optimizer makes up flow out of nothing by letting a user use its own return flow from the same priority, so the water balance aspect of the problem is solved;
- We _**do not**_ prevent that a user uses its own return flow at all, since the same user can still use the return flow from one priority as abstraction in a later priority.

# To do
- [ ] Chop the above up in small to do points
- [ ] Add docstrings to complex functions
- [ ] Update allocation documentation

core/src/allocation_init.jl
```diff
@@ -579,15 +622,67 @@ function add_constraints_flow_conservation!(
...
+            F[(node_id, outneighbor_id)] for
+            outneighbor_id in outflow_ids_allocation(graph, node_id)
+        ]) == sum([
+            F[(inneighbor_id, node_id)] for
+            inneighbor_id in inflow_ids_allocation(graph, node_id)
+        ]),
+        base_name = "flow_conservation_inlet"
+    )
+
+    # Nodes with flow demand
```
Huite, line 646: Similarly here, move it into a separate function.

core/src/allocation_init.jl
```diff
@@ -579,15 +622,67 @@ function add_constraints_flow_conservation!(
...
+        problem,
+        [node_id = node_ids_flow_demand],
+        F[(node_id, only(outflow_ids_allocation(graph, node_id)))] +
+        F_flow_buffer_in[node_id] ==
+        F[(only(inflow_ids_allocation(graph, node_id)), node_id)] +
+        F_flow_buffer_out[node_id],
+        base_name = "flow_conservation_flow_demand"
+    )
+
+    # Subnetwork inlets from main network
```
Huite, line 663: And also for this one.

Huite (commented):
Just drive-by review: would be good to break these into shorter functions.

core/src/util.jl
```diff
@@ -692,3 +706,28 @@ function set_is_pid_controlled!(p::Parameters)::Nothing
...
+    control_inneighbors = inneighbor_labels_type(graph, node_id, EdgeType.control)
+    for id in control_inneighbors
+        if graph[id].type == node_type
+            return true, id
+        end
+    end
+    return false, nothing
+end
+
+function Base.get(
```
visr, line 724: Did you intend to add a method to `getindex` instead so you can do `A[i]`?

core/src/config.jl
```diff
@@ -113,7 +113,7 @@ end
 end
 
 @option struct Allocation <: TableOption
-    timestep::Union{Float64, Nothing} = nothing
+    timestep::Union{Float64, Nothing} = 86400
```
visr, line 116: This change also needs to be made in config.py.

core/src/util.jl
SouthEndMusic: Looks like I don't use this at all anymore

python/ribasim/ribasim/config.py
```diff
@@ -210,7 +212,7 @@ class UserDemand(MultiNodeModel):
 class LevelDemand(MultiNodeModel):
     static: TableModel[LevelDemandStaticSchema] = Field(
         default_factory=TableModel[LevelDemandStaticSchema],
-        json_schema_extra={"sort_keys": ["node_id", "priority"]},
+        json_schema_extra={"sort_keys": ["node_id"]},
```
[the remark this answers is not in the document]
SouthEndMusic: I think we've been through this before :p Only `UserDemand` nodes can have multiple priorities

core/src/util.jl
SouthEndMusic: Ah no I use it, but the use case is that I want to be able to get a constraint by node ID or get `nothing` if the constraint does not exist.

core/src/allocation_init.jl
SouthEndMusic: Looks like this was a duplicate of the first codeblock, can be removed altogether

docs/core/allocation.qmd
```diff
@@ -84,8 +96,8 @@ A new graph is created from the subnetwork, which we call an allocation network.
 
 The allocation network consists of:
 
-- Nodes $V'_S \subset V_S$, where each Basin, source and demand in the subnetwork get a node in the allocation network.
-Also nodes that have FractionalFlow outneighbors get a node in the allocation network.
+- Nodes $V'_S \subset V_S$. Nodes that are represented in the allocation network are:
+nodes of type `Basin`, `UserDemand` and `Terminal`, nodes that have fractinal flow downstream neighbors, nodes that have a flow demand and nodes that are connected to a source edge.
```
visr, line 100:
```suggestion
nodes of type `Basin`, `UserDemand` and `Terminal`, nodes that have FractionalFlow downstream neighbors, nodes that have a flow demand and nodes that are connected to a source edge.
```

core/src/allocation_optim.jl
```diff
@@ -408,10 +556,12 @@ function save_demands_and_allocations!(
             user_demand_idx = findsorted(user_demand.node_id, node_id)
             demand = user_demand.demand[user_demand_idx]
             allocated = user_demand.allocated[user_demand_idx][priority_idx]
+            #NOTE: istantaneous
```
visr, line 559:
```suggestion
            #NOTE: instantaneous
```

visr (approved):

Issue #1098, opened 2024-02-08
Add `user.realized` and expose via BMI


SouthEndMusic:
**What**
https://github.com/Deltares/Ribasim/pull/1081 exposes a flat user demand vector via the BMI. As mentioned in https://github.com/Deltares/Ribasim/issues/893, a realized average abstraction from the users is also needed in the BMI for coupling.

**How**
Add a field `user.realized::Vector{Float64}` with the realized flow per user over the required timestep (?). To keep it general, return flow should be taken into account, so that the realized flow should be the average of the difference between the abstraction and the return flow.

visr:
We should be able to use the same method as the one required for #935.

Issue #1189, opened 2024-02-28
Lock BMI to saveat period


visr:
In `BMI.update_until`, check that the `saveat` is equal to the time since last BMI stop. This way we can use the same memory for both saving and BMI, at the cost of less flexibility.

One of the follow ups of #935.

EDIT: ln #1190 we are letting this requirement go in favor of separate vectors for BMI

PR #1190, opened 2024-02-28
add realized user demand and expose via BMI

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1098.

@visr @HendrikKok would it be OK if `realized` is not an average but a cumulative volume and resetting it to `0.0` is the responsibility of the coupler? Otherwise we need a discussion about where in the Ribasim code these things must happen in such a way that dividing by the time interval happens before `realized` is accessed and resetting afterwards. 

Also good to know that `demand` is per user per priority, but `realized` is only per user.

**For BMI variable name upgrade instructions see:**

https://github.com/Deltares/Ribasim/pull/1190/files#diff-88d699c34d26c16edb268bdacbda2afb8dad704b8db2931a83e04b9ca52dc31a.


visr:
> would it be OK if realized is not an average but a cumulative volume and resetting it to 0.0 is the responsibility of the coupler?

I think that would be ok. But realized should (in a separate PR) also be added to the results. That can only work if we save two copies of the vector in memories, one for BMI and one for results, to prevent setting it to zero every `saveat`. That is why we have #1189. With that one vector is enough, and we can directly serve averages over BMI.

core/src/callback.jl
```diff
@@ -123,6 +124,14 @@ function integrate_flows!(u, t, integrator)::Nothing
     @. flow_integrated += 0.5 * (flow + flow_prev) * dt
     @. flow_vertical_integrated += 0.5 * (flow_vertical + flow_vertical_prev) * dt
 
+    for user_demand_idx in eachindex(user_demand.node_id)
+        id = user_demand.node_id[user_demand_idx]
+        src_id = inflow_id(graph, id)
+        flow_idx = flow_dict[src_id, id]
+        user_demand.realized[user_demand_idx] +=
+            0.5 * (flow[flow_idx] + flow_prev[flow_idx]) * dt
+    end
```
visr, line 133:
```suggestion
    for (i, id) in enumerate(user_demand.node_id)
        src_id = inflow_id(graph, id)
        flow_idx = flow_dict[src_id, id]
        user_demand.realized[i] += 0.5 * (flow[flow_idx] + flow_prev[flow_idx]) * dt
    end
```

core/src/callback.jl
```diff
@@ -489,7 +498,7 @@ end
 "Solve the allocation problem for all demands and assign allocated abstractions."
 function update_allocation!(integrator)::Nothing
     (; p, t, u) = integrator
-    (; allocation) = p
+    (; allocation, user_demand) = p
```
visr, line 501:
```suggestion
    (; allocation) = p
```

visr (approved):

PR #1193, opened 2024-02-29
Disable the alternative object functions

Jingru923:
Now if user specify an object function that is NOT linear absolute, they will get an error "Type of object function is not supported". Relevant unit tests are modified. But the codes of alternative object function still exist.

### DIFF
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -856,6 +856,9 @@ function AllocationModel(
 
     # The JuMP.jl allocation problem
     problem = allocation_problem(config, p, capacity, allocation_network_id)
+    if (config.allocation.objective_type != "linear_absolute")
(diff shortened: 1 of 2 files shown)


core/src/allocation_init.jl
```diff
@@ -856,6 +856,9 @@ function AllocationModel(
 
     # The JuMP.jl allocation problem
     problem = allocation_problem(config, p, capacity, allocation_network_id)
+    if (config.allocation.objective_type != "linear_absolute")
+        error("Type of object function is not supported")
+    end
```
visr, line 861:
```suggestion
    if config.allocation.objective_type != "linear_absolute"
        error("Type of object function is not supported")
    end
```

We typically don't use brackets here if not needed.

visr (approved):
Thanks!

PR #1198, opened 2024-02-29
Use only explicit imports

visr:
We were mostly doing this already, but not entirely. ExplicitImports.jl just came out and I wanted to try it on this codebase. Works quite well! See this post for the advantages of explicit imports, especially for packages like Ribasim.

https://discourse.julialang.org/t/ann-explicitimports-jl-tooling-to-help-use-and-maintain-explicit-imports-in-your-package/110973

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -14,7 +14,6 @@ For more granular access, see:
 """
 module Ribasim
 
-import Arrow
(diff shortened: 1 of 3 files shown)


Hofer-Julian (approved):
I also think that we should avoid implicit imports outside of scripts.

PR #1215, opened 2024-03-05
Stop 'Julia auto update' from upgrading Julia version

visr:
I was wondering why the automated part of #1204 bumps the Julia version in the Manifest.toml. It is because we didn't run `juliaup default 1.10.0` yet, so 1.10.2 is the default. This fixes that. There is no need to run all of `pixi run install` for this action.


Hofer-Julian (approved):
Nice find!

PR #1220, opened 2024-03-08
QGIS plugin updates for `node_type`, remove `explode_and_connect`

### LINKED ISSUE #352: Adding edges in QGIS
Discussed with @Huite yesterday.

Right now after every edit of the edges, all polylines are exploded into their segments to form lines. Then for each line, each end has to lie on a node, such that the `from_node_id` and `to_node_id` can be automatically filled in into the attribute table.

This means that you can draw one feature for several edges since it will be exploded. However this doesn't go so well with the new `edge_type` attribute which can be "flow" or "control". Right now we always fill in "flow", but the user needs to get a dropdown of option just like the Node table. This only really makes sense if users draw edges one segment at a time. Doing it this way also allows edges to be a polyline rather than a single line segment, which is nice for when you want it to follow a water body on the map.

image

- [x] ask user to select the edge_type
- [x] stop exploding linestrings, just connect the end vertices of each feature

visr:
@jvanhouwelingen noticed that we don't create a `node_id` column instead of `fid`. To be able to edit tables in QGIS we always need `fid`, but its value no longer matters, so leaving it to autogenerate is normally best.

This removes `explode_and_connect` to automatically fill in part of the Edge attributes when creating a new geometry. It was no longer correct due to the schema updates, so better to remove it for now, and accept that users have to both draw the geometry and fill in the correct attributes. An issue with the desired behavior is already written in #352. In short, `derive_connectivity` would still be nice to have, but not `explode`.

- add `node_id` to the Node schema
- use `node_id` in the Node label
- add `from_node_type` and `to_node_type` to the Edge schema
- keep showing fid in the Edge label to be able to associate edge_id in timeseries widget
- use a dropdown selector for the Node and Edge editor widget, see screenshot below
- update schemas with `listen_node_type` as in #1110

image

Edit: with the updates from Huite this fixes #352


Huite (approved):
I've made the changes required so that automatic setting of the edge properties still works.

You can no longer draw the network in one go, you have to fill in one geometry per edge, but it may now consist of many segments.

I haven't looked at the tests...

PR #1223, opened 2024-03-08
Add plotting support for multiple line segment edges

### LINKED ISSUE #1218: Cannot plot model with edge geometries longer than a single line segment
We support LineString and MultiLineString geometries for edges. But cannot plot model with edge geometries longer than a single line segment in Ribasim Python.

image

There are some similar limitations regarding edge geometries described in #352.

Huite:
Fixes #1218

test_edge_plot:
image

And one of the testmodels I edited in QGIS to add additional vertices:
image

I also took the opportunity to use `.pop()` on the kwargs dicts.

The marker colors also had to be adjusted since there's no longer a 1:1 marked:edge_geometry relationship. I'm using a numpy array instead of generating the list, since the arrays support indexing.
Using boolean indexing for setting the colors doesn't work very well because the lengths of the color names might differ, and need to be supplied as the dtype. `np.where` is clever enough to take care of it, generating the right dtype on the fly.

### DIFF
--- a/python/ribasim/ribasim/geometry/edge.py
+++ b/python/ribasim/ribasim/geometry/edge.py
@@ -87,12 +87,10 @@ class Edge(SpatialTableModel[EdgeSchema]):
         angle = np.degrees(np.arctan2(dy, dx)) - 90
 
         # Set the color of the marker to match the line.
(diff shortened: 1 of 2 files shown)


visr (approved):
Great to have this, thanks!

PR #1239, opened 2024-03-13
Downgrade OteraEngine

### LINKED ISSUE #1238: `pixi run codegen` is broken on main
Probably since the merger of the Add API, `pixi run codegen` is broken on main. Specifically it fails on `gen_python.jl`.

Hofer-Julian:
Fixes #1238

Also add CI jobs to catch similar problems earlier


Project.toml
```diff
@@ -56,3 +56,6 @@ TestEnv = "1e6cf692-eddd-4d53-88a5-2d735e33781b"
 TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 TranscodingStreams = "3bb67fe8-82b1-5028-8e26-92a6c54297fa"
+
+[compat]
+OteraEngine = "^0.4"
```
visr, line 61:
```suggestion
OteraEngine = "0.4"
```

This means the same.

We don't use CompatHelper on the root project, so we'll need to remind ourselves to upgrade to 0.5.

visr (approved):

PR #1243, opened 2024-03-13
Enable read method again with the new add API

### LINKED ISSUE #1232: Add `Model.read` again to Ribasim Python
`Model.read` wasn't migrated when working on the add API.
Let's add it back again.

evetion:
Re-introduce database context (originally in removed Network node) and split Node tables per Node. Fixes #1232

### DIFF
--- a/python/ribasim/ribasim/geometry/edge.py
+++ b/python/ribasim/ribasim/geometry/edge.py
@@ -63,7 +63,7 @@ class EdgeTable(SpatialTableModel[EdgeSchema]):
             if geometry is None
             else [geometry]
         )
-        table_to_append = GeoDataFrame(
+        table_to_append = GeoDataFrame[EdgeSchema](
             data={
                 "from_node_type": pd.Series([from_node.node_type], dtype=str),
                 "from_node_id": pd.Series([from_node.node_id], dtype=int),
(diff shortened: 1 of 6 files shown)


python/ribasim/tests/test_model.py
```diff
@@ -155,7 +155,7 @@ def test_write_adds_fid_in_tables(basic, tmp_path):
     # for node an explicit index was provided
     nrow = len(model_orig.basin.node.df)
     assert model_orig.basin.node.df.index.name is None
-    assert model_orig.basin.node.df.index.equals(pd.Index(np.full(nrow, 0)))
+    # assert model_orig.basin.node.df.index.equals(pd.Index(np.full(nrow, 0)))
```
evetion, line 158: @Hofer-Julian I needed to reset the index after filtering the table to make write work, but that fails this test. Shall we remove this one?

Hofer-Julian: Yes, let's remove it.

Hofer-Julian (commented):
Can you please enable all tests with reason "Needs Model read implementation"?

python/ribasim/ribasim/geometry/edge.py
```diff
@@ -42,10 +42,6 @@ class Config:
 class EdgeTable(SpatialTableModel[EdgeSchema]):
     """Defines the connections between nodes."""
 
-    def __init__(self, **kwargs):
-        kwargs.setdefault("df", DataFrame[EdgeSchema]())
-        super().__init__(**kwargs)
```
Hofer-Julian, line 47: This is needed.

Ribasim Core always requires an Edge table, even if empty.
Without this, no Edge table is added if you don't have edges

evetion: Yep, but this interferes with the default Pydantic initialization, I fixed it in another way.

Hofer-Julian (approved):
After mypy is happy, this is ready to go from my side.
Thanks!

PR #1267, opened 2024-03-14
Add waterquality (substance/concentration) tables

### LINKED ISSUE #1125: Create external timeseries table specific for concentration timeseries results
The output from Delwaq is a concentration timeseries (possibly multiple per substance) per basin. We should be able to access it from a control node.

Relates to #520, but in discussion with Martijn, I propose to expand the Basin with an extra table, instead of having a new External node that somehow links to the same basin for a better user experience.

Added complexity, these fields would ideally be re-used in the online coupling via BMI later on.

evetion:
Fixes #1125. Split from #1137


evetion:
There can be some bike shedding on the table names, but the tables are all required. Ideally we merge this soon, so we prevent a huge PoC merge later on, as discussed in our retro.

core/src/schema.jl
```diff
@@ -96,6 +100,22 @@ end
     urban_runoff::Union{Missing, Float64}
 end
 
+@version BasinBoundaryconcentrationV1 begin
```
visr, line 103: BasinBoundaryConcentration

visr (commented):
Haven't had a look yet, but does this already incorporate the comments on #1125?

core/src/schema.jl
visr:
```julia
@version BasinConcentrationV1 begin
    node_id::Int32
    time::DateTime
    substance::String
    basin::Union{Missing, Float64}  # optional, for external calculations
    drainage::Union{Missing, Float64}
    precipitation::Union{Missing, Float64}
end
```

visr (approved):

evetion:
After an extensive discussion with @visr, we arrived at the current setup, with split out table for the basin.

For generating concentrations (whether with Delwaq, internal) we need an initial state and boundary conditions. For control on concentrations, we only need a timeseries. Catering to both needs is complex and confusing, so we now decided to split them all out. The External table can be refactored into an external node once the functionality and/or internal concentration calculations are there.

Issue #1251, opened 2024-03-13
Adding priorities to UserDemand / Time skips part of validation


Huite:
This may essentially be a duplicate of #1250, but I'm reporting it anyway.

If I set a single priority (1) in the UserDemand table, but forget to include return flow edges, I get the (very useful) validation error:

```
❯ c:\bin\ribasim\bin\ribasim.exe ribasim.toml

┌ Info: Starting a Ribasim simulation.
│   cli.ribasim_version = "2024.4.0"
│   starttime = 2017-01-01T00:00:00
└   endtime = 2019-01-01T00:00:00
┌ Error: UserDemand #118 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #119 must have at least 1 flow inneighbor(s) (got 0).
┌ Error: UserDemand #121 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #122 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #123 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #124 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #125 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #126 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
┌ Error: UserDemand #127 must have at least 1 flow inneighbor(s) (got 0).
└ @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\validation.jl:443
[...]

Issue #1250, opened 2024-03-13
Adding priorities to UserDemand/Time causes initialization to fail


Huite:
Running a model with all UserDemand at 1 runs fine, but assigning (random) priorities 1 to 3 results in this error below.

I haven't changed anything else in the model, so allocation isn't used.
In that case, the whole grouping per priority should maybe not happen at all?

```
❯ c:\bin\ribasim\bin\ribasim.exe ribasim.toml
┌ Info: Starting a Ribasim simulation.
│   cli.ribasim_version = "2024.4.0"
│   starttime = 2017-01-01T00:00:00
└   endtime = 2019-01-01T00:00:00
ERROR: BoundsError: attempt to access 0-element Vector{Float64} at index [1]
Stacktrace:
  [1] getindex
    @ .\essentials.jl:13 [inlined]
  [2] get_scalar_interpolation(starttime::Dates.DateTime, t_end::Float64, time::StructArrays.StructVector{Ribasim.UserDemandTimeV1, @NamedTuple{node_id::Vector{Int64}, time::Vector{Dates.DateTime}, demand::Vector{Float64}, return_fa
ctor::Vector{Float64}, min_level::Vector{Float64}, priority::Vector{Int64}}, Int64}, node_id::Ribasim.NodeID, param::Symbol; default_value::Float64)
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\util.jl:215
  [3] Ribasim.UserDemand(db::SQLite.DB, config::Ribasim.config.Config)
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\read.jl:703
  [4] Ribasim.Parameters(db::SQLite.DB, config::Ribasim.config.Config)
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\read.jl:893
  [5] Ribasim.Model(config::Ribasim.config.Config)
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\model.jl:51
  [6] run(config::Ribasim.config.Config)
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\main.jl:11
  [7] (::Ribasim.var"#206#208"{Ribasim.config.Config})()
    @ Ribasim D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\main.jl:66
[...]

PR #1268, opened 2024-03-14
Refactor function UserDemand in read.jl

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1250, Fixes https://github.com/Deltares/Ribasim/issues/1251.

@Huite it was a good call to ask for a refactor here, I think this cleans up the code nicely. Let me know whether this fixes the bugs you encountered.


Huite:
I'll check this afternoon, thanks!

core/src/read.jl
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
+    active::BitVector,
+    demand::Vector{Float64},
+    demand_itp::Vector{Vector{ScalarInterpolation}},
+    return_factor::Vector{Float64},
+    min_level::Vector{Float64},
+    static::StructVector{UserDemandStaticV1},
+    node_ids::Vector{NodeID},
+    priorities::Vector{Int32},
+)::Nothing
+    user_demand_idx = 1
```
Huite, line 648: Instead of user_demand_idx = 1 and the while nodes, I recommend this:

```julia
indices = searchsorted(node_ids, unique(static.node_id)
grouped = IterTools.groupby(row -> row.node_id, static)
for (user_demand_idx, group) in zip(indices, grouped)
     ...
end

File: core/src/read.jl [hunk 2]
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
-    time_node_id_vec = NodeID.(NodeType.UserDemand, time.node_id)
+        active[user_demand_idx] = coalesce(first_row.active, true)
+        return_factor[user_demand_idx] = first_row.return_factor
+        min_level[user_demand_idx] = first_row.min_level
 
-    if !valid
-        error("Problems encountered when parsing UserDemand static and time node IDs.")
+        priority_idx = 1
+        for row in group
+            while priorities[priority_idx] < row.priority
```
Huite, line 661: Does this work?

```julia
priority_idx = searchsortedfirst(priorities, row.priority)
```

The current thing is hard to read. If you really want to keep it, wrap it in some utility function.

core/src/read.jl
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
+                priority_idx += 1
+            end
+            demand_itp[user_demand_idx][priority_idx].u .= row.demand
+            demand[(user_demand_idx - 1) * length(priorities) + priority_idx] = row.demand
```
Huite, line 665: This really makes clear that a `Matrix{Float64}` would be nicer, but I'll open a separate issue on that.

core/src/read.jl
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
+    min_level::Vector{Float64},
+    time::StructVector{UserDemandTimeV1},
+    node_ids::Vector{NodeID},
+    priorities::Vector{Int32},
+    config::Config,
+)::Bool
     errors = false
-    trivial_timespan = [nextfloat(-Inf), prevfloat(Inf)]
     t_end = seconds_since(config.endtime, config.starttime)
+    user_demand_idx = 1
```
Huite, line 685: Also here, it becomes a lot more readable if you pre-search the indices, then zip over them.

Something like this:

```julia
grouped = IterTools.groupby(row -> (row.node_id, row.priority), time)
group_ids = [first(group).node_id for group in grouped]
indices = searchsorted(node_ids, group_ids)

for (user_demand_idx, group) in zip(indices, grouped)
```

core/src/read.jl
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
-                    push!(demand_itp_node_id, demand_p_itp)
-                end
-            end
-            push!(demand_itp, demand_itp_node_id)
+    # Initialize vectors for UserDemand fields
+    priorities = get_all_priorities(db, config)
+    n = length(node_ids)
+    active = BitVector(ones(Bool, n))
+    realized_bmi = zeros(n)
+    demand = zeros(n * length(priorities))
```
Huite, line 735: I'll open a separate issue, but to be clear, I think this would be nicer:

```julia
n_priority = length(priorities)
demand = zeros((n_priority, n_user)
allocated = fill(Inf, (n_priority, n_user)
```
This obviously means more changes in the code.

core/src/read.jl
```diff
@@ -635,111 +635,143 @@ function PidControl(db::DB, config::Config, chunk_sizes::Vector{Int})::PidContro
...
-                    end
-                else
-                    demand_p_itp = LinearInterpolation([0.0, 0.0], trivial_timespan)
-                    push!(demand_itp_node_id, demand_p_itp)
-                end
-            end
-            push!(demand_itp, demand_itp_node_id)
+    # Initialize vectors for UserDemand fields
+    priorities = get_all_priorities(db, config)
+    n = length(node_ids)
```
Huite, line 732: Maybe define this as `n_user` here, especially if we introduce `n_priority` with a follow up PR.

Huite (changes requested):

Huite:
I've opened #1277 for the Matrix things.

My suggestion is to merge this now with minimal changes related to the indices, and we can make a new PR for #1277.

core/src/read.jl
SouthEndMusic: Regarding this and the previous comment: my code was probably overly optimized, so that it essentially did the same as `searchsortedfirst` but without having to start at the beginning of `node_ids` or `priorities`.

core/src/read.jl
SouthEndMusic: Is there a particular reason why you want to pre-search the indices, and not search them within the loop?

Huite (approved):

PR #1275, opened 2024-03-15
Infer edge type in add API

### LINKED ISSUE #1261: Infer edge type in add API
Right now we always need to specify the edge type:

https://github.com/Deltares/Ribasim/blob/b5058d051817f6e0bdd8f07e68ec9a3428a0af4f/python/ribasim_testmodels/ribasim_testmodels/basic.py#L143

But QGIS infers it automatically. If it comes from DiscreteControl or PidControl, it is always a control edge, otherwise a flow edge:

https://github.com/Deltares/Ribasim/blob/b5058d051817f6e0bdd8f07e68ec9a3428a0af4f/ribasim_qgis/core/topology.py#L91-L95

I suppose we could do this in Ribasim Python as well.

Hofer-Julian:
Fixes #1261


Huite (approved):
Great work!

PR #1276, opened 2024-03-15
Add `model.node_table()`

visr:
With the add API the full Node table is never created in memory, but the rows are available per node type. Users that need just the network need a way to compute the table when needed.

Marking this as a draft because I'm stuck. The new function works fine, but I tried re-enabling the round trip test, and found that the read from #1243 creates empty node tables for each node type due to the way it filters. I wanted to fix that by setting `node.df = None` if the filtered Node table was empty (if the node type doesn't occur).

But that triggers behavior I do not understand.

```python
from ribasim import Model, Node
from ribasim.nodes import basin
from shapely.geometry import Point

model = Model(
    starttime="2020-01-01",
    endtime="2020-01-20",
)

model.basin.add(Node(5, Point(0, 0)), [basin.State(level=[1.0])])
# now both model.basin.node and model.basin.state have 1 row
model.basin.add(Node(8, Point(3, 0)), [basin.State(level=[2.0])])
# now model.basin.state has 2 rows, but model.basin.node only the last node 8
```

The `setattr` in the `add` method triggers this, so if you pass no tables like `basin.State`, `model.basin.node` contains both rows. I rewrote the if/else in `add` a bit but no real changes, the logic there seems ok. We just seem to trigger something pydantic that I find hard to debug, so any help welcome.

This is all triggered by my change to `filter`, so perhaps if there are better ways to fix that, by e.g. pruning empty node types at the end of model reading or something like that, I'm also happy.

### DIFF
--- a/python/ribasim/ribasim/config.py
+++ b/python/ribasim/ribasim/config.py
@@ -124,26 +124,33 @@ class MultiNodeModel(NodeModel):
             )
 
         for table in tables:
-            member_name = _pascal_to_snake(table.__class__.__name__)
-            existing_member = getattr(self, member_name)
-            existing_table = (
-                existing_member.df if existing_member.df is not None else pd.DataFrame()
-            )
             assert table.df is not None
             table_to_append = table.df.assign(node_id=node_id)
-            setattr(self, member_name, pd.concat([existing_table, table_to_append]))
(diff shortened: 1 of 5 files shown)


visr:
Closing in favor of a working PR #1279 and an issue #1278.

Issue #1278, opened 2024-03-15
Model.read creates empty node tables for each node type


visr:
Preferably these node tables are kept None, such that it is clear that they are not present, e.g. in `_nodes()` from #1270.

In #1276 I came across this issue and try to adapt the `filter` function, which caused several other issues.

PR #1279, opened 2024-03-15
Add `model.node_table()`, take two

visr:
This supersedes #1276 by working around #1278 and still enabling the round tripping tests.

Also simplifies `__assert_equal` a bit and stimulates `Model.read(toml_path)` usage over `Model(filepath=toml_path)`.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -3,6 +3,7 @@ from collections.abc import Generator
 from pathlib import Path
 from typing import Any
 
+import pandas as pd
 import tomli
 import tomli_w
 from matplotlib import pyplot as plt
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -133,11 +135,24 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
...
 
+    def node_table(self) -> NodeTable:
+        """Compute the full NodeTable from all node types."""
+        df_chunks = [node.node.df for node in self._nodes()]
+        df = pd.concat(df_chunks, ignore_index=True)  # type: ignore
+        node_table = NodeTable(df=df)
+        node_table.sort()
+        return node_table
+
     def _nodes(self) -> Generator[MultiNodeModel, Any, None]:
```
Huite, line 146: Is there really a reason to prefer a generator over an ordinary list here?

Huite (approved):

python/ribasim/ribasim/model.py
visr: Not really I guess. The implementation is slightly shorter. We were already using a generator for a similar function, and we only use this function to iterate over, so no `list(gen)`.

PR #1284, opened 2024-03-18
Add flags to pip install

Hofer-Julian:
Stolen from the pixi docs, so I assume that's the recommended way


visr (approved):
Seems fine to me. We are going to start using pixi's uv integration for this later on?

Hofer-Julian:
> We are going to start using pixi's uv integration for this later on?

Yes, indeed

Issue #1234, opened 2024-03-12
Plotting of listen (control) edges with Ribasim Python


Hofer-Julian:
We used to be able to plot control listen edges with Ribasim Python.
That was removed by moving to the add API.
We will want to add that functionality back.

This is the original code: https://github.com/Deltares/Ribasim/blob/c5ae83de3159ab77407773b02347fec6f1aa41c0/python/ribasim/ribasim/model.py#L391-L441

This is a render of the example: 
[Ribasim - Examples.pdf](https://github.com/Deltares/Ribasim/files/14572471/Ribasim.-.Examples.pdf)

PR #1288, opened 2024-03-18
Plot listen edges

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1234.

Probably not yet in the cleanest way though.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -3,13 +3,10 @@ from collections.abc import Generator
 from pathlib import Path
 from typing import Any
 
-import geopandas as gpd
-import numpy as np
 import pandas as pd
 import tomli
 import tomli_w
 from matplotlib import pyplot as plt
-from matplotlib.collections import LineCollection
 from pydantic import (
     DirectoryPath,
     Field,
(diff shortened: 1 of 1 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -232,7 +235,72 @@ def reset_contextvar(self) -> "Model":
...
+                    listen_nodes_geometry.y.to_numpy().reshape(-1, 1),
+                    control_nodes_geometry.x.to_numpy().reshape(-1, 1),
+                    control_nodes_geometry.y.to_numpy().reshape(-1, 1),
+                ]
+            ).reshape(-1, 2, 2),
+            color="gray",
+            ls="--",
+            label="Listen edge",
+        )
+        ax.add_collection(line_segments)
```
Hofer-Julian, line 302: Let's make a loop out of that (or watch @Huite optimize it into something that is fast and readable)

python/ribasim/ribasim/model.py
```diff
@@ -232,7 +235,72 @@ def reset_contextvar(self) -> "Model":
...
+            data={
+                "control_node_id": pd.Series([], dtype="int"),
+                "control_node_type": pd.Series([], dtype="str"),
+                "listen_node_id": pd.Series([], dtype="int"),
+                "listen_node_type": pd.Series([], dtype="str"),
+            }
+        )
+
+        # Listen edges from PidControl
+        for table in [self.pid_control.static.df, self.pid_control.time.df]:
```
Hofer-Julian, line 248:
```suggestion
        for table in (self.pid_control.static.df, self.pid_control.time.df):
```

Hofer-Julian (approved):

PR #1293, opened 2024-03-19
change int to string for PID listen_node_type

Jingru923:
In documentation of PID node, [Usage](https://deltares.github.io/Ribasim/core/usage.html#sec-pid-control) under Julia core, the listen_note_type is supposed to be "String" but in the document it says "Int32". It is incorrect, and therefore should be changed.


SouthEndMusic (approved):
Nice fix

Issue #661, opened 2023-10-10
Combined vertical fluxes on Basin


visr:
One limitation of the way we save flows in #644 is that we cannot separate the 4 basin fluxes (precipitation, evaporation, drainage, infiltration). We could make separate accommodations for those. Or perhaps these should really be separate node types with 1 flux each;

- Precipitation
- Evaporation (specify potential_evaporation and connect to Basin to get open water evaporation)
- Infiltration (combine with drainage? one node per modflow cell?)

This would also fix the timeseries duplication problem. Say you have 1000 Basins in a small area with a single precipitation timeseries. Now you have to copy this data 1000x, once per Basin. If you have a separate Precipitation node, you only have to define it once and connect it to all 1000 Basins, just like you can do already for e.g. a LevelBoundary. For visualization you'd probably want an easy way to hide these nodes and edges though.

_Originally posted by @visr in https://github.com/Deltares/Ribasim/issues/644#issuecomment-1745584689_

gijsber:
From a modeller's perspective I am not fond of this solution as it involves another connection to track that your timesries ends up in the correct basin. But if we can hide it from the modeller and julia or ribasim_python offers the methods to post/get these series using the basinId it probably is ok.

visr:
Yes I think it should be straightforward to create tooling to follow the indirection.

Huite:
I guess the benefit additional nodes is that you can always store a "flow" variable for a given node, so all the information is encoded in the network and the types. I'm not too fond it for the same reason @gijsber mentions, but also conceptually: I've always considered the nodes to be somewhat physically distinct entities, which is reflected by the network topology.

I would vote for what feels like the solution here: treat the edge flows separately in the output. I find it a bit clunky in how you need to select basin flows anyway. Why not store a separate table with a basin index and the various balance terms for that timestep?

MODFLOW 6 also has different format for internodal flows (flow-ja-face) and flows to the node itself.

In UGRID terms, similarly: the internodal flows would be stored on the edges. The node terms would be stored on the nodes.
Although in case of UGRID, you wouldn't be able to combine it nicely either, because the nodes of the topology are so dissimilar (a control node obviously has no precipitation term).
So in the case of writing to UGRID, you either need a lot of padding values (NaN for the precipitation variable of the control nodes e.g.), or you need to define separate dimensions per type of node.

[...]

Issue #1084, opened 2024-02-06
Create a (cumulative) integrated boundary flux in Ribasim and expose it via BMI


HendrikKok:
For the coupling to MODFLOW 6 we need to know how much of a requested infiltration flux could be realized by Ribasim over a given timespan. If Ribasim does a cumulative integration of its boundary fluxes, we can extract a realized volume per timestep by computing the difference over any given timespan.

gijsber:
linked to #935

visr:
Fixed by #1300

PR #1300, opened 2024-03-21
Split vertical fluxes on basin

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/661.


SouthEndMusic:
@visr let me know whether you think this is going in the right direction. The vertical fluxes get integrated now but they are not yet saved. Something about this also feels a bit silly, since this does an approximation of the integral while we could easily obtain the exact value.

SouthEndMusic:
In fact, for forward fill we don't need `vertical_flux_prev` at all, we just have to make sure the integration is computed before `vertical_flux` is updated.

SouthEndMusic:
Maybe https://github.com/Deltares/Ribasim/issues/1084 can also be picked up in this PR, just expose `vertical_flux_integrated` in stead of `vertical_flux` values. One thing we need to think about as discussed with @HendrikKok is what happens when a basin dries out.

core/src/callback.jl
```diff
@@ -496,6 +491,12 @@ function update_basin(integrator)::Nothing
         set_table_row!(table, row, i)
     end
 
+    for (i, id) in enumerate(basin.node_id)
+        update_vertical_flux!(basin, storage, i)
+    end
+
+    # Forget about vertical fluxes before basin update
+    copyto!(vertical_flux_prev, vertical_flux)
```
SouthEndMusic, line 499: To be most accurate, maybe an exception should be made for vertical fluxes which are not set by the `Basin \ time` table.

core/src/bmi.jl
```diff
@@ -36,10 +36,14 @@ function BMI.get_value_ptr(model::Model, name::AbstractString)
...
     elseif name == "basin.level"
         get_tmp(model.integrator.p.basin.current_level, 0)
-    elseif name == "basin.infiltration"
-        model.integrator.p.basin.infiltration
-    elseif name == "basin.drainage"
-        model.integrator.p.basin.drainage
+    elseif name == "basin.infiltration_instantaneous"
+        get_tmp(model.integrator.p.basin.vertical_flux, 0).infiltration
+    elseif name == "basin.drainage_instantaneous"
+        get_tmp(model.integrator.p.basin.vertical_flux, 0).drainage
```
visr, line 42:
```suggestion
    elseif name == "basin.infiltration"
        get_tmp(model.integrator.p.basin.vertical_flux, 0).infiltration
    elseif name == "basin.drainage"
        get_tmp(model.integrator.p.basin.vertical_flux, 0).drainage
```

core/test/bmi_test.jl
```diff
@@ -61,8 +61,10 @@ end
     for name in [
         "basin.storage",
         "basin.level",
-        "basin.infiltration",
-        "basin.drainage",
+        "basin.infiltration_instantaneous",
+        "basin.drainage_instantaneous",
```
visr, line 65:
```suggestion
        "basin.infiltration",
        "basin.drainage",
```

core/src/parameter.jl
```diff
@@ -171,26 +173,28 @@ struct Basin{T, C} <: AbstractParameterNode
...
         node_id,
-        precipitation,
-        potential_evaporation,
-        drainage,
-        infiltration,
+        vertical_flux_from_input::V1,
+        vertical_flux::V2,
+        vertical_flux_prev::V3,
+        vertical_flux_integrated,
+        vertical_flux_bmi,
```
visr, line 180:
```suggestion
        vertical_flux_integrated::V3,
        vertical_flux_bmi::V3,
```

core/src/solve.jl
```diff
@@ -52,34 +51,43 @@ end
...
+    return influx
+end
+
+function formulate_basins!(
+    du::AbstractVector,
+    basin::Basin,
+    storage::AbstractVector,
+)::Nothing
+    for (i, id) in enumerate(basin.node_id)
+        # add all precipitation that falls within the profile
```
visr, line 88:
```suggestion
    for i in eachindex(basin.node_id)
```

core/src/callback.jl
```diff
@@ -496,6 +493,12 @@ function update_basin(integrator)::Nothing
         set_table_row!(table, row, i)
     end
 
+    for (i, id) in enumerate(basin.node_id)
+        update_vertical_flux!(basin, storage, i)
+    end
```
visr, line 498:
```suggestion
    for i in eachindex(basin.node_id)
        update_vertical_flux!(basin, storage, i)
    end
```

core/src/callback.jl
```diff
@@ -496,6 +493,12 @@ function update_basin(integrator)::Nothing
...
+
+    # Forget about vertical fluxes before basin update
```
visr, line 500:
```suggestion
    # Forget about vertical fluxes to handle discontinuous forcing from basin_update
```

core/src/callback.jl
visr: Typically only some of the vertical fluxes will be in `Basin \ time`. Could you make an issue about this? I don't fully oversee the error introduced in this case.

core/src/util.jl
```diff
@@ -707,3 +707,25 @@ function Base.get(
         nothing
     end
 end
+
+"""
+Get the time interval between (flow) saves
+"""
+function get_Δt(integrator, graph)::Float64
```
visr, line 714: Probably better to make this accept only `integrator` and take the graph out of the parameters.

core/src/allocation_optim.jl
```diff
@@ -321,9 +321,15 @@ function get_basin_data(
...
 )
     (; graph, basin, level_demand) = p
+    (; vertical_flux) = basin
     (; Δt_allocation) = allocation_model
     @assert node_id.type == NodeType.Basin
-    influx = get_flow(graph, node_id, 0.0)
+    vertical_flux = get_tmp(vertical_flux, 0)
+    _, basin_idx = id_index(basin.node_id, node_id)
+    # NOTE: Instantaneous
+    influx =
```
visr, line 330: Is it possible to pull out this as a `get_flow` method and use it in `update_vertical_flux!` as well?

visr (approved):

docs/core/usage.qmd
```diff
@@ -705,21 +705,30 @@ derivative       | Float64  | -        | -
 
 ## Basin - `basin.arrow`
 
-The basin table contains results of the storage and level of each basin at every solver
-timestep. The initial condition is also written to the file.
+The basin table contains:
+
+- results of the storage and level of each basin, which are instantaneous values;
+- results of the vertical fluxes on each basin, which are mean values over the `saveat` intervals. In the time column the start of the period is indicated. This means that for the final time in the table the mean vertical fluxes are not computed, and thus left $0$.
```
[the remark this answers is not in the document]
SouthEndMusic: Ah missing makes more sense, this is just how it worked out in my implementation.

PR #1301, opened 2024-03-21
Unpin OteraEngine, switch to latest 0.5.1

### LINKED ISSUE #1240: Upgrade OteraEngine
We had to downgrade OteraEngine because of breaking changes: https://github.com/Deltares/Ribasim/pull/1239

We should investigate what's the problem and how to upgrade

visr:
Fixes #1240
It was not a breaking change but a regression, which I reported and then swiftly got fixed.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0"
 manifest_format = "2.0"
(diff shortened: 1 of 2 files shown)


Hofer-Julian (approved):
Awesome!

Issue #969, opened 2024-01-18
Show flows on the map in QGIS


visr:
**What**
Make it possible to see the flow rates over all edges on the map, that integrates with the QGIS temporal controller to select the moment in time to evaluate.

**Why**
To gain insight into the model behavior much more quickly. Currently users have to select edges and view the timeseries, but this does not make it easy to see what is happening overall at one moment in time.

**How**
Needs to be refined. For a first version it can be a description of a workflow that combines the edges and flow results and sets up this spatial view. We need flow rates as labels on the edges. Additionally we could use line color, thickness, or animated arrows to make it easy to interpret relative flow rates at a glance.

Huite:
This is arguably easiest to do with Mesh Layers rather than with Vector Layers, since it basically works out of the box.

A big downside of vector layers is that the Temporal Control integration is not great: you can set a datetime column, but it ends up filtering all the rows. The Mesh Layer data model just finds the required index instead, and fetches the data for that time axis index.

I would vote for UGRID netCDF. Though technically MDAL also seems to support network (edge) associated data with PLY: https://www.mdal.xyz/drivers/ply.html
I think binary PLY is also supported, but it doesn't seem to be documented.

visr:
Since I'm not currently working on it I put my work in progress here: https://gist.github.com/visr/a95ab57960bbc9aa3c080711f03084b7
This was also already shared with @evetion and used in #1137 here: https://github.com/Deltares/Ribasim/blob/4494cb0792192c9bddaa861e84732d57e7346887/coupling/delwaq/delwaq_util.py

The idea was to add this code to Ribasim Python with an optional dependency on xugrid, something like a class of Model called `to_xugrid`.

visr:
Fixed by #1369.

PR #1314, opened 2024-03-25
Add `model.to_xugrid()`

visr:
This adds an optional dependency on `xugrid`, to facilitate creating UGRID netCDFs, to support #969.

For now I did not yet add any results to the dataset, but this could be done as a follow-up. This is just the network.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -3,6 +3,7 @@ from collections.abc import Generator
 from pathlib import Path
 from typing import Any
 
+import numpy as np
 import pandas as pd
 import tomli
 import tomli_w
@@ -14,6 +15,11 @@ from pydantic import (
     field_serializer,
     model_validator,
 )
+
+try:
+    import xugrid
+except ImportError:
+    xugrid = None
 
 import ribasim
 from ribasim.config import (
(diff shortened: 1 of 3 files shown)


visr:
This code assumes globally unique node IDs, so I added it to the list at https://github.com/Deltares/Ribasim/issues/1318. An additional complication is that MDAL currently doesn't support (variable length?) strings as coordinate variables, so just adding `node_type::String` is maybe not possible. @Huite suggested encoding them as integers for now with an IntEnum.

python/ribasim/ribasim/model.py
```diff
@@ -342,3 +347,58 @@ def plot(self, ax=None, indicate_subnetworks: bool = True) -> Any:
         ax.legend(handles, labels, loc="lower left", bbox_to_anchor=(1, 0.5))
 
         return ax
+
+    def to_xugrid(self) -> xugrid.UgridDataset:
+        """Convert the network to a xugrid.UgridDataset."""
+        node_df = self.node_table().df
```
Hofer-Julian, line 353: I think that method should throw an Exception early if `xugrid` is `None`

python/ribasim/ribasim/model.py
```diff
@@ -342,3 +347,58 @@ def plot(self, ax=None, indicate_subnetworks: bool = True) -> Any:
...
+
+        # This will need to be adopted for locally unique node IDs,
+        # otherwise the `node_lookup` with `argsort` is not correct.
+        assert node_df.node_id.is_unique
```
Hofer-Julian, line 357: Maybe add a error message?

python/ribasim/ribasim/model.py
```diff
@@ -342,3 +347,58 @@ def plot(self, ax=None, indicate_subnetworks: bool = True) -> Any:
...
+
+        # from node_id to the node_dim index
+        node_lookup = pd.Series(
+            index=node_id,
+            data=node_id.argsort().astype("int32"),
+            name="node_index",
+        )
+
+        if node_df.crs is None:
+            # can be removed when CRS is required, #1254
```
Hofer-Julian, line 376:
```suggestion
            # TODO: can be removed when CRS is required, #1254
```

Hofer-Julian (approved):
Small comments, looks good to me otherwise

python/ribasim/ribasim/model.py
Huite: I use this in imod-python or xugrid:

```python
class MissingOptionalModule:
    """
    Presents a clear error for optional modules.
    """

    def __init__(self, name):
        self.name = name

    def __getattr__(self, name):
        raise ImportError(f"{self.name} is required for this functionality")
```

But here probably `if xugrid is None` suffices.

Huite (approved):

python/ribasim/ribasim/model.py
Hofer-Julian: Using `MissingOptionalModule` sounds better to me.
Less chance to overlook the assertion, since mypy doesn't seem to spot it.

visr: Done in d822065d3f1708940e973a7059d7075eaa6537de.

python/ribasim/ribasim/model.py
```diff
@@ -354,21 +355,21 @@ def to_xugrid(self) -> xugrid.UgridDataset:
 
         # This will need to be adopted for locally unique node IDs,
         # otherwise the `node_lookup` with `argsort` is not correct.
-        assert node_df.node_id.is_unique
+        assert node_df.node_id.is_unique, "node_id must be unique"
```
Huite, line 358: Should be a ValueError right?

visr: Ok yes that's better, done.

Huite (approved):

PR #1315, opened 2024-03-25
Use 32 bit signed integer on the Python side as well

visr:
Follow up of #1252

### DIFF
--- a/python/ribasim/ribasim/config.py
+++ b/python/ribasim/ribasim/config.py
@@ -3,6 +3,7 @@ from collections.abc import Sequence
 from enum import Enum
 from typing import Any
 
+import numpy as np
 import pandas as pd
 import pydantic
 from geopandas import GeoDataFrame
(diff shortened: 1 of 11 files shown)


visr:
Mypy is not happy with `Series[np.int32]`, even though it works and should work according to the docs.

```
python\ribasim\ribasim\geometry\edge.py:33: error: Type argument "signedinteger[_32Bit]" of "Series" must be a subtype of "bool | int | str | float | ExtensionDtype | Bool | Date | Timestamp | Decimal | Timedelta | Category | Float | Float16 
| Float32 | Float64 | Int | Int8 | Int16 | Int32 | Int64 | UInt8 | UInt16 | UInt32 | UInt64 | INT8 | INT16 | INT32 | INT64 | UINT8 | UINT16 | UINT32 | UINT64 | Object | String | STRING | Geometry"  [type-var]
```

Mypy is happy with this

```
from pandera.dtypes import Int32
Series[Int32]
```

https://pandera.readthedocs.io/en/stable/dtype_validation.html

Fix incoming

Hofer-Julian (approved):

PR #1322, opened 2024-03-26
Minimal pixi default environment

Hofer-Julian:
Only add dependencies to the default environment that are necessary to test Ribasim Python or Ribasim API. This should make it easier to keep track of which dependencies are necessary for what, and also reduce solve times.

### DIFF
--- a/docs/contribute/index.qmd
+++ b/docs/contribute/index.qmd
@@ -35,7 +35,7 @@ First, set up pixi as described on their getting started [page](https://prefix.d
 Then set up the environment by running the following commands:
 
 ```sh
-pixi run install
+pixi run --environment=dev install
 ```
 
 This will automatically install all required packages for development.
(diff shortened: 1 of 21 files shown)


docs/contribute/index.qmd
```diff
@@ -35,7 +35,7 @@ First, set up pixi as described on their getting started [page](https://prefix.d
 Then set up the environment by running the following commands:
 
 ```sh
-pixi run install
+pixi run --environment=dev install
```
Hofer-Julian, line 38: This is only necessary because otherwise the editable install installs into the wrong environment.
As soon as pixi supports editable installs this will not be necessary anymore.

visr: Perhaps it's good to create an issue for this, with a reminder to remove this.

visr:
> also reduce solve times

Do we now already use the new minimal default environment anywhere in CI, or do we use `dev` for everything?

Hofer-Julian:
> Do we now already use the new minimal default environment anywhere in CI, or do we use `dev` for everything?

Ribasim Python tests use it.

Hofer-Julian:
The lock file became a lot smaller:
image

Issue #1254, opened 2024-03-13
Support specifying coordinate reference system (CRS) of geometries


visr:
**What**
It is currently not possible to specify a CRS for our spatial tables. That means we cannot reliably place our geometries on top of a map.

**Why**
A clear and concise description of why this feature is needed and its context.

**How**
I think it'd be best to specify the CRS once when constructing the Model object, and then setting that preferably directly during GeoDataFrame construction. Setting it on `to_file` works for the GeoPackage, but it'd be nicer if we can also plot from Python on basemaps before writing to file.

~~CRS should be optional since maybe the true CRS is unknown, or there is no true CRS, like with most our test models. This unfortunately gives a warning from pyogrio that we currently suppress. Hopefully we can stop doing that. If this global suppression affects other Python code it would be best to only do that temporarily when calling `to_file` using [this](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).~~ EDIT: Let's make it required, see post below.

**Additional context**
https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.set_crs.html
https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_file.html
https://github.com/Deltares/Ribasim/blob/e1f8b81d861b62724d8fb9ccf8b828fa8e68ff32/python/ribasim/ribasim/input_base.py#L41

Huite:
I think you could seriously consider making the CRS required for Model initialization.
Preferably this doesn't introduce any additional attributes, but just forwards it immediately to the geodataframes.

The QGIS plugin always sets a CRS based on the current Map Canvas: https://github.com/Deltares/Ribasim/blob/1f20b68eb4a6cba52fda7b51fe3f5ba5cfae9edf/ribasim_qgis/widgets/ribasim_widget.py#L72

QGIS defaults to EPSG:4326 (WGS 84) -- although I personally set my default QGIS CRS to EPSG:28992.

Although in that case, it makes some sense to support `to_crs` and `set_crs` methods on the Model object as well (they just forward to the geodataframes). Otherwise you have to do it twice yourself on the node and edge dataframe.

visr:
Yes that is a good point, let's make it required input.

PR #1339, opened 2024-03-29
Support specifying coordinate reference system (CRS) of geometries

Hofer-Julian:
Fixes https://github.com/Deltares/Ribasim/issues/1254

Breaking changes:
- Adds new required TOML key "crs"
- Adds new required `Model` parameter "crs"

Note: we could avoid adding the TOML key by requiring `Model.read` to take `crs` as parameter. Not sure if that's an improvment though.

### DIFF
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -120,6 +120,7 @@ end
 @option @addnodetypes struct Toml <: TableOption
     starttime::DateTime
     endtime::DateTime
+    crs::String
     ribasim_version::String
     input_dir::String
     results_dir::String
--- a/core/test/docs.toml
+++ b/core/test/docs.toml
@@ -2,6 +2,7 @@
 # can also be set to a date-time like 1979-05-27T07:32:00
 starttime = 2019-01-01 # required
 endtime = 2021-01-01   # required
+crs = "EPSG:28992"     # required
 
 # input files
 input_dir = "."         # required
--- a/core/test/docs.toml
+++ b/core/test/docs.toml
@@ -2,6 +2,10 @@
 # can also be set to a date-time like 1979-05-27T07:32:00
 starttime = 2019-01-01 # required
 endtime = 2021-01-01   # required
+
+# Coordinate Reference System
+# The accepted strings are documented here:
+# https://proj.org/en/9.4/development/reference/functions.html#c.proj_create
 crs = "EPSG:4326"      # required
 
 # input files
--- a/core/test/docs.toml
+++ b/core/test/docs.toml
@@ -2,7 +2,7 @@
 # can also be set to a date-time like 1979-05-27T07:32:00
 starttime = 2019-01-01 # required
 endtime = 2021-01-01   # required
-crs = "EPSG:28992"     # required
+crs = "EPSG:4326"      # required
 
 # input files
 input_dir = "."         # required
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -401,12 +401,6 @@ class Model(FileModel):
             name="node_index",
         )
 
-        if node_df.crs is None:
-            # TODO: can be removed when CRS is required, #1254
-            projected = False
-        else:
-            projected = node_df.crs.is_projected
-
         grid = xugrid.Ugrid1d(
             node_x=node_df.geometry.x,
             node_y=node_df.geometry.y,
@@ -418,7 +412,7 @@ class Model(FileModel):
                 )
             ),
             name="ribasim",
-            projected=projected,
+            projected=node_df.crs.is_projected,
             crs=node_df.crs,
         )
 
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -182,6 +182,7 @@ class Model(FileModel):
             sub._save(directory, input_dir)
 
     def set_crs(self, crs: str) -> None:
+        self.crs = crs
         self.edge.df = self.edge.df.set_crs(crs)
         for sub in self._nodes():
             if sub.node.df is not None:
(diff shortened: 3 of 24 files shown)


visr (changes requested):
Yes I think adding the TOML key is fine. I added some commits. Notably a test that "Basin / area" also gets the CRS, but that fails, so I think that still needs to be fixed.

If it is hard to fix it is fine as-is, but currently the Node tables only gets a CRS when it is assembled by `node_table()`. That means if you just want the GeoDataframe of the Basin nodes with `model.basin.node` you get a GeoDataframe with no CRS set.

Hofer-Julian:
> If it is hard to fix it is fine as-is, but currently the Node tables only gets a CRS when it is assembled by node_table(). That means if you just want the GeoDataframe of the Basin nodes with model.basin.node you get a GeoDataframe with no CRS set.

Yeah, I don't know how that would be done. When we run `model.basin.add`, we don't have access to the model state and therefore no access to the crs.

visr:
Would it make sense to define `model.set_crs()` that sets the CRS of all spatial tables? Then we can call that on save, but also earlier if needed.

Huite:
Indeed. As I mentioned in #1254, I would probably implement both `set_crs` (to just assign a crs) and `to_crs` (to reproject).

Hofer-Julian:
> Would it make sense to define model.set_crs() that sets the CRS of all spatial tables? Then we can call that on save, but also earlier if needed.

Sounds good to me.

python/ribasim/ribasim/model.py
```diff
@@ -86,6 +88,7 @@ class Model(FileModel):
     user_demand: UserDemand = Field(default_factory=UserDemand)
 
     edge: EdgeTable = Field(default_factory=EdgeTable)
+    crs: str
```
evetion, line 91: Could be below `endtime` attribute, like it is in `config.jl`.

python/ribasim/ribasim/model.py
```diff
@@ -170,9 +181,18 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
...
+            if sub.node.df is not None:
+                sub.node.df = sub.node.df.set_crs(crs)
+            for table in sub._tables():
+                if isinstance(table, SpatialTableModel) and table.df is not None:
+                    table.df = table.df.set_crs(crs)
+
     def node_table(self) -> NodeTable:
         """Compute the full NodeTable from all node types."""
-        df_chunks = [node.node.df for node in self._nodes()]
+        df_chunks = [node.node.df.set_crs(self.crs) for node in self._nodes()]
```
evetion, line 195: Together with `_save` and `set_crs` this seems overkill? Essentially, when getting the node_table (used in _save), the crs is always set to self.crs, which follows from the toml. `_save` only needs to set the edge crs, and the `set_crs` method can go?

python/ribasim/ribasim/model.py
```diff
@@ -170,9 +181,18 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
         for sub in self._nodes():
             sub._save(directory, input_dir)
 
+    def set_crs(self, crs: str) -> None:
```
evetion, line 184: I would expect this method (only?) set `self.crs = crs`, but it doesn't. If I do model.crs = "something else", it would overwrite anything set in this method on _save.

core/src/config.jl
```diff
@@ -120,6 +120,7 @@ end
 @option @addnodetypes struct Toml <: TableOption
     starttime::DateTime
     endtime::DateTime
+    crs::String
```
evetion, line 123: Julia doesn't use this, so why not give it a default? That would help with older toml files as well.

core/test/docs.toml
```diff
@@ -2,6 +2,7 @@
 # can also be set to a date-time like 1979-05-27T07:32:00
 starttime = 2019-01-01 # required
 endtime = 2021-01-01   # required
+crs = "EPSG:4326"      # required
```
evetion, line 5: Maybe good to document what strings we support, just forwarding to the geopandas documentation page is good enough (indicating we can take proj strings, WKT etc).

python/ribasim/ribasim/model.py
evetion: I would propose a default here, which could even be the following:

> ENGCRS[\"Undefined Cartesian SRS with unknown unit\", EDATUM[\"Unknown engineering datum\"], CS[Cartesian,2], AXIS[\"X\",unspecified, ORDER[1], LENGTHUNIT[\"unknown\",0]], AXIS[\"Y\",unspecified, ORDER[2], LENGTHUNIT[\"unknown\",0]]]

python/ribasim/ribasim/model.py
```diff
@@ -170,9 +181,18 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
...
+        self.edge.df = self.edge.df.set_crs(crs)
+        for sub in self._nodes():
+            if sub.node.df is not None:
+                sub.node.df = sub.node.df.set_crs(crs)
+            for table in sub._tables():
+                if isinstance(table, SpatialTableModel) and table.df is not None:
```
evetion, line 190: Which tables are spatial that are not already covered by the `sub.node.df`?

evetion (changes requested):

python/ribasim/ribasim/model.py
Hofer-Julian: > Could be below endtime attribute, like it is in config.jl.

Done

> I would propose a default here

This has been discussed in the original issue and it was decided to not go for a default.

core/test/docs.toml
Hofer-Julian: Done

python/ribasim/ribasim/model.py
Hofer-Julian: You are correct, the latest iteration doesn't need that.

evetion (approved):

python/ribasim/ribasim/model.py
Hofer-Julian: Had to re-introduce after all, other functions depend on it.

PR #1345, opened 2024-04-03
CompatHelper: Don't bump equality constraints

visr:
This will avoid unwanted CompatHelper PRs like https://github.com/Deltares/Ribasim/pull/1342.


Hofer-Julian (approved):
I agree with sane compats

Issue #1317, opened 2024-03-26
Simplify Jacobian sparsity construction


SouthEndMusic:
The `tki_zutphen_berkel` model by @harm-nomden-sweco runs into some convergence issues, which can be partially be solved by setting `sparse = false`. This indicates that there is a problem in `sparsity.jl`. Some investigation of the jacobian from this model might point out where the problem is.

_Edit: simplifying the Jacobian sparsity construction might be the way to go, see [this comment](https://github.com/Deltares/Ribasim/issues/1317#issuecomment-2034058769)._

SouthEndMusic:
I had a look into this. When comparing the basin storages for `sparse = true` and `sparse = false` there are particular basins with notable differences:

![comparse_sparse_dense_Zupthen](https://github.com/Deltares/Ribasim/assets/74617371/7c7992af-76b9-4c00-801e-d70b883c2712)

However, I have not been able to deduce the cause of this.

SouthEndMusic:
One thing that does not solve this problem but might make some models a bit more stable: In `sparsity.jl`, `Outlet` belongs in the category of `LinearResistance` and `ManningResistance`. 

The sparsity structure now is quite highly optimized, which requires maintenance while we might not actually need it. It would make the code much simpler if we just say that any basin depends on itself and the basins it is connected to (possibly via `FractionalFlow`). This introduces some false positives but I don't think the performance drop will be significant. Extra care is still required for continuous control.

visr:
Yeah the simpler approach sounds easier to maintain.

PR #1349, opened 2024-04-03
Simplify sparsity.jl

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1317.

Obtaining the Jacobian takes a coarser (with some false positives) but easier to maintain approach. Special care must still be taken with e.g. continuous control concepts, where basins that are further apart (topologically) than 1 node can depend on each other. This does not fix all stability problems in the model by @harm-nomden-sweco, but achieves the same progress as before without the requirement `sparse = false`.

### DIFF
--- a/core/src/sparsity.jl
+++ b/core/src/sparsity.jl
@@ -1,5 +1,5 @@
 """
-Get a sparse matrix whose sparsity matches the sparsity of the Jacobian
+Get a sparse matrix whose sparsity matches (with some false positives) the sparsity of the Jacobian
 of the ODE problem. All nodes are taken into consideration, also the ones
 that are inactive.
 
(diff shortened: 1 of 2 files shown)


core/src/sparsity.jl
```diff
@@ -11,124 +11,34 @@ from the naming convention of this sparsity structure in the
...
-
-            if !has_fractional_flow_outneighbors
-                id_out = outflow_id(graph, id)
-                has_index_out, idx_out = id_index(basin.node_id, id_out)
-
-                if has_index_out
-                    jac_prototype[idx_in, idx_out] = 1.0
+    for (idx_1, id) in enumerate(basin.node_id)
+        for id_neighbor in inoutflow_ids(graph, id)
+            for id_neighbor_neighbor in inoutflow_ids(graph, id_neighbor)
```
visr, line 35: At first I thought this may take a bit long since it is neighbors of neigbors, but with the way our graphs are structured it should be fine, and is quite simple to read.

Otherwise I was thinking of neighborhood functions but I don't think that's worth it.
https://juliagraphs.org/Graphs.jl/stable/algorithms/connectivity/#Graphs.neighborhood_dists-Union{Tuple{T},%20Tuple{U},%20Tuple{AbstractGraph{T},%20Integer,%20Any},%20Tuple{AbstractGraph{T},%20Integer,%20Any,%20AbstractMatrix{U}}}%20where%20{U%3C:Real,%20T%3C:Integer}

core/src/sparsity.jl
```diff
@@ -11,124 +11,34 @@ from the naming convention of this sparsity structure in the
...
-    end
-    return nothing
-end
-
-"""
-If both the unique node upstream and the nodes down stream (or one node further
-if a fractional flow is in between) are basins, then the downstream basin depends
-on the upstream basin(s) and affect the Jacobian as many times as there are downstream basins
-Upstream basins always depend on themselves.
+Add nonzeros for basins connected to eachother via 1 node and possibly a fractional flow node
```
visr, line 26: This function is also responsible for putting nonzeros on the diagonal, basins depending on themselves, right? If so, good to make that explicit in the docstring.

visr (approved):
Nice code reduction!

PR #1357, opened 2024-04-04
No need to formulate flow for Terminal anymore

visr:
This was missed from #1300. We don't store vertical flows anymore. Doing this was the only reason for `formulate_flows!` for Terminal nodes.

What I don't fully understand is why tests did not fail on this. Any test model with a Terminal should have failed since the 3-arg `add_flow!(graph, id, -q)` it called was removed in #1300. This PR also removes the 4-arg version that is no longer used.

### DIFF
--- a/core/src/graph.jl
+++ b/core/src/graph.jl
@@ -180,15 +180,6 @@ function set_flow!(graph::MetaGraph, id_src::NodeID, id_dst::NodeID, q::Number):
 end
 
 """
-Add the given flow q to the existing flow over the edge between the given nodes.
-"""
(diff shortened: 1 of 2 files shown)


SouthEndMusic (approved):
I had a look at why the tests don't fail as you mentioned. It is because the way the test models with terminal nodes in them are initialized (without any data other than the node location), the `Terminal \ static` table is never initialized and thus never written, and so the `terminal` object is empty which is why the problematic code was never run. This would still have thrown an error for models in which the `Terminal \ static` table does exist.

Issue #1358, opened 2024-04-04
Add result state file


visr:
If a user wants to continue a simulation at time t based on the results of a previous simulation that ended at time t, they need to use a `Basin / state` table with the final storages from the previous simulation. We should add an output file `basin_state.arrow` that contains this information.

This can then also be used in Delft-FEWS for this purpose.

If PidControl nodes are also present they should also carry over state, but this could be added later, since we don't support setting this state as input either at the moment.

gijsber:
Hi @visr , if this is implemented we should document it. such that externals (such as Micha (Hydrologic)) can use it

PR #1359, opened 2024-04-04
Don't add the endtime to basin.arrow results

visr:
This reverts a part of 4ff9987710fac6642d723d97bceb574517b09f24

On Tuesday we discussed the `basin.arrow` with @SouthEndMusic and @Huite. A few issues/PRs will come out of this, but this already removes the endtime, such that we don't have `missing` vertical fluxes in `basin.arrow` anymore. The downside is that the final storage is not currently available, but we have #1358 for that, and plan to add `dS/dt (m3/s)` as well to be able to compute it.

### DIFF
--- a/docs/core/usage.qmd
+++ b/docs/core/usage.qmd
@@ -705,10 +705,11 @@ derivative       | Float64  | -        | -
 
 ## Basin - `basin.arrow`
 
-The basin table contains:
+The Basin table contains:
 
-- results of the storage and level of each basin, which are instantaneous values;
-- results of the vertical fluxes on each basin, which are mean values over the `saveat` intervals. In the time column the start of the period is indicated.
+- Results of the storage and level of each basin, which are instantaneous values;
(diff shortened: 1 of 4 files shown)


docs/core/usage.qmd
```diff
@@ -708,20 +708,20 @@ derivative       | Float64  | -        | -
 The basin table contains:
 
 - results of the storage and level of each basin, which are instantaneous values;
-- results of the vertical fluxes on each basin, which are mean values over the `saveat` intervals. In the time column the start of the period is indicated. This means that for the final time in the table the mean vertical fluxes are not computed, and thus are `missing`.
+- results of the vertical fluxes on each basin, which are mean values over the `saveat` intervals. In the time column the start of the period is indicated.
```
SouthEndMusic, line 711: Maybe also explicitly indicate here that the final state of the model is absent?

SouthEndMusic (approved):

PR #1360, opened 2024-04-05
Add `Model.to_crs` method in order to reproject

### LINKED ISSUE #1354: Add `Model.to_crs` method in order to reproject
https://github.com/Deltares/Ribasim/pull/1339 introduces `Model.set_crs` which sets CRS.
This is a follow-up issue that allows to reproject CRS with `Model.to_crs`

Hofer-Julian:
Fixes #1354

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -161,6 +161,7 @@ class Model(FileModel):
         return fn
 
     def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
+        # Set CRS of the tables to the CRS stored in the Model object
         self.set_crs(self.crs)
         db_path = directory / input_dir / "database.gpkg"
         db_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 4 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -182,14 +182,21 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
...
     def set_crs(self, crs: str) -> None:
-        self.crs = crs
-        self.edge.df = self.edge.df.set_crs(crs)
+        self._apply_spatial_function("set_crs", crs)
+
+    def to_crs(self, crs: str) -> None:
+        self.set_crs(self.crs)
+        self._apply_spatial_function("to_crs", crs)
+
+    def _apply_spatial_function(self, function_name: str, crs: str) -> None:
```
evetion, line 191:
```suggestion
    def _apply_crs_function(self, function_name: str, crs: str) -> None:
        """Apply `function_name`, with `crs` as the first and only argument to all spatial tables."""
```

python/ribasim/ribasim/model.py
```diff
@@ -182,14 +182,21 @@ def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
             sub._save(directory, input_dir)
 
     def set_crs(self, crs: str) -> None:
-        self.crs = crs
-        self.edge.df = self.edge.df.set_crs(crs)
+        self._apply_spatial_function("set_crs", crs)
+
+    def to_crs(self, crs: str) -> None:
+        self.set_crs(self.crs)
```
evetion, line 188: This should probably be removed, or at least be called *after* to_crs, if that doesn't set the crs of all geodataframes.

Now this would change any previous crs to this crs, and then ask to project to that crs, resulting in a nothing/identity transform?

evetion (changes requested):
Surprisingly little code. I dislike getattr stuff a bit, but not enough to block it.

python/ribasim/ribasim/model.py
Hofer-Julian: Done

python/ribasim/ribasim/model.py
Hofer-Julian: As discussed, there was some confusion about what this line does. 
I've added a comment and will open a tech-dept issue.

Hofer-Julian:
> I dislike getattr stuff a bit, but not enough to block it.

I am not attached to it. Happy to hear alternative suggestions @evetion

python/ribasim_testmodels/ribasim_testmodels/bucket.py
visr, line 1: How about you change the location of the existing bucket model instead? Then we don't have to maintain another test model for only this purpose.

Hofer-Julian: Done!

Issue #1271, opened 2024-03-15
Tooling for model diagnostic checks


gijsber:
**What**
Diagnostics tooling to better understand model performance bottlenecks
suggestions by Daniel Tollenaar:
•	Balance-error per basin: abs(Q * dt) - (dV)) / Vt-1, wherej Q is the nett in-/outflow
•	Volume-Courant (Cv) per basin: Q * dt / V -> Cv, wherej Q is the nett in-/outflow, Cv > 1 always creates issues (when using epxlicit solvers)
•	Relative-discharge change per edge: min([Qt+1, Qt]) / max([Qt+1, Qt]) -> large to unrealistic changes in flowrate over time highlight a larger risk for numerical instability

**Why**
Modellers would appreciate assistance and tooling to help in analysing and improving their model behaviour

**How**

A clear and concise description of how a technical implementation in code should look like.

**Additional context**

visr:
Related to #452

Issue #642, opened 2023-10-02
Add water balance output


visr:
The level and flow output we have right now are samples in a continuous signal.

To get a good sense of the total water balance, and to check if it closes, we need to integrate these over time, to get output in m3 over the past timestep that we can sum.

I think we should implement this using the [`IntegratingCallback`](https://docs.sciml.ai/DiffEqCallbacks/stable/integrating/).

Also we should clean up some leftover code from similar old functionality:
https://github.com/Deltares/Ribasim/blob/v2023.09.0/core/src/bmi.jl#L200
https://github.com/Deltares/Ribasim/blob/v2023.09.0/core/src/solve.jl#L1194

**EDIT**: This issue is now only about water balance calculation and output. Mean flows are #935. This can be done separately, e.g. the water balance below was calculated using sampled flows, ignoring the error for that.

visr:
The `IntegratingCallback` doesn't seem to really match our use case, though perhaps [some of the code](https://github.com/SciML/DiffEqCallbacks.jl/blob/9d72b90b3d5490c3b22a9d10533106c1fe566b00/src/integrating.jl#L197-L212) could be used: the Gaussian quadrature that interpolates between the timesteps. The main issue seems to be that it seems to want to integrate over all parameters (`integral = allocate_zeros(integrator.p)` fails for us). Though in the end these are all `DiscreteCallback`, so we can build them ourselves.

#644 takes an important first step by making sure the flow output table is complete, by including the gains and losses from non-conservative nodes. Here is a first quick script calculating the water balance error over the basic_transient test model.

```julia
using Plots
using DataFrames, Dates
using Ribasim

toml_path = normpath(@__DIR__, "generated_testmodels/basic_transient/basic_transient.toml")
model = Ribasim.run(toml_path)
# use unique since some timesteps are saved multiple times
basin = unique(DataFrame(Ribasim.basin_table(model)), [:time, :node_id])
flow = DataFrame(Ribasim.flow_table(model))
time = unique(Ribasim.datetimes(model))

function water_balance_error(basin, flow, time)
    wbal_errors_m3s = Float64[]
    wbal_errors_m3 = Float64[]
    storages = Float64[]
    for (t1, t2) in zip(time[begin:(end - 1)], time[(begin + 1):end])
        # we have to calculate the storage difference over time ourselves
        dt = 0.001 * Dates.value(t2 - t1)

        s_t1 = sum(filter(:time => t -> t == t1, basin).storage)
[...]

PR #1367, opened 2024-04-08
Add remaining water balance terms to basin.arrow

visr:
Technically these terms can be calculated from flow.arrow and the storage. But since calculating the water balance for a Basin is such a common operation, we include them directly. This also makes it easier to locate water balance errors in time and space, by including the error term.

At some point we may also want to add the instantaneous inflow and outflow to the Basin struct, and update them in `formulate_du!`. This would allow us to reject steps with too large of a balance error, using one of the metrics mentioned in #1271. But for now this only adds them to the output via a callback.

### DIFF
--- a/core/src/write.jl
+++ b/core/src/write.jl
@@ -134,8 +134,8 @@ function basin_table(
 
     for i in 1:nrows
         storage_flow = storage_rate[i]
-        storage_increase = storage_flow > 0.0 ? storage_flow : 0.0
-        storage_decrease = storage_flow > 0.0 ? 0.0 : -storage_flow
+        storage_increase = max(storage_flow, 0.0)
+        storage_decrease = max(-storage_flow, 0.0)
 
         total_in = inflow_rate[i] + precipitation[i] + drainage[i] - storage_increase
         total_out = outflow_rate[i] + evaporation[i] + infiltration[i] - storage_decrease
--- a/core/src/write.jl
+++ b/core/src/write.jl
@@ -115,9 +115,9 @@ function basin_table(
     relative_error = zeros(nrows)
 
     idx_row = 0
-    for vec in saved.vertical_flux.saveval
+    for cvec in saved.vertical_flux.saveval
         for (precipitation_, evaporation_, drainage_, infiltration_) in
-            zip(vec.precipitation, vec.evaporation, vec.drainage, vec.infiltration)
+            zip(cvec.precipitation, cvec.evaporation, cvec.drainage, cvec.infiltration)
             idx_row += 1
             precipitation[idx_row] = precipitation_
             evaporation[idx_row] = evaporation_
--- a/core/src/write.jl
+++ b/core/src/write.jl
@@ -84,31 +84,35 @@ function basin_table(
     node_id::Vector{Int32},
     storage::Vector{Float64},
     level::Vector{Float64},
+    inflow_rate::Vector{Float64},
+    outflow_rate::Vector{Float64},
+    storage_rate::Vector{Float64},
     precipitation::Vector{Float64},
     evaporation::Vector{Float64},
     drainage::Vector{Float64},
     infiltration::Vector{Float64},
+    error::Vector{Float64},
 }
     (; saved) = model
-    (; vertical_flux) = saved
-
     # The last timestep is not included; there is no period over which to compute flows.
     data = get_storages_and_levels(model)
     storage = vec(data.storage[:, begin:(end - 1)])
     level = vec(data.level[:, begin:(end - 1)])
+    Δstorage = vec(diff(data.storage; dims = 2))
 
     nbasin = length(data.node_id)
     ntsteps = length(data.time) - 1
     nrows = nbasin * ntsteps
 
+    inflow_rate = FlatVector(saved.flow.saveval, :inflow)
+    outflow_rate = FlatVector(saved.flow.saveval, :outflow)
     precipitation = zeros(nrows)
     evaporation = zeros(nrows)
     drainage = zeros(nrows)
     infiltration = zeros(nrows)
 
     idx_row = 0
-
-    for vec in vertical_flux.saveval
+    for vec in saved.vertical_flux.saveval
         for (precipitation_, evaporation_, drainage_, infiltration_) in
             zip(vec.precipitation, vec.evaporation, vec.drainage, vec.infiltration)
             idx_row += 1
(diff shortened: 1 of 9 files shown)


visr:
@gijsber @DanielTollenaar does this proposal and the change in usage.qmd look ok to you?

core/src/write.jl
```diff
@@ -120,17 +124,27 @@ function basin_table(
     end
 
     time = repeat(data.time[begin:(end - 1)]; inner = nbasin)
+    Δtime_seconds = 0.001 * Dates.value.(diff(data.time))
+    Δtime = repeat(Δtime_seconds; inner = nbasin)
     node_id = repeat(Int32.(data.node_id); outer = ntsteps)
+    storage_rate = Δstorage ./ Δtime
+    error =
+        inflow_rate - outflow_rate - storage_rate + precipitation - evaporation + drainage -
```
SouthEndMusic, line 132: I was a bit confused by the sign of the storage rate, maybe write this as `(all flows except storage_rate) - storage_rate` with the brackets to emphasise which difference you are calculating

docs/core/usage.qmd
```diff
@@ -707,23 +707,31 @@ derivative       | Float64  | -        | -
...
-The initial condition is also written to the file.
+- Results of the storage and level of each Basin, which are instantaneous values;
+- Results of the fluxes on each Basin, which are mean values over the `saveat` intervals.
+  In the time column the start of the period is indicated.
+- The initial condition is written to the file, but the final state is not.
+  It will be placed in a separate output state file in the future.
+- The `inflow_rate` and `outflow_rate` are the sum of the flows into and out of the Basin respectively.
+  The actual flows determine in which term they are counted, not the edge direction.
+- The `storage_rate` is flow that adds to the storage in the Basin, increasing the water level.
+- The `error` is the result of adding all terms of the water balance; `inflow_rate` - `outflow_rate` - `storage_rate` + `precipitation` - `evaporation` + `drainage` - `infiltration`.
```
SouthEndMusic, line 718: Same here

SouthEndMusic (approved):
Good stuff

SouthEndMusic:
> At some point we may also want to add the instantaneous inflow and outflow to the Basin struct, and update them in `formulate_du!`. This would allow us to reject steps with too large of a balance error, using one of the metrics mentioned in #1271. But for now this only adds them to the output via a callback.

Would it be an idea to integrate the absolute water balance error per basin as a state of the system of differential equations, and throw an error in the same way we do for negative storages if this gets too big? Now that I think of it, why don't we integrate the vertical fluxes this way, just let the solver do it?

DanielTollenaar:
Good stuff! What about considering a `relative_error` as well?

How I compute these in Python (think we have to think about something to avoid replace infinite (balance error devided by 0 in- or outflow_rate):
```
# possible positive error (if error not 0): inflow is larger than outflow
positive = balance_error_df["inflow_rate"] > balance_error_df["outflow_rate"]
 
# if positive error, we devide the error by the outflow_rate (x% too much inflow compared to outflow, or x% too little outflow compared to inflow)
balance_error_df.loc[positive, ["relative_error"]] = (
    balance_error_df[positive]["error"] / balance_error_df[positive]["outflow_rate"]
) * 100
 
# if negative error, we devide the error by the inflow_rate (x% too much outflow compared to inflow, or x% too little inflow compared to outflow)
balance_error_df.loc[~positive, ["relative_error"]] = (
    balance_error_df[~positive]["error"] / balance_error_df[~positive]["inflow_rate"]
) * 100
```

@SouthEndMusic, guess relative_error is something you may use in the solver?

visr:
In 12c77be4fce0838799c72f2be94433fc8db368a0 I renamed `error` to `balance_error` and added `relative_error`, as a fraction, not a percentage.

@DanielTollenaar I added your `relative_error`, though not sure if this is a good general metric. I agree that it is useful for Basins that are in a river with both inflow and outflow. But now if either is 0 then the `relative_error` is 0. Why this particular metric? Perhaps a fraction discrepancy of all inflows (horizontal and vertical) compared to all outflows is more useful. I think this is more similar to the MODFLOW output as posted at the bottom of https://github.com/Deltares/Ribasim/issues/642#issuecomment-1755654305.

@SouthEndMusic I think it is a good idea to let the integrator integrate the flows as well. Do you want to create an issue for it? We should probably do a small test first to get a feeling for it.

DanielTollenaar:
> In [12c77be](https://github.com/Deltares/Ribasim/commit/12c77be4fce0838799c72f2be94433fc8db368a0) I renamed `error` to `balance_error` and added `relative_error`, as a fraction, not a percentage.
> 
> @DanielTollenaar I added your `relative_error`, though not sure if this is a good general metric. I agree that it is useful for Basins that are in a river with both inflow and outflow. But now if either is 0 then the `relative_error` is 0. Why this particular metric? Perhaps a fraction discrepancy of all inflows (horizontal and vertical) compared to all outflows is more useful. I think this is more similar to the MODFLOW output as posted at the bottom of [#642 (comment)](https://github.com/Deltares/Ribasim/issues/642#issuecomment-1755654305).
> 
> @SouthEndMusic I think it is a good idea to let the integrator integrate the flows as well. Do you want to create an issue for it? We should probably do a small test first to get a feeling for it.

@metric all inflows; agreed, that's better.

core/src/write.jl
```diff
@@ -84,31 +84,38 @@ function basin_table(
...
     evaporation = zeros(nrows)
     drainage = zeros(nrows)
     infiltration = zeros(nrows)
+    balance_error = zeros(nrows)
+    relative_error = zeros(nrows)
 
     idx_row = 0
-
-    for vec in vertical_flux.saveval
+    for vec in saved.vertical_flux.saveval
```
SouthEndMusic, line 118: Maybe not use a variable name that is already the name of a function

core/src/write.jl
```diff
@@ -120,17 +127,39 @@ function basin_table(
...
 
     time = repeat(data.time[begin:(end - 1)]; inner = nbasin)
+    Δtime_seconds = 0.001 * Dates.value.(diff(data.time))
+    Δtime = repeat(Δtime_seconds; inner = nbasin)
     node_id = repeat(Int32.(data.node_id); outer = ntsteps)
+    storage_rate = Δstorage ./ Δtime
+
+    for i in 1:nrows
+        storage_flow = storage_rate[i]
+        storage_increase = storage_flow > 0.0 ? storage_flow : 0.0
```
SouthEndMusic, line 137:
```suggestion
        storage_increase = max(storage_flow, 0.0)
```

core/src/write.jl
```diff
@@ -120,17 +127,39 @@ function basin_table(
...
+        storage_decrease = storage_flow > 0.0 ? 0.0 : -storage_flow
```
SouthEndMusic, line 138:
```suggestion
        storage_decrease = min(-storage_flow, 0.0)
```

SouthEndMusic (approved):
Added some final comments

Issue #1346, opened 2024-04-03
Control based on difference condition


gijsber:
**What**
Allow control conditions based on an algebratic combination of observations (e.g. head difference or some statistical relation).

**Why**
In some situations, a head difference or othe algebraic/statistical relation could determine a control action (like opening/closing a gate).

**How**
_As paraphrased from a mail from Bart to Edwin and Peter._

We could extend the `DiscreteControl / condition` table with 2 columns:
- A condition identifier (integer? string?);
- A weight for each variable.

A compound condition is then built up by multiple rows with the same condition identifier, and specifies a linear combination of variables. Note thus that products of variables are not allowed (this would give more complications because then the order of operations starts to matter), but I imagine that the use case of that is really rare.

In terms of validation: sets of rows with the same condition identifier should have the same `greater_than` value.

PR #1371, opened 2024-04-10
Support conditions on linear combinations of variables for `DiscreteControl`

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1346.


core/src/callback.jl
```diff
@@ -272,16 +277,20 @@ function discrete_control_affect_upcrossing!(integrator, condition_idx)
...
+    listen_node_ids = discrete_control.listen_node_id[condition_idx]
+    is_basin =
+        length(listen_node_ids) == 1 ? id_index(basin.node_id, only(listen_node_ids))[1] :
+        false
+
     # NOTE: The above no longer works when listen feature ids can be something other than node ids
     # I think the more durable option is to give all possible condition types a different variable string,
     # e.g. basin.level and level_boundary.level
-    if variable[condition_idx] == "level" && control_state_change && is_basin
+    if variable[condition_idx][1] == "level" && control_state_change && is_basin
```
SouthEndMusic, line 288: This is a hacky mess. Dealing with `greater_than` 'touches' instead of crossings (as happens in [this example](https://deltares.github.io/Ribasim/python/examples.html#model-with-discrete-control)) already gave quite some complexity, but for compound variables the maintenance load is even worse. Even more motivation to switch to discrete condition checking

core/test/control_test.jl
```diff
@@ -54,7 +54,7 @@ end
     p = model.integrator.p
     (; discrete_control, flow_boundary) = p
 
-    Δt = discrete_control.look_ahead[1]
+    Δt = discrete_control.look_ahead[1][1]
```
evetion, line 57: Maybe better to use the name here (assuming namedtuples?)

core/src/validation.jl
```diff
@@ -552,29 +559,32 @@ function valid_discrete_control(p::Parameters, config::Config)::Bool
...
-                errors = true
-                @error "Look ahead supplied for non-timeseries listen variable '$var' from listen node $node_id."
-            else
-                if Δt < 0
+    for (look_aheads, variables, listen_node_ids) in
+        zip(look_ahead, variable, listen_node_id)
+        for (Δt, var, node_id) in zip(look_aheads, variables, listen_node_ids)
+            if !iszero(Δt)
+                node_type = node_id.type
+                # TODO: If more transient listen variables must be supported, this validation must be more specific
```
evetion, line 567: Did you make an issue?

evetion (changes requested):
Nice PR! 👍🏻  I wonder if we should change the input tables slightly (but it would be breaking).

Specifically, compoundvariable is now mostly a repeat of condition, with the change of only a single field (added weight, dropped greater_than). Furthermore, condition is changed that type can take "compound", while the node_id is ignored.

I wonder whether we should split condition into condition and variable, where variable can be compound if a id occurs multiple times. listen-*, variable and look ahead are then all dropped/moved.

core/test/control_test.jl
SouthEndMusic: It's not a named tuple, it's a look ahead per compound variable per singular variable

SouthEndMusic:
Some clarifications for the latest commits:

- We're now back at allowing multiple compound variables on the same `DiscreteControl` node using the `compound_variable_id`. This means that we're back in the situation where all the control of a whole model can be done with a single `DiscreteControl` node, so it's up to the modeller to do this in a sensible way;
- We also allow multiple `greater_than` values on a single compound variable, which all define their own condition. I did not want to copy the compound variable data for every condition, so I now store all `greater_than` values per compound variable in `discrete_control.greater_than`. This does require some translation between the `condition_idx` (which is for a single flat dense vector) and the compound variable and `greater_than` indices, see `get_discrete_control_indices`.

docs/core/usage.qmd
```diff
@@ -625,18 +626,35 @@ node_id               | Int32   | -            | sorted
...
+listen_node_type     | String   | -       | known node type
+listen_node_id       | Int32    | -       | sorted per node_id
+variable             | String   | -       | must be "level" or "flow_rate", sorted per listen_node_id
+weight               | Float64  | -       | (optional, default 1.0)
+look_ahead           | Float64  | $s$     | Only on transient boundary conditions, non-negative (optional, default 0.0).
+
 ## DiscreteControl / condition
 
-The condition schema defines conditions of the form 'the discrete_control node with this node id listens to whether the given variable of the node with the given listen feature id is grater than the given value'. If the condition variable comes from a time-series, a look ahead $\Delta t$ can be supplied.
+The condition schema defines conditions of the form 'the `discrete_control` node with this `node_id`listens to whether the variable given by the `node_id` and `compound_variable_id` is greater than `greater_than`'. If the condition variable comes from a time-series, a look ahead $\Delta t$ can be supplied. Multiple conditions with different `greater_than` values can be defined on the same compound_variable.
```
evetion, line 651: Sentence about look ahead can be removed.

docs/core/usage.qmd
```diff
@@ -625,18 +626,35 @@ node_id               | Int32   | -            | sorted
...
+$$
+\text{weight}_1 * \text{variable}_1 + \text{weight}_2 * \text{variable}_2 + \ldots,
+$$
+
+which can be for instance an average or a difference of variables. If a variable comes from a time-series, a look ahead $\Delta t$ can be supplied.
+
+column               | type     | unit    | restriction
+-------------------- | -------- | ------- | -----------
+node_id              | Int32    | -       | sorted
+compound_variable_id | String   | -       | sorted per node_id
```
evetion, line 642: Int32

core/src/util.jl
```diff
@@ -709,3 +709,19 @@ function get_influx(basin::Basin, basin_idx::Int)::Float64
     return precipitation[basin_idx] - evaporation[basin_idx] + drainage[basin_idx] -
            infiltration[basin_idx]
 end
+
+function get_discrete_control_indices(discrete_control::DiscreteControl, condition_idx::Int)
+    (; greater_than) = discrete_control
+    condition_idx_now = 1
+
+    for (compound_variable_idx, vec) in enumerate(greater_than)
```
evetion, line 717: This method can use a docstring and/or some inline documentation on why/what is happening.

docs/core/usage.qmd
```diff
@@ -625,18 +626,35 @@ node_id               | Int32   | -            | sorted
...
-node_id           | Int32    | -       | sorted
-listen_node_type  | String   | -       | known node type
-listen_node_id    | Int32    | -       | sorted per node_id
-variable          | String   | -       | must be "level" or "flow_rate", sorted per listen_node_id
-greater_than      | Float64  | various | sorted per variable
-look_ahead        | Float64  | $s$     | Only on transient boundary conditions, non-negative (optional, default 0)
+column               | type     | unit    | restriction
+-------------------- | -------- | ------- | -----------
+node_id              | Int32    | -       | sorted
+compound_variable_id | Int32    | -       | -
```
evetion, line 656: Why did we suport multiple compound variables per condition? Is there a specific use case? If so, let's put it in the PR description, so it ends up in the git commit history.

python/ribasim/ribasim/model.py
```diff
@@ -321,11 +321,12 @@ def plot_control_listen(self, ax):
...
         # Listen edges from DiscreteControl
-        condition = self.discrete_control.condition.df
-        if condition is not None:
-            to_add = condition[
+        df_variable = self.discrete_control.variable.df
+        if df_variable is not None:
+            to_add = df_variable[
                 ["node_id", "listen_node_id", "listen_node_type"]
             ].drop_duplicates()
+            to_add = to_add[to_add["listen_node_type"] != "compound"]
```
evetion, line 329: I don't think this line does anything anymore.

docs/core/usage.qmd
evetion: Hmm, I believe it follows logically from these tables right? If you can have multiple variables, you have to indicate which are compound, and which are not.

evetion (changes requested):

docs/core/usage.qmd
SouthEndMusic: We support multiple conditions per `DiscreteControl` node, which do not need to be on the same variable (and now all variables on which a condition is placed is called a compound variable, even if it is effectively a singular variable). Logic like 'do X if the level at basin 1 is at least 10 and the level at basin 2 is at least 10' is quite sensible.

python/ribasim/ribasim/model.py
SouthEndMusic: Excellent point

core/src/util.jl
SouthEndMusic: This function is already removed in the follow up PR https://github.com/Deltares/Ribasim/pull/1393.

core/src/validation.jl
SouthEndMusic: I removed the comment. If it becomes relevant it will be evident

evetion (approved):

Issue #1372, opened 2024-04-10
PID control example plot is broken


SouthEndMusic:
The plot now looks like this:

![cell-44-output-1](https://github.com/Deltares/Ribasim/assets/74617371/2cf231d1-b4f8-4722-817a-2cf4c4cfc3af)

It's probably not a problem in the core because when I plot directly from the basin table in Julia I get the correct plot:

![correct](https://github.com/Deltares/Ribasim/assets/74617371/b98e13c1-ccd0-4b9e-aae1-b61e73cbb64e)

PR #1373, opened 2024-04-10
Fix PID example

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1372. 

There were some mistakes in the schematization.


Jingru923 (approved):
Ran it locally and was successful. Code looks good to me.

PR #1374, opened 2024-04-11
Fix "add results to xugrid"

visr:
Small bug I didn't notice in #1369 because the basic example has the same number of nodes and edges.


evetion (approved):
Love it when bugs are this easy

Issue #1378, opened 2024-04-11
Add comprehensible error when starttime > endtime


SouthEndMusic:
The core now gives the following cryptic error:

`ERROR: AssertionError: integrator.tdir == sign(Δt)`

PR #1379, opened 2024-04-11
Add error starttime > endtime

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1378.

### DIFF
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -44,6 +44,11 @@ function Model(config::Config)::Model
         TimerOutputs.enable_debug_timings(Ribasim)  # causes recompilation (!)
     end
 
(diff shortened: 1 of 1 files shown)


core/src/model.jl
```diff
@@ -44,6 +44,11 @@ function Model(config::Config)::Model
         TimerOutputs.enable_debug_timings(Ribasim)  # causes recompilation (!)
     end
 
+    t_end = seconds_since(config.endtime, config.starttime)
+    if t_end < 0
+        error("starttime > endtime")
+    end
```
visr, line 50:
```suggestion
    if t_end <= 0
        error("Model starttime is not before endtime.")
    end
```

visr (approved):
One suggestion.

Issue #1383, opened 2024-04-15
HiGHS dimension mismatch


visr:
*What

HiGHS prints "dimension mismatch". @SouthEndMusic ran into this before. It comes from inside HiGHS:

https://github.com/ERGO-Code/HiGHS/blob/13363c9f1252b015cf6527132eb9cf8f4b5bf020/src/qpsolver/factor.hpp#L181

But it doesn't seem to be a failure. @SouthEndMusic is this related to #1366?

*Why
Error message can be confusing for modeller

SouthEndMusic:
The `dimension mismatch` thing doesn't fail any tests right? And indeed the failing test just needs an increased tolerance. I saw that on main the Julia tests fail on Windows but not on Linux

visr:
Yeah you are right, I already updated this issue to split out the error that should be fixed by https://github.com/Deltares/Ribasim/pull/1384 and make this just about the `dimension mismatch` print statement.

SouthEndMusic:
> But it doesn't seem to be a failure. @SouthEndMusic is this related to #1366?

Maybe there's still some redundant things left in the problems after removing most of the code for absolute value calculations, I'll have a look.

visr:
Possibly related, here something seems to be stuck in "Allocation with main network optimization problem".

```
  Worker 4531:  [4531] signal (15): Terminated
in expression starting at none:1


No Captured Logs for test item "Allocation with main network optimization problem" at test/allocation_test.jl:194 on worker 4528
┌ Error: Worker(pid=4528, terminated=true, termsignal=9) timed out running test item "Allocation with main network optimization problem" after 1800 seconds. Recording test error.
└ @ ReTestItems ~/.julia/packages/ReTestItems/ELM7z/src/ReTestItems.jl:557
```
From https://github.com/Deltares/Ribasim/actions/runs/8685702994/job/23815699761?pr=1384#step:6:515

PR #1384, opened 2024-04-15
Add tolerance to subnetwork demands test

visr:
Fixes CI on main. See also #1383.


SouthEndMusic (approved):
Looks good, but better wait until all checks are green 😋

Issue #1121, opened 2024-02-14
Use averaged source flows and basin forcings in allocation


SouthEndMusic:
**What**
https://github.com/Deltares/Ribasim/pull/1082 incorporates supply/demands from basins in allocation. However, it does this based on instantaneous horizontal and vertical fluxes. We want to transition to averaged values over the previous allocation timestep.

**How**
See https://github.com/Deltares/Ribasim/issues/935.

PR #1390, opened 2024-04-15
Run allocation first after BMI stop

visr:
This allows allocation over period `(t, t + dt)` to use variables set over BMI at time `t`.

We no longer run allocation as a callback, but call it ourselves, such that we can control that BMI runs allocation first before running the physical layer.

### DIFF
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -206,7 +206,7 @@ function SciMLBase.step!(model::Model, dt::Float64)::Model
     if ntimes > 0 && round(ntimes) ≈ ntimes
         update_allocation!(integrator)
     end
-    step!(integrator, dt; stop_at_tdt = true)
+    step!(integrator, dt, true)
     return model
 end
 
@@ -222,9 +222,14 @@ function SciMLBase.solve!(model::Model)::Model
         for _ in TimeChoiceIterator(integrator, times)
             update_allocation!(integrator)
         end
+
+        # https://github.com/SciML/SciMLBase.jl/issues/669
+        if integrator.sol.retcode != ReturnCode.Default
+            return integrator.sol
+        end
+        integrator.sol = SciMLBase.solution_new_retcode(integrator.sol, ReturnCode.Success)
     else
         solve!(integrator)
     end
-    check_error!(integrator)
     return model
 end
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -47,11 +47,6 @@ function Model(config::Config)::Model
         TimerOutputs.enable_debug_timings(Ribasim)  # causes recompilation (!)
     end
 
-    t_end = seconds_since(config.endtime, config.starttime)
-    if t_end <= 0
-        error("Model starttime is not before endtime.")
-    end
-
     # All data from the database that we need during runtime is copied into memory,
     # so we can directly close it again.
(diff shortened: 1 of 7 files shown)


visr:
Some pretty large allocation test differences suggest we don't call allocation correctly yet.

```
No Captured Logs for test item "Allocation level control" at test\allocation_test.jl:308 on worker 26268
Error in testset "Allocation level control" on worker 26268:
Test Failed at D:\Ribasim\core\test\allocation_test.jl:334
  Expression: ≈(storage[stage_1], u_stage_1.(t[stage_1]), rtol = 0.0001)
   Evaluated: [500.0, 672.7963463605041] ≈ [500.0, 543.2] (rtol=0.0001)
```

SouthEndMusic:
This might have to do with the fact that allocation uses the flows from the last `water_balance!` call, which depends (maybe among other things) on the used solver and might not correspond to the right state of the Model. To make this consistent, maybe always do a `water_balance!` call before allocation, which also would help with performing allocation at `t=0`.
Alternatively, we should first pick up the issue about using average flows and levels as input for allocation (https://github.com/Deltares/Ribasim/issues/1121, but this issue does not yet take basin levels into account).

visr:
It actually ran allocation on the wrong times, that is fixed now.

core/src/validation.jl
```diff
@@ -146,6 +146,17 @@ function sorted_table!(
     return table
 end
 
+function valid_config(config::Config)::Bool
```
SouthEndMusic, line 149: Now that you have this, can it be removed below?

https://github.com/Deltares/Ribasim/blob/e10f657bb56434881bfd3382ddce921b35cff9f5/core/src/model.jl#L50-L53

core/src/model.jl
```diff
@@ -188,10 +191,51 @@ function SciMLBase.successful_retcode(model::Model)::Bool
...
+"""
+function SciMLBase.step!(model::Model, dt::Float64)::Model
+    (; config, integrator) = model
+    (; t) = integrator
+    # If we are at an allocation time, run allocation before the next physical
+    # layer timestep. This allows allocation over period (t, t + dt) to use variables
+    # set over BMI at time t before calling this function.
+    # Also, don't run allocation at t = 0 since there are no flows yet (#1389).
+    ntimes = t / config.allocation.timestep
+    if ntimes > 0 && round(ntimes) ≈ ntimes
```
SouthEndMusic, line 206: I'd say the more common way to do this is

```julia
if t % config.allocation.timestep ≈ 0.0
   ... 
end
```

and the other check could just be `t > 0` if you actually need that.

SouthEndMusic (approved):
Looks good, left some small comments

core/src/model.jl
SouthEndMusic: Although the comparison tolerances work out differently

core/src/validation.jl
visr: Right, I forgot, thanks.

core/src/model.jl
visr: That would only work if the t is right after, but not right before, right?
```
julia> prevfloat(3.0) % 0.5
0.49999999999999956
```

SouthEndMusic:
> Oops

lol

PR #1392, opened 2024-04-16
Schemas: Only set default value for `node_id` not for all integer columns

Hofer-Julian:


Hofer-Julian:
Discussed with @visr and this PR doesn't make too much sense.

Will open a tech-dept issue instead.

PR #1396, opened 2024-04-16
Stop pinning pandas

### LINKED ISSUE #984: Wiki page of Test models has unwanted text output
In the page, there were supposed to be only brief introductions and graphs of the test models. Since we updated Pandas, many text messages have been generated.

```
/home/runner/work/Ribasim/Ribasim/.pixi/env/lib/python3.11/site-packages/pandera/backends/pandas/array.py:336: FutureWarning:

A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.
```

As this text is not wanted, we should go to the source (Ribasim Python) where these warning texts are from and fix the problem.

Hofer-Julian:
Fixes #984

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -218,7 +218,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/openssl-3.2.1-hd590300_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/orc-2.0.0-h1e5e2c1_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-24.0-pyhd8ed1ab_0.conda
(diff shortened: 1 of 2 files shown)


Jingru923:
I see that now we are using the newest version of Pandas again. 

I ran quarto-review locally and saw no unwanted text output. This changes looks good to me

Jingru923 (approved):

Issue #1388, opened 2024-04-15
Use same network for allocation and flow


visr:
Currently allocation is using a derived network that only takes in features that matter for allocation. That means we need to map back and forth. The allocation code will be more maintainable if we stick to a 1:1 mapping. This will also help with implementing #714.

For Terminals we can add a 0 flow constraint instead of pruning the branch.

PR #1399, opened 2024-04-17
Use the same network for allocation as for the physical layer

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1388.

This also fixes a bug I found: In allocation the `max_flow_rate` of a node connecting the main network to a subnetwork was not taken into account


core/src/allocation_init.jl
```diff
@@ -460,27 +243,29 @@ Only finite capacities get a constraint.
...
         if !isinf(c) && edge ∉ main_network_source_edges
             push!(edge_ids_finite_capacity, edge)
         end
     end
+
     problem[:capacity] = JuMP.@constraint(
         problem,
         [edge = edge_ids_finite_capacity],
-        F[edge] <= capacity[edge...],
+        F[edge] ≤ capacity[edge...],
```
visr, line 268: Do you use a font with coding ligature support? To me a `<` followed by a `=` already looks like one symbol. I use Cascadia Code, see https://github.com/microsoft/cascadia-code?tab=readme-ov-file#font-features.

If you use that perhaps we can just keep the ASCII version?

core/src/allocation_init.jl
```diff
@@ -666,37 +341,88 @@ The constraint indices are Basin node IDs.
...
+
+    edges_allocation = only(F.axes)
+
+    for node_id in node_ids
+
+        # No flow conservation constraint on sources/sinks
+        is_source_sink = node_id.type in
+        [NodeType.FlowBoundary, NodeType.LevelBoundary, NodeType.UserDemand]
+
+        # No flow conservation on nodes with fractional flow outneighbors
```
visr, line 368:
```suggestion
        # No flow conservation on nodes with FractionalFlow outneighbors
```

core/src/allocation_init.jl
```diff
@@ -666,37 +341,88 @@ The constraint indices are Basin node IDs.
...
+            if edge_in in edges_allocation
+                push!(inflows_node, F[edge_in])
+            end
+            edge_out = (node_id, neighbor_id)
+            if edge_out in edges_allocation
+                push!(outflows_node, F[edge_out])
+            end
+        end
+
+        # If the node is a basin, add basin in- and outflow
```
visr, line 395:
```suggestion
        # If the node is a Basin, add Basin in- and outflow
```

visr (approved):
Such a nice simplification!

core/src/allocation_optim.jl
```diff
@@ -829,63 +827,62 @@ function save_allocation_flows!(
...
+            edge_metadata = graph[edge_1...]
+        else
+            edge_1_reverse = reverse(edge_1)
+            flow_rate -= JuMP.value(F[edge_1_reverse])
+            sign_2 = 1.0
+            edge_metadata = graph[edge_1_reverse...]
+        end
+
+        # Check whether the next edge is the current one reversed
+        if edge_2 == reverse(edge_1)
```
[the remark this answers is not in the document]
SouthEndMusic: That's an excellent question. I actually caught a bug looking into this. If an edge is bidirectional, both allocation edges are created consecutively:

https://github.com/Deltares/Ribasim/blob/dccb31eec6ff4592fa3bfd54eaf48ff98e5deee8/core/src/allocation_init.jl#L64-L73

Also, the underlying structure of `capacity` as a `SparseAxisArray` is an `OrderedDict`, so I think this is safe?

PR #1400, opened 2024-04-17
`pixi run --environment=dev install` to `pixi run install`

### LINKED ISSUE #1324: Stop explicitly requiring `--environment=dev` when running pixi install tasks
When installing packages with `pixi run install`, Python packages are installed in the `default` environment instead of `dev`.
Because of that, we recommend running `pixi run --environment=dev install` at the moment.
https://github.com/Deltares/Ribasim/blob/0c02f9a356fd7e9db3bec53b598c37d5337ee270/docs/contribute/index.qmd#L38

As soon as pixi supports editable installs, this will not be necessary anymore and we can remove `--environment=dev` again.

Steps:

1. Add [editable dependency](https://pixi.sh/latest/reference/configuration/#path) and remove corresponding `install-*` tasks
2. Change all instances of  `pixi run --environment=dev install` to  `pixi run install`

Jingru923:
Fixes #1324


pixi.toml
```diff
@@ -13,15 +13,19 @@ documentation = "https://deltares.github.io/Ribasim/"
 repository = "https://github.com/Deltares/Ribasim"
 
 [tasks]
+install-ribasim-python ={ path = "python/ribasim", editable = true }
+install-ribasim-api ={ path = "python/ribasim_api", editable = true }
+install-ribasim-testmodels ={ path = "python/ribasim_testmodels", editable = true }
```
visr, line 18: With editable installs the installation is no longer done with a task, but automatically if listed under pypi dependencies. See https://github.com/Deltares/Ribasim-NL/pull/75 for an example.

Jingru923:
Small update: The pixi version is updated for the windows agents. Now we are waiting for @evetion to update it for the Linux ones

PR #1393, opened 2024-04-16
Discrete control with discrete callback

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/455, fixes https://github.com/Deltares/Ribasim/issues/495. The latter is made redundant by this PR since U/D truth values are removed.

PR #1403, opened 2024-04-17
Remove dutch_waterways model

SouthEndMusic:
As discussed with @gijsber, this model is not worth maintaining and it broke with https://github.com/Deltares/Ribasim/pull/1393.


Jingru923 (approved):

visr:
@Jingru923 was this the only part that was "broken" about 2024.7.0? If so, I wouldn't say the release itself is bad, and there is no need for a quick 2024.7.1.

Hofer-Julian:
> @Jingru923 was this the only part that was "broken" about 2024.7.0? If so, I wouldn't say the release itself is bad, and there is no need for a quick 2024.7.1.

I would have to be done manual then.
The TeamCity pipeline is set up to refuse to publish if tests are broken.

visr:
If that's not too much extra work, a manual action would be great. Should we update the release docs to manually trigger TeamCity builds on the version bump PRs like #1402? Or can we trigger it automatically if the branch name begins with `release`?

Hofer-Julian:
> If that's not too much extra work, a manual action would be great.

I will take a look later today

> Should we update the release docs to manually trigger TeamCity builds on the version bump PRs like https://github.com/Deltares/Ribasim/pull/1402? Or can we trigger it automatically if the branch name begins with release?

You mean to avoid situations like this?
Sure we can do this. But it would mean we have to open a PR, wait for an hour to see if it's green, make the release and wait another hour for it to actually publish. Sounds a bit annoying to me.

visr:
Thanks, that'd be great.

Indeed it is a bit annoying, but if there is any time that TeamCity should be green, it is when doing a release. This time it caught something harmless but the test surface is larger on TeamCity, and cleaning up broken releases is also not fun. Otherwise we'd need to only do releases first thing in the morning after a green nightly.

Hofer-Julian:
Fair enough, wanna open an issue for that?

Hofer-Julian:
One alternative approach: maybe we should only publish PyPI packages after the GitHub release is done?
That would avoid that problem as well.

visr:
This would help, but that would still give us broken tags right? Just not broken Python releases, and possibly Ribasim tags that won't get Python releases.

I'll make an issue.

PR #1427, opened 2024-04-24
Use mean input in allocation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1121.

To do:
- [x] Use average source flows
- [x] Use average vertical fluxes on basin for level demand
- [x] Update documentation


SouthEndMusic:
The last commit also removes all variables and constraints from the allocation problems for basins that do not have a level demand.

visr (approved):

Issue #1208, opened 2024-03-04
Avoid unconstrained sections in allocation networks


SouthEndMusic:
As far as I can tell, there are now at least 2 cases of unconstrained sections in the allocation problem left:
- [ ] `FractionalFlow` nodes which are discussed [below](https://github.com/Deltares/Ribasim/issues/1208#issuecomment-1989556081);
- [x] `UserDemand` return flow is essentially an extra source. This leads to unconstrained behavior when collecting subnetwork demands, because if there is still return flow left when starting demand collection, flow can come from either the user return flow or the main network inlet. When collecting subnetwork demands flow should only be allowed to come from the main network inlet, to get a proper value for how much flow is required from the main network. Solving this might be as easy as setting user return flow capacity to 0 before starting demand collection. Yes, this means that some flow capacity is lost, but I find having an algorithm that is as deterministic as possible more important.

This is also as good a place as any to start using `max(0, JuMP.value(F[edge]))` everywhere when accessing flows, since they can be (very small) negative numbers.

SouthEndMusic:
I forgot to bring up fractional flow in the meeting with the RTC-Tools team, but I realised that this can also be solved nicely with buffers. That is: use equality constraints for the relationship between the inflow and the outflow for each of the fractional flow branches, and store flow that cannot immediately be used in one of these branches in a buffer. 

Note that this can lead to situations where a lot of flow ends up unused in these buffers, but I would call that a schematization issue.

gijsber:
Hi @SouthEndMusic : are you saying.. if a user has two outgoing edges, one with fraction 0.5 and one with fraction 0.4, you basically buffer the 0.1 portion to keep your optimization problem sane.
It is the modellers problem that the fractions do not add up to 1.0.

Now I wonder two things:
1. what happens if the modeller has defined fractions of 0.5 and 0.6 (total > 1)
2. Do we have a validaton rule that actually detects that fractions do not add up to exactly 1 ?

gijsber:
Hi @SouthEndMusic , 

For `UserDemand` I agree with you solution as given in the issue description.
For `FractionalFlow` @SnippenE  and me concluded that we indeed have to follow the fractional flow division (as this is what the physical layer will follow), which results in B potentially missing out on water that is requested.

If modellers are unhappy, they need to consider if they have other solutions instead of fractional flows.
If we really end up in undesirable sitautaions, we may decide to update the algorithms at a later stage, e.g. as follows:
[...]

Issue #565, opened 2023-08-30
Optimize allocation per source in a fixed order


gijsber:
Use Case RWS Linge: Als regionaal watersysteem uit meerdere inlaatpunten kan worden gevoed is er een bepaalde volgordelijkheid. Wanneer bijvoorbeeld 1 inlaatpunt niet volledig kan worden benut vanwege een lagere waterstand in het hoofdwaterysteem, vindt aanvulling plaats.

If a regional watersystem/network can be fed from multiple inlets, a certain sequence of using inlets should be feasible. When the preferred inlet does not have the water available, the request/setpoint of the second inlet should be adjusted.

_This was discussed with members from the RTC tools team. They advocate the goal programming approach, i.e. solve for each source separately in order of their preference._

SouthEndMusic:
As discuseed with @gijsber: we first look at single inlets to subnetworks. When the allocation implementation is more mature, we can consider multiple inlets. The main question to answer at that point is: If subnetworks are viewed as users in the main water network optimization, how is this modelled when the subnetwork has multiple inlets? This can probably be taken into account in the optimization problem formulation, where the sum of the flows to the multiple inlets together are considered the flow to one user.

gijsber:
from @SouthEndMusic 
#565 moet ook nog gedaan worden. Mijn huidige aanpak ondersteunt nu wel meerdere edges vanuit het hoofdnetwerk naar een en hetzelfde subnetwerk, maar ik weet nog niet hoe een ordering over die edges kan worden meegenomen in optimalisatie, anders dan dat de flow vanaf elke inlet individueel wordt getrackt in het subnetwerk, wat het optimalisatieprobleem wellicht een stuk complexer maakt

gijsber:
Hi @SouthEndMusic, 
As long as we only track and add a objective function term (i.e. cost) to the inlet edges, the complexity increase will be limited.

We really will get expensive as soon as all edges in the allocation_subnetwork need to get a cost to enable prioritization over sources, as any small cost factors will make the optimal solution less pronounced

SouthEndMusic:
This is the first step in the new plan for allocation development as worked out on 27-5-2024. Steps:

- [ ] Make the source order per subnetwork a required input. This is a breaking change for models with allocation, but we take that for granted at this point. External modelers won't suffer much problems from this as allocation is hardly used yet.
- [ ] Implement the allocation optimization per source. Things to keep in mind:
[...]

PR #1429, opened 2024-04-25
Avoid unconstrained sections in allocation networks

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1208.

I fixed the `UserDemand` point more generally: no sources other than the main network should be available when collecting subnetwork demands. We do allow multiple inlets from the main network to a subnetwork. We do not test this yet, but this also allows for insufficiently constraint distribution of the subnetwork demands over these inlets. A preference ordering over sources (https://github.com/Deltares/Ribasim/issues/565) could resolve this.


SouthEndMusic:
Brainstorm:

61adfa53-ae94-49df-bb81-7ab17b818ebd

Issue #1428, opened 2024-04-25
How to handle basins without a level demand in allocation


SouthEndMusic:
Currently the allocation algorithm only takes local available water of basins (including their forcings) into account when these basins have a level demand. That is because the level demand defines when a basin is in a demand or a supply state. When the basin doesn't have a level demand, there is no reference for this. However, modellers probably still expect for instance that a large amount of precipitation on a basin is taken into account in allocation, even if that basin has no level demand. Should we actually implement something for this, or should we just thoroughly document that this is the behavior of basins in allocation?

@gijsber @visr

visr:
I think the default behavior is ok, to not allow abstractions from Basins unless we know there is surplus water. I can imagine some Basins don't have a demand for low water levels, but can have a surplus. In this case we can attach a LevelDemand node with only a `max_level`.

Perhaps for that it is good to make both `min_level` and `max_level` optional, with defaults `-Inf` and `Inf`?

https://deltares.github.io/Ribasim/core/usage.html#sec-level_demand

gijsber:
Proposal from Martijn is fine to me. If no LevelDemand is set, one can basically utilize all water from a basin in allocation.

visr:
> If no LevelDemand is set, one can basically utilize all water from a basin in allocation.

My proposal was to not allow any use in that case. I feel like that behavior is safer against over abstraction. Modelers need to opt in to Basin sources. Then again, local precipitation should perhaps always be a source. Should any storage increase be a source, but not all storage?

Not sure what most modelers would expect here. I feel like it is quite an important choice.

visr:
We discussed this together, and decided that the current behavior of not allowing Basin storage as a source by default is the conservative thing to do, and easy to explain. If modelers expect a particular Basin's storage to be used, they can attach a LevelDemand to indicate from which level it is considered surplus water. #1430 will help with this as well.

PR #1430, opened 2024-04-25
Make LevelDemand min_level and max_level optional

visr:
As proposed in https://github.com/Deltares/Ribasim/issues/1428#issuecomment-2076766380

This makes it easier to use LevelDemand in case you only care about a shortage or surplus situation.

### DIFF
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -93,7 +93,7 @@ function parse_static_and_time(
 
     errors = false
     t_end = seconds_since(config.endtime, config.starttime)
-    trivial_timespan = [nextfloat(-Inf), prevfloat(Inf)]
+    trivial_timespan = [0.0, prevfloat(Inf)]
 
     for (node_idx, node_id) in enumerate(node_ids)
         if node_id in static_node_ids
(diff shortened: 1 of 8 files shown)


core/src/read.jl
```diff
@@ -93,7 +93,7 @@ function parse_static_and_time(
 
     errors = false
     t_end = seconds_since(config.endtime, config.starttime)
-    trivial_timespan = [nextfloat(-Inf), prevfloat(Inf)]
+    trivial_timespan = [0.0, prevfloat(Inf)]
```
visr, line 96: Without this change we would get a NaN instead of Inf from DataInterpolations.

core/src/util.jl
visr, line 1: Returning a storage of 0 for levels below the bottom seems fine to me, and is needed to support -Inf min_level.

docs/core/usage.qmd
```diff
@@ -479,14 +479,17 @@ min_level     | Float64  | $m$          | -
 
 A `LevelDemand` node associates a minimum and a maximum level with connected basins to be used by the allocation algorithm.
 Below the minimum level the basin has a demand of the supplied priority,
-above the maximum level the basin acts as a source, used by all nodes with demands in order of priority.
-The same `LevelDemand` node can be used for basins in different subnetworks.
+above the maximum level the basin has a surplus and acts as a source, used by all nodes with demands in order of priority.
+The same `LevelDemand` node can be used for Basins in different subnetworks.
+
+Both `min_level` and `max_level` are optional, to be able to handle only the demand or surplus side.
+If both are missing `LevelDemand` won't have any effects on allocation.
```
SouthEndMusic, line 486:
```suggestion
If both are missing, `LevelDemand` won't have any effects on allocation.
```

SouthEndMusic (approved):
Good good

PR #1415, opened 2024-04-22
Use a Rust binary to expose Ribasim CLI

Hofer-Julian:
- libribasim and ribasim_cli can share the same libraries
- we get `ribasim.exe` instead of `ribasim.cmd`

PR #1434, opened 2024-04-28
Tighten edge validation

visr:
Since LevelDemand and FlowDemand were not AbstractParameterNodes, their connections were not being validated despite the constraints being set properly. I noticed since they don't show up in the tables in https://deltares.github.io/Ribasim/core/validation.html.

LinearResistance and ManningResistance already required 1 in and 1 outflow, just not in validation.

Basin already cannot have outgoing control edges.

Also removes an outdated comment about `ribasim.cmd`. We now only have one `ribasim.exe` since #1415.


SouthEndMusic (approved):
Good stuff 👍

Issue #889, opened 2023-12-12
Pre-calculate flow neighbor IDs per node


visr:
See #830 for a description of the current performance bottleneck of large models. #830 sped this up, but not enough, it still dominates the profile whereas before it was hardly visible.

The easiest way to address this that I see is to simply pre-calculate the flow neighbor IDs per node, and put that into the Parameters.

visr:
Partially implemented by #1436 and #1438 for the *Resistance and Basin node types.

Jingru923:
@visr Is it okay that @SouthEndMusic  take over this issue?

visr:
Yeah definitely

visr:
#1448 takes another chunk out of this. Some still remain, for instance DiscreteControl. Search for `flowid(` or `flowids(` in solve.jl.

Allocation also doesn't pre-calculate them. But I also haven't seen them show up in profiles, so probably better to leave that as-is for now.

PR #1435, opened 2024-04-28
Standardize neighbor ID naming

visr:
I like the `inflow_id` name since it is clear what the edge direction is, the edge type, and that it concerns an ID. It is both the name of the function and used as a variable. I think that's fine because you never want to mix both usages.

These are some preparations for #889.


visr:
Incorporated as part of #1436.

PR #1436, opened 2024-04-29
Precalculate resistance neighbors

visr:
On top of #1435.
This significantly speeds up the running time at the cost of using slightly more storage.


SouthEndMusic (approved):
Looks good, also some improved docstrings 👍

PR #1438, opened 2024-04-29
Precalculate Basin neighbors

visr:
Like #1436, but for Basin. Those have `n` inflow and `n` outflow nodes, so I'm using a Vector of Vectors.


SouthEndMusic (approved):
Noice noice

Issue #452, opened 2023-07-24
Show convergence bottleneck nodes to users


visr:
Related to #419, if a run doesn't converge or is slower than expected, we need good tools to investigate the issue.

One thing that will help is to look at the [solver stats](https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#Differential-Equation-Solver-Statistics-(destats)):

```
julia> using Ribasim
[ Info: Precompiling Ribasim [aac5e3d9-0b8f-4d4f-8241-b1a7a9632635]

julia> model = Ribasim.run("nl.toml")
┌ Warning: Newton steps could not converge and algorithm is not adaptive. Use a lower dt.
└ @ SciMLBase C:\Users\visser_mn\.julia\packages\SciMLBase\wvDeR\src\integrator_interface.jl:612
Model(ts: 2, t: 2020-01-01T00:00:00)

julia> model.integrator.sol.stats
DiffEqBase.Stats
Number of function 1 evaluations:                  8803
Number of function 2 evaluations:                  0
Number of W matrix evaluations:                    1
Number of linear solves:                           10
Number of Jacobians created:                       1
Number of nonlinear solver iterations:             10
Number of nonlinear solver convergence failures:   1
Number of rootfind condition calls:                0
Number of accepted steps:                          0
Number of rejected steps:                          0
```

Additionally it would be nice to locate the nodes in the network causing the issues easily. One thing @Huite suggested was to calculate some stability criterium, like the Basin retention time or courant number.

The goal of this issue is to collect some tools, some of which could be useful for the documentation.

Huite:
Another thought: we're currently relying on DifferentialEquations.jl machinery to do the timestepping for us.

As far as I know, it compares a higher and lower order to estimate whether time step sizes have to be adjusted. Essentially, I expected it do compare `du_lower` and `du_higher`. If we can investigate these a bit and get access to them, we should be able to output them and interpret them spatially.

This should help us and users to identify what the problematic nodes and connections are.

SouthEndMusic:
I asked about this on the Julia discourse:

https://discourse.julialang.org/t/find-out-which-state-s-cause-convergence-problems/113738

SouthEndMusic:
Chris (the man, the myth, the legend) has responded. Something like this works (for certain algorithms):

```julia
(; cache, p) = model.integrator
(; basin) = p
if hasproperty(cache, :atmp)
    storage_error = abs.(cache.atmp.storage)
[...]

PR #1440, opened 2024-04-29
Show convergence bottleneck nodes to users

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/452.

To do:
- [x] Implement https://github.com/Deltares/Ribasim/issues/452#issuecomment-2091466143,
- [x] Add some more basins to the model and test that the quickly emptying one is the bottleneck.


SouthEndMusic:
I've implemented this now, but the result is not what I expect. The problematic basins (which get drained extremely quickly) in the test model have error $0$, and the others have some positive error. We need to think about under which circumstances, say which values of `retcode` this method makes sense.

core/test/validation_test.jl
```diff
@@ -458,3 +458,19 @@ end
         dt,
     )
 end
+
+@testitem "Convergence bottleneck" begin
+    using Suppressor: @capture_out
```
visr, line 463: I think you can also use `using IOCapture: capture`, we use that already.

evetion (approved):
LGTM

PR #1445, opened 2024-04-30
Update julia from 1.10.0 to 1.10.3

visr:
The other patch releases had some issue affecting our build, specifically loading libribasim with Python / ribasim_api (https://github.com/JuliaLang/julia/pull/53643). These should all be resolved now, but I'll trigger a build to be sure.


visr:
TeamCity also likes 1.10.3 :)

Jingru923 (approved):

PR #1446, opened 2024-05-01
Re-enable validation in Ribasim python

### LINKED ISSUE #1333: Re-enable validation functions for Ribasim Python
Together with the add-API, some validation methods were disabled:
https://github.com/Deltares/Ribasim/blob/af6b1adb92b58f44edccb7ea3a7256c57ffcf399/python/ribasim/ribasim/model.py#L186-L190

We should re-implement this functionality again.
The original implementation can be found here:
https://github.com/Deltares/Ribasim/blob/c5ae83de3159ab77407773b02347fec6f1aa41c0/python/ribasim/ribasim/model.py#L257-L305

Jingru923:
Fixes #1333

### DIFF
--- a/python/ribasim/tests/test_model.py
+++ b/python/ribasim/tests/test_model.py
@@ -7,10 +7,12 @@ import pytest
 import xugrid
 from pydantic import ValidationError
 from pyproj import CRS
+from ribasim import Node
 from ribasim.config import Solver
 from ribasim.geometry.edge import NodeData
 from ribasim.input_base import esc_id
 from ribasim.model import Model
+from ribasim.nodes import pump
 from shapely import Point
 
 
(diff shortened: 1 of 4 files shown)


python/ribasim/tests/test_model.py
```diff
@@ -65,24 +67,15 @@ def test_exclude_unset(basic):
     assert d["solver"]["saveat"] == 86400.0
 
 
-@pytest.mark.xfail(reason="Needs implementation")
+# @pytest.mark.xfail(reason="Needs implementation")
 def test_invalid_node_id(basic):
     model = basic
```
visr, line 72: I think this test doesn't need to build on the basic model. Running `Node(-1, Point(7.0, 7.0))` by itself should be enough to trigger it. Can you try removing that argument, and just putting `Node(-1, Point(7.0, 7.0))` in the `pytest.raises` block?

visr (approved):
Looks good, thanks!

PR #1448, opened 2024-05-02
Pre-calculate flow neighbor IDs for Pump, Outlet, UserDemand, FractionalFlow, TabulatedRatingCurve

visr:
Takes another chunk out of #889
Also does some refactoring to make the solve methods more uniform.


visr:
This was a bit more involved. With the new additions to the structs we can no longer create Parameters with the wrong number of nodes, breaking the validation tests for this. So I had to move the neighbor validation earlier, directly after graph construction.

core/src/model.jl
visr, line 1: This wasn't called before.

core/src/util.jl
visr, line 1: Removed this since we already have a `nodetypes::Vector{Symbol}` for similar purposes. For that I also added a test.

python/ribasim_testmodels/ribasim_testmodels/invalid.py
visr, line 1: This model is not used to test neighbor counts, so I had to add a missing outflow edge to allow the construction of Parameters.

core/test/utils_test.jl
```diff
@@ -286,18 +286,52 @@ end
...
+        :Outlet,
+        :LinearResistance,
+        :FractionalFlow,
+    ])
+    for nodetype in nodetypes
+        NodeType.T(nodetype)
+        if nodetype != :Terminal
+            # It has a struct which is added to Parameters
+            T = getproperty(Ribasim, nodetype)
+            @test T <: AbstractParameterNode
```
SouthEndMusic, line 333: Nice, lack of this gave some problems before

SouthEndMusic (approved):

PR #1455, opened 2024-05-06
Redo the "stop explicitly requiring `--environment=dev` when running pixi install tasks" with new pixi release

### LINKED ISSUE #1324: Stop explicitly requiring `--environment=dev` when running pixi install tasks

Jingru923:
Fixes #1324 

And reimplement all the changes in PR #1400


Jingru923:
I have also changed the teamcity configuration. Now in the two test pipeline (linux - Test Ribasim Binaries and windows - Test Ribasim Binaries) have no `pixi run install-python` in the build step

.teamcity/Ribasim_Linux/buildTypes/Ribasim_Linux_TestRibasimBinaries.xml
```diff
@@ -29,7 +29,7 @@ pixi --version]]></param>
 source /usr/share/Modules/init/bash
 
 module load pixi
-pixi run --environment=dev install
+pixi run install
```
visr, line 32: The `install` task no longer exists, right? And `install-ci`, installs Julia, which isn't needed here. So I guess this line can be removed. Can you also trigger TeamCity builds on this branch to make sure that still works?

.teamcity/Ribasim/buildTypes/Ribasim_GenerateTestmodels.xml
```diff
@@ -19,7 +19,7 @@ source /usr/share/Modules/init/bash
 
 module load pixi
 pixi --version
-pixi run --environment=dev install-ci]]></param>
+pixi run install-ci]]></param>
```
visr, line 22: I don't think this task is needed anymore.

.github/workflows/qgis.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           pixi-version: "latest"
       - name: Prepare pixi
         run: |
-          pixi run --environment=dev install-ci
+          pixi run install-ci
```
visr, line 30: I don't think this task is needed anymore.

.github/workflows/python_codegen.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         with:
           pixi-version: "latest"
       - name: Prepare pixi
-        run: pixi run --environment=dev install-ci
+        run: pixi run install-ci
```
visr, line 23: I don't think this task is needed anymore.

.github/workflows/mypy.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         with:
           pixi-version: "latest"
       - name: Prepare pixi
-        run: pixi run --environment=dev install-ci
+        run: pixi run install-ci
```
visr, line 23: I don't think this task is needed anymore.

.teamcity/Ribasim_Windows/buildTypes/Ribasim_Windows_TestRibasimBinaries.xml
```diff
@@ -21,7 +21,7 @@
       </runner>
       <runner id="RUNNER_1503" name="Run tests" type="simpleRunner">
         <parameters>
-          <param name="script.content"><![CDATA[pixi run --environment=dev install
+          <param name="script.content"><![CDATA[pixi run install
```
visr, line 24: I don't think this task is needed anymore.

docs/contribute/core.qmd
```diff
@@ -37,7 +37,7 @@ You can let it start automatically by following these [instructions](https://tim
 There is a section on editors and IDEs for Julia on <https://julialang.org/>, scroll down to see it.
 We use and recommend Microsoft's free editor [Visual Studio Code](https://code.visualstudio.com/).
 When combined with the [Julia extension](https://www.julia-vscode.org/) it provides a powerful and interactive development experience.
-Make sure to have the correct environment when opening your IDE by running `open-vscode.bat`, or opening a `pixi --environment=dev shell` and then calling the command to open the editor of your choice.
+Make sure to have the correct environment when opening your IDE by running `open-vscode.bat`, or opening a `pixi shell` and then calling the command to open the editor of your choice.
```
visr, line 40: Good to check, but I think `pixi --environment=dev shell` is still needed here, just like `open-vscode.bat` also still using `pixi run --environment=dev code .`.

visr (commented):
Left some comments. Before this is merged we should manually trigger the TeamCity jobs on this PR, since it is likely to affect it.

.teamcity/Ribasim_Linux/buildTypes/Ribasim_Linux_TestRibasimBinaries.xml
Jingru923: I removed it and triggered both Linux and Windows Ribasim. They worked.

SouthEndMusic (approved):
I see no problems with this, because it's mainly deletions. You shouldn't only rely on my judgement for this though

Issue #1456, opened 2024-05-07
Increase performance by reducing lookups


SouthEndMusic:
https://github.com/Deltares/Ribasim/pull/1448 Gave some nice performance improvements. With these changes, a small number of utility functions are now dominating the profile (for the RWS HWS model):

- [x] `set_flow!` because of a dictionary lookup. To resolve this without much added complexity, I propose to not (just) store inflow and outflow IDs like in https://github.com/Deltares/Ribasim/pull/1448, but to store the `EdgeMetadata` of the inflow and outflow edges (which also includes the inflow and outflow IDs). Also add the flow index to the metadata (and make the distinction between that and the edge ID clear), so that it doesn't have to be retrieved by the aforementioned lookup.
- [x] `get_level` because of a basin index lookup. Per node type whose `formulate_flow!` uses this function:
  - `PidControl` will probably never be used heavily enough so that this becomes a performance bottleneck
  - `UserDemand`: store the upstream basin index per `UserDemand` node
  - `ManningResistance`: store the upstream and downstream basin index per `Manningresistance` node
  - `LinearResistance`, `Outlet`: store the upstream and downstream basin index per `LinearResistance` node (although both can be `LevelBoundary` nodes, not sure yet about this)
- [x] `basin_bottom` also because of a basin index lookup. This function is only used for `TabulatedRatingCurve`, so I propose storing upstream and downstream bottoms in the `TabulatedRatingCurve` struct.

PR #1457, opened 2024-05-07
Increase performance by reducing lookups

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1456. The main speedup is due to adding basin indices to the edge metadata, and precalculating vectors of the metadata of inflow and outflow edges of several node types.


core/src/graph.jl
```diff
@@ -65,7 +65,14 @@ function create_graph(db::DB, config::Config, chunk_sizes::Vector{Int})::MetaGra
...
             subnetwork_id = 0
         end
-        edge_metadata = EdgeMetadata(fid, edge_type, subnetwork_id, (id_src, id_dst))
+        edge_metadata = EdgeMetadata(
+            fid,
+            edge_type == EdgeType.flow ? flow_counter + 1 : 0,
+            edge_type,
+            subnetwork_id,
+            (id_src, id_dst),
+            (0, 0),
```
evetion, line 74: Might be good to set to -1, as 0 has a meaning already. Might also be nice to document inline that it is a placeholder which is overwritten later by `x`.

core/src/parameter.jl
```diff
@@ -116,16 +116,20 @@ end
...
 edge: (from node ID, to node ID)
+basin_idxs: Basin indices of source and destination nodes (0 if not a basin)
 """
 struct EdgeMetadata
     id::Int32
+    flow_idx::Int32
     type::EdgeType.T
     subnetwork_id_source::Int32
     edge::Tuple{NodeID, NodeID}
+    basin_idxs::Tuple{Int32, Int32}
```
evetion, line 132: I would advise to split this into basin_source_idx and basin_destination_idx attributes. Code later on makes repeated use of [1] or [2] which is less readable (especially when it's unclear that something returns EdgeMetadata). Or you could make it into a namedtuple?

core/test/validation_test.jl
```diff
@@ -86,7 +86,7 @@ end
     graph[NodeID(:Pump, 6)] = NodeMetadata(:pump, 9)
 
     function set_edge_metadata!(id_1, id_2, edge_type)
-        graph[id_1, id_2] = EdgeMetadata(0, edge_type, 0, (id_1, id_2))
+        graph[id_1, id_2] = EdgeMetadata(0, 0, edge_type, 0, (id_1, id_2), (0, 0))
```
evetion, line 89: Might set these added 0 to -1 because 0 already has meaning.

core/src/util.jl
```diff
@@ -374,12 +374,13 @@ storage: tells ForwardDiff whether this call is for differentiation or not
...
     p::Parameters,
     node_id::NodeID,
+    i::Integer,
     t::Number;
     storage::Union{AbstractArray, Number} = 0,
 )::Tuple{Bool, Number}
     (; basin, level_boundary) = p
     if node_id.type == NodeType.Basin
-        _, i = id_index(basin.node_id, node_id)
+        i = iszero(i) ? id_index(basin.node_id, node_id)[2] : i
```
evetion, line 383: If it's not a basin, it's the other node of an edge? This could use some (inline) documentation.

core/src/parameter.jl
```diff
@@ -233,17 +237,17 @@ Type parameter C indicates the content backing the StructVector, which can be a
...
 active: whether this node is active and thus contributes flows
 tables: The current Q(h) relationships
 time: The time table used for updating the tables
 control_mapping: dictionary from (node_id, control_state) to Q(h) and/or active state
 """
 struct TabulatedRatingCurve{C} <: AbstractParameterNode
     node_id::Vector{NodeID}
-    inflow_id::Vector{NodeID}
-    outflow_ids::Vector{Vector{NodeID}}
+    inflow_edge::Vector{EdgeMetadata}
```
evetion, line 249: Is it guaranteed that the node_id is always the first node of an edge? If not, we might want to store that, if so, might be good to document it?

core/src/solve.jl
```diff
@@ -345,12 +346,15 @@ function formulate_flow!(
...
+        inflow_edge = linear_resistance.inflow_edge[i]
+        outflow_edge = linear_resistance.outflow_edge[i]
+
+        inflow_id = inflow_edge.edge[1]
+        outflow_id = outflow_edge.edge[2]
 
         if active[i]
-            _, h_a = get_level(p, inflow_id, t; storage)
-            _, h_b = get_level(p, outflow_id, t; storage)
+            _, h_a = get_level(p, inflow_id, inflow_edge.basin_idxs[1], t; storage)
```
evetion, line 356: This pattern is now repeated a lot, might be good to refactor into `get_outflow_level(p, edge, t; storage)` and `get_inflow_level`? Similar to how you did for set/get_flow.

evetion (changes requested):
Nice to see such speedups! Most of my remarks are about readability, so I would also ask you to write in the description (which will turn into the commit message) why and how you did things. Can only be a few lines, but it should explain the changes to all nodes (now storing edges instead of node ids).

PR #1464, opened 2024-05-14
Move to DataInterpolations v5

visr:
This was breaking for us since it removed a type alias that we defined in our const alias ScalarInterpolation and VectorInterpolation.

This also tightens some return types to those concrete aliases, and changes a LevelDemand field type from the abstract `Vector{LinearInterpolation}` to the concrete `Vector{ScalarInterpolation}`, which should help the compiler.

Closes #1461.


SouthEndMusic (approved):
LGTM

PR #1468, opened 2024-05-14
Introduce custom `TrapezoidIntegrationCallback` to integrate flows and basin forcings

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1473.

To do:
- [ ] Update `TrapezoidIntegrandAffect.integrand_value` in `update_basin!` for integrands that contain the basin forcings (both for flow integration and bmi integration)
- [ ] `TrapezoidIntegrandAffect.integrand_value_prev` can be removed, if the integration is done in this order:
```julia
function (affect!::TrapezoidIntegrationAffect)(integrator)::Nothing
    (; integrand_func!, integrand_value, cache, integral) = affect!
    (; dt) = integrator

    cache .= integrand_value
    integrand_func!(integrand_value, integrator.p)
    cache .+= integrand_value
    cache .*= 0.5 * dt

    integral .+= cache
    return nothing
end
```
- [ ] Test the integration around a discontinuity of the basin forcings
- [ ] Add docstrings, inline comments


SouthEndMusic:
A drawback of the current way flows and basin forcings are computed is that a vector for all these values is allocated at each timestep, while they are later averaged over the `saveat` intervals. It would be nice if the `IntegratingCallback` would support passing the `saveats` to predefine the integration intervals, maybe that's worth making an upstream issue.

SouthEndMusic:
I just realized I made a wrong assumption about how `IntegratingCallback` and `IntegratingSumCallback` work. I thought that the integrand function was only evaluated at time points of the integrator, but it is in fact called at several different points:

https://github.com/SciML/DiffEqCallbacks.jl/blob/a35955b8480356503e3df492a9e02249cef00cd0/src/integrating.jl#L175C5-L177C55

The problem with this is that we don't know the value of the integrand at these in between points. The `t_temp` in the code linked above are between the `tprev` and `t`, so as a (silly) workaround we could cache the value of the integrand at `tprev` and do a linear interpolation.

There's probably also a bug in these functions, where it checks whether the ODEproblem is in-place instead of the integrand function.

SouthEndMusic:
Although this was a fun dive into creating a custom callback, it didn't reach its goal, namely reducing the complexity of the integration code.

PR #1469, opened 2024-05-14
Stricter TabulatedRatingCurve validation

visr:
Fixes a part of #279.

The four checks added here are:

- At least two datapoints are needed.
- The `flow_rate` must start at 0.
- The `level` cannot be repeated.
- The `flow_rate` cannot decrease with increasing `level`.

And additionally we now ensure that flow rates are kept at 0 for levels below the minimum.

~~I should still check the docs, they may need some updating.~~ The other part of #279 is best done in a separate PR as it relates to the Basin / profile table.

### DIFF
--- a/docs/core/usage.qmd
+++ b/docs/core/usage.qmd
@@ -368,24 +368,29 @@ fraction      | Float64 | -            | in the interval [0,1]
 
 # TabulatedRatingCurve {#sec-tabulated-rating-curve}
 
-This table is similar in structure to the Basin profile. The TabulatedRatingCurve gives a
-relation between the storage of a connected Basin (via the outlet level) and its outflow.
+This table is similar in structure to the Basin profile.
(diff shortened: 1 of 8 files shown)


docs/core/usage.qmd
```diff
@@ -368,24 +368,29 @@ fraction      | Float64 | -            | in the interval [0,1]
 
 # TabulatedRatingCurve {#sec-tabulated-rating-curve}
 
-This table is similar in structure to the Basin profile. The TabulatedRatingCurve gives a
-relation between the storage of a connected Basin (via the outlet level) and its outflow.
+This table is similar in structure to the Basin profile.
+The TabulatedRatingCurve gives a relation between the upstream level of a connected Basin and its outflow.
```
SouthEndMusic, line 372:
```suggestion
The TabulatedRatingCurve gives a relation between the level of a directly upstream Basin or Levelboundary and its outflow.
```

SouthEndMusic (approved):
Lookin good

PR #1486, opened 2024-05-21
Stronger 'Basin / profile' validation

### LINKED ISSUE #279: Extrapolating the LinearInterpolation

visr:
Fixes #279.
Follow up of #1469.

Most rules were already in place, this just adds a bit more validation.
I moved the validation out of the constructor to be more in line with the other validation taking place.

### DIFF
--- a/core/src/parameter.jl
+++ b/core/src/parameter.jl
@@ -186,44 +186,6 @@ struct Basin{T, C, V1, V2, V3} <: AbstractParameterNode
     demand::Vector{Float64}
     # Data source for parameter updates
     time::StructVector{BasinTimeV1, C, Int}
-
-    function Basin(
-        node_id,
-        inflow_ids,
-        outflow_ids,
-        vertical_flux_from_input::V1,
-        vertical_flux::V2,
-        vertical_flux_prev::V3,
-        vertical_flux_integrated::V3,
-        vertical_flux_bmi::V3,
-        current_level::T,
-        current_area::T,
-        area,
-        level,
-        storage,
-        demand,
-        time::StructVector{BasinTimeV1, C, Int},
-    ) where {T, C, V1, V2, V3}
-        is_valid = valid_profiles(node_id, level, area)
-        is_valid || error("Invalid Basin / profile table.")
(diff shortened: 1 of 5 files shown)


SouthEndMusic (approved):
Nice Nice

PR #1492, opened 2024-05-21
Fix github release script

### LINKED ISSUE #1467: Broken `github-release.py` needs to be fixed
In our automated release process, when a tag that starts with `v20` are made, TeamCity will start the automated release process.
When I was about to make a release, it didn't happen because the tag can't be detected. This is a result of previous changes in file `github-release.py`. This needs to be fixed

visr:
Fixes #1467


Jingru923 (approved):
Look good to me

Issue #1389, opened 2024-04-15
Run allocation at t = 0


visr:
Currently allocation doesn't run at t = 0. This is done because flow on the source edges is still unknown. This means that effectively there is a free-for-all situation for the allocation interval. It would be better to special case t = 0, to run the RHS once to get a first estimate for the instantaneous flows at t = 0, and then run allocation with that.

SouthEndMusic:
With https://github.com/Deltares/Ribasim/pull/1427 it is required to set these initial instantaneous flows as the mean flows.

PR #1496, opened 2024-05-22
Run allocation at `t = 0`

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1389.

### DIFF
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -405,6 +405,7 @@ function add_constraints_conservation_node!(
     end
 
     # Only the node IDs with conservation constraints on them
+    # Discard constraints of the form 0 == 0
     node_ids = [
         node_id for node_id in keys(inflows) if
         !(isempty(inflows[node_id]) && isempty(outflows[node_id]))
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -405,7 +405,10 @@ function add_constraints_conservation_node!(
     end
 
     # Only the node IDs with conservation constraints on them
-    node_ids = keys(inflows)
+    node_ids = [
+        node_id for node_id in keys(inflows) if
+        !(isempty(inflows[node_id]) && isempty(outflows[node_id]))
+    ]
 
     problem[:flow_conservation] = JuMP.@constraint(
         problem,
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -174,7 +174,7 @@ function Model(config::Config)::Model
         @show Ribasim.to
     end
 
-    if config.allocation.use_allocation
+    if config.allocation.use_allocation && is_active(parameters.allocation)
         set_initial_allocation_mean_flows!(integrator)
     end
 
@@ -211,7 +211,8 @@ function SciMLBase.step!(model::Model, dt::Float64)::Model
     # If we are at an allocation time, run allocation before the next physical
     # layer timestep. This allows allocation over period (t, t + dt) to use variables
     # set over BMI at time t before calling this function.
-    if t % config.allocation.timestep ≈ 0
+    ntimes = t / config.allocation.timestep
+    if round(ntimes) ≈ ntimes
         update_allocation!(integrator)
     end
     step!(integrator, dt, true)
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -174,6 +174,10 @@ function Model(config::Config)::Model
         @show Ribasim.to
     end
 
+    if config.allocation.use_allocation
+        set_initial_allocation_mean_flows!(integrator)
+    end
+
     return Model(integrator, config, saved)
 end
 
@@ -207,9 +211,7 @@ function SciMLBase.step!(model::Model, dt::Float64)::Model
     # If we are at an allocation time, run allocation before the next physical
     # layer timestep. This allows allocation over period (t, t + dt) to use variables
     # set over BMI at time t before calling this function.
-    # Also, don't run allocation at t = 0 since there are no flows yet (#1389).
-    ntimes = t / config.allocation.timestep
-    if t > 0 && round(ntimes) ≈ ntimes
+    if t % config.allocation.timestep ≈ 0
         update_allocation!(integrator)
     end
     step!(integrator, dt, true)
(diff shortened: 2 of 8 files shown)


SouthEndMusic:
I am struggling a bit to understand the behavior of the `allocation_example` model at the start of the simulation. Here is the plot of the allocation data:

14719993-635d-421f-ad9f-c74703f603e8

And here is the data for the first allocation solve:

```
 Row │ time     subnetwork_id  node_type   node_id  priority  demand   allocated  realized 
     │ Float64  Int32          String      Int32    Int32     Float64  Float64    Float64  
─────┼─────────────────────────────────────────────────────────────────────────────────────
   1 │     0.0              2  UserDemand        6         1      0.0        0.0       1.0
   2 │     0.0              2  UserDemand        3         1      1.5        1.5       1.5
   3 │     0.0              2  UserDemand       13         1      0.0       -0.0       1.2
   4 │     0.0              2  UserDemand        6         2      0.0        0.0       1.0
   5 │     0.0              2  UserDemand        3         2      0.0       -0.0       1.5
   6 │     0.0              2  UserDemand       13         2      1.2        0.2       1.2
   7 │     0.0              2  UserDemand        6         3      1.0       -0.0       1.0
   8 │     0.0              2  UserDemand        3         3      0.0       -0.0       1.5
   9 │     0.0              2  UserDemand       13         3      0.0       -0.0       1.2
```

The `abstracted` a.k.a. `realized` values at $t = 0$ can be ignored, as there is no realized flow at $t = 0$ yet. These are the things I find weird:

- `UserDemand` node 13 having a positive demand of 1.2 at priority 2. In the schematization this node is only assigned a demand at priority 3. Things are getting mixed up somewhere
- `UserDemand` node 3 having 0 abstraction after the first day, even though flow was allocated to this node

core/src/model.jl
```diff
@@ -207,9 +211,7 @@ function SciMLBase.step!(model::Model, dt::Float64)::Model
     # If we are at an allocation time, run allocation before the next physical
     # layer timestep. This allows allocation over period (t, t + dt) to use variables
     # set over BMI at time t before calling this function.
-    # Also, don't run allocation at t = 0 since there are no flows yet (#1389).
-    ntimes = t / config.allocation.timestep
-    if t > 0 && round(ntimes) ≈ ntimes
+    if t % config.allocation.timestep ≈ 0
```
visr, line 214: If `t` is a tiny bit less than a round number of allocation timesteps, this fails:

```
julia> prevfloat(3.0) % 1.0
0.9999999999999996
```

core/src/parameter.jl
```diff
@@ -509,14 +509,14 @@ outflow_edge: outgoing flow edge metadata
...
-demand: water flux demand of UserDemand per priority over time
+demand: water flux demand of UserDemand per priority ([node_idx, priority_idx])
     Each UserDemand has a demand for all priorities,
     which is 0.0 if it is not provided explicitly.
 demand_reduced: the total demand reduced by allocated flows. This is used for goal programming,
     and requires separate memory from `demand` since demands can come from the BMI
 demand_itp: Timeseries interpolation objects for demands
 demand_from_timeseries: If false the demand comes from the BMI or is fixed
-allocated: water flux currently allocated to UserDemand per priority
+allocated: water flux currently allocated to UserDemand per priority ([node_idx, priority_idx])
```
visr, line 519:
```suggestion
demand: water flux demand of UserDemand per priority (node_idx, priority_idx)
    Each UserDemand has a demand for all priorities,
    which is 0.0 if it is not provided explicitly.
demand_reduced: the total demand reduced by allocated flows. This is used for goal programming,
    and requires separate memory from `demand` since demands can come from the BMI
demand_itp: Timeseries interpolation objects for demands
demand_from_timeseries: If false the demand comes from the BMI or is fixed
allocated: water flux currently allocated to UserDemand per priority (node_idx, priority_idx)
```

core/src/allocation_init.jl
```diff
@@ -405,7 +405,10 @@ function add_constraints_conservation_node!(
     end
 
     # Only the node IDs with conservation constraints on them
-    node_ids = keys(inflows)
+    node_ids = [
+        node_id for node_id in keys(inflows) if
+        !(isempty(inflows[node_id]) && isempty(outflows[node_id]))
+    ]
```
visr, line 411: Can you explain this change? Perhaps as a comment.

core/src/util.jl
```diff
@@ -750,3 +750,27 @@ function get_basin_idx(edge_metadata::EdgeMetadata, id::NodeID)::Int32
         0
     end
 end
+
+function set_initial_allocation_mean_flows!(integrator)::Nothing
+    (; u, p, t) = integrator
+    (; allocation, graph, basin) = p
+    (; mean_flows, allocation_models) = allocation
+    (; Δt_allocation) = allocation_models[1]
```
visr, line 758: Does this work if `use_allocation = true` but there are no allocation models?

core/src/util.jl
```diff
@@ -750,3 +750,27 @@ function get_basin_idx(edge_metadata::EdgeMetadata, id::NodeID)::Int32
         0
     end
 end
+
+function set_initial_allocation_mean_flows!(integrator)::Nothing
```
visr, line 754: Could use a comment that we don't have mean flows at t=0, but use the instantaneous flow at t=0 as an approximation.

visr (commented):
Nice! Left a few comments.

core/src/model.jl
SouthEndMusic: Ah yes maybe not risk that ^^ I figured there was a reason you didn't do this before, but I didn't know what it was. I've used this pattern quite a lot though, but I guess mostly for integers

core/src/allocation_init.jl
SouthEndMusic: It makes no sense to create a constrant `0 == 0`, so if there are no terms on either side of the equation, no not make the constraint.

core/src/util.jl
SouthEndMusic: I added a check before the call of this function to prevent that situation

visr (approved):

Issue #1498, opened 2024-05-23
Add test of mean basin forcing computation involving forcing discontinuity


SouthEndMusic:
We currently have mean basin forcing computations for the output and also for the BMI. However, we do not test yet whether this goes well when new basin forcing data becomes available within a saveat interval. We should test that the mean basin forcing computation handles this discontinuity properly.

PR #1499, opened 2024-05-23
Test Integrating discontinuous flow

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1498.


Jingru923 (approved):
Code looks good to me. As long as the math part is correct, here is my approval.

PR #1503, opened 2024-05-28
Allocation refactoring

### LINKED ISSUE #1465: Refactor allocation code
These refactoring points came out of the knowledge sharing session (which was more of a code review session):

- [ ] Use function `is_main_network` here:
https://github.com/Deltares/Ribasim/blob/61b3a173155647b78735a73357908cf8c48ed9ce/core/src/allocation_optim.jl#L1035
- [ ] Introduce a separate `allocation_util.jl` file with allocation utility functions. This might need some discussion though, I'm not sure what counts as a utility function
- [ ] The function `allocate!`: Dispatch on `optimization_type`, so several `if` statements can be removed
- [ ] The functions `allocate!` and `allocate_priority!` need to be renamed, maybe simply to `optimize!` and `optimize_priority!`, as they are not always allocating, depending on the `optimization_type`
- [ ] For the functions in `allocate_priority`
- We now only have the `AbstractParameterNode`  type for nodes. We can add a more specific abstract node type in between, say `AbstractDemandNode`, for nodes that define a demand.
- For the `adjust_demands_*` functions, loop over the `AbstractDemandNode` subtypes and dispatch on them. Alternatively, for all the `adjust_*` functions, dispatch on the constraint set names.

Jingru923:
Fixes #1465

### DIFF
--- a/core/src/allocation_optim.jl
+++ b/core/src/allocation_optim.jl
@@ -524,29 +524,6 @@ Set the demand of the flow demand nodes. 2 cases:
 - Before an allocation solve, subtract the flow trough the node with a flow demand
   from the total flow demand (which will be used at the priority of the flow demand only).
 """
-# function adjust_demands_user!(
-#     allocation_model::AllocationModel,
-#     p::Parameters,
-#     priority_idx::Int,
-# )::Nothing
-#     (; problem, subnetwork_id) = allocation_model
-#     (; graph, user_demand) = p
-#     (; node_id, demand_reduced) = user_demand
-#     F = problem[:F]
-
-#     # Reduce the demand by what was allocated
-#     for (i, id) in enumerate(node_id)
-#         if graph[id].subnetwork_id == subnetwork_id
-#             d = max(
-#                 0.0,
-#                 demand_reduced[i, priority_idx] - JuMP.value(F[(inflow_id(graph, id), id)]),
-#             )
-#             demand_reduced[i, priority_idx] = d
-#         end
-#     end
-#     return nothing
-# end
-
 function adjust_demands!(
     allocation_model::AllocationModel,
     p::Parameters,
@@ -575,26 +552,6 @@ end
 Subtract the allocated flow to the basin from its demand,
 to obtain the reduced demand used for goal programming
 """
-# function adjust_demands_level!(
-#     allocation_model::AllocationModel,
-#     p::Parameters,
-#     ::LevelDemand,
-# )::Nothing
-#     (; graph, basin) = p
-#     (; node_id, demand) = basin
-#     (; subnetwork_id, problem) = allocation_model
-#     F_basin_in = problem[:F_basin_in]
-
-#     # Reduce the demand by what was allocated
-#     for id in only(F_basin_in.axes)
-#         if graph[id].subnetwork_id == subnetwork_id
-#             _, i = id_index(basin.node_id, id)
-#             demand[i] -= JuMP.value(F_basin_in[id])
-#         end
-#     end
-
-#     return nothing
-# end
 
 function adjust_demands!(
     allocation_model::AllocationModel,
@@ -669,27 +626,6 @@ function adjust_demands!(
     end
     return nothing
 end
-# function adjust_demands_flow!(allocation_model::AllocationModel, p::Parameters)::Nothing
-#     (; flow_demand, graph) = p
-#     (; problem, subnetwork_id) = allocation_model
-#     F = problem[:F]
-
-#     for (i, node_id) in enumerate(flow_demand.node_id)
-#         if graph[node_id].subnetwork_id != subnetwork_id
-#             continue
-#         end
-
-#         node_with_demand_id =
-#             only(outneighbor_labels_type(graph, node_id, EdgeType.control))
-
-#         flow_demand.demand[i] = max(
-#             0.0,
-#             flow_demand.demand[i] -
-#             JuMP.value(F[(inflow_id(graph, node_with_demand_id), node_with_demand_id)]),
-#         )
-#     end
-#     return nothing
-# end
 
 """
 Set the flow buffer of nodes with a flow demand to 0.0
@@ -998,9 +934,6 @@ function allocate_priority!(
             adjust_demands!(allocation_model, p, priority_idx, demand_node)
         end
     end
-    # adjust_demands_user!(allocation_model, p, priority_idx)
-    # adjust_demands_level!(allocation_model, p)
-    # adjust_demands_flow!(allocation_model, p)
     return nothing
 end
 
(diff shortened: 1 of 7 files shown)


pixi.lock
```diff
@@ -8458,28 +8458,6 @@ packages:
   license_family: BSD
   size: 15322
   timestamp: 1712542844430
-- kind: conda
```
SouthEndMusic, line 8461: Why are there changes to this file?

docs/core/allocation.qmd
```diff
@@ -14,13 +14,12 @@ within this *Allocation* section the main network is also considered to be a sub
 # The high level algorithm
 The allocation algorithm contains 3 types of optimization:
 
-- `internal_sources`, where flows are allocated within a subnetwork by only using sources inside the subnetwork;
```
SouthEndMusic, line 17: Maybe this documentation can remain unchanged, as it refers to the optimization types, not the names of the functions.

docs/contribute/addnode.qmd
```diff
@@ -16,6 +16,14 @@ struct NewNodeType <: AbstractParameterNode
     # Other fields
 end
 ```
+Another abstract class inherited from `AbstractParameterNode` is called `AbstractDemandNode`. For creating new node type used in allocation, define a struct:
```
SouthEndMusic, line 19:
```suggestion
Another abstract class which inherits from `AbstractParameterNode` is called `AbstractDemandNode`. For creating a new demand node type used in allocation, define a struct:
```

Nice that you thought of also adding this here (: I'm not sure 'inheritance' is the proper word to use here, as that word might be specific for object-oriented languages. Let's ask @visr.

core/test/allocation_test.jl
```diff
@@ -307,8 +306,7 @@ end
     # Collecting demands
     u = ComponentVector(; storage = zeros(length(basin.node_id)))
     for allocation_model in allocation_models[2:end]
-        Ribasim.allocate!(p, allocation_model, t, u, OptimizationType.internal_sources)
```
SouthEndMusic, line 310: Nice simplification here 👌

core/src/callback.jl
```diff
@@ -480,16 +480,15 @@ function update_allocation!(integrator)::Nothing
     # If a main network is present, collect demands of subnetworks
     if has_main_network(allocation)
         for allocation_model in Iterators.drop(allocation_models, 1)
-            allocate!(p, allocation_model, t, u, OptimizationType.internal_sources)
-            allocate!(p, allocation_model, t, u, OptimizationType.collect_demands)
+            collect_demands(p, allocation_model, t, u)
```
SouthEndMusic, line 483: Nice simplification!

SouthEndMusic (changes requested):
Nice work, I added some comments

pixi.lock
Jingru923: `pixi.lock` file is usually auto updated. Every one runs `pixi run ....`, pixi will check if our dependency has update and then update the lock file.

docs/contribute/addnode.qmd
visr: Yeah instead of "abstract class inherits", "abstract type subtypes" is perhaps more precise.

core/src/allocation_optim.jl
```diff
@@ -968,43 +983,72 @@ end
...
+    # When collecting demands, only flow should be available
+    # from the main to subnetwork connections
+    empty_sources!(allocation_model, allocation)
+
+    # Loop over the priorities
+    for priority_idx in eachindex(priorities)
+        optimize_priority!(allocation_model, u, p, t, priority_idx, optimization_type)
     end
+end
+
```
Jingru923, line 1030: This function is no longer mutable

core/src/allocation_optim.jl
```diff
@@ -968,43 +983,72 @@ end
...
+function allocate_demands(
```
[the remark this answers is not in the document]
Jingru923: This function is no longer mutable

core/src/allocation_optim.jl
```diff
@@ -968,43 +983,72 @@ end
...
     (; subnetwork_id) = allocation_model
     (; priorities, subnetwork_demands) = allocation
-    main_network_source_edges = get_main_network_connections(p, subnetwork_id)
 
-    if subnetwork_id == 1
-        @assert optimization_type == OptimizationType.allocate "For the main network no demands have to be collected"
+    ## Find internal sources
+    optimization_type = OptimizationType.internal_sources
+    set_initial_capacities_inlet!(allocation_model, p, optimization_type)
+
```
SouthEndMusic, line 999:
```suggestion
```

core/src/allocation_optim.jl
```diff
@@ -968,43 +983,72 @@ end
 Update the allocation optimization problem for the given subnetwork with the problem state
 and flows, solve the allocation problem and assign the results to the UserDemand.
 """
-function allocate!(
+function collect_demands(
```
SouthEndMusic, line 986:
```suggestion
function collect_demands!(
```

core/src/callback.jl
SouthEndMusic, line 483:
```suggestion
            collect_demands!(p, allocation_model, t, u)
```

core/test/allocation_test.jl
```diff
@@ -223,8 +223,7 @@ end
...
-        Ribasim.allocate!(p, allocation_model, t, u, OptimizationType.collect_demands)
+        Ribasim.collect_demands(p, allocation_model, t, u)
```
SouthEndMusic, line 226:
```suggestion
        Ribasim.collect_demands!(p, allocation_model, t, u)
```

core/test/allocation_test.jl
```diff
@@ -307,8 +306,7 @@ end
...
```
SouthEndMusic, line 309:
```suggestion
        Ribasim.collect_demands!(p, allocation_model, t, u)
```

docs/core/allocation.qmd
```diff
@@ -20,7 +20,7 @@ The allocation algorithm contains 3 types of optimization:
 
 The full algorithm goes through the following steps:
 
-1. Perform `internal_sources` followed by `collect_demands` for all subnetworks apart from the main network;
+1. Perform `collect_demands` for all subnetworks apart from the main network;
```
SouthEndMusic, line 23: I think this change is also not needed

SouthEndMusic (approved):
I added a few last nitpicks, but other than that it is good to go!

core/src/allocation_optim.jl
SouthEndMusic: nit

PR #1508, opened 2024-05-30
Smooth linear resistance flow rate bounding

SouthEndMusic:
I figured that the hard clamping of the flow rate from a linear resistance node could be hard on the solver, which is why I added a smoothing function for this. I figured this hard clamping could be the reason that @DanielTollenaar experienced problems with linear resistance nodes.


visr:
We should keep this in mind for when we run into this.

PR #1510, opened 2024-05-30
Fix a few issues found by JET

visr:
The first two are to handle `nothing`. The replacement of `outflow_ids` is just an optimization.


SouthEndMusic (approved):
I don't know what JET is, but looks good

PR #1513, opened 2024-05-30
Don't require unique node IDs

### LINKED ISSUE #1256: Validate uniqueness of Node table in core
Having a duplicate edge (#1255), results in a late error while writing results. We should add validation code to check for duplicates in the Node and Edge tables.

For Node this means the combination of `node_type` and `node_id` must be unique.
For Edge this means the combination of `from_node_type`, `from_node_id`, `to_node_type` and `to_node_id` must be unique.

### LINKED ISSUE #1318: Stop requiring globally unique node ids
We recently moved away from requiring globally unique node ids.
However, parts of our code still require it implicitly.
Because of that, we added checks to require it after all.

Tasks:
- [x] Make QGIS stop relying on `fid` for Node, #1306
- [x] Make QGIS stop relying on `fid` for Edge, #1310
- [x] Adapt optimization code to allow node ids that aren't globally unique #1262
- [ ] Adapt `to_xugrid`, https://github.com/Deltares/Ribasim/pull/1314#issuecomment-2019866629
- [ ] Revert https://github.com/Deltares/Ribasim/pull/1311
- [ ] Revert https://github.com/Deltares/Ribasim/pull/1316
- [x] Fix known bug in allocation: https://github.com/Deltares/Ribasim/pull/1330 
- [ ] Bring back tests: see what was reverted in https://github.com/Deltares/Ribasim/pull/1330/commits/d03526df8a4c62ce58edb7d67f13fd8ec6354dc1
- [ ] #1256

visr:
Fixes #1318
Fixes #1256

@evetion I didn't touch the Delwaq `node_lookup`. If only Basins are in the graph, then you can still use an ID as a unique index.

The #1256 fix in 72dcf24661fc82ee84d38666a43d113305a43feb is not tested, since Ribasim-Python stops us from creating such an invalid model. I manually confirmed it works by duplicating rows in QGIS.

The `main` methods were needed to fix `pixi run ribasim-core`.


core/src/main.jl
```diff
@@ -14,6 +14,9 @@ function run(config::Config)::Model
     return model
 end
 
+main(ARGS::Vector{String})::Cint = main(only(ARGS))
+main()::Cint = main(ARGS)
```
evetion, line 18: This doesn't seem part of this PR?

core/src/validation.jl
```diff
@@ -162,6 +162,21 @@ function valid_config(config::Config)::Bool
     return !errors
 end
 
+function valid_nodes(db::DB)::Bool
+    errors = false
+
+    sql = "SELECT node_type, node_id FROM Node GROUP BY node_type, node_id HAVING COUNT(*) > 1"
```
evetion, line 168: +1 for such nice SQL code

python/ribasim/ribasim/model.py
```diff
@@ -209,7 +199,7 @@ def _apply_crs_function(self, function_name: str, crs: str) -> None:
         self.crs = crs
 
     def node_table(self) -> NodeTable:
-        """Compute the full NodeTable from all node types."""
+        """Compute the full sorted NodeTable from all node types."""
```
evetion, line 202: Sorted on what? And does it matter for this function?

evetion (commented):
LGTM, some nitpicks. This will break the current assumptions in the Delwaq coupling scripts, so maybe not merge just yet?

core/src/main.jl
visr: Not really, but it was broken without, I needed it to run `pixi run ribasim-core ribasim.toml`. These lines were there before.

python/ribasim/ribasim/model.py
visr: `sort_keys = ["node_type", "node_id"]`. I thought that the sort order was documented for this table, but it isn't, so I'll add it to my doc PR.

It matters because if it is documented to be sorted we know we don't have to sort it ourselves after. I now rely on this behavior in the code.

evetion:
I'll make the changes so the coupling can handle the non-uniqueness. Would be good to also create a new coupling test that takes a horribly non-unique model to convert.

coupling/delwaq/generate.py
```diff
@@ -43,7 +43,7 @@ def generate(toml_path: Path) -> tuple[nx.DiGraph, set[str]]:
     for row in model.node_table().df.itertuples():
         if row.node_type not in ribasim.geometry.edge.SPATIALCONTROLNODETYPES:
             G.add_node(
-                row.node_id,
+                row.node_type + str(row.node_id),
```
visr, line 46: Not sure exactly what it's used for, but since it is a string representation, let's use the same one we use elsewhere, "Pump #5" instead of "Pump5".

```suggestion
                f"{row.node_type} #{row.node_id}",
```

coupling/delwaq/generate.py
```diff
@@ -52,7 +52,12 @@ def generate(toml_path: Path) -> tuple[nx.DiGraph, set[str]]:
             )
     for row in model.edge.df.itertuples():
         if row.edge_type == "flow":
-            G.add_edge(row.from_node_id, row.to_node_id, id=[row.Index], duplicate=None)
+            G.add_edge(
+                row.from_node_type + str(row.from_node_id),
+                row.to_node_type + str(row.to_node_id),
```
visr, line 57:
```suggestion
                f"{row.from_node_type} #{row.from_node_id},
                f"{row.to_node_type} #{row.to_node_id},
```

coupling/delwaq/generate.py
evetion: The string representation is only for uniqueness, and is removed soon after. Making it into the visualization equivalent implies that it is used as such later on, so I rather leave it as is.

Issue #1356, opened 2024-04-04
Use average flows over allocation intervals for `realized` in `allocation.arrow`


SouthEndMusic:
**What**
Use average flows over allocation intervals for `realized` in `allocation.arrow`.  In the case of level demand, use the difference in storage between the beginning and the end of the basin.

**How**
Integrate flows in `integrate_flows!` for the flows that are relevant for this allocation input, stored somewhere in the allocation data structure (make sure that this goes well in subnetworks where all optimizations are run twice). For the storage difference: use the current `u` and the previous one which can be taken from the `integrator` object (which is available in the `update_allocation!` function).

PR #1514, opened 2024-05-30
Compute mean realized flows

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1356.


SouthEndMusic:
Last to do: Add test for mean realized user demand

core/src/read.jl
```diff
@@ -1036,20 +1036,40 @@ function Allocation(db::DB, config::Config, graph::MetaGraph)::Allocation
...
+        level_demand_inflow =
+            (type == EdgeType.control) && (src_id.type == NodeType.LevelDemand)
+        flow_demand_inflow =
+            (type == EdgeType.flow) && has_external_demand(graph, dst_id, :flow_demand)[1]
+
+        if user_demand_inflow || flow_demand_inflow
+            mean_realized_flows[edge] = Ref(0.0)
+        elseif level_demand_inflow
+            mean_realized_flows[(dst_id, dst_id)] = Ref(0.0)
         end
```
visr, line 1073: Can you explain why you went for Ref here instead of a Float? Is it for the ergonomics of being able to do

```julia
value = dict[key]
value[] += x
value[] /= dt
```

```julia
value = dict[key]
value += x
value /= dt
dict[key] = value
```

i.e. having to put it back in? I worry a bit about the performance of this approach.

SouthEndMusic: I got rid of the use of `RefValue{Float64}`, but now there are more dictionary lookups, e.g.

https://github.com/Deltares/Ribasim/blob/8e337aa1c00b80ca1cbc3be52117bc3f903dacaf/core/src/callback.jl#L487-L500

Is there a way to circumvent this?

visr: Dict lookups are quite fast, so I doubt it's a serious issue. I haven't looked into it, but perhaps the compiler can figure out it can optimize this with a single lookup. Also I just tried this, and surprisingly it works. Could be of course that there are still two lookups behind this syntax:

```
julia> dict = Dict{Int, Int}(1 => 1)
Dict{Int64, Int64} with 1 entry:
  1 => 1

julia> dict[1] += 1
2

julia> dict
Dict{Int64, Int64} with 1 entry:
  1 => 2
```

visr (approved):

Issue #207, opened 2023-05-12
One documentation page per node type


visr:
Currently we have two pages that contains many headers with the different node types. The usage page focuses on the required input data, and the equations on the underlying math. Especially the equations is incomplete right now, it is easily forgotten. It would be better to have a page that lists all node types in a table or list with a one line description, symbol and other essentials. Then from there you can click the node type to reach a page with all information related to that node type. This should probably start with what it does and when to use it, followed by the input tables and equations.

https://deltares.github.io/Ribasim/core/usage.html
https://deltares.github.io/Ribasim/core/equations.html

When done the template of #205 should be updated as well.

- [ ] reorganization structure
- [ ] make template
- [ ] when to use what node type --> see also #1158 
- [ ] checks+improvements storylines

edit:
- [ ] mermaid diagrams on each of the node pages that shows all possible connections and flow directions
- [ ] put the styling icon from QGIS/testmodels to the individual nodepages

PR #1522, opened 2024-06-03
Split usage to separate pages

visr:
Part of https://github.com/Deltares/Ribasim/issues/207.

This moves the information out of the usage page to separate pages. For some node types I added information on the top of the page, others already had this, and I just moved it out of the schema section.

The pages are grouped together in the sidebar. By default they are not collapsed. This can be controlled by the collapse level in Quarto, see https://quarto.org/docs/websites/website-navigation.html#hybrid-navigation.

image

Since this is already quite big, I prefer to not add the equations in this step as well.


evetion:
Keep in mind the https://docs.divio.com/documentation-system/structure/ system to refactor the documentation.

evetion (approved):

visr:
Yeah I need to think a bit about that. I see these pages mainly as reference.

https://diataxis.fr/

evetion:
Based on our discussion yesterday, the validation table which node can connect to what would be greatly helped by mermaid diagrams on each of the node pages that shows all possible connections and flow directions.

Another idea, can we put the styling icon from QGIS/testmodels to the individual nodepages?

visr:
Yeah I think these are both good ideas, perhaps add them to #207?

My idea was that it would be nice that pretty much anything related to node X should be on the X page. Therefore users should not have to go to a separate page to figure out what they can connect it to.

Issue #1512, opened 2024-05-30
Performance assessment/improvement based on AGV model


gijsber:
HKV created a running model for the Amstel, Gooi, Vecht area (the AGV model) with many pumps and discrete controllers.

We should profile this model and improve performance where logical
https://deltares.thegood.cloud/apps/files/?dir=/Ribasim%20modeldata/AmstelGooienVecht/modellen&fileid=56373

PR #1529, opened 2024-06-04
Discrete Control performance improvements based on AGV model

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1512
Fixes https://github.com/Deltares/Ribasim/issues/1539


SouthEndMusic:
This could be the topic of the next knowledge sharing, but here's a quick view of the profiles of the AGV model (4 simulated days).

On the current main:

profile_main

On this branch:
profile_now

SouthEndMusic:
Here's a more elaborate breakdown of the changes to discrete control in this PR.

### Avoiding allocation

This is a straightforward one, there was memory allocation each time a truth state (a string of 'T' and 'F' characters) was determined:

https://github.com/Deltares/Ribasim/blob/89b8574a3ccf7e7ae8c669394a269c2569b9eca3/core/src/callback.jl#L295-L303

To get around this, I pre-allocate memory for the longest possible truth state and use that every time a truth state is determined:

https://github.com/Deltares/Ribasim/blob/77a2f0de9c51decd865ef92146f6910b68b6cd12/core/src/read.jl#L757-L763

I also changed the truth states from strings to Boolean vectors, which is possible now that we don't have the 'U' and 'D' truth values anymore. The input and output are still written as strings of 'T' and 'F'.

### Type stability

This is a more interesting one. This part of the was very detrimental to performance:

https://github.com/Deltares/Ribasim/blob/89b8574a3ccf7e7ae8c669394a269c2569b9eca3/core/src/callback.jl#L396-L412

That is because this code is type unstable; a priori the type of `node` is unknown, `new_state` is a tuple of arbitrary contents, and so the type of `vec` is unknown too.

To get around this, I:

1. Created a `ParameterUpdate` type, which contains the data for a control state:

https://github.com/Deltares/Ribasim/blob/77a2f0de9c51decd865ef92146f6910b68b6cd12/core/src/parameter.jl#L193-L214

2. Combined all control mappings from the different node types into one:

https://github.com/Deltares/Ribasim/blob/77a2f0de9c51decd865ef92146f6910b68b6cd12/core/src/read.jl#L1210-L1215

3. Created parameter update functions for all node types that support being controlled, which all get called when a control state change occurs but only the one for the right node type does something:

https://github.com/Deltares/Ribasim/blob/77a2f0de9c51decd865ef92146f6910b68b6cd12/core/src/callback.jl#L401-L526

### Notes on complexity and maintainability

The code above makes the discrete control code more explicit, which I think makes it easier to understand but also increases maintenance load. For instance, if we want to support discrete control for a new variable or node type, this needs to be added to `ParameterUpdate` and the corresponding `parameter_update!` method. The `parameter_update!` methods do allow for node/variable specific functionality regarding updating variables.

Maybe the discrete control code should also get its own script (again).

gijsber:
nice explanation. thanks!

evetion:
Funny thing to play with, slightly relevant:

```julia
julia> macro b_str(v); collect(v).=='T'; end
@b_str (macro with 1 method)

julia> b"TTF"
3-element BitVector:
 1
 1
 0
```

core/src/callback.jl
```diff
@@ -289,18 +289,23 @@ function discrete_control_affect!(integrator, compound_variable_idx)
...
-    )
-    truth_state = join(truth_values, "")
+    # Get the first index of the (compound) variables that this control node listens to
+    variable_idx = searchsortedfirst(discrete_control.node_id, discrete_control_node_id)
+    truth_value_idx = 1
+
+    # Build up the truth state from the truth values per (compound) variable per greater than
+    while variable_idx <= length(discrete_control.node_id) &&
+        discrete_control.node_id[variable_idx] == discrete_control_node_id
+        truth_values_variable = discrete_control.condition_value[variable_idx]
```
evetion, line 299: Can use a blank line to distinguish this code from the while clause

core/src/callback.jl
```diff
@@ -289,18 +289,23 @@ function discrete_control_affect!(integrator, compound_variable_idx)
...
-        ]...;
-        dims = 1,
-    )
-    truth_state = join(truth_values, "")
+    # Get the first index of the (compound) variables that this control node listens to
+    variable_idx = searchsortedfirst(discrete_control.node_id, discrete_control_node_id)
+    truth_value_idx = 1
+
+    # Build up the truth state from the truth values per (compound) variable per greater than
+    while variable_idx <= length(discrete_control.node_id) &&
```
evetion, line 297: I think it's much more readable to for loop over these nodes, instead of jumping with a searchsortedfirst and having a while loop?

core/src/callback.jl
```diff
@@ -393,22 +398,131 @@ function set_fractional_flow_in_allocation!(
...
+            flow_rate_scalar,
+            outlet.min_flow_rate[node_idx],
+            outlet.max_flow_rate[node_idx],
+        )
+    end
+    return nothing
+end
 
-        # Set new fractional flow fractions in allocation problem
-        if is_active(p.allocation) && node isa FractionalFlow && field == :fraction
```
evetion, line 408: So this was clearly a red flag in the old code, although I like the genericity of the rest.

core/src/read.jl
```diff
@@ -602,7 +657,7 @@ function parse_variables_and_conditions(compound_variable, condition)
     weight = Vector{Float64}[]
     look_ahead = Vector{Float64}[]
     greater_than = Vector{Float64}[]
-    condition_value = BitVector[]
+    condition_value = Vector{Bool}[]
```
evetion, line 660: Why the change? A `Vector{Bool}` is actually 8 times larger in memory than `BitVector`. Performance should also be equal or better (largely) because of that.

core/src/callback.jl
```diff
@@ -393,22 +398,131 @@ function set_fractional_flow_in_allocation!(
...
+    end
+    return nothing
+end
+
+function set_control_params!(p::Parameters, node_id::NodeID, control_state::String)::Nothing
+    parameter_update = p.discrete_control.control_mapping[(node_id, control_state)]
+
+    # Only the method corresponding to the type of the node_id actually updates parameters,
+    # but this way there is no runtime dispatch on node type
+    update_parameters!(p.pump, parameter_update)
```
evetion, line 518: If only we knew the type in `parameter_update`, so we don't need to call all these functions here ;)

core/src/callback.jl
```diff
@@ -393,22 +398,131 @@ function set_fractional_flow_in_allocation!(
     return nothing
 end
 
-function set_control_params!(p::Parameters, node_id::NodeID, control_state::String)
-    node = getfield(p, p.graph[node_id].type)
-    idx = searchsortedfirst(node.node_id, node_id)
-    new_state = node.control_mapping[(node_id, control_state)]
+function update_parameters!(pump::Pump, parameter_update::ParameterUpdate)::Nothing
+    if parameter_update.node_type != NodeType.Pump
```
evetion, line 402: This looks like a red flag to me, ideally we only call this with the correct data, so ParameterUpdate is too generic/uniony?

core/src/callback.jl
```diff
@@ -393,22 +398,131 @@ function set_fractional_flow_in_allocation!(
...
+        return nothing
+    end
+    (; node_idx, active, flow_rate_scalar) = parameter_update
+    pump.active[node_idx] = active
+    if !isnan(flow_rate_scalar)
```
evetion, line 407: Why do we need to check for isnan? Not sure where it comes from, but ideally we don't even trigger when we have nan data?

core/src/read.jl
```diff
@@ -230,6 +232,57 @@ function initialize_allocation!(p::Parameters, config::Config)::Nothing
...
 end
 
+"""
+Constructor for the ParameterUpdate object. Scalar Float64 fields have default value
+NaN and ScalarInterpolation values have default value LinearInterpolation(Float64[], Float64[])
+(an empty interpolation object). These default values get ignored in DiscreteControl.
+Non-supported key word inputs do not raise an error but get ignored via kwargs....
+"""
+function ParameterUpdate(
+    node_type::NodeType.T,
```
evetion, line 242: Might I suggest some bedtime reading material? https://docs.julialang.org/en/v1/manual/types/#Parametric-Types

I think reading that might trigger a slight refactor of this struct and related code.

core/test/control_test.jl
```diff
@@ -11,16 +11,16 @@
...
-        (NodeID(:DiscreteControl, 6), "F") => "active",
-        (NodeID(:DiscreteControl, 5), "TF") => "off",
-        (NodeID(:DiscreteControl, 5), "FF") => "on",
-        (NodeID(:DiscreteControl, 5), "FT") => "off",
-        (NodeID(:DiscreteControl, 6), "T") => "inactive",
+    @test pump_control_mapping[(NodeID(:Pump, 4), "off")].flow_rate_scalar == 0
+    @test pump_control_mapping[(NodeID(:Pump, 4), "on")].flow_rate_scalar == 1.0e-5
+
+    logic_mapping::Dict{Tuple{NodeID, Vector{Bool}}, String} = Dict(
+        (NodeID(:DiscreteControl, 5), [true, true]) => "on",
```
evetion, line 18: You might want to use my suggested string macro to keep the "TT" in tests for readability. Up to you.

core/src/callback.jl
```diff
@@ -393,22 +398,131 @@ function set_fractional_flow_in_allocation!(
...
+    return nothing
+end
+
+function set_control_params!(p::Parameters, node_id::NodeID, control_state::String)::Nothing
+    parameter_update = p.discrete_control.control_mapping[(node_id, control_state)]
+
+    # Only the method corresponding to the type of the node_id actually updates parameters,
+    # but this way there is no runtime dispatch on node type
+    update_parameters!(p.pump, parameter_update)
+    update_parameters!(p.outlet, parameter_update)
```
evetion, line 519: Consider renaming either the function/argument/struct to indicate the parameter update happen because of control.

evetion (changes requested):
Nice to see another performance increase! What are the absolute runtimes/relative part of control of/in the overall runtime?

Seems like some good changes, but I wonder what you think of your solution using the `ParameterUpdate` struct. I see two red flags in its design, that you have to check for nan, and the check for the correct type. Let me know if you want help on refactoring it further.

core/src/read.jl
SouthEndMusic: I know about parametric types, that could at least resolve the `flow_rate` problem. I did actually make a parametric version of the `ParameterUpdate` struct at some point, but eventually I wanted it to be non-parametric so that I could put all their instances into a single type stable dictionary as mentioned above. I realize now that that is not needed for a type stable implementation as we can access a node type specific object containing the parameters of a control state  within the corresponding `update_parameters!` method. 

In response to your comment that `ParameterUpdate` is too uniony: I agree. I actually tried using the same field of `ParameterUpdate` for different variables for different node types, but that makes the code unnecessarily hard to understand. So the better solution is to make a `ParameterUpdate` struct for each node type. 

The point of indicating that not all controllable parameters are part of the control state remains, or we have to rethink that possibility. We could also say that unspecified optional parameters in a control state always get their default value, I'm not sure how that works in the code at the moment actually.

core/src/callback.jl
```diff
@@ -319,7 +324,7 @@ function discrete_control_affect!(integrator, compound_variable_idx)
 
         push!(record.time, integrator.t)
         push!(record.control_node_id, Int32(discrete_control_node_id))
-        push!(record.truth_state, truth_state)
+        push!(record.truth_state, convert_truth_state(truth_state))
```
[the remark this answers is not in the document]
SouthEndMusic: Those pushes aren't that bad actually, but `convert_truth_state` is quite slow

core/src/callback.jl
```diff
@@ -289,18 +289,23 @@ function discrete_control_affect!(integrator, compound_variable_idx)
...
+        truth_values_variable = discrete_control.condition_value[variable_idx]
+        n_greater_than = length(truth_values_variable)
+        discrete_control.truth_state[truth_value_idx:(truth_value_idx + n_greater_than - 1)] .=
+            truth_values_variable
+        variable_idx += 1
+        truth_value_idx += n_greater_than
+    end
+
+    # The maximum truth state length is possibly shorter than this truth state
+    truth_state = view(discrete_control.truth_state, 1:(truth_value_idx - 1))
```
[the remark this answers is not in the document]
SouthEndMusic: doctest?

core/src/callback.jl
```diff
@@ -297,31 +297,50 @@ function discrete_control_affect!(integrator, compound_variable_idx)
...
-            error(
-                "No control state specified for $discrete_control_node_id for truth state $truth_state.",
-            )
-        end
+    control_state_new = get(
+        discrete_control.logic_mapping,
+        (discrete_control_node_id, truth_state),
+        nothing,
+    )
+    isnothing(control_state_new) && error(
```
SouthEndMusic, line 341: Nice lookup reduction

core/src/callback.jl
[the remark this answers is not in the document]
evetion: Yes, providing an example (and testing it) as part of the docstring https://documenter.juliadocs.org/stable/man/doctests/.

core/src/callback.jl
evetion: @SouthEndMusic Looking through the code, I don't think this can ever be nan? While it looks like it because of the ParameterUpdate struct, it is actually set for the correct nodetype using `parse_static_and_time`. In the AGV model any asserts didn't trigger on nan.

evetion:
I have refactored several smaller things several times, but have not found significant improvements, apart from the commit I did. Things that do not pan out:
- Threading in our control loop(s)
- NodeID to Int32 for the control state dictionary lookup 
- Symbol instead of String for variable for lookups and comparisons
- Parametric struct with NamedTuple field. These caused some runtime dispatch, but were not really slower, but it still required most of the boilerplate in this PR. Not sure if the extra complexity of such a struct is worth it, so I kept it as is.

This PR is 2-3x faster than the current main, mostly because we went from 35M to 2M allocations. Latest flamegraph with two observations:
Screenshot 2024-06-07 at 08 42 47

Control takes roughly a third of the runtime (originally 70%). Of that time, a third is spent in purely looping over all control conditions (`discrete_control_condition`), of which only a small part is spent in looking up values(!). Might be improved by a different memory structure. The other highlight is the get function, spending 12% of the control time retrieving the new control state from the dict with a complex key.

Wrapping up, getting this faster requires a more significant refactor. Besides, I believe the current code *needs* a refactor, as it introduced a lot of boilerplate, with repetition and fields also existing elsewhere. This is discussed partially in #1539. Depending on the priority of getting the AGV model faster, we can either merge this, and promise to follow up soon, or keep this open until we refactor.

evetion:
Based on the discussion of today, let's refactor while we can! We see several steps, each one slightly more complex than the last, would be good to implement them one by one, and see how performance/complexity behaves. It's fine to stop after step 1 or 2 if this takes too long for too little benefit.

1. Refactor each Node to contain change `control_mapping::Dict{Tuple{NodeID, String}, ParameterUpdate}` to a `control_mapping::Dict{Tuple{NodeID, String}, NamedTuple{specific to nodetype}`. This will remove the ParameterUpdate struct, and would simplify several `discrete_control_parameter_update` to a single `discrete_control_parameter_update(node::AbstractParameterNode, i)` (and one specific to fractionalflow).
2. In `discrete_control_affect!`, we can shortcut before the dict lookup by checking whether the `truth_state` is identical to the current truth state?
3. Think of a way to reduce the ifelses of `set_control_params!`. Sorry, we didn't discuss this. I feel that each NodeID should have a direct access to its Node struct (@visr a structview?), instead of the weird lookups we do now.
4. Remove the need for Dict(s) like `logic_mapping` and `control_mapping`. This requires refactoring in how we `read` the initial structure, but could be akin to how previous perfomance by caching. For example, the control state, now a string and part of the dict that gives you a state could be replaced by a vector of states, and storing the index in that vector instead of the string. The same thing could be done on a NodeID level, the `control_mapping` could be a Vector{Vector{ParameterUpdate}}, aligning with the NodeIDs in the `node_id` field. On `read` we could store the logical index to the NodeID instead of the NodeID itself.
5. Test changing the memory layout of `discrete_control_condition!` from SoA to AoS. SoA should be faster in single variable operations, but since we access the whole struct all the time, might not be. This might be informative on the topic: https://viralinstruction.com/posts/hardware/.

SouthEndMusic:
@evetion I also now implemented your point 2, now I am down to a runtime of ~2.7 s. I'm now also wondering whether I need this whole part at all:

https://github.com/Deltares/Ribasim/blob/ace0b10fe4080f4657adab896759fa5052c2605a/core/src/callback.jl#L304-L341

That is, why have separate truth values per variable and then copy them to the full truth state. I think this was relevant when we considered that variables could be listened to by multiple discrete control nodes, but now (compound) variables are discrete control node specific so we don't need that anymore. Probably the change in truth state can already be detected in `discrete_control_condition!`.

SouthEndMusic:
Tests now probably fail because a control state change check should be done at the beginning of the simulation regardless of truth state change.

Some thoughts on refactoring the `logic_mapping` and `control_mapping` dicts: split them out by node ID so that this doesn't have to be part of the keys. For `logic_mapping`: the boolean vector input can be converted to an integer index via binary. This index would be for a sparse vector of strings (because not every possible truth state must have a control state). That's probably still slow, so the strings should be converted to an index too, which has to be coordinated properly between the logic mapping and control mappings. The string control state name must still be available for the output data.

SouthEndMusic:
Another big step in the refactor: the functions `discrete_control_affect!` and `discrete_control_condition!` have been merged into one function and one loop, which also got rid of the need to copy groups of truth values. Data for a single compound variable has been clustered in the `CompoundVariable` struct, which makes clear which things belong together and I think makes the code a bit more readable. Current runtime of the AGV model is ~2.4s.

I propose to leave it here for now regarding this PR and the refactoring of discrete control. One thing left to do is a round of updating inline documentation.

core/src/callback.jl
```diff
@@ -195,22 +195,31 @@ function save_vertical_flux(u, t, integrator)
 end
 
 """
-Apply the discrete control logic. Each DiscreteControl node has one or more variables associated with it,
-and these variables are a linear combination of one or more state/time derived parameters of the model.
-The values of these variables are compared against 'greater_than' thresholds. These conditions yield a 'truth_state'
-consisting of 'truth_values' for a DiscreteControl node. This truth_state then corresponds with a 'control_state',
-and nodes controlled by a DiscreteControl node have parameter values associated with this control_state.
+Apply the discrete control logic. There's somewhat of a complex structure:
+- Each DiscreteControl node can have one ore multiple compound variables it listens to
```
evetion, line 199:
```suggestion
- Each DiscreteControl node can have one or multiple compound variables it listens to
```

evetion (approved):
Well done, it's more readable now. Besides, you coded even more subtle performance improvements.

PR #1525, opened 2024-06-04
Add subgrid levels results `subgrid_level.arrow` to docs

Jingru923:
Fixes #1126

PR #1538, opened 2024-06-06
Don't write empty result tables

visr:
Came up in https://github.com/Deltares/Ribasim/pull/1525#discussion_r1627515435

At first I thought it would be better to write e.g. `subgrid_levels.arrow` if `Basin / subgrid` is defined. That would still write an empty file if the input table is empty or there were no timesteps. But just never writing empty tables is much easier to implement and explain, and I expect the difference not to matter in practice.

### DIFF
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -176,7 +176,7 @@ function Model(config::Config)::Model
     end
 
     model = Model(integrator, config, saved)
-    write_results(model)  # check permissions
+    write_results(model)  # check whether we can write results to file
     return model
 end
 
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -174,7 +174,9 @@ function Model(config::Config)::Model
         set_initial_allocation_mean_flows!(integrator)
     end
 
-    return Model(integrator, config, saved)
+    model = Model(integrator, config, saved)
+    write_results(model)  # check permissions
+    return model
 end
 
 "Get all saved times in seconds since start"
--- a/core/src/write.jl
+++ b/core/src/write.jl
@@ -330,6 +330,10 @@ function write_arrow(
     table::NamedTuple,
     compress::Union{ZstdCompressor, Nothing},
 )::Nothing
+    # Don't write empty tables
+    if isempty(table.time)
+        return nothing
+    end
     # ensure DateTime is encoded in a compatible manner
     # https://github.com/apache/arrow-julia/issues/303
     table = merge(table, (; time = convert.(Arrow.DATETIME, table.time)))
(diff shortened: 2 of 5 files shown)


core/src/write.jl
```diff
@@ -328,14 +329,31 @@ end
...
     # ensure DateTime is encoded in a compatible manner
     # https://github.com/apache/arrow-julia/issues/303
     table = merge(table, (; time = convert.(Arrow.DATETIME, table.time)))
     metadata = ["ribasim_version" => string(pkgversion(Ribasim))]
     mkpath(dirname(path))
-    Arrow.write(path, table; compress, metadata)
+    try
+        Arrow.write(path, table; compress, metadata)
+    catch
+        @error "Failed to write results, it may be locked." path
```
evetion, line 354: nit: Re-use string from l342? And maybe (lazy) interpolate the string and include the (now not-catched) error message? Like: `"Failed to write results to $(path): $e"`?

core/src/model.jl
```diff
@@ -175,7 +175,9 @@ function Model(config::Config)::Model
         set_initial_allocation_mean_flows!(integrator)
     end
 
-    return Model(integrator, config, saved)
+    model = Model(integrator, config, saved)
+    write_results(model)  # check permissions
```
evetion, line 179:
```suggestion
    write_results(model)  # check whether we can write results to file
```

core/src/write.jl
```diff
@@ -328,14 +329,31 @@ end
 function write_arrow(
     path::AbstractString,
     table::NamedTuple,
-    compress::Union{ZstdCompressor, Nothing},
+    compress::Union{ZstdCompressor, Nothing};
+    init::Bool,
```
evetion, line 333: Could be false by default?

core/src/write.jl
```diff
@@ -328,14 +329,31 @@ end
...
 )::Nothing
+    # At the start of the simulation, write an empty table to ensure we have permissions
+    # and fail early.
+    # At the end of the simulation, write all non-empty tables, and remove existing empty ones.
+    if !init && isempty(table.time)
```
evetion, line 338: The order of simulation and the inline documentation are different from the code order. This confused me for a bit. Partially because the init keyword is only used inversely, to indicate the end of the simulation. Maybe rename the keyword to `remove_empty_table`?

evetion (changes requested):

core/src/write.jl
visr: That is a slightly different error, so I prefer a different message. Other comments are all fixed in db1ed9201e9221980caa1d7bb71f0dc02c8be466

Issue #1491, opened 2024-05-21
Update pixi to 0.22+ on TeamCity


visr:
TeamCity still has pixi 0.20.1, and the current pixi 0.22 has a higher lockfile format 5, so we cannot merge this until TeamCity is updated for Windows and Linux. @Jingru923 can you set that process in motion?

```
  pixi 0.20.1
13:49:27     Ã— Failed to load lock file from `D:
13:49:27     â”‚ \buildAgent\work\ecd2b8f9b25b1609\ribasim\pixi.lock`
13:49:27     â•°â”€â–¶ found newer lockfile format version 5, but only up to including version
13:49:27         4 is supported.
```

I'll do an update with an older version manually since we do seem to need some updates.

_Originally posted by @visr in https://github.com/Deltares/Ribasim/issues/1487#issuecomment-2122542973_

visr:
I see pixi v0.22 in the Windows TeamCity logs now.

PR #1542, opened 2024-06-07
Add ipykernel to pixi deps

evetion:
Not sure if manifest version 5 will cause trouble.

I use VS Code for interactive development, which requires ipykernel. Since VS Code uses the pixi env, I always end up adding ipykernel in the past months. Seems like a good developer addition.

Similarly, I always limit the pyarrow version, as I otherwise get a 
```python
  File "/Users/evetion/code/Ribasim.jl/.pixi/envs/dev/lib/python3.12/site-packages/pandas/compat/pyarrow.py", line 10, in <module>
    _palv = Version(Version(pa.__version__).base_version)
                            ^^^^^^^^^^^^^^
AttributeError: module 'pyarrow' has no attribute '__version__'
```

I'm not sure why the CI doesn't reproduce this.


evetion:
Ah the manifest needs #1491

visr (approved):
Seem fine to me. I'm a bit surprised about ipykernel, I use interactive Python in VSCode all the time but don't run into this. But the [VSCode docs](https://code.visualstudio.com/docs/datascience/jupyter-kernel-management#_python-environments) do mention it needs an ipykernel install, so let's add it.

I've also run into the pyarrow version issue. So far re-installing the pixi environment worked for me. But capping the version is probably more robust for now.

Issue #1482, opened 2024-05-18
Use `FindFirstFunctions` for lookups in sorted lists


SouthEndMusic:
Functions from [this package](https://docs.sciml.ai/FindFirstFunctions/dev/) are internally used in `DataInterpolations`, so if lookups in sorted lists become a performance bottleneck for us we can use them too.

Issue #566, opened 2023-08-30
Smoother interpolation of basin profile and Q(h) relation


SouthEndMusic:
`DataInterpolations.jl`, which we currently use for linear interpolation of Q(h) relations, supports all sorts of other interpolation types which are explained very minimally in [the docs](https://docs.juliahub.com/DataInterpolations/GwjmX/4.0.1/). We can modify the code so that it becomes very easy to test different types of interpolation, currently primarily for the stability/convergence of large models.

For Q(h) relations the interpolation type can probably easily be changed, although care has to be taken with extrapolation behavior.

The interpolation of the basin profile can be done much cleaner than it is now, getting rid of `get_area_and_level`. I propose that we use an interpolation object for the level as a function of the storage $h(S)$, and then the area for a given level can be obtained as $\frac{1}{h'(S)}$, either by the `derivative` function of `DataInterpolations.jl` or via `ForwardDiff.jl`. Also here care has to be taken with extrapolation.

SouthEndMusic:
Also update `numerics.qmd` (which has not been merged at the time of writing).

visr:
This may help: https://github.com/SouthEndMusic/SmoothInterpolation.jl

Fati-Mon:
It mentions that this will be updated once this issue is resolved, is that still correct? 
If so, can we add more explanation in https://deltares.github.io/Ribasim/concept/numerics.html#basin-profiles ?

PR #1543, opened 2024-06-10
Incorporate SmoothInterpolation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1482
Fixes https://github.com/Deltares/Ribasim/issues/566

- Currently this doesn't work remotely yet, because I've installed `SmoothInterpolation` with a local path. `SmoothInterpolation` can be installed with the package manager once it is registered: [New package: SmoothInterpolation v0.1.0 by JuliaRegistrator · Pull Request #108572 · JuliaRegistries/General (github.com)](https://github.com/JuliaRegistries/General/pull/108572)
- If we actually use this, it would probably make sense to give you admin rights on [SouthEndMusic/SmoothInterpolation.jl (github.com)](https://github.com/SouthEndMusic/SmoothInterpolation.jl)
- Whether it is actually useful to use smoothing of the linear interpolations remains to be seen. Currently the implementation uses a sampled `SmoothedLinearInterpolation` which is a `LinearInterpolation`. The only new interpolation type used in the simulation is `LinearInterpolationIntInv`, which is a much more concise implementation of what was in `get_area_and_level`
- One takeaway of this PR regardless of the use of `SmoothedLinearInterpolation` is that we can get rid of quite some utility code, also partly because of making more use of derivatives and integrals via `DataInterpolations.jl`


SouthEndMusic:
I smelled something fishy when I found out that on this branch the AGV model takes 3-4x longer than on main. No tests are failing, and noting remarkable shows up in the profile. So I investigated a bit, and I found that the profile for certain basins behaves weird in the old code:

weird_plot

I haven't figured out what triggers this yet, but apparently it doesn't happen in our test models, at least not where tests depend on it. I'm pretty sure the new code (in `SmoothInterpolation`) is more correct and robust because it is tested more elaborately.

visr (approved):
Great that we can get rid of so much utility code, and have a more solid basis to work on.

PR #1546, opened 2024-06-10
Convert TeamCity xml configuration files to Kotlin

Manangka:
Fixes #

This commit replaces the the TeamCity configuration files written in xml with the newer format written in Kotlin.
This has several advantages:
- Make the configuration files easier to read
- Makes it easier to recognize duplicate code and extract them to templates
- Makes it possible to share template between teams that also uses Kotlin
- Makes it possible to write more complex configurations because the configuration files can now contain logic

The current versioned settings of Ribasim are set to the xml format. Therefor I couldn't test these changes in the existing Ribasim project. Instead I created a separate Ribasim_Sandbox project that is set to use the Kotlin format. The build seems ok. The reason I say ok is because I'm not able to test everything due to missing credentials which you cant copy between projects

When merging this commit to main we have to set the settings format in the versioned settings tab to Kotlin otherwise the builds will fail


evetion (approved):
This makes more code readable, before I could just skip it because it was xml...

.teamcity/Ribasim/buildTypes/Ribasim_MakeGitHubRelease.kt
```diff
@@ -0,0 +1,108 @@
...
+    cleanup {
+        keepRule {
+            id = "KEEP_RULE_10"
+            keepAtLeast = allBuilds()
+            applyToBuilds {
+                withStatus = successful()
+            }
+            dataToKeep = everything()
+            applyPerEachBranch = true
+            preserveArtifactsDependencies = true
```
evetion, line 102: @Manangka This is the culprit in my issues (as I copied it stupidly 😄).  This rule deletes everything that is not successful. The previous XML seemed to have just disabled the cleanup, so I'm not sure why you introduced this extra rule.

Manangka: I've never seen this code snippet before :open_mouth:.
I'ts not in the imod-python so I wonder how it got here

Manangka: Ah I see, this is the PR in which the exisitng TeamCity project is converted to Kotlin. So the change was already there in the existing project

PR #1541, opened 2024-06-07
Refactor docs to follow diataxis concept

evetion:
From the refactor it's quite clear we're missing tutorials, and some API documentation. I also noticed slight overlaps between usage/model concepts.

I didn't write *new* documentation, only moved pages/paragraphs around and fixed the links I could find. Furthermore, I copied the README.md (but removing the prototype sentence) to serve as a documentation intro, making it slightly like the quarto website [itself](https://quarto.org/), as I found the current homepage quite dry.

This will incur merge conflicts for any documentation PR, so I say we merge this quickly, and create new documentation in subsequent PRs.

PR #1548, opened 2024-06-11
Restructure docs a bit more

visr:
Follow-up to #1541
Fixes #1548


docs/_quarto.yml
```diff
@@ -9,66 +9,55 @@ website:
...
-      - text: "Guides"
-        file: guides/index.qmd
-      - text: "Background"
-        file: background/concept.qmd
+      - text: "Tutorials"
+        file: tutorial/index.qmd
+      - text: "How-to guides"
+        file: guide/index.qmd
+      - text: "Explanation"
+        file: explanation/index.qmd
```
evetion, line 17: I don't mind changing any of the titles, but note I consciously chose different names than how diataxis names/splits its system (which is for writers of documentation, not end-users). Indeed, this is fine per [diataxis](https://diataxis.fr/explanation/#explanation-and-understanding):

> Explanation by any other name:
>
>	Your explanation documentation doesn’t need to be called Explanation.
>
>	Alternatives include:
>		
>	- Discussion 
>	- Background
>	- Conceptual guides
>	- Topics

In the different names, I tried to find names that (I think) are clear to end-users, and took inspiration from a page like https://quarto.org/. Maybe @gijsber can comment on the preferred names (feel free to suggest others)?

`Tutorials` or `Get Started`
`Guides` or `How-to guides`
`Background` or `Explanation`

This PR removes also the About header, which is the same as the logo, but for me it's not clear the logo is a link to the home page. I prefer to add it back (can be another name, such as Home).

evetion (changes requested):

docs/_quarto.yml
gijsber: My suggestions would be
`Getting Started`
`How-to`
`Background`
`Reference`

visr: I'm fine with adding back the home page, though perhaps "Overview" like in https://quarto.org/ is more accurate and less internet-lingo, especially if we want to make PDF docs out of this as well.

I should've included my reasoning for renaming in the PR description, but I didn't have time anymore. They are:

- Tutorials: Diátaxis says they are for learning. This may include advanced topics in tutorial format. For that "Getting started" does not seem the right title. Quarto.org uses this and then combines a How-to install guide with some tutorials. Installation is a more of a guide, not a tutorial.
- How-to guides: They are guides that tell you how to do something. I think it's clearer to write it out, it's not long.
- Explanation: I feel like "Background" is much narrower that "Explanation". Some of the content there is background information, but not all that is needed for understanding is background information. Also background seems less important.

gijsber: ok. Fine with your explanation to go for `Tutorials`, `How-to guides`.
Although I am fine with `Explanation`, we might also consider if `Understanding` is an appropriate alternative. Other words, like `Concepts` can be too narrow and `Background` may indeed not trigger people to read and understand.

visr: Understanding is the goal. But we cannot do that, only provide a good explanation and hope it leads to understanding. I think that is why it is not in the alternative names list that @evetion linked to.

gijsber:
Should the Validation part in Concepts/implementation not be moved to the refernce section, possibly under a different name `Network connectivity`or `Network connectivity validation rules`

Issue #1439, opened 2024-04-29
Python reference docs are almost empty


visr:
https://deltares.github.io/Ribasim/reference/python/
Could be due to changes in https://github.com/machow/quartodoc

Possibly related to #219

Edit: I think it is working fine, but we just don't have many docstrings anymore. And currently methods that have docstrings don't seems to get included, so we should look into that. ~~Perhaps best done after #1298.~~

- [x] Fix bug in docstrings
- [x] add info to pages (focus on most important items:

Jingru923:
I added some docstring under Model() class, and quartodoc can show those docstring now, the bug is no longer there.

Jingru923:
When I was working on this issue, I hope that quardoc can collect all the self-defined functions in Model class and show it in API. 

I tried `include_functions` option in `content`, it does nothing. This turns out to be impossible. 

My current conclusion is that we need to list all the functions manually, and later if new functions are added, keep in mind to also add them to API page. I think  we can generate a list of functions in python, and copy paste that list to `quartodoc` in `_quarto.yml`. I couldn't succeed it yet but I believe we can investigate that

evetion:
It might be good to link to an open issue on quartodoc about this, and link to the configuration page, where you got the `include_functions` from.

evetion:
FWIW, after checking the [documentation](https://machow.github.io/quartodoc/get-started/basic-content.html#all-content-options), and [this issue](https://github.com/machow/quartodoc/issues/275#issuecomment-1733945469), we could do this:

```
quartodoc:
  style: pkgdown
  dir: python/reference
  title: API Reference
  package: ribasim
  options:
      include_inherited: true
      include_empty: true
      include_attributes: true
      include_classes: true
      include_functions: true
  sections:
    - title: ribasim
      contents:
        - name: ribasim
          package: null
          include_imports: true
```

The options will make sure to render everything (classes/attributes/etc), while the ribasim section would render everything we export.

Jingru923:
Continue work of this issue is here #1571 

This issue thus can be closed

PR #1553, opened 2024-06-13
Include Ribasim Python API reference

visr:
In the recent doc restructuring these pages were no longer included.
Related to #1439, but does not fix that issue.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,8 +1,7 @@
 /.quarto/
 /_site/
 /site_libs/
-python/reference/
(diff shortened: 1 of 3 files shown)


evetion (approved):
Maybe also include a first (albeit empty) Julia reference docs?

visr:
> Maybe also include a first (albeit empty) Julia reference docs?

I removed it intentionally in https://github.com/Deltares/Ribasim/pull/1451, because it isn't good and not public API.

PR #1298, opened 2024-03-20
Render PDF docs

Hofer-Julian:
Fixes #666

Issue #666, opened 2023-10-13
include docs PDF in release assets


visr:
Currently we only use the Quarto HTML output which we host online. For offline use it would be good to build and include the PDF output of our documentation.

https://quarto.org/docs/output-formats/pdf-basics.html

- [ ] get PDF output working, I believe https://github.com/Deltares/iMOD-Documentation already has it
- [ ] add it to CI for our nightly and release builds
- [ ] update the release process docs

Having this would alleviate #665, giving users a way to read old documentation.

visr:
I tried adding a `quarto-render-pdf` task in pixi.toml, which is the same as `quarto-render` but with `--to pdf`.

First it complained that no TeX installation was detected, so I ran `quarto install tinytex`. After that it the amsmath package failed on some equations on allocations.qmd: "Package amsmath Error: Erroneous nesting of equation structures;"

visr:
With #735 and #737 we are closer to PDF output. However, there is a bug that is fixed upstream in Quarto for including remote images, that was discussed in #737.

It's easiest to put this on hold until Quarto 1.4 is released, that is already available as a pre-release, but not in conda-forge / pixi.

visr:
With the update to Quarto 1.4 in #1139 this is no longer Paused.

Hofer-Julian:
This is currently blocked by https://github.com/conda-forge/quarto-feedstock/issues/44

visr:
This is currently blocked by https://github.com/Deltares/Ribasim/pull/1298#issuecomment-2085709572

PR #1555, opened 2024-06-13
Quarto PDF docs take 2

visr:
Part of #666, does not automatically build yet or is included in release assets. Supersedes #1298.

For this we need to wait on the conda-forge release of Quarto 1.5, which is currently a release candidate. Otherwise it fails as mentioned [here](https://github.com/Deltares/Ribasim/pull/1298#issuecomment-2085709572).

Use this to render:

```
pixi run --environment=dev quarto render docs --profile manual --to pdf
```

Only since the Quarto 1.5 RC is not in conda-forge, I have to remove `pixi run`, causing issues of not being able to find Python. So I didn't test if this worked on all content, just some parts.

Some more references in case someone wants to try this:
https://quarto.org/docs/books/
https://quarto.org/docs/reference/formats/pdf.html
https://quarto.org/docs/reference/projects/books.html
https://github.com/Deltares/iMOD-Documentation
https://github.com/Deltares/iMOD-Documentation/blob/main/docs/_quarto-manual.yml


asinghvi17:
https://github.com/PumasAI/QuartoNotebookRunner.jl/pull/135 may be interesting to you if using the Julia engine, we might start using it in https://github.com/geocompx/geocompjl pretty soon as well.  

It basically allows the code in the cell to determine the number and contents of outputs, so it can even output things like a tabset.  Useful if you want to show parameters programmatically without having to manually create the tabset...

visr:
This is too outdated. If we still need this we need a new PR. It seems quite tricky to do though, and we'd need quite some text and formula modifications for Latex support. Perhaps once Typst book support lands in Quarto we should revisit this https://github.com/quarto-dev/quarto-cli/discussions/6979.

PR #1565, opened 2024-06-18
Auto-generate framework for Python API reference documentation

Jingru923:
Part of #1439 .

I added the public functions (the one that doesn't start with a "_") to the documentation. Give some short description on some of them.

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -108,45 +108,9 @@ quartodoc:
   dir: reference/python
   title: Python API
   package: ribasim
-  sections:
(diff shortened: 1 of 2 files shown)


Jingru923:
My new commit allow us to have a nice framework for API reference. All the classes and their functions are included.

`quartodoc` also display all the docstring that are in the classes and functions. However, at the moment there are not that many of docstrings.

To have more text in the API, I suggest that we create a new issue for it.

visr (approved):
Nice that this can now be done automatically.

PR #1566, opened 2024-06-19
Use keywords for creating parameter structs

visr:
By putting the [kwdef](https://docs.julialang.org/en/v1/base/base/#Base.@kwdef) macro in front of our struct definitions we can create them with keywords and set defaults if applicable. This helps with readability and makes it less likely to pass the wrong argument due to a wrong order. E.g. compare this test code before and after:

```julia
basin = Ribasim.Basin(
    NodeID.(:Basin, [1], 1),
    [NodeID[]],
    [NodeID[]],
    zeros(1),
    zeros(1),
    zeros(1),
    zeros(1),
    zeros(1),
    zeros(1),
    zeros(1),
    [storage_to_level],
    [level_to_area],
    demand,
    StructVector{Ribasim.BasinTimeV1}(undef, 0),
)
```

```julia
basin = Ribasim.Basin(;
    node_id = NodeID.(:Basin, [1], 1),
    storage_to_level = [storage_to_level],
    level_to_area = [level_to_area],
    demand,
    time = StructVector{Ribasim.BasinTimeV1}(undef, 0),
)```

### DIFF
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -544,5 +544,5 @@ function AllocationModel(
     capacity = get_capacity(p, subnetwork_id)
     problem = allocation_problem(p, capacity, subnetwork_id)
 
-    return AllocationModel(subnetwork_id, capacity, problem, Δt_allocation)
+    return AllocationModel(; subnetwork_id, capacity, problem, Δt_allocation)
 end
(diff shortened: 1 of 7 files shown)


Jingru923:
Looks good to me. In terms of which struct to make use of this macro and to have default parameter, I trust you. For the unit tests, I don't see any unit tests that can make use of this being left out. Approve
(I also got your failing test on TC passed)

Jingru923 (approved):

PR #1568, opened 2024-06-20
Allow for meta_ prefix in spatial tables.

### LINKED ISSUE #1560: Support `meta_` kwargs for Node and `model.edge.add`
For most data tables we accept and carry extra columns as long as they are prefixed with `meta_`. This is tested [here](https://github.com/Deltares/Ribasim/blob/4e7ea0e87aa65852195208cf8ac2d22d0dd0211a/python/ribasim/tests/test_io.py#L86) and implemented on `TableModel` [here](https://github.com/Deltares/Ribasim/blob/4e7ea0e87aa65852195208cf8ac2d22d0dd0211a/python/ribasim/ribasim/input_base.py#L160).

With the add API redesign the Node and Edge table no longer inherit this behavior, and thus don't support extra columns. However metadata is often most relevant for these two tables.

I expect these to work and write these columns to the database. If they are not prefixed with `meta_` it should error. Right now for Node it silently ignores any unknown kwargs.

```python
from ribasim import Model, Node
from shapely.geometry import Point

Node(1, Point(3, 1), meta_a = 1)

model.edge.add(
    model.tabulated_rating_curve[1],
    model.terminal[1],
    meta_b = 2,
)
```

evetion:
Fixes #1560 

Requires just passing the extra kwargs to the relevant tables, and the existing validation errors actually trigger. 👍🏻

### DIFF
--- a/python/ribasim/ribasim/config.py
+++ b/python/ribasim/ribasim/config.py
@@ -97,18 +97,20 @@ class Node(pydantic.BaseModel):
     name: str = ""
     subnetwork_id: int | None = None
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="allow")
 
     def __init__(self, node_id: int, geometry: Point, **kwargs) -> None:
         super().__init__(node_id=node_id, geometry=geometry, **kwargs)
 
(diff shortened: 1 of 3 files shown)


visr (approved):
Awesome

PR #1572, opened 2024-06-21
change the names of artifacts

Jingru923:
Fixes our release pipeline


visr (approved):
This should fix the script, but I think there are a few more old names in e.g. the readme.

Issue #1545, opened 2024-06-10
For `UserDemand`, first use the available capacity in the directly connected basin


SouthEndMusic:
We have an ongoing discussion about the default behavior of basins within a subnetwork. One of the relatively simple things to implement is that `UserDemand` nodes always first get flow allocated from the basin they are abstracting from (also when this basin doesn't have a level demand). This can happen outside and before the actual JuMP optimization.
Then there is still the question of to which level te basin can be drained by the user. This is already regulated by the `min_level` of the `UserDemand`, but we have to consider whether a `LevelDemand` node connected to this basin also affects this.

PR #1581, opened 2024-06-25
Allocate to UserDemand from directly connected basin if possible

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1545

Some notes on this PR:

- This still only works when the `Basin` upstream of a `UserDemand` has a level demand to define the source capacity of the basin
- It makes sense that this flow from the `Basin` to the `UserDemand` also shows up in `allocation_flow.arrow`, which requires some refactoring. That is fine though, as that refactor is needed anyway for https://github.com/Deltares/Ribasim/issues/565, where we probably want to aggregate the allocation flows for output over the optimizations for all sources in stead of saving them per source 
- I'm not sure how to directly test this


core/src/allocation_optim.jl
```diff
@@ -853,6 +852,46 @@ function save_allocation_flows!(
...
+function allocate_to_users_from_connected_basin!(
+    allocation_model::AllocationModel,
+    flow_priority,
+    p::Parameters,
+    priority_idx::Int,
+)::Nothing
+    (; problem) = allocation_model
+    (; graph, user_demand) = p
+
+    # Get al UserDemand nodes from this subnetwork
```
visr, line 864:
```suggestion
    # Get all UserDemand nodes from this subnetwork
```

visr:
> This still only works when the Basin upstream of a UserDemand has a level demand to define the source capacity of the basin

Link to discussion: https://github.com/Deltares/Ribasim/issues/1428#issuecomment-2077601391

We should directly test this since it will have a large impact on results. Would a `UserDemand - Basin - LevelDemand` model suffice, with only initial storage?

This behavior probably also needs to be in the user docs, probably https://deltares.github.io/Ribasim/reference/node/user-demand.html, maybe https://deltares.github.io/Ribasim/concept/allocation.html.

Issue #720, opened 2023-10-31
Support continuous controller


SnippenE:
Currently with DiscreteControl we can set values when thresholds are crossed. If say we want to gradually increase the `flow_rate` of an Outlet based on a water level further upstream (think Driel), we can currently approximate this with a stepped function by specifying many thresholds.

A more general fix would be to allow a LinearInterpolation from a scalar to another scalar, such that the target is continuously set based on the input. We need to think about how to implement this. Adding it to DiscreteControl may not be logical. Might go better with #455.

SouthEndMusic:
Note that this will influence the Jacobian sparsity structure. Pick up after https://github.com/Deltares/Ribasim/pull/1349.

SouthEndMusic:
Some thoughts on this:

This works well of the outcome of the continuous control can be expressed as an explicit expression in terms of basin storages (not that we do _not_ need to know this exact expression). From this expression the effect on the Jacobian sparsity is also immediately clear. Note here that the flow through connector nodes can also be expressed in terms of the storages of the connected basins.

PR #1602, opened 2024-07-02
Add `ContinuousControl` node

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/720.


SouthEndMusic:
For those wondering why the diff is so big: this is a large PR, but the majority now actually comes from changes to pixi.lock. I don't know what is happening there, the file keeps being updated automatically. I saw I am behind on my pixi version, so I now upgraded to 0.25.0.

visr (commented):
Left some comments.
As discussed, let's remove `min_output` and `max_output` as well. I also noticed we need to update `SPATIALCONTROLNODETYPES` in ribasim_qgis.

visr (approved):

PR #1606, opened 2024-07-03
Symbolic Jacobian sparsity

visr:
This doesn't work yet, but I quickly tried this out yesterday, thought I might as well put it up to look if we should pursue this.
I think the main advantage is that we don't have to maintain our own code for determining the sparsity, which can more easily get out of sync. It does add some dependencies.

The error I get is with DataInterpolations not supporting `::SymbolicUtils.BasicSymbolic{Real}`, which looks like this closed issue: https://github.com/SciML/DataInterpolations.jl/issues/168. DataInterpolations has an [extension on Symbolics](https://github.com/SciML/DataInterpolations.jl/blob/master/ext/DataInterpolationsSymbolicsExt.jl) for this purpose, so perhaps we just need to create a minimal example and file an issue.

```
ERROR: MethodError: no method matching (::DataInterpolations.LinearInterpolation{Vector{…}, Vector{…}, Float64})(::SymbolicUtils.BasicSymbolic{Real})

Closest candidates are:
  (::DataInterpolations.AbstractInterpolation)(::Symbolics.Num)
   @ DataInterpolationsSymbolicsExt C:\Users\visser_mn\.julia\packages\DataInterpolations\NoIUa\ext\DataInterpolationsSymbolicsExt.jl:15
  (::DataInterpolations.AbstractInterpolation)(::Number)
   @ DataInterpolations C:\Users\visser_mn\.julia\packages\DataInterpolations\NoIUa\src\DataInterpolations.jl:22
  (::DataInterpolations.AbstractInterpolation)(::Number, ::Integer)
   @ DataInterpolations C:\Users\visser_mn\.julia\packages\DataInterpolations\NoIUa\src\DataInterpolations.jl:23
  ...

Stacktrace:
  [1] get_area_and_level(basin::Ribasim.Basin{…}, state_idx::Int64, storage::Symbolics.Num)
    @ Ribasim d:\repo\ribasim\Ribasim\core\src\util.jl:57
  [2] set_current_basin_properties!(basin::Ribasim.Basin{…}, storage::SubArray{…})
    @ Ribasim d:\repo\ribasim\Ribasim\core\src\solve.jl:43
  [3] water_balance!(du::ComponentArrays.ComponentVector{…}, u::ComponentArrays.ComponentVector{…}, p::Ribasim.Parameters{…}, t::Float64)
```

It could be that there are a bunch more issues of this kind once this is resolved.

If it starts to work but we don't want it, we should perhaps use it to test our manual Jacobian sparsity.


SouthEndMusic:
https://github.com/SciML/DataInterpolations.jl/issues/310

SouthEndMusic:
I did some digging. I found out these things:
- The problem you state above is likely not caused by `DataInterpolations.jl`, but by `SmoothInterpolation.jl`, because the problematic type is the output of a function call of that package and the problem goes away when I remove that call. That will be solved by https://github.com/Deltares/Ribasim/pull/1649
- Code that branches out is incompatible, for instance in `reduction_factor`. I made a non-branching version of it
- `PreallocationTools` is not directly compatible with `Symbolics`, I found a workaround here: https://discourse.julialang.org/t/type-issue-to-get-preallocationtools-diffcache-work-with-symbolics-jacobian-sparsity/88542/3?u=bdekoning
- To trace trough the code to obtain the jacobian sparsity, the type parameter `T` in `Parameters` and its fields must (probably) be `Vector{Symbolics.Num}`, so the previous point might be irrelevant. I have not yet thought of a non-intrusive way to achieve this

SouthEndMusic:
There seems to be a problem with `DataInterpolations` after all:

https://github.com/SciML/DataInterpolations.jl/issues/317

SouthEndMusic:
One thing to keep in mind here is that some basin dependencies are temporary and not 'active' at initialization. To get a proper sparsity pattern, all nodes should be active

core/src/util.jl
```diff
@@ -348,13 +354,18 @@ Function that goes smoothly from 0 to 1 in the interval [0,threshold],
...
+        0.5 * (
+            0.5 *
+            (sign(threshold - x) + one(T)) *
+            (sign(x) + one(T)) *
+            ((-2 * x_scaled + 3) * x_scaled^2) +
+            sign(x - threshold) +
+            one(T)
+        )
     end
 end
```
visr, line 370: When is the input infinite? Can you add a test for infinite input in the "reduction_factor" testitem?

core/src/util.jl
```diff
@@ -809,8 +822,21 @@ function set_previous_flows!(integrator)
...
 
-    flow = get_tmp(flow, 0)
-    vertical_flux = get_tmp(vertical_flux, 0)
+    flow = flow[parent(u)]
+    vertical_flux = vertical_flux[parent(u)]
     copyto!(flow_prev, flow)
     copyto!(vertical_flux_prev, vertical_flux)
 end
+
+function wrap_forcing(vector)
```
visr, line 831: Could use a short docstring.

core/src/sparsity.jl
```diff
@@ -13,7 +13,8 @@ Note: the name 'prototype' does not mean this code is a prototype, it comes
 from the naming convention of this sparsity structure in the
 differentialequations.jl docs.
 """
-function get_jac_prototype(p::Parameters)::SparseMatrixCSC{Float64, Int64}
+
+function get_jac_prototype(p::Parameters, t0, du0, u0)::SparseMatrixCSC{Float64, Int64}
```
visr, line 17: If we go with this, we can remove this.

core/src/parameter.jl
```diff
@@ -74,13 +74,18 @@ end
...
+    Vector{Float64},
+    Vector{Float64},
     Vector{Float64},
     Vector{Float64},
     Float64,
 }
 
+set_zero!(v) = fill!(v, zero(eltype(v)))
+const LBC = LazyBufferCache{Returns{Int}, typeof(set_zero!)}
+FixedSizeLazyBufferCache(len::Int)::LBC =
```
visr, line 86: Capitalizing this confused me a bit thinking it is a type. How about just calling this function `cache`? Could also use a docstring, to explain why you need the initializer.

`Return` is neat, hadn't used it before.

core/src/parameter.jl
```diff
@@ -74,13 +74,18 @@ end
...
-    ReadOnlyVector{Float64, Vector{Float64}},
+    Vector{Float64},
+    Vector{Float64},
     Vector{Float64},
     Vector{Float64},
     Float64,
 }
 
+set_zero!(v) = fill!(v, zero(eltype(v)))
+const LBC = LazyBufferCache{Returns{Int}, typeof(set_zero!)}
```
visr, line 85: May I suggest renaming `LBC` to `Cache` for readability? With a docstring explaining what we cache.

core/src/Ribasim.jl
```diff
@@ -24,7 +24,7 @@ import TranscodingStreams
 using Accessors: @set
 using Arrow: Arrow, Table
 using CodecZstd: ZstdCompressor
-using ComponentArrays: ComponentVector
+using ComponentArrays: ComponentVector, Axis, parent
```
visr, line 27:
```suggestion
using ComponentArrays: ComponentVector, Axis
```

I'm assuming `ComponentArrays.parent` === `Base.parent`, so you don't need to import it.

visr (commented):
Left some comments. Let's have a look at the impact performance and binary size.

visr:
Calculating the Jacobian seems fast, but runtime performance went down a lot. The `wrap_forcing` function leads to GC and runtime dispatch. One possible way to resolve this that we discussed is to switch the forcing ComponentVector to four separate Vectors. The issue is probably that the ComponentVector does not play well with `similar_type` used in LazyBufferCache, since the Axis of the components are part of the type.

visr:
We should also have a look at https://github.com/adrhill/SparseConnectivityTracer.jl as an alternative, lighter dependency.

core/src/util.jl
SouthEndMusic: The input can be infinite because of certain parameters that default to `+/Inf` (e.g. reference levels).

SouthEndMusic:
`SparseConnectivityTracer.jl` needs some work: https://github.com/adrhill/SparseConnectivityTracer.jl/issues/145

SouthEndMusic:
`SparseConnectivityTracer` (the unsafe version) now works, only `Aqua` complains about type piracy on a custom overload, but that should be solved upstream

visr (commented):
Looks good to me! You'll have to approve yourself ;)

SouthEndMusic (approved):
https://nl.wikipedia.org/wiki/Wij_van_Wc-eend_adviseren_Wc-eend

PR #1607, opened 2024-07-03
Validate level with upstream basin

### LINKED ISSUE #1591: Validate `min_crest_level` (Outlet) and `level` (TabulatedRatingCurve) with upstream `Basin / profile`
**What**
When specifying an `Outlet` or `TabulatedRatingCurve` one can specify `min_crest_level` or `level` that is lower than the upstream `Basin / profile`. This is one of the most common causes of model instability/crashes. 

**Why**
Validating the `min_crest_level` and `level` column with the minimum `level`  upstream `Basin / profile` prior to modelling would speed-up model debugging. 

**How**
Compare `min_crest_level` and `level` columns in `Outlet` or `TabulatedRatingCurve` with upstream `Basin / profile` and give a validation error when `min_crest_level` and `level` are lower than the minimum level in the basin-profile

Jingru923:
Fixes #1591 

An error will be given when the `level` in tabulated rating curve is lower than upstream bottom. 

An error will be given when the `min_crest_level` in outlet is lower than upstream bottom. In the validation function, it will set the `min_crest_level` to the basin bottom if this parameter was not specified.


core/src/validation.jl
```diff
@@ -441,6 +441,39 @@ function valid_demand(
...
 
+function valid_outlet_crest_level(graph::MetaGraph, outlet::Outlet, basin::Basin)::Bool
+    errors = false
+    for (id, crest) in zip(outlet.node_id,outlet.min_crest_level)
+        id_in = inflow_id(graph, id)
+        if id_in.type == NodeType.Basin
+            #validation:
+            basin_bottom_level = basin_bottom(basin, id_in)
+            if crest == -Inf
+                crest == basin_bottom_level[2]
```
visr, line 452: `crest == basin_bottom_level[2]` checks if they are equal, perhaps you meant `=`? Note though that that would not change the value in the `outlet.min_crest_level`, only `outlet.min_crest_level[i] = ...` would do that.

Jingru923: Ops, it should be `=`

core/test/validation_test.jl
```diff
@@ -438,6 +438,70 @@ end
     )
 end
 
+@testitem "tabulated rating curve validation" begin
```
visr, line 441:
```suggestion
@testitem "TabulatedRatingCurve upstream level validation" begin
```

core/test/validation_test.jl
```diff
@@ -438,6 +438,70 @@ end
...
+        normpath(@__DIR__, "../../generated_testmodels/level_range/ribasim.toml")
+    @test ispath(toml_path)
+    invalid_level = -0.1
+
+    config = Ribasim.Config(toml_path)
+    model = Ribasim.Model(config)
+
+    db_path = Ribasim.input_path(config, config.database)
+    db = SQLite.DB(db_path)
+    parameters = Ribasim.Parameters(db, config)
```
visr, line 456: It looks like this is not needed, the parameters are already in the model and can be retrieved as `model.integrator.p`. Same below.

core/src/validation.jl
```diff
@@ -441,6 +441,39 @@ function valid_demand(
...
+    errors = false
+    for (index, (id, crest)) in enumerate(zip(outlet.node_id, outlet.min_crest_level))
+        id_in = inflow_id(graph, id)
+        if id_in.type == NodeType.Basin
+            #validation:
+            basin_bottom_level = basin_bottom(basin, id_in)
+            if crest == -Inf
+                outlet.min_crest_level[index] = basin_bottom_level[2]
+            elseif crest < basin_bottom_level[2]
+                @error "Minimum crest level of $id is lower than bottom of upstream $id_in"
```
visr, line 454:
```suggestion
                @error "Minimum crest level of $id is lower than bottom of upstream $id_in" crest basin_bottom_level
```

This shows the values directly, also good to do below. It will show the variable name as well as the value, so the variable name should be clear, otherwise you can use `@error "asda" clear_name = a[idx][2]` to rename.

core/src/validation.jl
```diff
@@ -441,6 +441,39 @@ function valid_demand(
     return !errors
 end
 
+function valid_outlet_crest_level!(graph::MetaGraph, outlet::Outlet, basin::Basin)::Bool
+    errors = false
+    for (index, (id, crest)) in enumerate(zip(outlet.node_id, outlet.min_crest_level))
+        id_in = inflow_id(graph, id)
+        if id_in.type == NodeType.Basin
+            #validation:
+            basin_bottom_level = basin_bottom(basin, id_in)
```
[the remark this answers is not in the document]
Jingru923: For now I will use basin_bottom(basin, id_in)[2]. Later we can always change it back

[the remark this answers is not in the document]
Jingru923: Since we have been using the basin bottom in many validation, can the basin bottom be part of the basin node?

Jingru923:
In the validation of tabulated rating curve level, currently many models that has tabulated rating curve with upstream basin fails. The reason is that in the Ribasim python, the level is [0.0 1.0], but during initialization, we add 1.0 to avoid negative extrapolation for Q (line 121)
image

Once an idea of during validation before the parameter initialization came up, but it is found to be impossible because the validation request the graph information.

For now, I add the 1.0 back during the validation

image

But this make us question the robustness of the extrapolation

Huite opened an issue to research this robustness problem #1621

core/src/validation.jl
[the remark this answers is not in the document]
visr: Since we'd be duplicating data, I prefer to do that if either it leads to a significant performance improvement, or simpler code. I'm not sure either would be the case here.

visr (approved):

Issue #1071, opened 2024-02-05
Urban runoff is not used simulation? And what about non-urban runoff?


Huite:
The current Basin definition:

```
struct Basin{T, C} <: AbstractParameterNode
    node_id::Indices{NodeID}
    precipitation::Vector{Float64}
    potential_evaporation::Vector{Float64}
    drainage::Vector{Float64}
    infiltration::Vector{Float64}
    # cache this to avoid recomputation
    current_level::T
    current_area::T
    ...
```

does not have a vector for the urban runoff values.

Secondly, should it maybe just be runoff? Not clear whether we need the distinction.

For the coupling, MetaSWAP will just return runoff. Urban areas are separated in terms of their SVAT unit, but it's the same water balance term.

gijsber:
distinction of urban runoff would be a good nice-to-have as it allows future distinction at the WWTP between dry-weather and wet-weather waste loads. If it becomes hard to make this seperation between urban and non-urban runoff, we should do it now, else we can postpone.

visr:
To resolve this, we will rename `urban_runoff` to `runoff` and implement this similar to the way we implement `drainage`. Also needs doc updates.

evetion:
This is needed for the LWKM project (to have concentrations for runoff).

PR #1611, opened 2024-07-04
Remove urban_runoff from Basin

visr:
This takes care of the removing `urban_runoff` part of #1071. Technically breaking if people had this unused column in their Basin tables.


Huite:
Shouldn't e.g. the MetaSWAP runoff be received in a runoff column?

Jingru923 (approved):

PR #1612, opened 2024-07-05
Add the image of cascade polders back

Jingru923:
The image of the cascade polder example was broken. 
image

Our documentation page has no access to our the picture on our GitHub. So I uploaded the picture again. Please don't change it.

PR #1613, opened 2024-07-05
Fix cascading polders image in docs

visr:
@Jingru923 I suggest this as an alternative to #1612. We try to keep this repository as small as possible and therefore prefer to not add images to it. So far we've uploaded images into #1 and used those. Recently GitHub changed something:

> Uploading images here won't work anymore for displaying them in our docs, due to GitHub changes, see https://github.blog/changelog/2023-05-09-more-secure-private-attachments/

So using those directly in Markdown no longer seems to work (for newly uploaded images). But @evetion noticed that putting them in an `img` HTML tag works fine.

### DIFF
--- a/docs/guide/examples.ipynb
+++ b/docs/guide/examples.ipynb
@@ -1684,7 +1684,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


Jingru923:
Looks good.

Jingru923 (approved):

PR #1618, opened 2024-07-09
Use the latest greatest Python deps

### LINKED ISSUE #1573: NumPy 2 support
In https://github.com/conda-forge/ribasim-feedstock/pull/17 I had to add `numpy <2.0`. We should add that as well in the pyproject.toml, or look into supporting NumPy 2 directly.

We should use the ruff rule as mentioned in https://numpy.org/doc/stable/numpy_2_0_migration_guide.html.

evetion:
Fixes #1573

This only fixes us to the first version of pandera that supports numpy 2, the rest should follow automatically.

I took longer than I want to admit to find that the previous `[None, 1]` is now silently coerced to `[0,1]` and therefore passed validation, and failed the test.

I also (still?) see
```
tests/test_model.py::test_xugrid[model0]
  <frozen importlib._bootstrap>:488: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
```

Which is probably due to xugrid/netcdf/meshkernel(?) using something compiled against a different numpy version.


visr:
I looked into the pyarrow situation, and filed https://github.com/prefix-dev/pixi/issues/1594.

In short, nothing is broken on pyarrow 16, it is just that the update from pyarrow 15 to 16 breaks the 16 installation.

visr (approved):
I upgraded pyarrow and updated pyproject.toml.
In trying this out I also discovered that we need to unset juliaup overrides if they are already set, so I changed that and filed https://github.com/JuliaLang/juliaup/issues/987.

evetion:
> I upgraded pyarrow and updated pyproject.toml. In trying this out I also discovered that we need to unset juliaup overrides if they are already set, so I changed that and filed [JuliaLang/juliaup#987](https://github.com/JuliaLang/juliaup/issues/987).

I ran into the pyarrow __version__ attribute version as before, hence why I set it back to <15 in my second commit. Can you reproduce that locally?

evetion:
> I looked into the pyarrow situation from #1542, and filed [prefix-dev/pixi#1594](https://github.com/prefix-dev/pixi/issues/1594).
> 
> In short, nothing is broken on pyarrow 16, it is just that the update from pyarrow 15 to 16 breaks the 16 installation. So let's update, but we may need to delete our .pixi folder.

Your error seemed different from mine, but cool that you found a possible source for the `__version__` error. 👍🏻

visr:
I got the same error as #1542, until I realized I didn't need pandas to reproduce.

PR #1624, opened 2024-07-15
Remove `Terminal / static` table

### LINKED ISSUE #1209: Remove `Terminal / static` as it is not needed
When updating models in #1110 I noticed that I can run a model with a Terminal node with or without the 'Terminal / static' table. Then I realized, especially with the new API, there is no real information in that table that isn't already in the Node table. We should look into if that is true, and if it is, perhaps we should just remove `Terminal / static`.

https://deltares.github.io/Ribasim/core/usage.html#sec-terminal

visr:
This also removes the Terminal type from the core, as it is not needed for the same reason. That meant I had to adapt a bit of code that relied on `getfield(p, :terminal)` working.

Fixes #1209


Jingru923 (approved):
lgtm

PR #1627, opened 2024-07-16
Require QGIS 3.34 (LTR); remove activation scripts

visr:
Currently the only thing we use from a newer QGIS (3.32) is `Qgis.AttributeFormSuppression.On`. So we could technically still support older versions. Though since 3.34 is the new Long Term Release, and we have yet to release v1.0, it is best to make it clear what we'll support now.

This also removes the activation scripts, including `open-vscode.bat`. With [this VSCode Python release](https://github.com/microsoft/vscode-python/releases/tag/v2024.10.0), it now correctly locates pixi environments. I added instructions on this in the dev docs.

The mypy settings still affected 4 issues in the QGIS Python code. All are of this type:
```
ribasim_qgis\ribasim_qgis.py:11: error: Class cannot subclass "QgsDockWidget" (has type "Any")  [misc]
```
I added type ignores for those four, I don't think this disables much effective functionality since mypy still finds issues inside the class.


Huite:
I think moving to the newest LTR is good, also as a precedent. Although changes are generally minor (e.g. `AttributeFormSuppression`), it makes our life easier to have "lastest LTR" as a policy.

The only thing I'm not sure about is whether people who older versions of QGIS will be notified about newer versions of the plugin in their older QGIS version. It would be good to add this very explicitly in the documentation.

The current Install docs don't mention QGIS very explicitly: https://deltares.github.io/Ribasim/install.html

Maybe add a section there?

I would also suggest adding a note in the README.md of the repo. There's a link to download the plugin there.

Huite (approved):

visr:
> The only thing I'm not sure about is whether people who older versions of QGIS will be notified about newer versions of the plugin in their older QGIS version.

I don't know the answer, but this would only apply if we register the QGIS plugin, and we don't do that yet.

I listed your other suggestions in #1630.

PR #1628, opened 2024-07-16
Fix DataInterpolations to 0.5.2

visr:
CI is currently failing on main since the test environment brings in https://github.com/SciML/DataInterpolations.jl/releases/tag/v5.3.0, which is breaking for us probably because we use DataInterpolations internals. This is a quick fix to get CI green again.

Our Manifest.toml still has 5.2.0 as well.


SouthEndMusic (approved):
Yess

PR #1629, opened 2024-07-17
`level_demand` can now be without max_level or min_level

### LINKED ISSUE #1589: Allow `LevelDemand` nodes with neither a `min_level` nor a `max_level`
Currently, if for instance in the `level_demand` testmodel you remove all `min_level` and `max_level` in the `LevelDemand / static` table, you get the following very unhelpful error:

```
ERROR: Encountered an error in HiGHS (Status -1)
``` 

So probably something has to be adjusted in the construction of the optimization problem for `LevelDemand` nodes that have neither of these reference levels defined.

Jingru923:
Fixes #1589


SouthEndMusic (commented):
It's nice that we have this now, but you should also add a test where you run a model with the default `min_level` and `max_level` and check that there us no crash.

core/test/allocation_test.jl
```diff
@@ -550,3 +550,22 @@ end
...
+    @test ispath(toml_path)
+    model = Ribasim.Model(toml_path)
+    (; p, u, t) = model.integrator
+    (; allocation_models) = p.allocation
+    (; basin, level_demand, graph) = p
+
+    fill!(level_demand.max_level[1].u, Inf)
+    fill!(level_demand.max_level[2].u, Inf)
+
+    @test Ribasim.get_basin_capacity(allocation_models[1], u, p, t, basin.node_id[1]) == 0.0
```
SouthEndMusic, line 568: Might be nice to explain here that we expect the basin capacity to be $0$ because it is not possible for the basin level to be > Inf.

SouthEndMusic (approved):
Looks good, I have one single suggestion.

PR #1636, opened 2024-07-18
Always show convergence bottlenecks

### LINKED ISSUE #1593: Report on convergence bottlenecks always
**What**
As a modelbuilder I want to find convergence bottlenecks **always**. Even if my model converges during simulation I want to be able to check where there might be room for improvements in simulation efficiency and/or accuracy. The current report on convergence bottlenecks is very helpful, the key-info we navigate on in getting models running.

A database with two example-tomls can be found here: https://we.tl/t-aPHw399bTf. We suggest two improvements:
- running `hws_sturing_maxiters_metbasins.toml` results in a great report. Would be better if we would find a similar report even without setting `maxiters=30000` in the toml:
image
- setting `maxiters=30005` (see `hws_sturing_maxiters_metbasins.toml`) does produce a crash, but doesn't provide basin-numbers:
hws_sturing-maxiters_zonderbasins We can get a basin-list for this run, but then we we have to play with `abstol` and `reltol` to force the model in a crash wíth a report on convergence bottlenecks. 

**Why**
The feature would make debugging models much more straight forward and less time-consuming without having knowledge/experience in how to make solvers crash :-)

**How**
Printing in the console and/or log would be a super-handy. It may be even more user-friendly if you can inspect in QGIS to spatially find the causes of performance bottlenecks in the database, usually faulty profiles and or flow-nodes as described in #1591.

Follow-up of #1440.

visr:
Fixes #1593. Follow-up of https://github.com/Deltares/Ribasim/pull/1440.

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -39,7 +39,7 @@ using ForwardDiff: pickchunksize
 using Graphs:
     DiGraph, Edge, edges, inneighbors, nv, outneighbors, induced_subgraph, is_connected
 using Legolas: Legolas, @schema, @version, validate, SchemaVersion, declared
-using Logging: with_logger, LogLevel, AbstractLogger
+using Logging: with_logger, @logmsg, LogLevel, AbstractLogger
 using MetaGraphsNext:
     MetaGraphsNext,
     MetaGraph,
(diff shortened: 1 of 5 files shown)


SouthEndMusic (approved):
Good stuff

PR #1641, opened 2024-07-22
Avoid `curl_easy_setopt` error logs

Manangka:
Due to an issue between the curl and openssl libraries the `pixi run build` tasks failed.
In https://github.com/JuliaLang/Downloads.jl/issues/244 the cause of the issue is gfound to be environment variables being set that conflict.
This commit adds an environment variable to the tasks failing, clearing the SSL variables


visr:
@Manangka closed this since the codegen failed, looks like juliaup fails to download with this. The error log this fixes is harmless, but looks scary, so I'm re-opening to see if we can use this.

visr (approved):

PR #1647, opened 2024-07-22
QGIS: Update minimal TOML with new required fields

visr:
For CRS it uses the current QGIS project CRS.
The minimal TOML used in QGIS tests is now also checked by Ribasim Python to avoid it getting outdated.

### DIFF
--- a/ribasim_qgis/widgets/dataset_widget.py
+++ b/ribasim_qgis/widgets/dataset_widget.py
@@ -93,8 +93,8 @@ class DatasetTreeWidget(QTreeWidget):
 
         # Start deleting
         elements = {item.element for item in selection}  # type: ignore[attr-defined] # TODO: dynamic item.element should be in some dict.
-        qgs_instance = QgsProject.instance()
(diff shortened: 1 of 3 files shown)


ribasim_qgis/widgets/dataset_widget.py
```diff
@@ -268,8 +268,10 @@ def _write_new_model(self) -> None:
                 [
                     f"starttime = {datetime(2020, 1, 1)}\n",
                     f"endtime = {datetime(2021, 1, 1)}\n",
+                    f'crs = "{self.ribasim_widget.crs.authid()}"\n',
                     'input_dir = "."\n',
                     'results_dir = "results"\n',
+                    'ribasim_version = "2024.9.0"\n',
```
deltamarnix, line 274: This seems rather hardcoded. Can we get this version from a different place? Some python file that we generate, or a pyproject file?

deltamarnix: Cries in python 🐍 💧

Issue #1625, opened 2024-07-16
Get rid of dependency `SmoothInterpolation.jl` in favor of the new `DataInterpolations.jl` release


SouthEndMusic:
All features of `SmoothInterpolation` and more have been merged and released in `DataInterpolations.jl`:

https://github.com/SciML/DataInterpolations.jl/releases/tag/v5.3.0

Most notably, we can now do integral inversion via `DataInterpolations.jl`.

SouthEndMusic:
I had to pin the `DataInterpolations.jl` version to `5.2.0` in https://github.com/Deltares/Ribasim/pull/1602 as the new version breaks some things.

Jingru923:
Issue #1621 can be addressed

PR #1649, opened 2024-07-22
Remove `SmoothInterpolation` as a dependency

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1625. Some upstream issues came out of this:

https://github.com/SciML/DataInterpolations.jl/issues/301
https://github.com/SciML/DataInterpolations.jl/issues/302
https://github.com/SciML/DataInterpolations.jl/issues/303


visr (approved):

SouthEndMusic:
With https://github.com/SciML/DataInterpolations.jl/pull/315 the `ReadOnlyArray` wrappers are removed again, and we should do `cache_parameters = true` to get the performance benefit. For `LinearInterpolation` appending data is still supported in that case, see https://github.com/SciML/DataInterpolations.jl/blob/master/src/online.jl.

visr:
Ok, how about we fix to 5.3.1 with f470d2484a39aa931ccee5bef597cc9eb23e3002. Then we can merge this, if you agree with my commits, and implement your comments once a new release is out.

SouthEndMusic:
Looks like you snuck some unrelated changes in there, but it's fine by me to merge this now

visr:
Oh you mean `JULIA_SSL_CA_ROOTS_PATH `? I actually needed that to pass CI when checking out the master branch of DataInterpolations (or any non-released package), otherwise there was a download error. So that's good to have even if we currently don't need unreleased packages.

PR #1656, opened 2024-07-23
Use install-ci on CI

visr:
These are the last two occurrences in our TeamCity config. This avoids installing QGIS plugins and pre-commit, both of which are not needed on TeamCity.

```toml
install = { depends_on = [
    "install-ci",
    "install-qgis-plugins",
    "install-pre-commit",
] }
```


Jingru923:
The change looks good to me. Can I review this PR?

visr:
Yes please :)

PR #1657, opened 2024-07-23
Update release process a bit

visr:
Mainly regarding the changelog, that we keep the same both in [our docs](https://deltares.github.io/Ribasim/changelog.html) and [GitHub releases](https://github.com/Deltares/Ribasim/releases).


Jingru923 (approved):
Lgtm (;

Issue #105, opened 2023-03-13
distribute ribasim executables


visr:
Currently we build nightly executables of the `ribasim` command line tool on TeamCity. It would be nice if we could make these artifacts public, or to automatically upload builds to a public place. This way users can quickly test out fixes.

gijsber:
Delft-FEWS never pushed automatically from Teamcity to external FTP site.

Candidatee route to investigate:
Teamcity-artifact --> \\dfs-trusted.directory.intra\dfs\repository\ --> outside FTP

gijsber:
Under discussion by Dimitri Tzoumas and Floris Langeraert

visr:
Thanks, I also asked Edward about if this would be possible with download.deltares.nl.

evetion:
We're now automatically pushing the Teamcity artefacts to s3, which are publicly available. I'm working on getting a nightly link to work as well.

gijsber:
No news from Dimitri.
For a centralized Deltares solution one is looking into:
https://www.sonatype.com/products/nexus-repository

@evetion : can you provide the link so we can communicate this coming Tuesday to the partners

visr:
Here is the link of a recent build, build number is in the link:

https://ribasim.s3.eu-west-3.amazonaws.com/teamcity/Ribasim_Ribasim/BuildRibasimCliWindows/2304861/ribasim_cli.zip

We should also put that in the docs.

evetion:
Now live: [ribasim.s3.eu-west-3.amazonaws.com/teamcity/Ribasim_Ribasim/BuildRibasimCliWindows/latest/ribasim_cli.zip](https://ribasim.s3.eu-west-3.amazonaws.com/teamcity/Ribasim_Ribasim/BuildRibasimCliWindows/latest/ribasim_cli.zip) 🎉

PR #1659, opened 2024-07-25
Regression test 1/2

Jingru923:
Fixes the test part of  #462 , the update of CI/CD will come in a different PR (will be named "Regression test 2/2")
Now if you enter `pixi run test-ribasim-regression` terminal, regression tests are run.


visr:
I want to avoid checking the reference results into this repository. For these models they may still be small, but not for the larger models. And they will need to be upgraded every now and then.

@deltamarnix do you know a good solution? It's a bit like the hosting of images for the docs of #1. I guess we could use the hack of uploading assets into https://github.com/visr/ribasim-artifacts/releases. Or use S3 like we did before for hosting nightlies: https://github.com/Deltares/Ribasim/issues/105#issuecomment-1482532408.

Jingru923:
In my opinion, having the reference result somewhere that can be accessed by TeamCity seems reasonable.

visr:
@Manangka suggests requesting self hosted [MinIO](https://publicwiki.deltares.nl/display/GIT/MinIO) storage for this.

deltamarnix:
> @Manangka suggests requesting self hosted [MinIO](https://publicwiki.deltares.nl/display/GIT/MinIO) storage for this.

I agree. Test data should be in our MinIO buckets, just like Delft3D does.

PR #1661, opened 2024-07-25
Allow 2 control edges to Pump and Outlet

visr:
In case you want to have ContinuousControl of these nodes, but switch them off under certain conditions with DiscreteControl. Put this up for testing for @rbruijnshkv.


visr:
Not needed for now, we can add this later. If we do this we probably need to validate that the two controls don't clash; they shouldn't control the same variable.

Issue #1097, opened 2024-02-08
Add tests for all names in `BMI.get_value_ptr`


SouthEndMusic:
**How**
Add tests in `test_bmi.py` similar to the existing one `test_get_value_ptr` which is only for volume.

PR #1665, opened 2024-07-29
test `libribasim.get_value_ptr` with all names

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1097.

I wasn't able to let `basin.subgrid_level` become anything other than `nan` 🤔


visr (approved):
Very nice. Could you make a bug report on the subgrid levels?

PR #1667, opened 2024-07-29
Ruff: apply all fixes

visr:
This is the default. I'm not sure why we only fixed `isort` rules before.


Jingru923 (approved):
I am also surprised :o

PR #1670, opened 2024-07-31
Fix development tools

deltamarnix:
* Restore open-vscode.bat to use with Quarto and Julia
* Fix mypy paths for qgis
* Insert new line on save

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,5 +23,6 @@
     "julia.lint.disabledDirs": [
         ".pixi"
     ],
-    "julia.lint.run": true
+    "julia.lint.run": true,
+    "files.insertFinalNewline": true
 }
--- a/mypy.ini
+++ b/mypy.ini
@@ -10,6 +10,7 @@ extra_checks = True
 disallow_subclassing_any = True
 disallow_untyped_decorators = True
 disallow_any_generics = True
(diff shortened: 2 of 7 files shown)


visr (approved):
This looks good, thanks! Looking at the changes in the lock file, this downgrades a lot of packages, e.g. numpy back to 1, pyarrow back to 14. I guess that is all because the QGIS LTR build is not compatible with those yet? I do hope that new LTR builds will fix this because otherwise we won't know when new downstream packages break Ribasim Python. We can also consider therefore sticking with the latest QGIS release in dev, and perhaps testing LTR separately.

deltamarnix:
> This looks good, thanks! Looking at the changes in the lock file, this downgrades a lot of packages, e.g. numpy back to 1, pyarrow back to 14. I guess that is all because the QGIS LTR build is not compatible with those yet? I do hope that new LTR builds will fix this because otherwise we won't know when new downstream packages break Ribasim Python. We can also consider therefore sticking with the latest QGIS release in dev, and perhaps testing LTR separately.

If that is the case, maybe it's better to resolve them in separate environments. Developing QGIS is different than developing Ribasim python. It will even make us use a different python version for that specific piece of code, right?

visr:
Yeah perhaps a separate environment would be better. I didn't check the python version in this lockfile, but the conda-forge QGIS builds do allow a much newer Python compared to the qgis.org builds.

deltamarnix:
@visr I have removed the QGIS fixation, as QGIS tests and Delwaq tests were failing.

Issue #1674, opened 2024-08-01
Save solver stats per saveat


SouthEndMusic:
Currently motivated by https://github.com/Deltares/Ribasim/issues/1414, I'm looking for ways to visualize how the solver is performing. To this end I want to obtain solver stats (`model.integrator.sol.stats`) per saveat as an output table.

PR #1677, opened 2024-08-01
Write output table with solver stats per saveat

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1674.

Not sure how to test this?


visr (commented):
Does this look backwards over time, e.g. that a record for t_i, is over the period t_i-1 to t_i? It would be consistent with the other tables if it is over the coming period.

I think `water_balance_calls` can be a bit confusing, perhaps `rhs_calls` or `f_calls` would be better. Otherwise we encode our internal function name in our data model.

Jingru923 (approved):
The changes look good to me.

Issue #1690, opened 2024-08-02
Go back to globally unique node IDs?


deltamarnix:
The edge table should only consist of (fid), name, from_node_id, to_node_id, edge_type, subnetwork_id.

But now it also contains from_node_type and to_node_type. This information is superfluous, because the information is already in the Node table.

We should get rid of it, because it makes changing the node type (e.g. from DiscreteControl to PidControl) hard in applications like QGIS. I want to go back to globally unique node ids.

That's the opposite of #1513

visr:
We didn't really document the reasoning why you now need both node ID and type (`Basin #5`) to uniquely specify a node.

The reason was to simplify the design and implementation of the Python API for adding nodes.

```python
model.basin.add(Node(1, Point(0.0, 0.0)), basin_data)
```

The Node table for all Basins is in model.basin, not model. Only on writing out the model or calling `model.node_table()` is the whole Node table combined. So when adding a Basin we only need to check locally rather than keeping track of all IDs.

That doesn't mean we cannot change this though.

Issue #1689, opened 2024-08-02
Edge snapping is performed after saving layer in QGIS


deltamarnix:
After you create two nodes and snap an edge in between, you receive a popup where you can input a lot of information. It is not clear that you should actually leave everything empty and first save the layer. That is when the code kicks off and snaps the information from the nodes onto the edges.

Issue #1688, opened 2024-08-02
Adding nodes in QGIS lets users add the same node_id multiple times


deltamarnix:
If you add a node, in the feature attributes table you can set the same node_id twice.
This could be detected by QGIS before pressing OK.

See: https://gis.stackexchange.com/questions/339737/populating-text-field-with-predefined-attributes-in-qgis

deltamarnix:
I edited the attribute forms via QGIS, but it is saved in the project file. I am going to guess that editing the TABLE information on creation can be detected by QGIS, but the other way around it doesn't seem to be altered into the geopackage by QGIS.

deltamarnix:
The same issue holds for node_type. The default is NULL and therefore you get white dots instead of basins (or something else).

![image](https://github.com/user-attachments/assets/08d7433a-efa9-4a26-a692-4a2fdc735aaa)

visr:
Just want to add that it is currently the combination of node_id and node_type that needs to be unique. For that combination this issue remains. Name can be duplicated.

Issue #1685, opened 2024-08-02
Error when adding the same table twice in QGIS


deltamarnix:
- Open QGIS and ensure that the Ribasim plugin is installed and enabled.
- Open the application via the Ribasim button on the QGIS toolbar: _Ribasim panel opens_.
- Press the "New" button in the Model tab: _file navigation window pops up_.
- Fill in a name ("test1").
- Press OK: _A toml and database file are created on the given location_.
- Go to the Nodes tab: _All buttons are enabled_.
- Press the Basin / time button: _A layer is added to the layers panel with that name_.
- Press the Basin / time button again: _Error occurs_.

> An error has occurred while executing Python code: 
> 
> ValueError: Name already exists in geopackage: Basin / time 
> Traceback (most recent call last):
>   File "C:\Users\X\Source\Repos\Ribasim\.pixi\qgis_env\profiles\default/python/plugins\ribasim_qgis\widgets\nodes_widget.py", line 61, in new_node_layer
>     node = klass.create(
>            ^^^^^^^^^^^^^
>   File "C:\Users\X\Source\Repos\Ribasim\.pixi\qgis_env\profiles\default/python/plugins\ribasim_qgis\core\nodes.py", line 86, in create
>     raise ValueError(f"Name already exists in geopackage: {cls.input_type()}")
> ValueError: Name already exists in geopackage: Basin / time

Expected: The button gets disabled once a layer with that name was there.

Huite:
Disabling is one option, but it's difficult to implement. I would raise an error here and show a QMessageBox instead.

The reason that it's difficult to implement is because there's all kinds of ways in which people may delete the layer from the underlying geopackage. Through the "Remove from Dataset" button is one, but another is through the Browser Panel. Another is running an SQLite query through QGIS's processing toolbox, or maybe via a script in the background. This means you cannot rely on the Qt signals to inform you that the button should be re-enabled (granted, >99% of deletion would occur through "Remove from Dataset").

What easier to implement is prior to creating the layer, check the layers in the geopackage. Then error message if the name already exists.
To be fair this check already occurs and you could just try-except this specific ValueError and just show this error in a QMessageBox.

PR #1691, opened 2024-08-02
Initial qgis test plan

### LINKED ISSUE #320: Test QGIS plugin

deltamarnix:
An initial QGIS test plan for the plugin. There are still some tests to be written.
I have also performed all these tests and created quite some new reports.
There is also a known issues page for all of Ribasim, as we have some things that simply don't work as expected.

New issues:

* #1678 
* #1679 
* #1681 
* #1682 
* #1683 
* #1684 
* #1685 
* #1688 
* #1689 
* #1690 

Fixes #320


visr:
We should also include these: https://github.com/Deltares/Ribasim/pull/1748#issuecomment-2306368298

visr (approved):

Issue #1414, opened 2024-04-22
Convergence of large outflows from small Basins


visr:
If you have a small Basin with e.g. a large infiltration, the model is not very stable. We need to build a test model for this case to make sure it is able to handle such a case well. In principle we use a `reduction_factor` to reduce the infiltration, but for sufficiently large outflows it will step over this smooth landing within one timestep. We need to look into this case to understand the behavior, and how it behaves with both adaptive and fixed timestepping, under different solvers. Perhaps we need to use better tolerances or error norms to handle these cases better.

Related to #1406.

SouthEndMusic:
This PR proposes a method to handle basins drying out using `VectorContinuousCallback`, which we should test with a large model containing basins that are drying out: https://github.com/Deltares/Ribasim/issues/1575

evetion:
Commenting here instead of on the PR, it's good to read https://docs.sciml.ai/DiffEqDocs/stable/basics/faq/#My-ODE-goes-negative-but-should-stay-positive,-what-tools-can-help? and related the [PositiveDomain](https://docs.sciml.ai/DiffEqCallbacks/stable/step_control/#DiffEqCallbacks.PositiveDomain) callback, which specifically states its a more performant alternative to `isoutofdomain`.

However while isoutofdomain is "blind" (requiring many iterations before arriving at a very small timestep), using a callback might not solve it, as I can imagine rootfinding is difficult when we have a incredibly smooth reduction factor applied to q (from some arbitrary threshold onwards). This leads to the question best answered by @visr, why/when did we introduce the reduction factor (for basins) and more generally can we do without (and clamp or something when under a certain threshold/level)?

evetion:
PS. When someone implements a quick (e.g. GPU) D-Flow FM alternative, there's someone who always asks (not me), but can it handle "droogval" (waterlevel becoming zero)?! It's a harder and more common problem than you'd think.

visr:
In #639 I commented about PositiveDomain:

> One alternative is to use the [PositiveDomain](https://docs.sciml.ai/DiffEqCallbacks/stable/step_control/#DiffEqCallbacks.PositiveDomain) callback, but this only works for adaptive=false, and is not easily adapted to enforce only the storage part of the state to be positive.

Though I think also `isoutofdomain` only really makes sense for adaptive timesteps, since you can reject those. We had to implement an erroring fallback in case we still end up with negative storages in #1406.

[...]

PR #1693, opened 2024-08-05
Convergence large outflow from small basin

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1414


SouthEndMusic:
Interesting results for this PR with the new `Crystal` test model provided by @fati-mon (now part of our test models):

Runtime new branch: ~3.1 seconds
Run current main: Several minutes

evetion (commented):
Looking good already based on all the previous discussions! I would advise to keep it as simple as possible first, which IMHO would be using storage (instead of level) everywhere and setting both thresholds to zero. From there on we can experiment if needed, but probably first investigate why other models are unstable. 

Also, we probably should remove `isoutofdomain`, not sure how that interferes with the root finding of this continuouscallback.

SouthEndMusic:
Closed in favor of https://github.com/Deltares/Ribasim/pull/1697

visr:
This code could be useful in the future, but #1697 will do for a little while :)

Issue #1365, opened 2024-04-08
Consistently add axes labels in examples notebook


SouthEndMusic:
A lot of them are missing in the plots in `examples.ipynb`.

PR #1699, opened 2024-08-07
Plot labels + clean up running instructions in examples notebook

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1365.

In addition to the issue description I noticed that in the examples notebook only the first section contains elaborate running instructions. So I decided to create a link to there in all other sections.


Jingru923 (approved):
I made a tiny change to it and approve

PR #1701, opened 2024-08-08
Update GitHub action for updating lockfiles

visr:
This now uses the newly documented way to do this: https://pixi.sh/latest/advanced/updates_github_actions/

The main benefit for us is that it outputs a human readable diff.

### DIFF
--- a/.github/workflows/pixi_auto_update.yml
+++ b/.github/workflows/pixi_auto_update.yml
@@ -1,4 +1,8 @@
-name: Pixi auto update
+name: Update Pixi lockfile
+
+permissions:
+  contents: write
+  pull-requests: write
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


visr:
Hmm this doesn't seem to work yet, I triggered the action from this branch manually: https://github.com/Deltares/Ribasim/actions/runs/10301265855/job/28512425240

```
  × Cannot update pypi dependencies without first installing a conda prefix
  │ that includes python.
```

.github/workflows/pixi_auto_update.yml
```diff
@@ -8,23 +12,26 @@ on:
...
-          pixi-version: "latest"
-          cache: false
-      - name: Update pixi lock file
-        run: pixi update
-      - uses: peter-evans/create-pull-request@v6
+          run-install: false
+      - name: Update lockfiles
+        run: |
+          set -o pipefail
+          pixi update --json --no-install | pixi exec pixi-diff-to-markdown >> diff.md
```
Hofer-Julian, line 26:
```suggestion
          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
```

Seems like `--no-install` doesn't work (always?) if pypi dependencies are involved

visr: That makes sense, thanks! It works now: https://github.com/Deltares/Ribasim/pull/1708

Shall I add a line about this to https://pixi.sh/latest/advanced/updates_github_actions/?

Hofer-Julian: https://github.com/prefix-dev/pixi/issues/1773

PR #1710, opened 2024-08-09
Use pixi to fix julia package permissions on Windows

### LINKED ISSUE #1603: Work around Julia artifacts permission errors
Sometimes we get a [randomly failing build](https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_Windows_TestDelwaqCoupling/4097452), of the form:

> ERROR: IOError: rm("C:\\Users\\svc-teamcity-ansible\\.julia\\artifacts\\jl_wq7J7S\\bin"): directory not empty (ENOTEMPTY)

The upstream issue and workaround is https://github.com/JuliaLang/julia/issues/52272#issuecomment-2093625309.

```ps1
icacls .julia/artifacts /q /c /t /reset
```

Of course it is a lot of work to fix permissions manually on TeamCity agents, so we need something better.

@evetion suggested earlier to make a TeamCity job which can fix this error. Or perhaps we can just try fixing the permissions on each run, to never leave behind a broken agent.

evetion:
Fixes #1603

Given the recent changes in Teamcity, the original project is lost-ish https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_Ribasim_Windows_FixJuliaArtifactPermissions?branch=continuous_control&buildTypeTab=overview&mode=builds? As in automatic (?), greyed-out, no build steps or builds visible. Furthermore, the new Template(s) used is/are platform-agnostic, so harder to integrate a Windows specific fix.

So the easier way is just to integrate it into pixi, and always running it at CI time, also useful when it irrevocably happens on Windows laptops. However, there seems to be no way to have platform specific tasks (@Hofer-Julian is that correct and maybe worth an issue? I know you proposed the deno shell 😉), so I now just append an `exit`, ignoring any exit code of the previous command.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -21,8 +21,13 @@ test-ribasim-api = "pytest --basetemp=python/ribasim_api/tests/temp --junitxml=r
 [feature.dev.tasks]
 # Installation
 install-julia = "juliaup add 1.10.4 && juliaup override unset && juliaup override set 1.10.4"
+fix-julia-dir = "icacls $homedrive/$homepath/.julia/artifacts /q /c /t /reset;exit" # ignore errors (on non-Windows platforms)
 install-pre-commit = "pre-commit install"
-install-ci = { depends_on = ["install-julia", "update-registry-julia"] }
+install-ci = { depends_on = [
+    "fix-julia-dir",
+    "install-julia",
+    "update-registry-julia",
+] }
 install = { depends_on = [
     "install-ci",
     "install-qgis-plugins",
@@ -77,7 +82,7 @@ lint = { depends_on = [
 build = { "cmd" = "julia --project build.jl", cwd = "build", depends_on = [
     "generate-testmodels",
     "initialize-julia",
-], env = {JULIA_SSL_CA_ROOTS_PATH = ""} }
+], env = { JULIA_SSL_CA_ROOTS_PATH = "" } }
 remove-artifacts = "julia --eval 'rm(joinpath(Base.DEPOT_PATH[1], \"artifacts\"), force=true, recursive=true)'"
 # Tests
 test-ribasim-cli = "pytest --numprocesses=4 --basetemp=build/tests/temp --junitxml=report.xml build/tests"
--- a/pixi.toml
+++ b/pixi.toml
@@ -21,10 +21,12 @@ test-ribasim-api = "pytest --basetemp=python/ribasim_api/tests/temp --junitxml=r
 [feature.dev.tasks]
 # Installation
 install-julia = "juliaup add 1.10.4 && juliaup override unset && juliaup override set 1.10.4"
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -21,8 +21,13 @@ test-ribasim-api = "pytest --basetemp=python/ribasim_api/tests/temp --junitxml=r
 [feature.dev.tasks]
 # Installation
 install-julia = "juliaup add 1.10.4 && juliaup override unset && juliaup override set 1.10.4"
+fix-julia-dir = "icacls $homedrive/$homepath/.julia/artifacts /q /c /t /reset;exit 0" # ignore errors (on non-Windows platforms)
```
visr, line 24:
```suggestion
# Workaround rare issue, only on Windows, hence uses exit 0 to ignore errors
# Upstream issue: https://github.com/JuliaLang/julia/issues/52272
reset-artifact-permissions = "icacls $homedrive/$homepath/.julia/artifacts /q /c /t /reset;exit 0"
```

pixi.toml
```diff
@@ -21,8 +21,13 @@ test-ribasim-api = "pytest --basetemp=python/ribasim_api/tests/temp --junitxml=r
...
 install-pre-commit = "pre-commit install"
-install-ci = { depends_on = ["install-julia", "update-registry-julia"] }
+install-ci = { depends_on = [
+    "fix-julia-dir",
```
visr, line 27:
```suggestion
    "reset-artifact-permissions",
```

visr (approved):
Looks good to me. Technically the DEPOT_PATH can also be different, but this should in practice be sufficient.

Hofer-Julian:
> However, there seems to be no way to have platform specific tasks

I think there could be a use case for dispatching which would also fulfill your use case.
Like `pixi run task_x` would run something else on linux than on windows. If a linux task doesn't exist it could be ignored like in your case. Feel free to tag me if you open an issue for this.

In your case, I would recommend a Python script, instead of your `exit 0` hack.

evetion:
> I think there could be a use case for dispatching which would also fulfill your use case. Like `pixi run task_x` would run something else on linux than on windows. If a linux task doesn't exist it could be ignored like in your case. Feel free to tag me if you open an issue for this.

Thanks, I will make an issue.

> In your case, I would recommend a Python script, instead of your `exit 0` hack.

I considered this, but I disliked the extra script and the layer of indirection.

evetion:
> @evetion turns out platform specific tasks is already a thing: [pixi.sh/dev/reference/project_configuration#the-target-table](https://pixi.sh/dev/reference/project_configuration/#the-target-table)

Ah great stuff! I made a PR. While I remembered there being overrides (had to use them for mac), I never connected these to tasks, and couldn't find examples in the documentation.

PR #1713, opened 2024-08-12
Save QGIS styles in geopackage and remove them from QGIS plugin

### LINKED ISSUE #610: Write QGIS style to GeoPackage
Currently the Ribasim QGIS plugin applies styling to the layers on loading a model.

It would be nice if we could get the styling information out of the plugin and into the GeoPackage itself so users can start with only Ribasim Python to build a model from skratch and inspect the network. 

That way the plugin does not need to be used for a good looking label. QGIS supports saving styling information to the GeoPackage. I think this is QGIS specific, and not sure how/if the GeoPackage spec says anything about it (it seems not).

It's worth looking into if we can get ribasim-python to write this information to the GeoPackage.

evetion:
Fixes #610

I saved the styles to disk, and load them when writing a new geopackage. Furthermore, the styling part is removed from the QGIS plugin, which now always loads the style files (after symlinking them in the install step) as well.

@DanielTollenaar I started from the code in Ribasim-NL and hence made you co-author. However, I did change the code significantly (use parameterized sql, no regex needed on the styles, and required registration of the styles table in gpkg_contents).


python/ribasim/ribasim/styles/README.md
```diff
@@ -0,0 +1,6 @@
+# Ribasim QGIS Styles
+
+One can update the Ribasim styles in QGIS itself, saving the style to the geopackage, and retrieve it from there.
```
visr, line 3:
```suggestion
Ribasim Python always adds layer styles to the GeoPackage, such that the node types can be recognized also without the Ribasim QGIS plugin installed.

One can update the Ribasim styles in QGIS itself, saving the style to the GeoPackage, and retrieve it from there.
```

python/ribasim/ribasim/styles/README.md
```diff
@@ -0,0 +1,6 @@
...
+See https://github.com/Deltares/Ribasim/issues/610#issuecomment-1729511312.
+
+Here we store the Node and Edge styles, saved in both the qml and sld format, from the `styleQML` and `styleSLD` attributes in the `layer_styles` in the geopackage, respectively.
```
visr, line 6:
```suggestion
Here we store the Node and Edge styles, saved in both the QML and SLD format, from the `styleQML` and `styleSLD` attributes in the `layer_styles` in the GeoPackage, respectively.
```

visr (approved):

PR #1727, opened 2024-08-14
Support transient user demand return factor

### LINKED ISSUE #1230: Add support for transient `UserDemand` return factor
@Fati-Mon requested for one of her models that we support transient `UserDemand` return flow. 

This is currently not supported, but it is already possible to put it in the `UserDemand \ time` table. https://github.com/Deltares/Ribasim/issues/1229 is for avoiding confusion about this for `time` tables in general.

Jingru923:
Fixes #1230


core/src/read.jl
```diff
@@ -932,7 +932,12 @@ function user_demand_static!(
         user_demand_idx = searchsortedfirst(ids, first_row.node_id)
 
         active[user_demand_idx] = coalesce(first_row.active, true)
-        return_factor[user_demand_idx] = first_row.return_factor
+        return_factor_old = return_factor[user_demand_idx]
+        return_factor[user_demand_idx] = LinearInterpolation(
+            fill(first_row.return_factor, 2),
+            return_factor_old.t;
+            extrapolate = true,
+        )# first_row.return_factor
```
SouthEndMusic, line 940:
```suggestion
        )
```

core/src/read.jl
```diff
@@ -1022,7 +1042,8 @@ function UserDemand(db::DB, config::Config, graph::MetaGraph)::UserDemand
     ]
     demand_from_timeseries = fill(false, n_user)
     allocated = fill(Inf, n_user, n_priority)
-    return_factor = zeros(n_user)
+    return_factor =
+        [LinearInterpolation(zeros(2), trivial_timespan) for i in eachindex(node_ids)]#zeros(n_user)
```
SouthEndMusic, line 1046:
```suggestion
        [LinearInterpolation(zeros(2), trivial_timespan) for i in eachindex(node_ids)]
```

SouthEndMusic (commented):
Looking good! You should test it by having a simple model without allocation with a transient return factor

core/test/run_models_test.jl
```diff
@@ -395,8 +395,8 @@ end
     @test only(model.integrator.sol(0day)) == 1000.0
     # constant UserDemand withdraws to 0.9m/900m3
     @test only(model.integrator.sol(150day)) ≈ 900 atol = 5
-    # dynamic UserDemand withdraws to 0.5m/509m3
-    @test only(model.integrator.sol(180day)) ≈ 509 atol = 1
+    # dynamic UserDemand withdraws to 0.5m/503m3
```
SouthEndMusic, line 398: This is still too coarse a test. You should have a look at `DataFrame(Ribasim.flow_table(model))`, and check whether the outflow to inflow proportion matches your transient return factor

Jingru923: What is the relation between out-to-inflow proportion and this test?

Jingru923: Timepoint at 179day show a return ratio of 0.197ish and return factor is 0.2

SouthEndMusic:
I made some modifications to the test model so that:
- The different user demand nodes abstract flow at different times so that they can be tested individually
- The interpolations are not so steep that averaging flows makes testing hard

core/src/read.jl
```diff
@@ -985,10 +1003,10 @@ function user_demand_time!(
         )
         demand[user_demand_idx, priority_idx] = demand_p_itp(0.0)
 
-        if is_valid
+        if is_valid_demand
             demand_itp[user_demand_idx][priority_idx] = demand_p_itp
         else
-            @error "The demand(t) relationship for UserDemand $node_id of priority $p from the time table has repeated timestamps, this can not be interpolated."
+            @error "The demand(t) relationship for UserDemand $(first_row.node_id) of priority $priority_idx from the time table has repeated timestamps, this can not be interpolated."
```
SouthEndMusic, line 1009:
```suggestion
            @error "The demand(t) relationship for UserDemand $(first_row.node_id) of priority $(first_row.priority) from the time table has repeated timestamps, this can not be interpolated."
```

`priority_idx` is not exposed to the modeller, this is something that is only used internally, so that provided priorities do not have to be consecutive

SouthEndMusic (approved):
I had one last remark, other than that this is good to if you agree with my changes.

Issue #1413, opened 2024-04-22
Error on large water balance error


visr:
**What**
If the water balance error exceeds a certain fraction, we want the simulation to stop with a clear error message.

**How**
We should probably add a solver option for this with a reasonable default fraction like 0.01.

**Additional context**
- #1406 is a similar error that we should also throw at runtime.
- #1367 calculated the water balance error for the results, after the simulation finished.
- #452 is relevant for pointing to the problematic nodes in an error message.

evetion:
We could consider implementing a continuity tracer for this (which gives a fraction around 1 per basin?) as well during the run? 

Or did you plan on moving the logic in https://github.com/Deltares/Ribasim/pull/1367 to a callback?

visr:
Yeah the latter.

visr:
Done in #1819, in the function `check_water_balance_error`.

PR #1735, opened 2024-08-19
Refactor flow integration

SouthEndMusic:
The flow integration is now refactored so that integrated flows are computed based on `water_balance!` calls after the timestep but before parameter updates, so that the flows are calculated with the values of the discrete parameters that they had over the last timestep.

I also made sure that the `db` is closed properly in the tests.

This PR was originally intended to fix https://github.com/Deltares/Ribasim/issues/1413, but that was put on hold because the resulting water balance errors were quite large, which will possibly be fixed by https://github.com/Deltares/Ribasim/issues/1808.


SouthEndMusic:
TL;DR for @HendrikKok : there are probably still problems with flow integration in the current Ribasim main.

In the interval of a single timestep, discrete parameters (from DiscreteControl,  basin forcings) are fixed. That means that for accurate flow integration over that interval, all flows must be formulated with these correct parameters set. This is particularly important on the boundary point with the previous timestep, which we currently do wrong on main. To solve this systematically, flow integration should only happen with flows formulated in the single `integrate_flows!` call, which happens after a timestep and before the updates of discrete parameters for the next timestep.

I already know this isn't the holy grail for fixing the water balance error computations, because I experimented with integration schemes that satisfy the above and the large water balance errors are still there.

SouthEndMusic:
This has been made redundant by https://github.com/Deltares/Ribasim/pull/1819.

PR #1737, opened 2024-08-19
Make edge_id user configurable

### LINKED ISSUE #1718: Use of `fid` column in Geopackage
The `fid` column is mandatory (by SQLite) as a unique key for each row. By default it is set to autoincrement if not defined.

Our usage in Ribasim is inconsistent, we always set it (1:n) on writing, but it does not equal the node_id. However, for the edge table, it *is* used (implicitly) as our edge_id key.

Possibly solutions;
- Use fid as node_id key
- Add edge_id back

I'm inclined towards the latter. fid is a file format specific thing, and depending on the tooling is hidden/ignored (geopandas silently ignores any existing fid column).

Relates to #1717

evetion:
Fixes #1718

Sets node_id and edge_id as primary key fields in the geopackage. Also sets fid indexes for all non-node/edge tables. This now sorts the Node table by node_id only (used to be node_type first).

I triggered some bugs by the amount of code I touched, so I changed some more:
- Split contextvars into reading and writing (arrow file was trying to be read on writing a model).
- Simplified sorting
- Moved all the DataFrame(dict(**kwargs)) boilerplate into a single validator on TableModel.


visr:
Am I correct that this is only breaking in the sense that the new version expects `edge_id` in Edge, whereas old models don't have that. But reading and writing with Python will fix that? Can we read old models with Python or will it fail on the presence of a `fid` column?

evetion:
> Am I correct that this is only breaking in the sense that the new version expects `edge_id` in Edge, whereas old models don't have that. But reading and writing with Python will fix that? Can we read old models with Python or will it fail on the presence of a `fid` column?

Correct on both. You can read older models with Python, as pyogrio hardcodes the index to fid anyway, we rename it afterwards. 😄

visr (approved):

visr:
Regarding this point from #1718:

> The non-spatial tables (`Basin / static` etc) also get an `fid`, that falls outside our data model but was needed to make the tables editable in QGIS. This was done in https://github.com/Deltares/Ribasim/pull/873. To avoid confusion, Marnix tried to hide these columns in https://github.com/Deltares/Ribasim/issues/841#issuecomment-1842885072 but ran into a QGIS bug that has since been fixed. So we should try again to hide these columns for non-spatial tables.

Should we create an issue to start hiding `fid` columns in QGIS again? Am I correct that `fid` is now carried over, an no longer reset each time, by Ribasim Python? But it still isn't really part of the data model.

A bit of a nit, but in QGIS `node_id` shows up as the first column, but `edge_id` as the last. Can we make `edge_id` the first as well?

image

This id column is hardcoded in the iMOD QGIS plugin to `fid`, but that doesn't affect anything other than a confusing name:

image

https://github.com/Deltares/imod-qgis/blob/afa0eb694d367999c328e0dffc45f0bafb8230d1/imodqgis/timeseries/timeseries_widget.py#L471

python/ribasim/ribasim/input_base.py
```diff
@@ -216,8 +216,13 @@ def tablename(cls) -> str:
     @model_validator(mode="before")
     @classmethod
     def _check_dataframe(cls, value: Any) -> Any:
+        # Enable initialization with a Dict.
+        if isinstance(value, dict) and len(value) > 0 and "df" not in value:
+            value = DataFrame(dict(**value))
```
visr, line 221: What does this look like when used, and why was this done? Is it tested?

evetion: This replaces the `__init__` boilerplate for each nodetable (in the nodes folder, old code below). It's tested everytime you do something like `basin.Time(time=[])`.

```python
class Time(TableModel[BasinTimeSchema]):
    def __init__(self, **kwargs):
        super().__init__(df=DataFrame(dict(**kwargs)))
```

evetion:
> Should we create an issue to start hiding `fid` columns in QGIS again? Am I correct that `fid` is now carried over, an no longer reset each time, by Ribasim Python? But it still isn't really part of the data model.

I tried to make things consistent by having the indexes always there in Python with the same name as well, not just in geopackage. While the `fid` index resets during normal use (sorting when writing destroys the original fid), it should carry over to and from QGIS. We can make an issue, if you dislike seeing the fid, but I don't care that much either way.
 
> A bit of a nit, but in QGIS `node_id` shows up as the first column, but `edge_id` as the last. Can we make `edge_id` the first as well?

Not sure why that happens, in the table schema it shows up as the first column.

python/ribasim/ribasim/input_base.py
visr: Oh right I see, that's much nicer to define that once.

visr:
Great, let's merge then. The `fid` in the tables don't bother me that much either. `edge_id` on the left would be nice, but perhaps not under our control then.

PR #1742, opened 2024-08-20
Add CNAME file for docs deployment

### LINKED ISSUE #1740: Custom domain setting gets reset
We now use ribasim.org as the domain for our docs here, and ribasim.nl as the domain for https://github.com/Deltares/Ribasim-NL.

In the GitHub setting we specify this like so, under https://github.com/Deltares/Ribasim/settings/pages:
image

Both @evetion and me have already observed multiple times in the past few days that this custom domain is removed and the field is empty. When this happens we get a 404 on our docs. When filling in the settings again like above, it is back online.

Why does this setting get removed? Could it be a GitHub Enterprise policy or such?

evetion:
Fixes #1740

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+ribasim.org
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -1,8 +1,7 @@
 project:
   type: website
   resources:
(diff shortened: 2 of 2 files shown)


visr (approved):
Aha, excellent.

PR #1745, opened 2024-08-21
Give error when priorities parameter is missing and allocation is active

### LINKED ISSUE #1732: Priority parameter should be non-optional
**What**
Recently in solving issue #1724 , it has come to me that even thought the priority of user demand is not given, the whole program can still be run.

As this is not a desire behavior, investigation into the reason is needed.

Jingru923:
Fixes #1732 

Since the priorities parameter is needed for allocation, we now give an error if priorities parameter is missing and allocation is active


core/src/util.jl
```diff
@@ -407,7 +407,7 @@ end
 
 function get_all_priorities(db::DB, config::Config)::Vector{Int32}
     priorities = Set{Int32}()
-
+    valid_p = true
```
visr, line 410: I'd go with `valid` or `is_valid` for consistency with the other code.

core/src/util.jl
```diff
@@ -417,9 +417,21 @@ function get_all_priorities(db::DB, config::Config)::Vector{Int32}
...
     ]
-        union!(priorities, load_structvector(db, config, type).priority)
+        if valid_priorities(
+            load_structvector(db, config, type).priority,
+            config.allocation.use_allocation,
+        )
+            union!(priorities, load_structvector(db, config, type).priority)
+        else
+            valid_p = false
+            @error "Missing priority parameter(s) for a $type node in the allocation problem."
```
visr, line 427: I don't want to expose users to the internal name `Ribasim.UserDemandStaticV1`. I think you can use the `nodetype` function to get an actual name.

python/ribasim_testmodels/ribasim_testmodels/invalid.py
```diff
@@ -162,3 +165,67 @@ def invalid_unstable_model() -> Model:
         model.edge.add(model.basin[1 + id_shift * i], model.pump[2 + id_shift * i])
         model.edge.add(model.pump[2 + id_shift * i], model.terminal[3 + id_shift * i])
     return model
+
+
+def invalid_priorities_model() -> Model:
+    """Model with allocation active but missing priority parameter(s)."""
```
visr, line 171: Could you make this model as small as possible? That will make it easier to maintain the test models.

visr (approved):

PR #1748, opened 2024-08-22
Load qml styles in QGIS

evetion:
instead of SLDs, which we didn't have anymore (and didn't work even when they were there).

### DIFF
--- a/ribasim_qgis/core/nodes.py
+++ b/ribasim_qgis/core/nodes.py
@@ -69,6 +69,10 @@ class Input(abc.ABC):
         return False
 
     @classmethod
+    def fid_column(cls) -> str:
+        return "fid"
+
+    @classmethod
     def nodetype(cls):
         return cls.input_type().split("/")[0].strip()
 
@@ -83,6 +87,8 @@ class Input(abc.ABC):
             raise ValueError(f"Name already exists in geopackage: {cls.input_type()}")
         instance = cls(path)
         instance.layer = instance.new_layer(crs)
+        # Load style from QML file
+        instance.load_default_style()
         return instance
 
     def new_layer(self, crs: QgsCoordinateReferenceSystem) -> QgsVectorLayer:
@@ -128,16 +134,20 @@ class Input(abc.ABC):
         self.layer = QgsVectorLayer(
             f"{self._path}|layername={self.input_type()}", self.input_type()
         )
+        # Load style from database if exists
+        self.layer.loadDefaultStyle()
+        # Connect signal to save style to database when changed
+        self.layer.styleChanged.connect(self.save_style)
         return self.layer
 
     def from_geopackage(self) -> tuple[QgsVectorLayer, Any]:
         self.layer_from_geopackage()
-        self.load_style()
         return (self.layer, self.labels)
 
     def write(self) -> None:
-        self.layer = geopackage.write_layer(self._path, self.layer, self.input_type())
-        self.load_style()
+        self.layer = geopackage.write_layer(
+            self._path, self.layer, self.input_type(), fid=self.fid_column()
+        )
         self.set_defaults()
 
     def remove_from_geopackage(self) -> None:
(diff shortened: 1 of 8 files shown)


ribasim_qgis/core/nodes.py
```diff
@@ -125,19 +123,7 @@ def set_read_only(self) -> None:
 
     @property
     def renderer(self) -> QgsFeatureRenderer | None:
```
visr, line 125: Do we still need this function? Do we need to update the return type annotation to None only?

ribasim_qgis/core/nodes.py
```diff
@@ -151,7 +137,9 @@ def layer_from_geopackage(self) -> QgsVectorLayer:
 
     def from_geopackage(self) -> tuple[QgsVectorLayer, Any, Any]:
         self.layer_from_geopackage()
-        return (self.layer, self.renderer, self.labels)
+        fn = STYLE_DIR / f"{self.input_type().replace(' / ', '_')}Style.qml"
```
visr, line 140: I see `STYLE_DIR = Path(__file__).parent / "styles"`. I don't see the styles there, only in Ribasim Python. How does that work?

evetion: In the pixi task where we install the qgis plugin, it symlinks the styles folder.

Huite: Somebody else might just install the plugin as a zip, then the symlink would be missing? Or do I miss something?

evetion: I've now added the pixi (symlink) command to our Teamcity CI, so release zips contain the styles.

visr:
How does it work now with #1713 and this? Is this correct?

- We write Ribasim default QML styles inside the GeoPackage when writing out a model with Ribasim Python
- We write Ribasim default QML styles inside the GeoPackage when creating a new model with the plugin
- The GeoPackage embedded QML styles are shown when loading in QGIS outside of the plugin
- The Ribasim default QML styles are shown when loading with the QGIS plugin

I'd like to support a workflow where modelers can modify the styling once, and this is then retained and overrules the Ribasim default styling. For instance someone wants to disable Edge labels, and not have them pop up again each time the load it in QGIS, and also not after they modify the model in Ribasim Python. What would be a good way to support that? Does a sidecar QML file work like that and overrule the GeoPackage styling? Ideally a user can use QGIS itself to restore the Ribasim default QML style.

Supporting this workflow can also be moved to a follow up issue. Though perhaps slightly modifying this PR also enables this already. We could rename `load_style` into `load_default_style` and only apply it if there is no modified style.

image

ribasim_qgis/core/nodes.py
```diff
@@ -164,6 +147,10 @@ def set_editor_widget(self) -> None:
         # Calling during new_layer doesn't have any effect...
         pass
 
+    def load_style(self):
+        fn = STYLE_DIR / f"{self.input_type().replace(' / ', '_')}Style.qml"
+        self.layer.loadNamedStyle(str(fn))
```
visr, line 152: Is this loaded only in the layer, so only for the QGIS session? That means the plugin currently doesn't put the Ribasim default style in the GeoPackage for new tables right?

evetion:
> We write Ribasim default QML styles inside the GeoPackage when writing out a model with Ribasim Python ✅ 
> We write Ribasim default QML styles inside the GeoPackage when creating a new model with the plugin 🟥 
> The GeoPackage embedded QML styles are shown when loading in QGIS outside of the plugin ✅ 
> The Ribasim default QML styles are shown when loading with the QGIS plugin ✅ 

So user defined styles are not saved, and are overridden if you load them via the plugin. It's pretty much moving the goalposts territory now, but let's see what I can do.

evetion:
Ok @visr, this took a bit longer than I wanted, but in QGIS we now:
- Load a qml style on layer creation (from scratch)
- Save any current style to the geopackage when writing (also used on first creation, so this saves the qml style to db), and load the style from the geopackage when loading it again.
- For any layer loaded from a geopackage (thus with any styling), connect a signal that saves the style when the layer symbology is changed (otherwise it is never saved).

Over at Python I only write a style to geopackage when it doesn't exist yet.

visr (approved):

visr:
This is great, thanks! This can go in, I made some follow-up issues not to move the goalposts again ;)

PR #1749, opened 2024-08-22
Remove TimerOutputs

### LINKED ISSUE #1741: Remove TimerOutputs
We currently don't use it, but it is a dependency. Let's just remove it, including the `config.logging.timing` setting that was used for this.

visr:
Fixes #1741.

Note that this is still an indirect dependency.

```
(root) pkg> why TimerOutputs
  Ribasim → DiffEqCallbacks → NonlinearSolve → TimerOutputs
  Ribasim → OrdinaryDiffEq → NonlinearSolve → TimerOutputs
```


evetion:
I'm sorta sorry to see this go, but if we don't use it, why not.

evetion (approved):

visr:
Yeah same, we can always bring it back. Though I feel it is hard to get good enough timing information this way. Some of the stuff we care most about is happening in dependencies. Profiling is much more informative. And we have also added `results/solver_stats.arrow` since.

PR #1753, opened 2024-08-22
refactor `get_influx` for performance

SouthEndMusic:
This is a follow-up of https://github.com/Deltares/Ribasim/pull/1606.

Hopefully fixes the runtime of the HWS model.

### DIFF
--- a/core/src/util.jl
+++ b/core/src/util.jl
@@ -541,9 +541,10 @@ end
 
 function get_influx(basin::Basin, basin_idx::Int; prev::Bool = false)::Float64
     (; node_id, vertical_flux, vertical_flux_prev, vertical_flux_from_input) = basin
-    if prev
+    influx = if prev
         vertical_flux_prev.precipitation[basin_idx] -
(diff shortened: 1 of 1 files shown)


SouthEndMusic:
It does look like the HWS model is still quite a bit slower than before.

core/src/util.jl
```diff
@@ -540,12 +540,19 @@ function get_influx(basin::Basin, node_id::NodeID)::Float64
...
+        vertical_flux_prev.evaporation[basin_idx] + vertical_flux_prev.drainage[basin_idx] -
+        vertical_flux_prev.infiltration[basin_idx]
+    else
+        n = length(node_id)
+        vertical_flux = vertical_flux[parent(vertical_flux_from_input)]
+        vertical_flux[basin_idx] - # precipitation
+        vertical_flux[n + basin_idx] + # evaporation
+        vertical_flux[2n + basin_idx] - # drainage
+        vertical_flux[3n + basin_idx] # infiltration
+    end
```
Jingru923, line 555: Can you add the keyword return? So it makes the function more clear

Jingru923 (approved):

PR #1756, opened 2024-08-25
Add relationships between tables

### LINKED ISSUE #1648: QGIS: Show data for selected nodes
In QGIS you can select one or more nodes:

image

This can be used to show the Node table for those nodes:

image

However, if you want to see the data associated with this node, you need several extra steps. For instance, open the "Basin / profile" table, apply a field filter on `node_id`, and enter the node ID you selected manually.

image

Ideally a user can directly see and edit all data for a selected node. Perhaps we can for a node of type X, automatically apply a field filter on all tables `X / *` to show this. Or perhaps some other way that integrates nicely with existing QGIS functionality exists? Like perhaps automatically doing some kind of joins between these tables.

It looks like this would also be needed for #790.

### LINKED ISSUE #1688: Adding nodes in QGIS lets users add the same node_id multiple times

### LINKED ISSUE #1755: Edge labels in style not rendered by QGIS without plugin
Loading a GeoPackage directly looks the same as loading via the plugin, except for the Edge labeling. The Edge labels are set to `concat("name", ' #', "edge_id")`. QGIS does not show the ID for Edge, only for Node.

This is the `trivial` testmodel loaded outside the plugin:

image

And this is within the plugin:

image

It seems that the QGIS plugin is still setting the labeling explicitly. Ideally this code can be removed since it is part of the style. This should work since it works for Node as well.

I noticed this when reviewing #1748.

evetion:
Fixes #1648
Fixes #1755 (by updating the edge style)
Fixes #1688 (node_id has to be unique)

Also sets the node_id and edge_id to be unique, so they are autogenerated in the edit forms.

The current changes also will make the snapping + edge creation based on coordinates obsolete, as the form now looks like this:

Screenshot 2024-08-25 at 13 45 11

There are buttons to open a form with the Node, a highlighter to blink the node on the map, and a map selection panel, where one can click the desired node.


ribasim_qgis/widgets/ribasim_widget.py
```diff
@@ -106,7 +106,7 @@ def create_groups(self, name: str) -> None:
         assert project is not None
         root = project.layerTreeRoot()
         assert root is not None
-        self.group = root.addGroup(name)
+        self.group = root.insertGroup(0, name)  # insert at the top
```
evetion, line 109: This was requested by @DanielTollenaar

visr (approved):
This will make large model analysis so much easier!

> The current changes also will make the snapping + edge creation based on coordinates obsolete

The code for this is still there, with `self.edge_layer.editingStopped.connect(self.connect_nodes)`. How do you mean obsolete? That people will get geometries from IDs rather than IDs from snapped geometries? But the geometries are not automatically created right? If I fill in just IDs in the windows from your screenshot, I get `ValueError: Null geometry cannot be converted to a polyline.`.

This will need some updates of https://ribasim.org/tutorial/qgis.html and the test plan https://github.com/Deltares/Ribasim/pull/1691 as well..

evetion:
> That people will get geometries from IDs rather than IDs from snapped geometries?

Mostly that we will get the IDs from the user, and not from the snapped geometries. We need to think about that workflow. Geometry creation is still needed, but maybe snap the first and last points to the chosen node_ids in the form?

Anyway, that discussion is not blocking for this PR. I will update the documentation though.

ribasim_qgis/core/nodes.py
```diff
@@ -123,6 +123,15 @@ def set_dropdown(self, name: str, options: set[str]) -> None:
         )
         layer.setEditorWidgetSetup(index, setup)
 
+    def set_unique(self, name: str) -> None:
+        layer = self.layer
+        index = layer.fields().indexFromName(name)
+        setup = QgsEditorWidgetSetup(
```
deltamarnix, line 129: I tried to keep a separation between Qgis things (in the widgets folder), and core things that can be unit tested. It feels like this is being compromised with this new code as codecode also tells us there are no unit tests.

ribasim_qgis/widgets/dataset_widget.py
```diff
@@ -208,6 +212,32 @@ def add_selection_to_qgis(self) -> None:
         for item in selection:
             self.add_item_to_qgis(item)
 
+    @staticmethod
+    def add_relationship(from_layer, to_layer_id, name, fk="node_id") -> None:
+        rel = QgsRelation()
+        rel.setReferencingLayer(from_layer.id())
+        rel.setReferencedLayer(to_layer_id)
+        rel.setName(name)
+        rel.setStrength(rel.RelationStrength.Composition)  # type: ignore
```
deltamarnix, line 221: Why is the type ignored in this case? Version mismatch between the stubs?

evetion: The stubs do not seem to be aware of enums linked in a class.

docs/guide/qgis.qmd
```diff
@@ -91,6 +96,15 @@ Now leave the edit mode and save the results to the layer. Your model is now rea
 - In your model directory there is now a `results/` folder with `basin.arrow` and
   `flow.arrow` output files.
 
+# Inspect a (large) model {#sec-inspect}
+
+For larger models the node tables---especially the transient ones---can grow quite large. To facilitate inspection,
```
visr, line 101:
```suggestion
For larger models the node tables can grow quite large. To facilitate inspection,
```

Really long timeseries are better fit for storage in Arrow.

PR #1757, opened 2024-08-25
Try to fix missing import of pandera in Delwaq testing

evetion:
To fix failing Delwaq pipelines.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -254,8 +254,8 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/packaging-24.1-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/pandas-2.2.2-py312h1d6d2e6_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/pandas-stubs-2.2.2.240807-pyhd8ed1ab_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/pandera-0.20.3-hd8ed1ab_0.conda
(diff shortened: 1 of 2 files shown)


visr:
Hmm, wrapt was already in our environment hence the lock file didn't change, so it's not surprising this doesn't do anything, the same error still shows:

```
  .pixi\envs\dev\Lib\site-packages\pandera\__init__.py:31: in <module>
17:52:40       from pandera.decorators import check_input, check_io, check_output, check_types
17:52:40   .pixi\envs\dev\Lib\site-packages\pandera\decorators.py:25: in <module>
17:52:40       import wrapt
```

evetion:
Yeah, I had a sinking feeling when I saw no changes to the lock file. I'm at a loss of what's happening, my gut now says that the wrapt installation is corrupt on some Windows agents (without a clear pattern though) and that pixi doesn't detect that. How it would become corrupt (pandera nor wrapt had an update in recent times) is a mystery, but I am more suspicious of the Teamcity agents than pixi?

visr:
Can this be closed?

evetion:
Hmm, it's seems to be one agent failing intermittently. And most PRs have this as a failing test, so this could still be useful?

visr (approved):
Fine with me to merge this if you prefer, though I don't see Delwaq failing on the latest PRs, We would need an issue to remind us to remove this again though, to avoid a situation like #1766 where a dependency we don't use blocks things.

evetion:
Ok, let's close for now, since main has been green for a bit. It's mostly agent c-teamcity22092, I've cleaned the source on that agent, let's see what happens.

Issue #1705, opened 2024-08-08
Look into BackTracking effect on new test model


visr:
#1697 increased the performance considerably, and passed the tests with only slightly higher tolerances.
Surprisingly @Fati-Mon observed this behavior on a Basin running dry:

![image](https://github.com/user-attachments/assets/c7eaf0cd-862b-489f-ae65-5179a2183896)

![image](https://github.com/user-attachments/assets/04a8c81d-1353-4255-be4b-94a20402f09f)

We need to look into what is going on here.

Huite:
I guess it's possible that storage still ends up just barely negative somehow and one of the `du` terms blows up when it sees a negative storage?

Huite:
Hmm good point. It shouldn't be too late: if it converges with a negative storage, it should still be rejected.
Both the callback and `isoutdomain` should be triggered. So in this case I guess there's really no negative storage involved.

Maybe some term that isn't smooth enough around S = 0?

Does any other water balance term jump?

visr:
Yeah perhaps we are stepping completely over a reduction factor here. The inflow is 29 m3/s and the outflow 62 m3/s. The outflow should go down to the inflow due to the reduction factors, but all I see is a brief minor dip in the outflow over the Outlet (50 m3/s), and no reduced flow to the UserDemand (12 m3/s).

The complete water balance at the start of the anomaly:

| time                    | node_id | storage | level   | inflow_rate | outflow_rate | storage_rate | precipitation | evaporation | drainage |
|-------------------------|---------|---------|---------|-------------|--------------|--------------|---------------|-------------|----------|
| 2023-11-22 | 3       | 5.3E+06 | 6.1E+01 | 2.9E+01     | 6.2E+01      | -3.3E+01     | 0.0E+00       | 0.0E+00     | 0.0E+00  |
| 2023-11-23 | 3       | 2.4E+06 | 4.1E+01 | 2.9E+01     | 6.2E+01      | -2.7E+01     | 0.0E+00       | 0.0E+00     | 0.0E+00  |
| 2023-11-24 | 3       | 9.5E+04 | 7.9E+00 | 2.9E+01     | 6.2E+01      | 1.1E+04      | 0.0E+00       | 0.0E+00     | 0.0E+00  |
| 2023-11-25 | 3       | 9.4E+08 | 8.2E+02 | 2.9E+01     | 6.2E+01      | 3.8E+04      | 0.0E+00       | 0.0E+00     | 0.0E+00  |
[...]

PR #1761, opened 2024-08-26
Re-enable `BackTracking`

SouthEndMusic:
`BackTracking` as relaxation is now enabled again, with a thin wrapper to reject it when the residual gets worse. Upstream issue: 

https://github.com/SciML/OrdinaryDiffEq.jl/issues/2442


visr:
Since the sensible gradients don't fix the Crystal model with BackTracking, I guess we should not enable that yet?

How do these gradients compare with Shampine's advice in https://github.com/Deltares/Ribasim/issues/1705#issuecomment-2283312301, should we bring that back as well in this PR?

SouthEndMusic:
I added Shampine's advice, but it doesn't seem to change anything

core/src/solve.jl
```diff
@@ -51,9 +51,20 @@ function water_balance!(
     # Formulate du (controlled by PidControl)
     formulate_du_pid_controlled!(du, graph, pid_control)
 
+    # https://github.com/Deltares/Ribasim/issues/1705#issuecomment-2283293974
+    adapt_negative_storage_du!(du, u)
+
     return nothing
 end
 
+function adapt_negative_storage_du!(du, u)
```
visr, line 60:
```suggestion
    stop_declining_negative_storage!(du, u)

    return nothing
end

function stop_declining_negative_storage!(du, u)
```

core/src/util.jl
```diff
@@ -887,3 +896,66 @@ end
...
+get_area_and_level(basin::Basin, state_idx::Int, storage::GradientTracer) = storage, storage
+adapt_negative_storage_du!(du, u::ComponentVector{<:GradientTracer}) = nothing
+
+@kwdef struct MonitoredBackTracking{B, V}
+    linesearch::B = BackTracking()
+    dz_tmp::V = []
+    z_tmp::V = []
+end
+
+function resid(z, integrator, nlsolver, f)
```
visr, line 908: Needs a docstring.

evetion: Or a rename, now I think of it as res_id, not residual. 🤷🏻‍♂️

core/src/config.jl
```diff
@@ -230,7 +230,7 @@ const algorithms = Dict{String, Type}(
 )
 
 "Create an OrdinaryDiffEqAlgorithm from solver config"
-function algorithm(solver::Solver)::OrdinaryDiffEqAlgorithm
+function algorithm(solver::Solver, u0 = [])::OrdinaryDiffEqAlgorithm
```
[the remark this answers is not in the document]
SouthEndMusic: The default value is only there to not have to specify `u0` in tests where it's not relevant.

visr (approved):
Very nice find!

PR #1764, opened 2024-08-26
Enable migrations from older database schemas.

evetion:
Fixes https://github.com/Deltares/Ribasim-NL/issues/138

This has no schema version (stored in the geopackage) yet, but this is something we could implement in the future.


deltamarnix:
I would like to opt to add a geopackage version in this PR and let's call everything that doesn't have it version 0, and from now on we will write a `modelversion: 1`. Otherwise we might get conflicts when we figure out that more backwards compatibility is required in the coming months. Then while writing migration, first check if the version is correct.

- [ ] change QGIS writing
- [ ] change Ribasim python writing
- [ ] Write the geopackage version during migration if it was successful? (Maybe not, since it's only in memory migration)
- [ ] Check in your migration code what the version is to migrate

For now, let's keep it as simple as you have it, although I would rather see a pipeline that migrates step by step. But that's for another moment in time.

visr:
This doesn't yet include removing `listen_node_type` columns:

```
ValidationError: 1 validation error for TableModel[DiscreteControlVariableSchema]
df
  Value error, Unrecognized column 'listen_node_type'.
```

evetion:
@deltamarnix This now writes the schema version to the geopackage (also from QGIS), which is used to check whether to migrate or not. @visr This now also migrates the other tables with `listen_node_type`.

python/ribasim/ribasim/db_utils.py
```diff
@@ -0,0 +1,66 @@
+from contextlib import closing
+from pathlib import Path
+from sqlite3 import Connection, connect
+
+
+def esc_id(identifier: str) -> str:
+    """Escape SQLite identifiers."""
+    return '"' + identifier.replace('"', '""') + '"'
```
deltamarnix, line 8: Would something like this be more readable?

```suggestion
    identifier = identifier.replace('"', '""')
    return f'"{identifier}"'
```

python/ribasim/ribasim/db_utils.py
```diff
@@ -0,0 +1,66 @@
...
+    with closing(connect(db_path)) as connection:
+        if exists(connection, "ribasim_metadata"):
+            with closing(connection.cursor()) as cursor:
+                cursor.execute(
+                    "SELECT value FROM ribasim_metadata WHERE key='schema_version'"
+                )
+                result = int(cursor.fetchone()[0])
+        else:
+            result = 0
+    return result
```
deltamarnix, line 53:
```suggestion
        if not exists(connection, "ribasim_metadata"):
          return 0
        with closing(connection.cursor()) as cursor:
            cursor.execute(
                "SELECT value FROM ribasim_metadata WHERE key='schema_version'"
               )
                return int(cursor.fetchone()[0])
```

python/ribasim/ribasim/migrations.py
```diff
@@ -0,0 +1,71 @@
...
+        warnings.warn("Migrating outdated Basin / Static table.", UserWarning)
+        df.drop("urban_runoff", inplace=True, axis=1)
+
+    return df
+
+
+def basintimeschema_migration(df: DataFrame) -> DataFrame:
+    if "urban_runoff" in df.columns:
+        warnings.warn("Migrating outdated Basin / Time table.", UserWarning)
+        df.drop("urban_runoff", inplace=True, axis=1)
```
deltamarnix, line 41: Like codecov is saying, tests would be great.

ribasim_qgis/core/geopackage.py
```diff
@@ -50,6 +51,25 @@ def layers(path: Path) -> list[str]:
     return layers
 
 
+def write_schema_version(path: Path, version: int = 1) -> None:
```
deltamarnix, line 54: I would also love to see tests for this function.

deltamarnix (changes requested):
I want to see tests for the migration and the writing of the version files (both qgis and ribasim python)

python/ribasim/ribasim/db_utils.py
deltamarnix: It looks like a lot, but it's a simple if-inversion actually.

deltamarnix (approved):
If @SnippenE agrees that loading an old model in QGIS does not do the migration, but crashes instead, I am okay with the changes so far.

PR #1765, opened 2024-08-27
Python edge validation

### LINKED ISSUE #1752: Validate node types and number of edges when adding a new edge.
By either linking (templating) or copying from the core code.

DoD:
When run the model in python, error is given.

Jingru923:
Fixes #1752


python/ribasim/ribasim/geometry/edge.py
```diff
@@ -81,6 +82,11 @@ def add(
             the allocation algorithm, and should thus not be set for every edge in a subnetwork.
         **kwargs : Dict
         """
+        if not can_connect(from_node.node_type, to_node.node_type):
+            raise ValueError(
+                f"Node {to_node.node_type} cannot connect with upstream node {from_node.node_type}"
```
evetion, line 87:
```suggestion
                f"Node of type {from_node.node_type} cannot be upstream of node of type {to_node.node_type}"
```

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,46 @@
+# Table for connectivity
+# "basin": ["linear_resistance"] means that the downstream of basin can be linear_resistance only
+connectivity = {
```
evetion, line 3: Maybe give it a type hint?

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,46 @@
...
+        "ManningResistance",
+        "TabulatedRatingCurve",
+        "Pump",
+        "Outlet",
+    ],
+}
+
+
+# Function to validate connection
+def can_connect(node_up: str, node_down: str) -> bool:
```
evetion, line 43: Maybe also give back the allowed Node types, so we can give a hint in the ValueError?

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,46 @@
+# Table for connectivity
```
evetion, line 1: Are you planning on generating this dict automatically?

python/ribasim/ribasim/validation.py
evetion: Well that one is deprecated. I meant `connectivity: dict[str, list[str]] = {...`

python/ribasim/ribasim/validation.py
Jingru923: Since the current change are a bit large, is it okay to make it into an new issue?

evetion: I think it would be good to do it in this PR, as it would otherwise just be the next issue to pick up? I can review the rest in between.

Jingru923:
#1779 as a follow up issue

python/ribasim/ribasim/validation.py
Jingru923: Auto-generating is a different scope from validation. I still think it would be better to do it in a different issue.

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,83 @@
...
+# "Basin": ["LinearResistance"] means that the downstream of basin can be LinearResistance only
+connectivity: dict[str, list[str]] = {
```
evetion, line 3: Might be good to rename to `node_type_connectivity`.

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,83 @@
...
+        "ManningResistance",
+        "TabulatedRatingCurve",
+        "Pump",
+        "Outlet",
+    ],
+}
+
+
+# Function to validate connection
+def can_connect(node_up: str, node_down: str) -> bool:
```
evetion, line 43: Maybe `can_node_type_connect`? And for the arguments, make clear it's about the type, e.g. `node_type_up`?

python/ribasim/ribasim/validation.py
```diff
@@ -0,0 +1,83 @@
...
+    if node_up in connectivity:
+        return node_down in connectivity[node_up]
+    return False
+
+
+flow_edge_amount: dict[str, list[int]] = {
```
evetion, line 49: Maybe add neighbor into this name, and a comment on what the integers in the list[int] mean?

python/ribasim/ribasim/geometry/edge.py
```diff
@@ -112,13 +123,62 @@ def add(
...
+        self._validate_edge(to_node, from_node, edge_type)
+
         self.df = GeoDataFrame[EdgeSchema](pd.concat([self.df, table_to_append]))
         if self.df.duplicated(subset=["from_node_id", "to_node_id"]).any():
             raise ValueError(
                 f"Edges have to be unique, but edge with from_node_id {from_node.node_id} to_node_id {to_node.node_id} already exists."
             )
         self._used_edge_ids.add(edge_id)
 
+    def _validate_edge(self, to_node: NodeData, from_node: NodeData, edge_type: str):
```
evetion, line 135: You don't use edge_type now? Might be good to actually use it, so you can reduce the repetition here, and then you also don't need to do `& (
            edge_type == "flow"` in the final checks.

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,107 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+            df_node[["node_type"]], on="from_node_id", how="left", rsuffix="_from"
+        )
+        df_graph = df_graph.rename(columns={"node_type": "from_node_type"})
+
+        df_graph = df_graph.join(
+            df_node[["node_type"]], on="to_node_id", how="left", rsuffix="_to"
+        )
+        df_graph = df_graph.rename(columns={"node_type": "to_node_type"})
+
+        if not self._check_neighbors(
```
evetion, line 305: I would raise the ValueError in the function, and only call the function here. Same for the next statement.

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,107 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+                }
+                df_result = pd.concat(
+                    [df_result, pd.DataFrame([new_row])], ignore_index=True
+                )
+        for _, row in df_result.iterrows():
+            # from node's outneighbor
+            try:
+                if row["from_node_count"] < edge_amount[row["from_node_type"]][2]:
+                    is_valid = False
+                    raise ValueError(
```
evetion, line 352: Why raise an error if you always catch it? You can just pass your message to logging.error?

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,107 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+        if not self._check_neighbors(
+            df_graph, flow_edge_amount, "flow", df_node["node_type"]
+        ):
+            raise ValueError("Minimum flow inneighbor or outneighbor unsatisfied")
+        if not self._check_neighbors(
+            df_graph, control_edge_amount, "control", df_node["node_type"]
+        ):
+            raise ValueError("Minimum control inneighbor or outneighbor unsatisfied")
+
+    def _check_neighbors(
```
evetion, line 314: Can you add more inline comments or docstrings on the what and why? There's a lot going on, including some repetition, but I can't follow completely.

I think that two groupby operations (like your from_node_count) should be enough to check this?

evetion (changes requested):
Seems like a good approach, nice to see so many tests.

python/ribasim_testmodels/ribasim_testmodels/invalid.py
```diff
@@ -215,7 +220,7 @@ def invalid_priorities_model() -> Model:
     )
     model.edge.add(model.tabulated_rating_curve[2], model.basin[3])
     model.edge.add(model.basin[3], model.user_demand[4])
-    model.edge.add(model.basin[3], model.level_demand[6])
```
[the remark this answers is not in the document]
Jingru923: Switched the position of basin[3] and level_demand[6] because it was a mistake

python/ribasim/ribasim/validation.py
Jingru923: I think after changing the arguement names, it is very clear that this function is returning if two node type can be connected. I also update the comment. So maybe the function name can stay `can_connect`?

python/ribasim/ribasim/geometry/edge.py
Jingru923: Good point, I am using the edge_type yes. :D

python/ribasim_testmodels/ribasim_testmodels/invalid.py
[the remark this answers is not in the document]
evetion: But this is an invalid model, so that might've been on purpose?

python/ribasim/ribasim/model.py
Jingru923: I wanted to give customized error message, I didn't know that logging.error() can already show the error message on the screen. I just checked and now I do know

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,137 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+        self, nodes, node_info: pd.DataFrame, direction: str
+    ) -> pd.DataFrame:
+        """Loop over node table.
+
+        Add the nodes whose id are missing in the from_node and to_node column in the edge table because they are not the upstream or downstrem of other nodes.
+
+        Specify that their occurrence in from_node table or to_node table is 0.
+        """
+
+        if direction == "from":
```
evetion, line 391: You could create a column variable `f"{direction}_node_id"`, and use it to remove the if/else.

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,137 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+        for _, row in to_node_info.iterrows():
+            if row["to_node_count"] < edge_amount[row["to_node_type"]][0]:
+                is_valid = False
+                logging.error(
+                    f"Node {row['to_node_id']} must have at least {edge_amount[row['to_node_type']][0]} inneighbor(s) (got {row['to_node_count']})"
+                )
+
+        return is_valid
+
+    def _add_source_sink_node(
```
evetion, line 381: Not sure what source_sink means here? I'd suggest something like `_get_isolated_nodes`?

python/ribasim/ribasim/model.py
```diff
@@ -280,6 +285,137 @@ def write(self, filepath: str | PathLike[str]) -> Path:
...
+            df_node[["node_type"]], on="from_node_id", how="left", rsuffix="_from"
+        )
+        df_graph = df_graph.rename(columns={"node_type": "from_node_type"})
+
+        df_graph = df_graph.join(
+            df_node[["node_type"]], on="to_node_id", how="left", rsuffix="_to"
+        )
+        df_graph = df_graph.rename(columns={"node_type": "to_node_type"})
+
+        if not self._check_neighbor_amount(
```
evetion, line 305: nit: Ideally Python reads like a normal text. In that sense, `if not self._has_valid_neighbor_amount` is more readable than using `check`. With `check` in the name I'd expect the error to be thrown inside the check function.

evetion (changes requested):

python/ribasim/ribasim/geometry/edge.py
evetion: You could still further reduce the code here by only using the if/else to determine whether you need flow_edge_neighbor_amount or control_edge_neighbor_amount, and using those in 2 checks insteads of the 4 now.

python/ribasim_testmodels/ribasim_testmodels/invalid.py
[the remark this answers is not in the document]
Jingru923: No, I didn't do it on purpose. I only wanted to use this model to test the error of priority. But now it gave me two errors because I accidently connected node types that is not valid. So I corrected

python/ribasim/ribasim/model.py
Jingru923: In graph theory "source" is a node that doesn't have "in" edge and "sink" is a node that doesn't have "out" edge.

Maybe we have some more comprehensive names for these type of nodes?

evetion: Well sure, but something that's both a source and a sink, is neither. A source or a sink would be present in the edge table. You're looking for vertices with degree zero right? Those are often called isolate(d) (see for example this method https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.isolate.number_of_isolates.html), hence my suggestion for `_get_isolated_nodes`.

Jingru923: The main purpose of this function is to check the nodes that don't have one of the in- or out edge. In every model there are nodes like this. If a node doesn't have an in edge, it will not be checked because it doesn't end up in the to_node_count column. Same for the nodes that don't have out edge and don't appear in from_node_count column. So this function adds those nodes and specifies that there is 0 in/out edge.

This function can handle isolated node as well. However the case of isolated node is much less common.

Jingru923: Already did ^o^

python/ribasim/ribasim/geometry/edge.py
```diff
@@ -112,13 +123,46 @@ def add(
             index=pd.Index([edge_id], name="edge_id"),
         )
 
+        self._validate_edge(to_node, from_node, edge_type)
```
deltamarnix, line 126: Can this line go up before creating the `table_to_append` and maybe even further? You only seem to need to_node, from_node, edge_type.

python/ribasim/ribasim/geometry/edge.py
```diff
@@ -112,13 +123,46 @@ def add(
...
+            & (self.df["edge_type"] == edge_type)
+        ].shape[0]
+
+        out_neighbor: int = self.df.loc[
+            (self.df["from_node_id"] == from_node.node_id)
+            & (self.df["edge_type"] == edge_type)
+        ].shape[0]
+        # validation on neighbor amount
+        if edge_type == "flow":
+            if in_neighbor >= flow_edge_neighbor_amount[to_node.node_type][1]:
```
deltamarnix, line 148: In all these lines below I see quite some code duplication. Would be good to give `flow_edge_neighbor_amount[to_node.node_type][1]` a variable and reuse it on the line below. And do that for every of these checks.

python/ribasim/ribasim/model.py
```diff
@@ -272,8 +275,10 @@ def write(self, filepath: str | PathLike[str]) -> Path:
         filepath : str | PathLike[str]
             A file path with .toml extension.
         """
-        # TODO
-        # self.validate_model()
+
+        if self.use_validation:
```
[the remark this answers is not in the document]
Jingru923: Conclusion: public

PR #1773, opened 2024-08-28
Error when the toml or database can't be found.

### LINKED ISSUE #1652: Error if model database is not found on `ribasim.Model.read`
```python
import ribasim
from pathlib import Path

toml_path = Path(r"d:\repo\ribasim\Ribasim\ribasim_qgis\tests\data\simple_valid.toml")
model = ribasim.Model.read(toml_path)
```

`simple_valid.toml` is a TOML file. There is no `database.gpkg` associated with this file. That should probably throw an error. Instead, it creates a model object, and writes an empty `database.gpkg`.

I noticed this after adding the test in #1647. For now I just .gitignore that empty database.

evetion:
Fixes #1652

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -253,6 +253,8 @@ class Model(FileModel):
         filepath : str | PathLike[str]
             The path to the TOML file.
         """
+        if not Path(filepath).is_file():
+            raise FileNotFoundError(f"File '{filepath}' does not exist.")
         return cls(filepath=filepath)  # type: ignore
 
     def write(self, filepath: str | PathLike[str]) -> Path:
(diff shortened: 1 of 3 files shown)


evetion:
Interesting, these tests pass locally. It seems I have an empty gitignored database.gpkg in ribasim_qgis tests data folder. @visr can you confirm the existence of said file?

evetion:
> This was created when we started testing if that TOML is valid in Ribasim Python, which produced the GeoPackage. I assume that should be fixed now that you split the reading and writing context var?

Hmm, I don't follow. An empty file was created previously? Anyway, I propose to either commit the empty geopackage file or to touch the file from the test?

visr:
Ok sounds good. It's a bit odd that read works with an empty database but not no database.

evetion:
> Ok sounds good. It's a bit odd that read works with an empty database but not no database.

There's some complexity in the FileModel, it can use a spring cleaning as I mentioned in another issue. Initialization also passes some of these codepaths, so an empty db is almost identical to initializing an empty model. Which actually makes some sense 😉

visr (approved):

PR #1781, opened 2024-08-30
Use pyarrow dtype_backend

### LINKED ISSUE #1394: Mark table columns as optional as described in the docs
Looking at the Ribasim [docs](https://deltares.github.io/Ribasim/core/usage.html#sec-node) the optional table columns are well documented.

These docs don't really reflect what is happening on the Ribasim Python side.

## Expectation

If a column is non-optional and it isn't set, Ribasim Python should error with a clear error message.
The information shouldn't be duplicated too often. Ideally it would be shared between docs, Ribasim Core and Ribasim Python, but at least it should be shared between Ribasim Python and Core.

## Reality

- Ribasim Python floats and string are nullable and therefore optional if Core allow `missing` for them
https://github.com/Deltares/Ribasim/blob/cd011cb19c9787d614c9b34f02ce2d58cbb52554/core/src/schema.jl#L68
- Ribasim Python integers are always optional with default value 0, because `node_id` needs to have a default because of Ribasim Python implementation details
https://github.com/Deltares/Ribasim/blob/cd011cb19c9787d614c9b34f02ce2d58cbb52554/utils/templates/schemas.py.jinja#L18

## Options

- Switch to the pandas arrow backend which allows null for every data type
- Use pandas nullable integer data type https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html

### LINKED ISSUE #1721: Use Pandas Arrow backend
Currently we use the Pandas NumPy backend. Nowadays pandera also supports the Pandas [Arrow backend](https://pandas.pydata.org/docs/user_guide/pyarrow.html), as well as [polars](https://pola.rs/), which is also based on Arrow.

Given that we make a lot of use of Arrow, and the documented advantages of these representations, we should consider switching.

I think Polars and especially https://github.com/geopolars/geopolars is a bit too new at this point, and Pandas Arrow is probably a smaller step.

https://pandera.readthedocs.io/en/stable/dtype_validation.html#pyarrow-data-types

visr:
Fixes #1721
Fixes #1394

The `dtype_backend` part of this is not breaking. The only part that is technically breaking is that we specify a unit of milliseconds for the Arrow time type. Previously we used the default nanosecond precision, which was then truncated to milliseconds in the core. I think it is better to disallow precision higher than milliseconds if we cannot distinguish them in the core.


evetion:
I'd like to see some extra tests with adding/setting a default pandas dataframe, without the correct dtypte and time.

DanielTollenaar:
Tested it on De Dommel workflow and get a model that gets written with the Python-API and read and simulated without ValidationErrors or Exceptions. Didn't test/check any results, but the model + results are here: https://we.tl/t-Uxk9xra8i1

However, I do repeatedly get this FutureWarning which is messing up my logging quite alot:

```
(file:///D:/repositories/Ribasim/python/ribasim/ribasim/model.py:239): FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
  pd.concat(df_chunks)

```

core/src/util.jl
visr, line 1: Previously the `np.int32` priority columns didn't support missing, so there was no need to coalesce the missing.

core/src/validation.jl
visr, line 1: Non-functional change.

core/test/validation_test.jl
visr, line 1: Non-functional change.

docs/guide/examples.ipynb
visr, line 1: `pd.read_feather` also works fine without `dtype_backend="pyarrow"`, but for consistency with the input tables, and because it directly maps to the Feather (Arrow) format, I changed them all.

pixi.toml
visr, line 1: Unrelated, I got a cache hit after updating `generate-testmodels.py`.

python/ribasim/ribasim/delwaq/generate.py
```diff
@@ -70,17 +70,18 @@ def _make_boundary(data, boundary_type):
...
+    # Convert Arrow time to Numpy to avoid needing tzdata somehow
+    piv.time = piv.time.astype("datetime64[ns]").dt.strftime("%Y/%m/%d-%H:%M:%S")
     boundary = {
         "name": bid,
         "substances": list(map(_quote, piv.columns[1:])),
-        "df": piv.to_string(
-            formatters={"time": _quote}, header=False, index=False, na_rep=-999
-        ),
+        "df": piv.to_string(formatters={"time": _quote}, header=False, index=False),
     }
```
visr, line 85: These are possibly pandas bugs:

- `.dt.strftime` crashes trying to find tzdata, even though it is tz naive
- `.to_string`'s `na_rep` didn't work, I moved it up to `pivot_table`. The only difference is that in the textfile now `-999.0` rather than `-999` is written, but I think that is harmless.

python/ribasim/ribasim/input_base.py
```diff
@@ -374,7 +378,13 @@ def _from_db(cls, path: Path, table: str):
...
             if exists(connection, table):
                 # pyogrio hardcodes fid name on reading
-                df = gpd.read_file(path, layer=table, fid_as_index=True)
+                df = pyogrio.read_dataframe(
+                    path,
+                    layer=table,
+                    fid_as_index=True,
+                    use_arrow=True,
+                    arrow_to_pandas_kwargs={"types_mapper": pd.ArrowDtype},
+                )
```
visr, line 387: Only pyogrio supported this directly, geopandas not yet.

python/ribasim_testmodels/ribasim_testmodels/continuous_control.py
```diff
@@ -25,7 +25,9 @@ def outlet_continuous_control_model() -> Model:
         Node(1, Point(0, 0)),
         [
             level_boundary.Time(
-                time=pd.date_range(start="2020-01-01", end="2021-01-01", periods=100),
+                time=pd.date_range(
+                    start="2020-01-01", end="2021-01-01", periods=100, unit="ms"
+                ),
```
visr, line 30: This needed `unit="ms"` otherwise the nanosecond precision times get a ValidationError now.

python/ribasim/ribasim/input_base.py
evetion: Best to place an inline comment to the specific kwarg that's not supported yet?

python/ribasim_testmodels/ribasim_testmodels/continuous_control.py
evetion: Can't we automatically convert such timestamps? I fear that users might hit this and have to find this fix manually.

evetion (changes requested):
Looking good, but I have my doubts on how we handle the defaults of pandas (both for time as the arrow dtype).

In terms of dtype, I think it's automatically converted, so we should be fine there. However, in terms of time I'd rather we convert (maybe with a warning), rather than erroring out. That still seems an improvement over silently dropping precision. That said it's mostly a usability concern for me here.

python/ribasim/ribasim/utils.py
```diff
@@ -70,6 +71,21 @@ def _time_in_ns(df) -> None:
     df["time"] = df["time"].astype("datetime64[ns]")
 
 
+def _concat(dfs, **kwargs):
```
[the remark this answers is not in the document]
visr: I initially set out to fix it, but it is a lot more work, and as far as I can tell the behavior change in Pandas 3 will not affect us at all, so I went for the easy route instead.

python/ribasim/ribasim/schemas.py
```diff
@@ -28,287 +30,499 @@ def migrate(cls, df: Any, schema_version: int) -> Any:
...
+        nullable=True
+    )
 
 
 class BasinProfileSchema(_BaseSchema):
     fid: Index[Int32] = pa.Field(default=1, check_name=True, coerce=True)
-    node_id: Series[Int32] = pa.Field(nullable=False, default=0)
-    area: Series[float] = pa.Field(nullable=False)
-    level: Series[float] = pa.Field(nullable=False)
+    node_id: Series[Annotated[pd.ArrowDtype, pyarrow.int32()]] = pa.Field(nullable=True)
```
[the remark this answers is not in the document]
visr: Yes 0 was a dummy-default since `np.int32` doesn't support missings, but `pyarrow.int32` does.

docs/guide/examples.ipynb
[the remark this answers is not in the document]
visr: No these changes are not needed, so it will work if users omit it. Still considering it is Arrow data, this should be quite a bit faster, and more direct. I think we should promote best practice in the docs, even if it means an extra keyword argument.

python/ribasim_testmodels/ribasim_testmodels/continuous_control.py
visr: The only point where we got nanosecond precision was with the `data_range` with periods pattern. I feel like this pattern will rarely be used for real models, maybe only synthetic ones like our docs. I don't think I've ever come across hydrological data with nanosecond precision.

Nanosecond units will automatically convert to millisecond as long as it doesn't lose data. The ValidationError is also quite good. I prefer not to throw a warning on this as it usually doesn't matter. If you know an easy way to round or truncate to the nearest millisecond I'd be ok with that as well. But I doubt many users will run into this.

python/ribasim/ribasim/input_base.py
visr: Actually I misunderstood that all extra kwargs were passed from `gpd.read_file` to `pyogrio.read_dataframe` if `engine="pyogrio"`, so I went back to `gpd.read_file` and added a comment on `arrow_to_pandas_kwargs`, which ends up in `pyarrow`, not `pyogrio`.

python/ribasim/ribasim/input_base.py
```diff
@@ -294,7 +293,10 @@ def _from_db(cls, path: Path, table: str) -> pd.DataFrame | None:
             if exists(connection, table):
                 query = f"select * from {esc_id(table)}"
                 df = pd.read_sql_query(
-                    query, connection, parse_dates={"time": {"format": "ISO8601"}}
+                    query,
+                    connection,
+                    parse_dates={"time": {"format": "ISO8601"}},
```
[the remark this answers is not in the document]
visr: Done

python/ribasim/ribasim/schemas.py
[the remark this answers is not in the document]
evetion: Well its also because the node_id is a required field and can't be missing schema wise?

Issue #1778, opened 2024-08-29
Get rid of function `residual`


SouthEndMusic:
https://github.com/Deltares/Ribasim/pull/1761 introduced the function `residual` which  makes use of `OrdinaryDiffEq` internals. This should be removed when possible, probably when the upstream issue is fixed: https://github.com/SciML/OrdinaryDiffEq.jl/issues/2442.

PR #1784, opened 2024-09-02
Remove `OrdinaryDiffEq` internals

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1778.

See also https://github.com/SciML/OrdinaryDiffEq.jl/pull/2443.


SouthEndMusic:
Things I changed:
- Replace `LineSearches.jl` by `LineSearch.jl` (part of `SciML`)
- Use the latest version of `OrdinaryDiffEq.jl` (I had quite some problems installing directly from the GitHub repository, so I now use a path to the repo locally as a POC). Using `NonlinearSolveAlg` now works with finite difference, but using `ForwardDiff` is still broken. OrdinaryDiffEqRosenbrock is also broken.

It must be noted that even though `NonlinearSolveAlg` with finite difference and `BackTracking` from `LineSearch.jl` works, I observed a huge drop in performance for the HWS model

Issue #1771, opened 2024-08-26
Consider adding `min_upstream_level` and `max_downstream_level` to Pump and Outlet


DanielTollenaar:
Currently a Pump will drain the upstream Basin until it's empty. I don't know a surface water example where this occurs (in sewers it does). The idea of a pump is in >90% of the cases to maintain a certain `upstream_target_level` and switch off if that level is reached.

Some pumps also switch off if a `downstream_target_level` is exceeded in case of flooding (maalstop) or water supply.

If this could be implemented in Ribasim we could drop quite some PID or Discrete Control Nodes.

@rbruijnshkv, what is your experience, opinion with/on this?

Edit:

- [ ] Pump: `min_upstream_level`
- [ ] Pump: `max_downstream_level`
- [ ] Outlet: `min_upstream_level` (rename `min_crest_level`)
- [ ] Outlet: `max_downstream_level`
- [ ] TabulatedRatingCurve: `max_downstream_level` (moved to #1789)

SnippenE:
Discussed with Martijn: saves time in development Ribasim-NL and fit within the initial sprint planning

visr:
Here's a proposal, would like to hear from @DanielTollenaar and @rbruijnshkv if they think this sufficiently covers this use case:

Pumps and Outlets both need to stop as the upstream level gets too low, and stop as the downstream level gets too high.

I don't think using `target` in the new parameter names is a good idea, that to me suggests that it tries to reach this level from both sides, which it doesn't. PidControl has `target`.

Currently only Outlet supports upstream too low via `min_crest_level`. This is done via a reduction factor, such that it smoothly goes to 0 as it approaches `min_crest_level`. This smooth reduction factor will probably help the solver compared to a discrete step to 0 flow as DiscreteControl would do this. Also since Outlet is gravity-based it supports stopping if the downstream level exceeds the upstream level. Adding a maximal downstream level on top of that makes it more flexible.

How about these parameters:

Pump: `intake_level`
Pump: `max_outlet_level`
Outlet: `crest_level` (rename `min_crest_level` since IMO the `min_` is superfluous)
Outlet: `max_downstream_level`

I'm avoiding using the term `downstream` and `upstream` for Pump since it can be confusing if it Pumps water back upstream.

Are other parameters needed? I can imagine TabulatedRatingCurve: `max_downstream_level` could also be useful. Like Outlet the TabulatedRatingCurve is also gravity-based.

DanielTollenaar:
[...]

PR #1788, opened 2024-09-02
Rename Outlet's `min_crest_level` to `min_upstream_level`

visr:
The breaking part of https://github.com/Deltares/Ribasim/issues/1771, with added automatic migration.


evetion:
@visr I've added some migration specific changes here in [`6ebc208` (#1788)](https://github.com/Deltares/Ribasim/pull/1788/commits/6ebc208456bd416da17800bfffa5974055ea1c63), feel free to review. In the future we should update schemas incrementally (from 0-1, 1-2, instead of 0-2, 1-2), but this works for now.

PR #1796, opened 2024-09-04
Update reduction factors of resistance nodes

SouthEndMusic:
These changes are needed for running the `De Dommel` model for Ribasim NL and were requested by Neeltje. The main change is that `ManningResistance` has low storage factors now, which is important to have if the connected basins have different bottom levels.

I made an unified approach for low storage factors of `LinearResistance` and `ManningResistance`, where the former already had low storage factors for both connected basins. That however leads to weird behavior where water cannot flow into an empty basin via the resistance node, so now I made it so that the reduction factor is based on the storage of the upstream basin as defined by the flow direction. This feels a bit discontinuous but I think it's fine.

I also fixed an upstream issue in `FindFirstFunctions`, for now the manifest points at the branch with the fix: https://github.com/SciML/FindFirstFunctions.jl/pull/26.


visr:
The Ribasim builds fail with

```
WARNING: Method definition (::FindFirstFunctions.Guesser{T<:(AbstractArray{T, 1} where T)})(Any) in module FindFirstFunctions at C:\Users\svc-teamcity-ansible\.julia\packages\FindFirstFunctions\2w5qZ\src\FindFirstFunctions.jl:233 overwritten in module Ribasim at D:\buildAgent\work\ecd2b8f9b25b1609\ribasim\core\src\util.jl:69.
09:53:18   ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

Looking at https://discourse.julialang.org/t/error-method-overwriting-is-not-permitted-during-module-precompilation-use-precompile-false-to-opt-out-of-precompilation/109191, you can try out if we can override from `__init__`? Possibly you need eval for that. Or is there another way to work around this without overwriting methods?

visr:
With my previous comment and a unit test of `low_storage_factor_resistance_node` this should be good to go, nice find!

visr (approved):
I see your PR just got merged as well

PR #1797, opened 2024-09-04
Autogenerate the validation tables for Python

### LINKED ISSUE #1779: Auto-generate validation tables in python
We have already done this for `schema.py` which is auto-generated by jinja from core. We can probably do the same for the three table in the `validation.py`. This way when adding new nodes or changing the connection between nodes, python and core are always in sync.

Jingru923:
Fixes #1779


python/ribasim/ribasim/validation.py
```diff
@@ -1,85 +1,124 @@
...
-    "Outlet": [1, 1, 1, 1],
-    "Terminal": [1, int(1e9), 0, 0],
-    "DiscreteControl": [0, 0, 0, 0],
-    "ContinuousControl": [0, 0, 0, 0],
     "PidControl": [0, 0, 0, 0],
+    "LevelBoundary": [0, 9223372036854775807, 0, 9223372036854775807],
+    "Pump": [1, 1, 1, 1],
     "UserDemand": [1, 1, 1, 1],
-    "LevelDemand": [0, 0, 0, 0],
+    "TabulatedRatingCurve": [1, 1, 1, 1],
```
evetion, line 95: This is different than before. Did we make a mistake previously?

core/test/config_test.jl
```diff
@@ -82,3 +82,9 @@ end
     @test Ribasim.snake_case("ABCdef") == "a_b_cdef"
     @test Ribasim.snake_case("snake_case") == "snake_case"
 end
+
+@testitem "camel_case" begin
```
evetion, line 86: Also (or only?) test for the symbol input, as indicated by codecov

python/ribasim/ribasim/validation.py
```diff
@@ -1,85 +1,124 @@
...
+    "Terminal": [],
 }
 
 
-# Function to validate connectivity between two node types
-def can_connect(node_type_up: str, node_type_down: str) -> bool:
-    if node_type_up in node_type_connectivity:
-        return node_type_down in node_type_connectivity[node_type_up]
+# Function to validate connection
+def can_connect(node_up: str, node_down: str) -> bool:
```
evetion, line 84: This has lost the original argument names, which were better imho.

utils/gen_python.jl
```diff
@@ -55,3 +55,30 @@ open(normpath(@__DIR__, "..", "python", "ribasim", "ribasim", "schemas.py"), "w"
...
+    [
+        (
+            name = T,
+            connectivity = Set(
+                Ribasim.config.camel_case(x) for x in Ribasim.neighbortypes(T)
+            ),
+            flow_neighbor_bound = Ribasim.n_neighbor_bounds_flow(T),
+            control_neighbor_bound = Ribasim.n_neighbor_bounds_control(T),
+        ) for T in keys(Ribasim.config.nodekinds)
+    ]
```
evetion, line 72: Maybe we want to sort this list? That makes looking through the code a bit simpler?

evetion (approved):

python/ribasim/ribasim/validation.py
visr: No I had to remove support for multiple outflows for TabulatedRatingCurve and Pump to do max downstream level.

core/test/config_test.jl
Jingru923: Good idea to add the symbol input tests. Since this function also has an overloading that take strings, removing the string tests may trigger the codecov indication again. We should keep it

core/src/config.jl
```diff
@@ -49,6 +49,15 @@ end
 
 snake_case(sym::Symbol)::Symbol = Symbol(snake_case(String(sym)))
 
+"Convert a string from snake_case to CamelCase."
+function camel_case(snake_case::AbstractString)::String
```
deltamarnix, line 53: Take a look at titlecase

deltamarnix: image

deltamarnix: If it should become CamelCase, I would probably go for: `titlecase()`, and then `replace('_' => '')`

utils/templates/validation.py.jinja
```diff
@@ -0,0 +1,31 @@
...
+
+# Table for connectivity
+# "Basin": ["LinearResistance"] means that the downstream of basin can be LinearResistance only
+
+node_type_connectivity: dict[str, list[str]] = {
+{% for n in nodes %}
+    '{{n[:name]}}': [
+    {% for value in n[:connectivity] %}
+        '{{ value }}',
+    {% end %}],
```
deltamarnix, line 11: https://mommawatasu.github.io/OteraEngine.jl/dev/tutorial/#White-Space-Control

core/src/config.jl
Jingru923: I want this function to return the same string when the input is already camel case. 
image

Seems titlecase() is not able to do it because it treats CamelCase as one word and return Camelcase.

The old way is the same amount of lines but more robust

utils/gen_python.jl
Jingru923: I don't quite understand which list that you meant. But I will mark this down so I can resolve it when you are back!

utils/templates/validation.py.jinja
Jingru923: Good source! In this case the white space trimming doesn't do much, because after ruff formatting the code, the code looks the same.

PR #1798, opened 2024-09-04
update `FindFirstFunctions`

SouthEndMusic:
The new version contains my upstream fix https://github.com/SciML/FindFirstFunctions.jl/pull/26.

Is this the way to go, or should we keep it as a direct dependency to be able to set the compat?


visr (approved):
This looks good to me. It's very unlikely that we will suddenly get an older v1 version when updating packages.

PR #1801, opened 2024-09-05
Fix `min_crest_level` migration

visr:
The migration function from #1788 was wrong because I didn't know that Ribasim Python would already add the new optional column by the time it got to the migration. This now drops that one first, to avoid ending up with `min_upstream_level` twice in one dataframe.

This also makes the other functions a bit simpler by allowing errors when dropping columns that are no longer needed.


SouthEndMusic (approved):
This looks good to me as far as I can tell

PR #1803, opened 2024-09-05
Disallow FlowBoundary -> Terminal

visr:
Because that's just cruel.

image


visr:
On second thought, I think this could actually be useful, for instance as a way to represent known fluxes to the sea. You may want to have this in the model such that the flow to the sea is complete.

PR #1805, opened 2024-09-06
Fix the images in the quick start guide documentation

### LINKED ISSUE #1800: Storing image in our documentation in MinIO
**What**
The image we store in #1 , will be stored in MinIO.

**Why**
The old approach sometime return broken images. And it's hard to organized the images because they are posted in comments and you can't group them.

**How**
We can store the image in the MinIO, and retrieve them when needed. We normally build documentation in the following cases:
- during development, documentation part needs to be modified and previewed
- doc pipeline is run

The solution should be able retrieve pictures in abovementioned cases and make sure that documentation can be built with the pictures. 

Since at the moment only the images in the [quick start guide](https://ribasim.org/guide/quickstart.html) are broken, this issue will focus on recover the broken image in this page. It acts as a proof of concept.

Jingru923:
Fixes #1800

I updated the policy in MinIO and now we have a public folder especially for storing images for documentation called `doc-image`. All the image in quick start guide are in the `doc-image/quickstart`.

### DIFF
--- a/docs/guide/quickstart.qmd
+++ b/docs/guide/quickstart.qmd
@@ -2,7 +2,7 @@
 title: "Quick start guide"
 ---
 
-![](https://github.com/user-attachments/assets/467e71da-def4-4723-91de-f76bf78758fe){fig-align="left"}
+![](https://s3.deltares.nl/ribasim/doc-image/quickstart/cover.png){fig-align="left"}
 
 # Introduction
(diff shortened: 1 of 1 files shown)


visr (approved):
Nice, let's try this. If you don't mind, I'd replace the spaces in the filenames with dashes, then we don't need to encode them %20, that is more readable also.

PR #1810, opened 2024-09-09
Fix geopandas CRS warning

### LINKED ISSUE #1799: Geopandas concat CRS warning
When adding a node, @DanielTollenaar sees this warning:

d:\repositories\Ribasim-NL\.pixi\envs\dev\Lib\site-packages\geopandas\array.py:1638: UserWarning: CRS not set for some of the concatenation inputs. Setting output's CRS as Amersfoort / RD New (the single non-null crs provided).  return GeometryArray(data, crs=_get_common_crs(to_concat))

We need to set the CRS for both parts.

visr:
Fixes #1799
This looks into the parent class to set the CRS of the Node table directly after it is created.
That means we also don't have to do this (and create an extra copy) each time we call `node_table`.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -178,8 +178,6 @@ class Model(FileModel):
         return fn
 
     def _save(self, directory: DirectoryPath, input_dir: DirectoryPath):
-        # Set CRS of the tables to the CRS stored in the Model object
-        self.set_crs(self.crs)
         db_path = directory / input_dir / "database.gpkg"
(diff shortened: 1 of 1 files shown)


evetion:
Thanks for fixing this, I forgot to make an issue after @DanielTollenaar mentioned this warning. LGTM, but this means we can remove the set_crs from the write method, and from https://github.com/Deltares/Ribasim/blob/634714677cff6ef91ed37acd429949b2273b988e/python/ribasim/ribasim/model.py#L297 as well.

evetion (approved):

PR #1816, opened 2024-09-10
Fix bench

Jingru923:


.teamcity/Templates/RegressionTest.kt
```diff
@@ -59,7 +59,7 @@ open class RegressionTest (platformOs: String) : Template() {
                 scriptContent = header +
                         """
                 pixi run python utils/get_benchmark.py %MiniO_credential_token% "benchmark/"
-                pixi run python utils/get_benchmark.py %MiniO_credential_token% "hws_2024_7_0/"
+                pixi run python utils/get_benchmark.py %MiniO_credential_token% "hws_migration_test/"
```
evetion, line 62: Don't you need to checkout both models now?

evetion (approved):

.teamcity/Templates/RegressionTest.kt
Jingru923: No, we separated the hws model that is used for migration test and the one used for integration test. Because the one for integration test is now updated and has version number and newest parameters, but we still need an "old" model for the migration test.

Issue #1804, opened 2024-09-06
model.edge.add() starts with existing edges starts with edge_id=1, causing unique validation error


DanielTollenaar:
Below an example on how I add new edges using the add-API in current main (pulled 06-09-2024). Adding edges will add edge_id=1, going up one int every time I edge. Problem is, edge_id 0 till 1956 are already used causing a unique validation error on the edge_id column while writing

![image](https://github.com/user-attachments/assets/af1af229-05e6-410a-a43e-b07bbb87e153)

evetion:
Fixed in #1818

PR #1818, opened 2024-09-10
Update used ids on reading an existing model.

### LINKED ISSUE #1806: model.basin.add() in model with existing nodes will cause AssertionError next-time model.node_table() is is accessed
While not to explicitly check available node_id in model with existing nodes, I run this code:

```
        geometry = basin_to_point(basin_polygon=basin_polygon, tolerance=10)
        node = Node(
            meta_categorie="bergend",
            geometry=geometry,
        )
        data = [basin.Profile(level=[0, 1], area=[5, 100]), basin.Area(geometry=[basin_polygon])]
        basin_node = model.basin.add(node=node, tables=data)
```

Inspecting the `model.basin.node.df` I notice node_id =1 has been assigned. Does not exist in the basin-table, but does exist as `FlowBoundary`:
image

Thereby accessing `model.node_table()` will raise an `AssertionError: node_id must be unique`:
image

evetion:
Fixes #1806 and #1804

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -124,6 +124,16 @@ class Model(FileModel):
         self.edge.df.set_geometry("geometry", inplace=True, crs=self.crs)
         return self
 
+    @model_validator(mode="after")
+    def _update_used_ids(self) -> "EdgeTable":
+        # Only update the used node IDs if we read from a database
+        if "database" in context_file_loading.get():
+            df = self.node_table().df
+            if len(df.index) > 0:
+                self._used_node_ids.node_ids.update(df.index)
+                self._used_node_ids.max_node_id = df.index.max()
+        return self
+
     @field_serializer("input_dir", "results_dir")
     def _serialize_path(self, path: Path) -> str:
         return str(path)
@@ -224,7 +234,11 @@ class Model(FileModel):
     def node_table(self) -> NodeTable:
         """Compute the full sorted NodeTable from all node types."""
         df_chunks = [node.node.df for node in self._nodes()]
-        df = pd.concat(df_chunks)
+        df = (
+            pd.concat(df_chunks)
+            if df_chunks
+            else pd.DataFrame(index=pd.Index([], name="node_id"))
+        )
         node_table = NodeTable(df=df)
         node_table.sort()
         assert node_table.df is not None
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -224,7 +235,11 @@ def _apply_crs_function(self, function_name: str, crs: str) -> None:
     def node_table(self) -> NodeTable:
         """Compute the full sorted NodeTable from all node types."""
         df_chunks = [node.node.df for node in self._nodes()]
-        df = pd.concat(df_chunks)
+        df = (
+            pd.concat(df_chunks)
+            if df_chunks
+            else pd.DataFrame(index=pd.Index([], name="node_id"))
```
visr, line 241: This is to handle empty models, without any nodes right? Did you need this for anything? I don't think we test empty models.

I think this fails, just like before, but now with:

```
>>> ribasim.model.NodeTable(pd.Index([], name="node_id"))
  File "<stdin>", line 1, in <module>
TypeError: BaseModel.__init__() takes 1 positional argument but 2 were given
>>> ribasim.model.NodeTable(df=pd.Index([], name="node_id"))
  File "<stdin>", line 1, in <module>
  File "D:\repo\ribasim\Ribasim\.pixi\envs\default\Lib\site-packages\pydantic\main.py", line 193, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for NodeTable
df
  Value error, Expected <class 'pandas.core.series.Series'> to have name 'node_id', found 'None' [type=value_error, input_value=Index([], dtype='object', name='node_id'), input_type=Index]
    For further information visit https://errors.pydantic.dev/2.8/v/value_error
```

Wouldn't we need all columns with the right dtypes? With this we get an object dtype.

So perhaps best to leave out this change unless needed.

visr (approved):

python/ribasim/ribasim/model.py
evetion: Your examples are not what's happening in the code, you need to pass the dataframe as the df argument to the nodetable.

That said, this might not be needed in this PR, it's for the edge case when you read an empty geopackage. And when you call nodetable on a new model, which previously failed.

visr: Oh sorry I pasted two prompts, the first once was missing the `df` arguments, but the second one is not.

evetion: Your second one passes an index, not a dataframe 😉

visr: Ha my bad. Let's merge!

Issue #1825, opened 2024-09-19
Save all results in SavingCallback


visr:
Right now we use a combination of the SavingCallback and the `saveat` that goes into solve/init and hence into the `integrator.sol`. For the latter we need to take extra care that `sol.t` matches exactly the times that the SavingCallback is called. For instance to avoid duplicate `t`, we need to set `save_positions` in callbacks.

It is probably simpler to just do all saving with SavingCallbacks instead, so it is more consistent. Also we wouldn't store e.g. the PidControl error state that we don't currently use.

https://docs.sciml.ai/DiffEqDocs/stable/basics/common_solver_opts/#CommonSolve.solve-Tuple%7BSciMLBase.AbstractDEProblem,%20Vararg%7BAny%7D%7D

PR #1829, opened 2024-09-19
Save all results via `SavingCallback`

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1825.


SouthEndMusic:
Left to do: currently both `check_water_balance_error` and `basin_table` compute water balance errors over the `saveat` intervals. We should save the computations done in `check_water_balance_error` in `SavedFlow`.

visr (approved):

PR #1842, opened 2024-09-23
Fix internal link to component overview

visr:
Saw this in a warning of `pixi run quarto-render`.


visr:
I wish Quarto render would error by default, also for external links, but last I checked I didn't see a way to enable that.

But yes, we should look into external tooling for this, feel free to make an issue.

PR #1843, opened 2024-09-23
Add default UserDemand boundary concentrations.

### LINKED ISSUE #1841: Mass balance is off in HWS Delwaq model
With the new formulation, the continuity tracer never exceeds 1 significantly (previously 1.02-1.03, so 3% error). However, I see some very large drops (to 0.52, so double the water in than out) over time.

My hunch is that this is due some bug in the generation of the simplified network, and on first glance it seems this happens around nodes with UserDemands.

If you want to check this yourself, node #3543 Amer and #10634 Twentekanaal are good candidates.

evetion:
Fixes #1841

The fix is indeed to UserDemand, and @visr's hunch about the concentration not being set on the boundary turned out to be right.

Seems we're in business now 🎉:
Screenshot 2024-09-23 at 17 38 26

### DIFF
--- a/python/ribasim/ribasim/delwaq/template/B5_bounddata.inc.j2
+++ b/python/ribasim/ribasim/delwaq/template/B5_bounddata.inc.j2
@@ -18,6 +18,10 @@ ITEM 'Precipitation'
 CONCENTRATIONS 'Continuity' 'Precipitation'
 DATA 1 1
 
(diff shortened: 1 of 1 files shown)


visr (approved):
Awesome. I guess those colors are so close to 1 they just decided to call it 1 on both the low and high end!

PR #1852, opened 2024-09-30
Remove support for resetting cumulatives from BMI

visr:
The coupler is switching to an approach where they don't need to reset the cumulatives to 0. Since this complicated the Ribasim code, let's get rid of it.

The reason for the coupler to move away from resetting cumulatives is that this possibly broke with #1819.

The dedicated BMI arrays like `vertical_flux_bmi` and `user_demand.realized_bmi` are removed. The `BMI.get_value_ptr` now just points to the cumulative state, or if it is not a state (drainage, precipitation) to `basin.cumulative_drainage` or `basin.cumulative_precipitation`.

This is breaking because I also renamed these BMI variables to be more consistent with the current code, where we use the term cumulative internally.

```
user_demand.realized -> user_demand.inflow
basin.drainage_integrated -> basin.cumulative_drainage
basin.infiltration_integrated ->  basin.cumulative_infiltration
```

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -31,24 +31,25 @@ function BMI.update_until(model::Model, time::Float64)::Model
 end
 
 function BMI.get_value_ptr(model::Model, name::AbstractString)::AbstractVector{Float64}
+    (; u, p) = model.integrator
     if name == "basin.storage"
-        model.integrator.p.basin.current_storage[parent(model.integrator.u)]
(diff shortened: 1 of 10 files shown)


evetion (approved):
Nit: Realized seemed better than inflow. And it seems integrated vs cumulative is math vs non-math coders (I used similar wordings in my concentration PR), but we should come to an agreement for future internal documentation.

Issue #1851, opened 2024-09-30
[bug] Terminals sometimes "leak" water


evetion:
I hit this during the implementation of #1849, where edges to Terminals have negative flows when checked in `update_cumulative_flows!` using the flow from `flow_update_on_edge`. For example in the HWS model I get the following:

```
┌ Warning: Terminal #5415 has flow of -338.54177272319794 at 6.862248444895375e6 for a timestep of 91.35735275302858
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -85.64921969175339 at 6.862339802248128e6 for a timestep of 91.35735275302858
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -8.970432549715042 at 6.862431159600881e6 for a timestep of 91.35735275302858
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -0.8940373957157135 at 6.862489481604243e6 for a timestep of 58.32200336178765
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -0.9066413938999176 at 6.862868324295173e6 for a timestep of 378.8426909304897
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -0.1415431797504425 at 6.863247166986104e6 for a timestep of 378.8426909304897
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -0.06280365586280823 at 6.8643238840039205e6 for a timestep of 1076.7170178166448
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -0.009804785251617432 at 6.865400601021737e6 for a timestep of 1076.7170178166448
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
┌ Warning: Terminal #5415 has flow of -3.9677118062973022 at 6.882244989985872e6 for a timestep of 0.059130550789141804
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/callback.jl:205
[...]

PR #1856, opened 2024-10-01
Check leaky Terminals

visr:
https://github.com/Deltares/Ribasim/issues/1851#issuecomment-2384910255


evetion:
Integration tests shows the same warnings as I've found locally:

https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_IntegrationTest_Linux/4605553?mode=branches&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&buildTab=log&focusLine=737&logView=linear&linesState=123

So this is either an edge case (or misunderstanding) of `flow_update_on_edge` or some previously unknown solver behavior. Note that the results, water balance and all seem fine.

visr:
With #1874 this is now never larger than 1e-8, and short. Closing this, it was never meant to be merged.

Issue #1853, opened 2024-09-30
Water balance error when saving each timestep


visr:
`check_water_balance_error` runs on `saveat`, by default once per day. We support writing output for each timestep with

```toml
[solver]
saveat = 0
```

For e.g. the HWS model this fails the water balance check at t=0, whereas it works even with a tiny `saveat = 1` second.

```
┌ Error: Too large water balance error
│   id = Basin #1340
│   balance_error = 84387.34177215189
│   relative_error = 2.0
└ @ Ribasim d:\repo\ribasim\Ribasim\core\src\callback.jl:317
```

Perhaps the check should be disabled for `saveat = 0`? One workaround now is to set large water balance tolerances:

```toml
[solver]
saveat = 0
water_balance_abstol = 1e30 # optional, default 1e-3
water_balance_reltol = 1e20 # optional, default 1e-2
```

Also I see that currently we fail the check if **both** the absolute and relative tolerance is exceeded. Shouldn't this be **either**?

Huite:
> Shouldn't this be either?

Indeed, should be either.

visr:
This may also be related to the water balance error the iMOD Coupler team sees when running the backwater model with groundwater exchange through BMI. #1896 may already help with this.

SouthEndMusic:
I cannot reproduce this error on the current main with the latest version of the RWS model (or De Dommel)

PR #1858, opened 2024-10-01
Fail water balance check on either tolerance

visr:
Part of #1853; https://github.com/Deltares/Ribasim/issues/1853#issuecomment-2383488129

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -310,7 +310,7 @@ function check_water_balance_error(
         mean_flow_rate = (total_in + total_out) / 2
         relative_error = iszero(mean_flow_rate) ? 0.0 : balance_error / mean_flow_rate
 
-        if abs(balance_error) > water_balance_abstol &&
+        if abs(balance_error) > water_balance_abstol ||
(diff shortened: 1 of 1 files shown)


evetion:
Good luck with finding new tolerances, nice PR just before 1.0! ❤️

visr:
Doesn't need to be merged before 1.0. Could have if it didn't give issues.
This is on `pump_discrete_control`, relative error is too high, but absolute error is tiny:

```
┌ Error: Too large water balance error
│   id = Basin #1
│   balance_error = -4.1119371282413245e-20
│   relative_error = -0.10526315789473695
└ @ Ribasim ~/work/Ribasim/Ribasim/core/src/callback.jl:316
```

Huite:
Uh, oops, I might've shooting from the hip there. I was thinking of acceptance rather than rejection!

What I discussed with @SouthEndMusic earlier: if the absolute error is tiny, you probably want to accept, right?
Hence the earlier implementation. Since `&&` short-circuits, it won't look at the relative error if the absolute error is already okay.

PR #1859, opened 2024-10-01
Simplify setting juliaup override

visr:
Possible since https://github.com/JuliaLang/juliaup/pull/1002


visr:
Hmm it seems that 1.17.5 and 6 are pre-releases that are only available in conda-forge. Closing for now.

Issue #1863, opened 2024-10-03
Error in exchange waterbalance coupler Ribasim-MF-MS  


SnippenE:
After the refactor in waterbalance error #1808 test in coupler are giving unexpected results. Could be related to the changes in how fluxes are integrated on Ribasim side.

Possibly the same as #1851.

visr:
@rleander made a reproducer in Python that I translated into Julia so I can directly test changes in the core. Here it is: https://gist.github.com/visr/78a864974619f9d445a3a0e1226310b0. It outputs figures like the following:

![mwe-1-main](https://github.com/user-attachments/assets/8cb42ec5-2511-4163-8c33-4fe95b3f4599)

These lines should both stay at 0 and 25 where they start. The error gets worse over time. Resetting the cumulatives like in https://github.com/Deltares/Ribasim/pull/1864 seems to help. But then how often should we do this and who should do this in BMI context? It is probably not a real solution, just like how forcing smaller timesteps or lower tolerances also helps reduce the problem.

Changing the solver algorithm to Rodas5P does seem to help, showing no divergence even if I run the bucket model for 30.000 days, hence #1869.
![mwe-4](https://github.com/user-attachments/assets/56862d67-021d-4708-9ae3-6a2eda14bd4c)

It seems like the cumulative flow states are difficult for some algorithms. Though I also wouldn't rule out that we can do more in the core to alleviate this.

visr:
This is with QNDF, only changing the reltol from 1e-5 to 1e-7. PR to make that the default incoming.
![image](https://github.com/user-attachments/assets/06c7ddac-6a5a-4e17-b572-721327324819)

Huite:
Is this without or without resetting to 0?

It kinda seems like it's showing growing oscillations:

![image](https://github.com/user-attachments/assets/f7f9400d-96d5-4ecf-bcb5-200cc2abe722)

SouthEndMusic:
With https://github.com/Deltares/Ribasim/pull/1912 divergence is very slow (I subtracted 25 from the bottom plot):

![Image](https://github.com/user-attachments/assets/597bbb5f-7a02-462b-a855-02069688fd38)

PR #1864, opened 2024-10-03
Offer BMI option to reset cumulatives

visr:
Looking into #1863.
Hopefully we don't need to merge this.


visr:
We don't need this yet.

PR #1871, opened 2024-10-04
Try pixi v0.31.0 with disabled cache

visr:
https://github.com/prefix-dev/pixi/issues/2196#issuecomment-2393044611


visr:
This didn't help. The issue is fixed upstream, we can upgrade to the next release after v0.31.0.

PR #1874, opened 2024-10-07
Decrease relative tolerance from 1e-5 to 1e-7

### LINKED ISSUE #1851: [bug] Terminals sometimes "leak" water

### LINKED ISSUE #1863: Error in exchange waterbalance coupler Ribasim-MF-MS

visr:
With #1819 the state goes further away from 0 with time. That means we rely more on the relative tolerance than the absolute tolerance, see the [SciML docs](https://docs.sciml.ai/DiffEqDocs/stable/basics/faq/#What-does-tolerance-mean-and-how-much-error-should-I-expect) on the topic. Hence I found that reducing the `abstol` had no effect, but reducing the default `reltol` by two orders of magnitude is enough to get only negligible leaky Terminals with occasional very brief leaks of 1e-8 m3/s. It is also enough to minimize the infiltration error to acceptable levels as can be seen in https://github.com/Deltares/Ribasim/issues/1863#issuecomment-2397302808.

~~For HWS this slows down simulation from 9.1s to 12.3s, which isn't too bad for a two orders of magnitude reduction in relative tolerance.~~ Rechecking HWS performance gives a speedup! 12.4s for 1e-8, 8.1s for 1e-7, 10.2s for 1e-6. So 1e-7 seems optimal. Not sure what went wrong last time, but this seems pretty consistent. This is with both absolute and relative set to the same number.

Fixes #1851
Fixes #1863


visr:
Perhaps for simplicity/consistency we should set the abstol to 1e-7 as well.

Huite:
> Hence I found that reducing the abstol had no effect, but reducing the default reltol by two orders of magnitude is enough to get only negligible leaky Terminals with occasional very brief leaks of 1e-8 m3/s.

This seems very obvious in retrospect, indeed.
It's a bit unintuitive coming from a regular head/storage formulation, since those are clearly more bounded.
Resetting to 0 (maybe not every `saveat`, but maybe every change of boundary condition), since the solver is getting a shock in those cases anyway.

visr:
It's hard to find a good time for resetting though that doesn't cause issues for BMI, which currently assumes there is never any resetting. Also saveat can be infinite. Not resetting also makes it easy to get the total water balance over time.

visr:
In what way meaningless? Do you mean floating point accuracy issues? Perhaps a yearly reset would be predictable enough for BMI to work around, although the BMI driver would have to be calendar aware.

Huite:
Let's say you have a flow boundary constantly pumping in water and the flow on the edge accumulates monotonically, with 1000 m3/d; let's further assume that flow is (nigh) steady-state.

At t=1 day, flow = 1000; relative tolerance value is `reltol * 1000`;
At t=1000 days, flow = 100 000 days; relative tolerance value is `reltol * 1_000_000`.

I.e. the tolerance has effectively become a 1000 times relaxed, right? While the physical meaning of the model remains the same.

PR #1875, opened 2024-10-07
Allow Polygon geometry for Basin / Area.

### LINKED ISSUE #1850: Accept Polygons for Basin / Area in add-api
**What**
`Basin / Area` will come from datasets as Polygons ánd MultiPolygons. Current Ribasim.main only accepts Polygons, which is quite uncommon/user-unfriendly

**Why**
To add a polygon, I would need to do something like this:

```
from shapely.geometry import MultiPolygon, Polygon

if isinstance(basin_polygon, Polygon):
   basin.Area(geometry=[MultiPolygon([basin_polygon])])
else: # we assume our set only contains Polygons or MultiPolygons, the only valid combination in polygon feature-files
   basin.Area(geometry=[basin_polygon])
```

**How**
I want to add a `Polygon` / `MultiPolygon` in the exact-same way:  `basin.Area(geometry=[basin_polygon])` and expect the Ribasim Python-API will do the conversion

evetion:
Fixes #1850

### DIFF
--- a/python/ribasim/ribasim/geometry/area.py
+++ b/python/ribasim/ribasim/geometry/area.py
@@ -2,7 +2,7 @@ import pandera as pa
 from pandera.dtypes import Int32
 from pandera.typing import Index, Series
 from pandera.typing.geopandas import GeoSeries
-from shapely.geometry import MultiPolygon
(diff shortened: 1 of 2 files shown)


visr (approved):
Thanks!

PR #1876, opened 2024-10-08
Go back to latest Pixi release

visr:
v0.32.0 fixes the issue we ran into with v0.31.0.
https://github.com/prefix-dev/pixi/releases/tag/v0.32.0


evetion:
Better to pin 0.32 then?

visr:
We pin everything except pixi, so we can keep in touch with @Hofer-Julian ;)

evetion (approved):

PR #569, opened 2023-08-31
allow overwriting open GeoPackage

visr:
This works fine having the file open in QGIS on Windows. We write to another file, and at the end move this file over the destination file.

Fixes #563

PR #1881, opened 2024-10-08
Allow overwriting a model that is open in QGIS

visr:
Redoing #569 which got lost in the add API changes.
Goes well with https://github.com/evetion/Reloader/


evetion (approved):
Nasty hack, I like it.

PR #1882, opened 2024-10-08
Remove database from struct Toml

visr:
It is not configurable, so it should not be in the struct.


evetion:
We should have the v1 release party, and during it, go over all things that could be nice to change and release a v2 the following week.

evetion (approved):

PR #1884, opened 2024-10-08
SparseConnectivityTracer v0.6.5 ⇒ v0.6.6

visr:
See if this hits https://github.com/Deltares/Ribasim/issues/1883. https://github.com/adrhill/SparseConnectivityTracer.jl/releases/tag/v0.6.6

PR #1885, opened 2024-10-08
SparseConnectivityTracer v0.6.5 ⇒ v0.6.4

visr:
Alternative to #1884, that brings us back to before #1860.
https://github.com/adrhill/SparseConnectivityTracer.jl/releases/tag/v0.6.5 doesn't seem to contain anything relevant for us though, so I'd be surprised if this would help.


visr:
It was Julia 1.11

Issue #1883, opened 2024-10-08
SparseConnectivityTracer: non-boolean used in boolean context


visr:
I just saw this randomly pop up in an unrelated PR: https://github.com/Deltares/Ribasim/actions/runs/11237297981/job/31239411451?pr=1881#step:5:629

```
ERROR: LoadError: TypeError: non-boolean (SparseConnectivityTracer.GradientTracer{SparseConnectivityTracer.IndexSetGradientPattern{Int64, BitSet}}) used in boolean context
Stacktrace:
  [1] clamp(x::SparseConnectivityTracer.GradientTracer{SparseConnectivityTracer.IndexSetGradientPattern{Int64, BitSet}}, lo::Float64, hi::Float64)
    @ Base ./intfuncs.jl:1266
  [2] formulate_flow!
    @ Ribasim ~/work/Ribasim/Ribasim/core/src/solve.jl:668
```

Strange because main is green. Is this non-deterministic? Doesn't seem like it, #1882 now also hits it.

visr:
@evetion noticed the Julia tests were running the just released julia 1.11.0. I made #1888 to address that. Testing locally with 1.11 and an updated Manifest.toml I can reproduce the issue running e.g. the HWS model. Might need fixing in https://github.com/adrhill/SparseConnectivityTracer.jl since it is surprising that `clamp` doesn't work.

PR #1888, opened 2024-10-08
Don't rely on juliaup override for version selection

visr:
We have

```
install-julia = "juliaup add 1.10.5 && juliaup override unset && juliaup override set 1.10.5"
```

But in #1883 we can see (@evetion can see) that the Julia tests are still run using Julia 1.11, which is the new stable release that came out today.

Rather than trying to find out what exactly goes wrong with the override, this just sets it when we start julia, so we can release.

Works around #1883.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -30,9 +30,9 @@ install = { depends_on = [
     "install-pre-commit",
 ] }
 # Julia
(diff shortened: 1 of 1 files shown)


visr:
https://github.com/Deltares/Ribasim/issues/1883 and https://github.com/Deltares/Ribasim/pull/1890 are better.

PR #1889, opened 2024-10-08
Add missing "initialize-julia" in pixi tasks

visr:
This allowed it to pick up Julia 1.11, leading to #1883.


visr:
I'll close this in favor of #1890. Since `Pkg,test` takes care of environment initialization anyway, and uses different precompile files, this does a bunch of extra work that we don't need. For the `quarto-render` job it could be useful, but that is mainly used in CI and there it is preceded by `install-ci` already.

PR #1891, opened 2024-10-08
Use `hatch build` rather than `python -m build`

visr:
I got an error that the build module was not found, when doing `pixi run publish-ribasim-python`. I tried `pixi add build` but it didn't work and I saw the last release was 3 years ago. Since we already use hatchling it makes sense to use https://hatch.pypa.io/1.12/build/ as well. I already tried it to publish v2024.11.0 and it seems to work fine.


evetion (approved):
I have a strong déjà vu that I had to do something similar recently. Not sure why/where.

Huite:
The build feedstock has been archived: see https://github.com/conda-forge/build-feedstock

It has been replaced by `python-build` and that's the one that will work as a pixi dependency: https://github.com/conda-forge/python-build-feedstock

That one is up to date, latest change 2 days ago.

Issue #1837, opened 2024-09-23
Avoid re-computing the water balance error


visr:
> Left to do: currently both `check_water_balance_error` and `basin_table` compute water balance errors over the `saveat` intervals. We should save the computations done in `check_water_balance_error` in `SavedFlow`.

_Originally posted by @SouthEndMusic in https://github.com/Deltares/Ribasim/issues/1829#issuecomment-2361205527_

PR #1896, opened 2024-10-10
Save all results via saving callback (follow up)

SouthEndMusic:
This is a follow up of https://github.com/Deltares/Ribasim/issues/1825 (and a few unrelated experiments).

Fixes https://github.com/Deltares/Ribasim/issues/1837.

Other changes in this PR:
- Storages from the cumulative flows are computed as a sparse matrix-vector product
- Storage rates are computed without involving the initial storage
- Manning resistance is more stable around $\Delta h = 0$.


core/test/utils_test.jl
visr, line 1: Could you add some tests on the value of `flow_to_storage`?

visr (approved):
This looks good to me. If you could add some unit tests for `flow_to_storage`, this can go in.

evetion:
> * Manning resistance is more stable around Δ h = 0 .

IMHO, that is an (important! 🎉) change that should not be committed by a squashed commit called `save all results via saving callback`.

PR #1903, opened 2024-10-15
Fix duplicate boundaries in Delwaq generation.

### LINKED ISSUE #1901: Delwaq Rhine tracer concentrations in Maas/Waal canal
Where there should be none, as the canal purely has an (small) outflow into the Waal.

Image

Will look into solver options of Delwaq, and how we can prevent this "diffusion".

### LINKED ISSUE #1902: Delwaq UserDemand edges/flows give warnings
It seems there's a bug in generating them. UserDemands are special, as they are the only boundaries connected with two edges.

Warnings in Delwaq
```
WARNING, there is no flow associated with boundary nr:      40
WARNING, there is no active segment associated with boundary nr:      40
WARNING, there is no flow associated with boundary nr:      42
WARNING, there is no active segment associated with boundary nr:      42
WARNING, there is no flow associated with boundary nr:      44
WARNING, there is no active segment associated with boundary nr:      44
 
Vervolgens krijg ik in Delwaq een foutmelding doordat er meer boundaries (-5628) worden doorgegeven dan in de pointer (-5626) staan. Daar had DELWAQ meteen aan de rem moeten trekken...
 
Tot slot, sommige boundary names zijn niet uniek (ribasim_bndlist.inc):
 
UserDeman_362       Boundary name      24                   UserDemand          
UserDeman_362       Boundary name      25                   UserDemand          
WARNING: boundary ID is not unique:UserDeman_362       
...
UserDeman_372       Boundary name      35                   UserDemand          
UserDeman_372       Boundary name      36                   UserDemand          
WARNING: boundary ID is not unique:UserDeman_372       
...        
UserDeman_376       Boundary name      40                   UserDemand          
UserDeman_376       Boundary name      41                   UserDemand          
WARNING: boundary ID is not unique:UserDeman_376       
```

evetion:
Fixes #1901 
Fixes #1902

This disables any dispersion in the Delwaq model, fixing some tracer intrusions. 

#1902 was actually two issues:
- In the graph simplification, we didn't anticipate boundary to boundary connections (UserDemand to Terminal, in use at IJmuiden). We now just remove the Terminal node/edge from it.
- UserDemands are the only boundaries that occur twice (when not connecting back to the same Basin, those are merged to one edge), and that results in a duplicate boundary name, as it was based on type + node_id.

### DIFF
--- a/python/ribasim/ribasim/delwaq/generate.py
+++ b/python/ribasim/ribasim/delwaq/generate.py
@@ -163,6 +163,20 @@ def _setup_graph(nodes, edge, use_evaporation=True):
                 merge_edges.extend(edge_ids)
                 G.remove_edge(*loop)
 
+    # Remove boundary to boundary edges
+    remove_double_edges = []
+    for x in G.edges(data=True):
+        a, b, d = x
+        if G.nodes[a]["type"] == "Terminal" and G.nodes[b]["type"] == "UserDemand":
+            print("Removing edge between Terminal and UserDemand")
+            remove_double_edges.append(a)
+        elif G.nodes[a]["type"] == "UserDemand" and G.nodes[b]["type"] == "Terminal":
+            remove_double_edges.append(b)
+            print("Removing edge between UserDemand and Terminal")
+
+    for node_id in remove_double_edges:
+        G.remove_node(node_id)
+
     # Relabel the nodes as consecutive integers for Delwaq
     # Note that the node["id"] is the original node_id
     basin_id = 0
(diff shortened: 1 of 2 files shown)


evetion:
I will make the prints debug, and the warnings errors, that should trigger one such a thing happens again in the tests.

evetion:
> Should we for instance read the log you linked in #1902 in `test_delwaq.py` and ensure that WARNING does not occur or something like that?

There are several warnings (that we can safely ignore) in the output file, so that's hard to test. I'd say if it runs, it runs 🤷🏻 (and we can parse the output). Let's revisit this when we bundle/wrap Delwaq.

visr (approved):

Issue #1900, opened 2024-10-14
Avoiding negative flows where they should not be


SouthEndMusic:
After the refactor https://github.com/Deltares/Ribasim/pull/1819 we found that flows that should be non-negative are sometimes negative for a brief period of time. I discussed this on Slack:

![Image](https://github.com/user-attachments/assets/6bf1e399-9cc7-4633-a7b2-0e869217c989)

Several solutions were proposed above depending on where the problem originates:

- Add a `stage_limiter` to the nonlinear solver (@visr had some comments about this, it would be nice if you add them below)
- Switch to QR factorization in the linear solver

If the above options don't work we could also implement a callback for enforcing non-negative flows. Some notes about this:

- This shouldn't affect the water balance error, as it just nudges the physics in the right direction
- This involves changing the state in the callback, which might lower the performance of multi-step methods. Also, our experiments (with @visr) with using `u_modified!` weren't very promising:

![Image](https://github.com/user-attachments/assets/e231dd57-574d-4d29-a4c2-ce5561eb7696)

Issue #1897, opened 2024-10-10
Taming cumulative flows


visr:
Since #1819 our states are cumulative flows, getting further from 0 all the time. This formulation helped the water balance error and mean flows a lot, but came at the cost of errors occuring after months of simulations. Decreasing the `reltol` in #1874 helped reduce the issues that came to light, though as mentioned in https://github.com/Deltares/Ribasim/pull/1874#issuecomment-2399205377 it still suffers from reduced effective tolerance over time.

One option to address this is by resetting cumulative flows. A draft function to reset is in #1864, though the difficult part is coming up with a good approach for when to call it, such that it works independently of `saveat` settings and BMI usage, see https://github.com/Deltares/Ribasim/pull/1874#issuecomment-2397423444.

We can also look into using custom error norm functions that are more robust to these cumulative flows, somehow keeping the effective errors constant over time.

#1874 was only tested manually, so as a part of this issue we should include tests to show the reduction in error over time. The manual checks consisted of running HWS with https://github.com/Deltares/Ribasim/pull/1856, and the gist linked in https://github.com/Deltares/Ribasim/issues/1863#issuecomment-2392310526.

Huite:
Another option to consider is to solve for flows AND storage at the same time, and add them both to the unknowns (maybe in ComponentArray form?). My feeling is that the overhead is negligible and it would likely simplify the implementation -- you can assume there's an up-to-date storage available at any time.

It would also bring us more in line with something like a 1D St. Venant simulator, since as far as I can tell, these solve for level (∝ storage) and flow velocity (∝ flow) at the same time too.

This in turn might allow a natural extension for certain basins / connectors (or parts of the network) that do incorporate a momentum balance.

SouthEndMusic:
I did an experiment with both storage and flow states before, then it was quite detrimental to performance. That was quite a while ago though

SouthEndMusic:
Regarding the custom norm: we should look into how the error is estimated again:

![Screenshot_20241014_093718_Chrome.jpg](https://github.com/user-attachments/assets/1907e289-8ce4-4efa-bee4-0d5988f3d460)

It's never clear to me from the docs what the order of operations is, I.e. what is per element of the state and when norms are applied. This formula at least indicates a decreasing error with an increasing state value because of the reltol term, maybe for us it makes more sense if that max is a difference.

evetion:
[...]

PR #1904, opened 2024-10-15
Add flow limiter (`step_limiter!`)

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1900.

#### Main branch

plot

#### This branch

plot

#### Conclusions

Having zero forcing when zero forcing is expected is now enforced properly. In the same way, I'm pretty sure forcing will now never exceed the given flux, and flow should be non-negative where expected.

Having total forcing when the basin is not in the low storage factor regime is harder to enforce, because there is no exact way to infer whether the basin storage passed through the reduction factor regime in the last timestep. A conservative heuristic might suffice here though. 

#### Main branch
- HWS: 9.837521 seconds
- De Dommel: 51.496409 seconds

#### this branch
- HWS: 9.554516 seconds
- De Dommel: 64.802680 seconds


core/src/config.jl
```diff
@@ -249,6 +249,18 @@ const algorithms = Dict{String, Type}(
     "Euler" => Euler,
 )
 
+"""
+Check whether the given function has a method that accepts the given kwarg.
+Note that it is possible that methods exist that accept :a and :b individually,
+but not both.
+"""
+function function_accepts_kwarg(f, kwarg)::Bool
```
evetion, line 257: Very meta, but nice improvement. Is `kwarg_decl` documented (i.e. public?)

SouthEndMusic: https://julialang.slack.com/archives/C7T968HRU/p1728982500911379?thread_ts=1728982500.911379&cid=C7T968HRU

SouthEndMusic: I don't think `Base.kwarg_decl` is documented, I can't find it and it doesn't have a docstring.

SouthEndMusic:
It is indeed a fix for #1900, which is one way in which the flows have to be 'tamed' as discussed in #1897. Even though it is not a global fix, I think this PR is a helpful one, as it enforces properties of the solution which we want to be strictly true for any set of solver settings, e.g.
- Non-negative flows where expected
- Realized user demand should not exceed demand

With the ever increasing norm of the state there is still the possibility that the solution slowly drifts away from the physics in ways we cannot capture in the `step_limiter`. If we want to solve that and keep using the fast solver `QNDF`, we either have to find a way to tame the state size (for instance getting resetting to `0` to work), or document that longer simulations require smaller relative errors.

core/src/config.jl
evetion: Well that's bad, but then again, it seems this method has been there for at least 5 years, so let's use it. Nice find.

evetion (approved):

SouthEndMusic:
I've added the `Average flow states` commit for the record since I was working on it on this branch, but I'm not going to use it. I feel like it's not worth the added complexity, both conceptually and in the code. My goal now is to fix urgent things with the step limiter, and either stop using the relative error (i.e. hardcoding it to be `0`) or documenting that it should be very small if users note growing inaccuracies for long simulations.

SouthEndMusic:
Closed in favor of https://github.com/Deltares/Ribasim/pull/1911.

PR #1908, opened 2024-10-17
Minor edits of comments

visr:


evetion (approved):
You should write Martijn's style guide on how fill! is bad.

visr:
Nothing bad about fill, we just use the other way everywhere else in the code.

Issue #1838, opened 2024-09-23
Document known issue: (small) negative flows from e.g. Terminals


visr:
- Add explanation of callback check
- In `usage.qmd`: add explanation of water balance tolerance options (too large error probably indicates error in the code or floating point errors)

_Originally posted by @SouthEndMusic in https://github.com/Deltares/Ribasim/issues/1828#issuecomment-2361129601_

evetion:
Add limitations of current version.

evetion:
#1899 Did not address that users can see (small) negative flows from Terminals.

visr:
Ah right, reopening for that. Unless this becomes fixed altogether with #1897.

PR #1912, opened 2024-10-21
Finetune `step_limiter!`

SouthEndMusic:
Follow-up of https://github.com/Deltares/Ribasim/pull/1911.

Fixes https://github.com/Deltares/Ribasim/issues/1897 (?).

Fixes https://github.com/Deltares/Ribasim/issues/1838 (how could we test that negative flows where they should not be no longer occur? Maybe add a callback that explicitly checks for non-decreasing states)

I focus in this PR on `UserDemand` inflow and infiltration because of their importance in coupling.


SouthEndMusic:
@evetion you might appreciate this first commit; it gets rid of many of the magic hardcoded numbers for threshold values.

SouthEndMusic:
@rleander this is an interesting one for you, this one together with https://github.com/Deltares/Ribasim/pull/1911 should fix the problems you noticed, in particular: infiltration and realized user demand should no longer drift away from their demands as the simulation goes on.

core/src/util.jl
```diff
@@ -1143,3 +1130,42 @@ function get_demand(user_demand, id, priority_idx, t)::Float64
         demand[id.idx, priority_idx]
     end
 end
+
+"""
+Estimate the minimum reduction factor achieved over the last time step by
```
evetion, line 1135: Might be good to mention why we want an underestimation (we don't want to be too strict in clamping the flow right?).

core/src/model.jl
```diff
@@ -118,6 +118,8 @@ function Model(config::Config)::Model
 
     # Synchronize level with storage
     set_current_basin_properties!(du0, u0, parameters, t0)
+    parameters.basin.level_prev .=
```
evetion, line 121: Would be good to inline comment (or in the PR description) as to why we now need to keep track of `level_prev`.

core/src/parameter.jl
```diff
@@ -1,3 +1,6 @@
+const LOW_STORAGE_THRESHOLD = 10.0
```
evetion, line 1: Please comment on why/where this is needed. Like `# below this storage threshold the reduction factor is triggered`.

core/src/parameter.jl
```diff
@@ -1,3 +1,6 @@
...
+const USER_DEMAND_MIN_LEVEL_THRESHOLD = 0.1
```
evetion, line 2: Same here.

core/src/parameter.jl
```diff
@@ -300,6 +303,21 @@ In-memory storage of saved instantaneous storages and levels for writing to resu
     t::Float64
 end
 
+"""
+Caches of current basin properties
+"""
+struct CurrentBasinProperties
```
evetion, line 309: Nice refactor :)

core/src/solve.jl
```diff
@@ -79,17 +84,19 @@ function set_current_basin_properties!(
...
         current_cumulative_precipitation,
         current_cumulative_drainage,
-        cumulative_precipitation,
-        cumulative_drainage,
-        vertical_flux,
-    ) = basin
+    ) = current_properties
+
     current_storage = current_storage[parent(du)]
+    current_low_storage_factor = current_low_storage_factor[parent(du)]
```
evetion, line 99: This parameter should have a comment (in the PR description, or at the paramter.jl level) as it's new functionality.

core/src/solve.jl
```diff
@@ -721,6 +759,11 @@ function limit_flow!(
         allocation,
     ) = p
 
+    du = get_du(integrator)
+    set_current_basin_properties!(du, u, p, t)
+    current_storage = basin.current_properties.current_storage[parent(u)]
+    current_level = basin.current_properties.current_level[parent(u)]
+
```
evetion, line 766: Why is this now needed?

core/test/utils_test.jl
```diff
@@ -261,28 +260,6 @@ end
     @test reduction_factor(-Inf, 2.0) === 0.0
 end
 
-@testitem "low_storage_factor" begin
```
evetion, line 264: Removing a test and not adding anything back seems bad. Is the BMI coupler test not a good one to add, and make sure we that at least a part of the `step_limiter` functionality works?

core/src/solve.jl
```diff
@@ -763,28 +806,58 @@ function limit_flow!(
...
+            max_flow_rate,
             active,
             dt,
         )
     end
 
-    # Evaporation is in [0, ∞) (a stricter upper bound would require also estimating the area)
-    # Infiltration is in [0, infiltration]
+    # Evaporation is in [0, ∞) (stricter bounds would require also estimating the area)
+    # Infiltration is in [f * infiltration, infiltration] where f is a rough estimate of the smallest low storage factor
```
evetion, line 847: There needs to be minimum infiltration? I thought the low storage could influence the maximum outflow, not the other way around?

core/src/solve.jl
```diff
@@ -763,28 +806,58 @@ function limit_flow!(
...
+                current_level,
+                basin.level_prev,
+                user_demand.min_level,
+                id,
+                inflow_id,
+            )
+            allocated_total =
+                is_active(allocation) ? sum(user_demand.allocated[id.idx, :]) :
+                sum(user_demand.demand[id.idx, :])
+            factor_basin_min * factor_level_min * allocated_total, allocated_total
```
evetion, line 833: Missing docstring for the bounds here (why is the minimum raised?)

core/src/util.jl
```diff
@@ -1143,3 +1130,42 @@ function get_demand(user_demand, id, priority_idx, t)::Float64
...
+            min(storage_now[id.idx], storage_prev[id.idx]) - 2LOW_STORAGE_THRESHOLD,
+            LOW_STORAGE_THRESHOLD,
+        )
+    else
+        one(T)
+    end
+end
+
+"""
+Estimate the minimum minimum level reduction factor achieved over the last time step by
```
evetion, line 1151: `minimum minimum`? If not a typo, please rewrite for clarity?

evetion (commented):
Some general comments:

For the next time, please refactor in separate PR's, mixing it makes the PR and the git history in the future hard to read. I do like the refactor, tackling the magic numbers and the basin complexity 👍🏻.

Could you provide a comment in the PR description on what you introduce and why? I'll give more detailed comments in the code below.

After reading it completely, my main confusion is that we clamp the minimum flow out of a basin and not the maximum flow? I thought the reduction factors trigger so a (maximum) flow cannot empty a basin.

core/src/solve.jl
SouthEndMusic: This limiter is not used to enforce the effect of the low storage factor. It does the opposite: it ensures that full infiltration is realised if enough storage is available

evetion (approved):

PR #1915, opened 2024-10-21
Added DVC, teamcity-messages and Julia TC integration

### LINKED ISSUE #1879: Save benchmark results over time in readable/plottable format
Minimum is to fix the current benchmark on teamcity

evetion:
Fixes #1879 

This does three things
- Adds teamcity-messages as a Python test dependency, enabling TC to keep track of pytest tests (number of tests, number of failures, linking to specific logs), making debugging much easier.
Screenshot 2024-10-21 at 22 35 54

- Adds teamcity specific logging in Julia (much like teamcity-messages) to keep track of the integration test time and differences with a previous model. These parameters are tracked by TC (and can be failed on), and can be plotted over time.

Screenshot 2024-10-21 at 22 38 31

- Finally, and biggest, this adds [DVC](https://dvc.org/) to our project, with the minio repo tracked, and the integration test as stage. It tracks a new default toml for that model (models/integration.toml), and like TC in 2, the time and differences. Once someone changes either the hws model, or the core code, or the integration.toml, the results are outdated and `dvc repro` will rerun the test. We can save these experiments over time, and compare results with the VS Code plugin:
Screenshot 2024-10-21 at 22 09 27

This disables the threshold test in the integration test, and opts for monitoring results (could be done automatically) instead. I've also opted for keeping track of results and settings threshold _outside_ of Julia.

This is a backwards compatible setup PR. Changes/additions for benchmark tuning can be made in another PR.


visr:
Wonderful!

evetion:
> Wonderful!

This is now good to go.

.dvcignore
```diff
@@ -0,0 +1,5 @@
+# Add patterns of files dvc should ignore, which could improve
+# the performance. Learn more at
+# https://dvc.org/doc/user-guide/dvcignore
+utils/juliaup
```
visr, line 4:
```suggestion
```

This hasn't been used for a while. If you still have stuff there, you can delete it.

core/test/utils.jl
```diff
@@ -0,0 +1,23 @@
...
+            :(println(
+                "##teamcity[buildStatisticValue key='",
+                $key,
+                "' value='",
+                $value,
+                "']",
+            )),
+        )
+    else
+        return esc(:(println($key, "=", $value)))
```
visr, line 21:
```suggestion
        return esc(:(println($key, '=', $value)))
```

pixi.toml
```diff
@@ -192,6 +192,7 @@ tomli-w = "*"
 xarray = "*"
 xmipy = "*"
 xugrid = "*"
+teamcity-messages = ">=1.32,<2"
```
visr, line 195:
```suggestion
teamcity-messages = "*"
```

Otherwise we currently have no way to get breaking updates. And with the readable diff like https://github.com/Deltares/Ribasim/pull/1873 we can easily see and double check major version updates.

visr (approved):
Looks good to me! Left some small suggestions.

pixi.toml
evetion: I didn't know about that fancy diff, very cool!

evetion:
Note that I have reset the benchmarks to the 1.0 release, before #1904, so any changes are currently from that PR and any followups that touch the core.

PR #1917, opened 2024-10-23
[WQ] Drop duplicate boundaries instead of making them unique.

evetion:
Found while debugging the LHM model. A tracer was added, but didn't show up in Delwaq because the boundary list (bndlist) was shifted after duplicate UserDemand edges.

Also added more information to the generated.csv files for debugging.


SouthEndMusic (approved):
I was convinced that this is good (:

Issue #1472, opened 2024-05-16
Derive Julia call stacks


evetion:
For problems such as allocation. As discussed during a knowledge sharing session.

With a call stack it's easier to understand how a computational feature works, and how new code can be designed/integrated, or old code refactored.

edit: I envision something like https://juliadocs.org/DocumenterMermaid.jl/stable/
![Image](https://github.com/user-attachments/assets/e26da8f4-a155-4554-a4f1-66425b9006bd)

evetion:
If this is not doable (in a sprint), let's just describe the different parts of the codebase.

evetion:
Ok, so ideally we get a call stack in Julia that implements the AbstracTreeInterface, which we then could easily put into a markdown with mermaidjs backticks.

However, I've tested the [gist](https://gist.github.com/mkborregaard/81825c3d370bb4d8dbfe59c3b2ae4b33) partially based on the old TraceCalls and this crashed on a Ribasim.run (after ~20min). I've tested TraceFuns, and it only prints, not stores, and even that crashed (but printed a lot of nice output).

So the proper way of doing this is editing TraceCalls, or following [this documentation of Cassette.jl](https://julia.mit.edu/Cassette.jl/stable/contextualdispatch/) to actually keep track of all the calls.

However, that seems out of scope here, so I would propose to do a manual documentation of the call stack for:
- toml/DB loads
- all the different callbacks
- water_balance! specific calls

PR #1922, opened 2024-10-30
Callstack visualization

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1472.


SouthEndMusic:
Obtaining the call graph seems to work well. We should just restrict it to smaller parts of the code, not an entire run. For visualization I went with a `Makie` approach, here is an example for `Parameters`:

image

We can debate very long about what the best way would be to visualize the data, but because it gets very dense very quickly I think this plot is best used interactively. I do think the color coding by script is quite nice.

docs/dev/callstacks.quarto_ipynb
evetion, line 1: Can be removed as it is generated by quarto

Project.toml
```diff
@@ -20,6 +21,7 @@ DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
 FindFirstFunctions = "64ca27bc-2ba2-4a57-88aa-44e436879224"
 FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
+GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
```
evetion, line 24: This is useful for interaction, but I can imagine CairoMakie is enough for quarto static plots?

evetion (commented):
Great effort! 💯 

Would it be useful to depend on https://graph.makie.org/stable/generated/depgraph/? The current layout seems quite constrained, ideally you have an automatic graph layout.

For interactivity, a script in utils might be a better fit? Or a new util folder in docs, so the quarto notebook is a bit shorter?

Did you test this with a single calculation timestep as well?

Finally, given your slack question, any progress on an mermaidjs translation?

SouthEndMusic:
Improved formatting, generated functions cut out:

image

SouthEndMusic:
The graph for a run of the allocation algorithm:

image

I also want to show allocation initialization but for some reason the call trace code crashes on that at the moment.

SouthEndMusic:
Ah I forgot `DiscreteControl`

docs/dev/callstacks.qmd
```diff
@@ -0,0 +1,413 @@
+# Call stacks
```
evetion, line 1: Maybe move all this code into a separate .jl file and `include` that here. That makes re-use easier, and also paves the way for creating CallGraph.jl in the future.

docs/dev/callstacks.qmd
```diff
@@ -0,0 +1,413 @@
...
+    squash_methods = [
+        :n_neighbor_bounds_flow,
+        :n_neighbor_bounds_control,
+        :sort_by_function,
+        :neighbortypes
+    ],
+    xlims = (-0.4, 5.6)
+)
+```
+
```
evetion, line 361: Maybe add the update_cumulative_flows callback here as well? That should cover most of the core code.

docs/dev/callstacks.qmd
```diff
@@ -0,0 +1,413 @@
...
+# | code-fold: true
+using OrdinaryDiffEqCore: get_du
+model = Ribasim.Model(toml_path)
+du = get_du(model.integrator)
+(; u, p, t) = model.integrator
+graph, verts = tracecall((Ribasim,), Ribasim.water_balance!, (du, u, p, t))
+plot_graph(graph, size = (1700, 1000), xlims = (-0.4, 4.5))
+```
+
+## Allocation initialization
```
evetion, line 374: Can you provide a sentence or two on each figure? Ideally (only) on something that stands out, or is not immediately clear.

evetion (approved):
Looking good 🚀. Two remarks before we merge:
- Maybe good to make some links from the equations/implementation page to each specific plot?
- In the plots all the Ribasim. prefixes should be removed.

edit:
Maybe remove the spines/axes labels/ticks with [`hidedecorations!`](https://docs.makie.org/dev/reference/blocks/axis#Hiding-Axis-spines-and-decorations) and/or the [minimal theme](https://docs.makie.org/dev/explanations/theming/predefined_themes#theme-minimal).

docs/dev/callstacks.qmd
SouthEndMusic: I decided against adding this because `update_cumulative_flows!` calls very few other (Ribasim) functions.

Issue #1709, opened 2024-08-09
Allocation should return same strategy if a model is run twice


Jingru923:
**What**
When building up benchmark for Ribasim, it has come to attention that if I run a model with allocation twice, I will get different water balance. This is because some optimization problems have unlimited solutions and the solver returns a random one. Therefore the output of allocation model is always different. This should be investigated.

**How**

- [ ] Investigate JuMP and its solvers. Given an optimization problem with non-unique solution, the solver should return the same solution every time.
- [ ] Run model with allocation twice, and see if the output is the same
- [ ] Update the benchmark for allocation models.

Jingru923:
In the regression test, these values should be benchmarked and tested against the benchmark.
```julia 
                # TODO @test all(q -> abs(q) < 12.0, basin.storage - basin_bench.storage)
                # TODO @test all(
                #     q -> abs(q) < 100.0,
                #     basin.storage[(end - 16):end] - basin_bench.storage[(end - 16):end],
                # )
                # TODO @test all(
                #     q -> abs(q) < 2.0,
                #     basin.level[(end - 16):end] - basin_bench.level[(end - 16):end],
                # )
                # TODO @test all(q -> abs(q) < 1e-1, basin.balance_error)
                # TODO @test all(q -> abs(q) < 30.0, basin.relative_error)
```

Huite:
> Given an optimization problem with non-unique solution, the solver should return the same solution every time

Wouldn't it be better to make sure you have a unique solution? I.e. we should make it so that the optimization problem is well defined.

SouthEndMusic:
Yes, that's what upcoming issues for allocation are for, especially optimizing per source individually. Still, it's weird that the optimizer is not deterministic.

Issue #1504, opened 2024-05-28
Don't allow flow to enter buffers when collecting subnetwork demands


SouthEndMusic:
When https://github.com/Deltares/Ribasim/issues/565#issuecomment-2134720174 is implemented, we can make modifications to the optimization problems for each source individually. Specifically, we can use specific constraints so that when collecting the demands of a subnetwork, i.e. optimizing for a source that is a main network to subnetwork inlet, we can prohibit flow from entering these buffers, so that this doesn't count towards the subnetwork demand.

PR #1927, opened 2024-11-04
Allocation: optimize per source

SouthEndMusic:
# Optimizing per source

Fixes https://github.com/Deltares/Ribasim/issues/565

What happens now in optimizing per subnetwork per priority (see `optimize_priority!`):

- `set_capacities_flow_demand_outflow!`: Set the capacity of outflow edge of a node with a flow demand (Inf if `priority == flow demand priority`, 0 otherwise)
- `set_objective_priority!`: Set the equal ratio objective function (a term per demand of the current priority)
- `allocate_to_users_from_connected_basin!`: Do allocation without optimization where users are directly connected to basins that can supply flow
- `allocate_priority!`: solve the optimization problem
- `assign_allocations!`: copy the allocated values to the proper places (either for UserDemand or a subnetwork treated as a UserDemand node)
- `save_demands_and_allocations!`: Write allocation result data to the allocation record (`record_demand`) for the `allocation.arrow` output
- `save_allocation_flows!`: Write allocation flow data to the allocation record (`record_flow`) for the `allocation_flow.arrow` output

What has to change when optimizing per source:

- [x] Add vector of sources in subnetwork in desired order to `AllocationModel` struct and add this to `AllocationModel` constructor in `allocation_init.jl`. Types of sources (in order used for now, 'use flow available within subnetwork first'): edge designated as source, basin with level demand, main network to subnetwork, connector node with flow demand (buffer)
- [x] Add loop over sources in `allocate_priority!`
- [x] Within this loop, set the capacity of the current source as the capacity of this source and the other sources 0, after which the optimization is run
- [x] Adapt all `adjust_*` functions for use within this loop and make sure `flow_priority` is cumulative over all sources
- [x] Call `save_demands_and_allocarions!` and `save_allocation_flows!` after the loop
- [x] Adjust the allocation docs to solving per source (both `allocation.qmd` instances)

# Limiting flow demand buffer inflow

Fixes https://github.com/Deltares/Ribasim/issues/1504
Fixes https://github.com/Deltares/Ribasim/issues/1709 (hopefully)

A bit of background here: nodes that have a flow demand have a flow buffer, which is implemented so that if the flow that goes from a source to a node with a flow demand has nowhere to go afterwards, it can be stored in the flow buffer to potentially be allocated downstream in subsequent optimizations. 

Currently the flow into the flow buffer is unconstrained, meaning that for each optimization an arbitrary amount of flow can end up in the buffer. The objective only cares about the flow over the edge into the node with the flow demand, not about the flow into the buffer. Changes:

- [x] Remove flow demand outflow `0` or `Inf` constraints
- [x] Use buffer inflow instead of flow to node with flow demand in objective function
- [x] Flow that goes through the flow demand node but not into the buffer should still count towards what's allocated to the flow demand
- [x] Fix tests
- [x] Update inline docs
- [x] Update online docs (both `allocation.qmd` instances)


SouthEndMusic:
Note (@visr @gijsber): this PR changes (again) how `UserDemand` returnflow (and flow buffers) is used as a source. Previously, when some flow was allocated to a `UserDemand` node, the return flow fraction was available as a source for the next priority (also to the same `UserDemand` node if the graph topology permits). Now that we optimize per source, when some flow is allocated to a `UserDemand` node from one source, the return fraction is already available as a source when optimizing for the next source within the same priority. We could go back to the same behavior as before but that requires some extra book keeping.

gijsber:
Hi @SouthEndMusic , I prefer the new behaviour. The old behaviour did not have my preference but seemed like unavoidable at that time.

SouthEndMusic:
Another point of attention: I think I already assumed this before, but I will rely on it now/again: for flow demands only flows in the direction of the edge count, not the other way. I think this is a reasonable assumption and requires less complexity

SouthEndMusic:
Maintaining [the allocation developer docs](https://ribasim.org/dev/allocation.html) seems a bit of a chore, especially with [the call stacks](https://ribasim.org/dev/callstacks.html) in place in combination with inline documentation. @FaridAlavi, I'd be interested to hear what you think the value of this documentation page is from your perspective.

core/src/allocation_optim.jl
```diff
@@ -214,32 +214,66 @@ function set_initial_capacities_source!(
...
+            AllocationSourceType.edge,
+            AllocationSourceType.main_to_sub,
+            AllocationSourceType.user_return,
         )
+            JuMP.value(problem[:F][edge])
+        elseif source.type == AllocationSourceType.basin
+            JuMP.value(problem[:F_basin_out][edge[1]])
+        elseif source.type == AllocationSourceType.buffer
+            JuMP.value(problem[:F_flow_buffer_out][edge[1]])
+        end
```
visr, line 243:
```suggestion
            JuMP.value(problem[:F_flow_buffer_out][edge[1]])
        else
            error("Unknown source type")
        end
```

core/src/allocation_optim.jl
```diff
@@ -717,42 +641,43 @@ function save_demands_and_allocations!(
 )::Nothing
     (; graph, allocation, user_demand, flow_demand, basin) = p
     (; record_demand, priorities, mean_realized_flows) = allocation
-    (; subnetwork_id, problem) = allocation_model
+    (; subnetwork_id, problem, sources, flow) = allocation_model
     node_ids = graph[].node_ids[subnetwork_id]
     constraints_outflow = problem[:basin_outflow]
```
visr, line 646:
```suggestion
```

Unused

core/src/allocation_optim.jl
```diff
@@ -791,13 +716,13 @@ function save_allocation_flows!(
     priority::Int32,
     optimization_type::OptimizationType.T,
 )::Nothing
-    (; flow_priority, problem, subnetwork_id, capacity) = allocation_model
+    (; flow, problem, subnetwork_id, capacity) = allocation_model
```
visr, line 719:
```suggestion
    (; flow, problem, subnetwork_id) = allocation_model
```

Unused

visr (approved):

PR #1928, opened 2024-11-05
Some small tweaks to the callstack plots.

evetion:
- Added opacity to arrows
- Removed the axes and borders
- Inverted the stroke/color of the labels
- Removed source path in legend

Screenshot 2024-11-05 at 11 04 03


SouthEndMusic (approved):
I'm in favor of changes improving readability. I do think that having a border around the legend makes it look cleaner, and keeping the depth indication on the x axis could aid in referring to parts in the data dense plots

evetion:
> I'm in favor of changes improving readability. I do think that having a border around the legend makes it look cleaner, and keeping the depth indication on the x axis could aid in referring to parts in the data dense plots

Fair enough:

Screenshot 2024-11-06 at 17 11 50

I've also simplified the plot command a bit, I wanted to go to automatic sizing, but it seems the depth is not what I think it is (branches are not pruned, but cut, so stuff is still left beyond max_depth). Could use a rethink when making it into a package (proper tree structure).

Issue #1905, opened 2024-10-16
Refactor Basin Parameter (constructor)


evetion:
Also, this basin `Basin` constructor is getting rather monstrous, could be divided into several parts. That is out of the scope of this PR though, so could be worth a new issue.

_Originally posted by @SouthEndMusic in https://github.com/Deltares/Ribasim/pull/1849#discussion_r1799336397_

evetion:
> Maybe it would be a simple development quality of life improvement of we make a simple Base.show method for Parameters.

PR #1929, opened 2024-11-07
Refactor of concentration code

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1905


docs/guide/delwaq.ipynb
SouthEndMusic, line 1: 👀

evetion (approved):
I had my doubts about this refactor because it now becomes a separate callback, and I thought that cumulative flow was updating some states that we also needed for concentrations, but that doesn't seem to be the case. Let's merge the other PR, retest here, and then merge.

Issue #1919, opened 2024-10-28
Smoothing the `Q(h)` relation


SouthEndMusic:
In the grand scheme of boosting performance by making the whole RHS of the Ribasim ODE problem C1 smooth, doing this for `TabulatedRatingCurve` is one of the steps. Currently the $Q(h)$ relation is given by piecewise linear interpolation of the input data. To make this more smooth, we can use a cubic polynomial that rounds out the corners. As proposed by @Huite, this is easiest to understand by hydrologist by taking some fixed small interval centered around each data point where the linear interpolation is replaced by a polynomial (some care must be taken half this interval is bigger than the distance between points in the $x$ direction). 

For the nicest integration this could be incorporated into `DataInterpolations.jl`. For some inspiration see [here](https://southendmusic.github.io/CachedInterpolations.jl/stable/construction/construction_smoothed_linear_interpolation/), which uses an interval around the data points whose size is relative to the distance between points instead of absolute as proposed here.

Huite:
Commenting this for completeness, this paper

> Model smoothing strategies to remove microscale discontinuities and spurious secondary optima in objective functions in hydrological calibration
[Dmitri Kavetski](https://agupubs.onlinelibrary.wiley.com/authored-by/Kavetski/Dmitri), [George Kuczera](https://agupubs.onlinelibrary.wiley.com/authored-by/Kuczera/George)
First published: 08 March 2007 https://doi.org/10.1029/2006WR005195

primarily suggest using logistic smoothing. This has the upside of being better than C2, but I'd say with a significant cost since our Qh tables contain quite a lot of rows.

Basically, you need a logistic function for each jump of dQ/dh. For integration, to keep it continous, you need to sum as many integrals (a softplus) as there are jumps (because the function is only 0 or 1 in the limit as x to ±infinity). For e.g. 20 points, that means 20 softplus evaluations.

![Image](https://github.com/user-attachments/assets/650d79c3-bbad-48a7-b139-34188f7945ae)

Issue #1920, opened 2024-10-28
Smooth forcing


Huite:
It's not uncommon to see this behavior in the output:
Forcing (precipitation or evaporation) changes abruptly (at t_i, t_i+1, t_i+2), then we do a forward fill. But Ribasim has to stomach the sudden change, and is thereby forced to create a lot more timesteps (red dots) to accurately simulate the effect of the shock.

![Image](https://github.com/user-attachments/assets/586ee4fc-f591-462f-ad03-4777f0a49562)

But obviously, we don't care about the dynamics of the shock at all: it's just an artifact, a result of having the forcing as daily sums. We get **sub-second** (or worse) dynamics because the solver is trying its darnedest to represent the large jumps in the **daily** sums. That feels very disproportionate.

I've discussed this with @visr before, and @SouthEndMusic mentioned it now in relationship to #1918 and #1919: the forcing is VERY discontinuous (not even C0).

## Smoothing forcing time series

It's plausible that we could substantially reduce the number of timesteps required by lessening the effects of these shocks. We've been hesitant to do so because it basically requires interpolation which makes the input difficult to interpret: interpolation might not conserve the total inflow, or it spreads it out in unforeseen ways.
(I reckon that interpolating for e.g. level boundaries is probably a lot less controversial.)

However, there might be an easier way out: make it an explicit optional pre-processing step. If we were to add a function e.g. to ribasim-python that takes pandas Series of precipitation and returns a (slightly) smoother one, I think that would be perfectly acceptable to most modelers (especially if you get a substantial run time reduction). You run the function, you plot the smoothed timeseries together with the original daily sums, and you check it's still close enough / preserves the thing you care about.

I think linear interpolation would already make a significant difference. Maybe the primary challenge is that making sure that the solver doesn't skip too much of the "up- and off-ramp"? To preserve cumulative inflow, it should interpolate for `t + 0.5 * dt` where `dt` is the proposed timestep (so it requires knowledge of `dt` next to `t`).

## Coupler flows

The coupler flows are also forward filled. Unlike forcing, these cannot simply be pre-processed, nor can precipitation be (slightly) redistributed over time. We have access to a previous flow and the flow for the next period; however, we are more less free (?) to distribute the flow internal of the flow coupler timestep. Maybe a scheme like this would be somewhat effective:

![Image](https://github.com/user-attachments/assets/14e10de3-ded8-4f8e-8b56-ee997e5fdf39)

[...]

Issue #1918, opened 2024-10-24
Smooth clamp function


Huite:
I've noticed `clamp` is used four times in `solve.jl`. The problem with clamp is that it is non-smooth so it won't play nice with the derivatives for Newton-Raphson.

A straightforward `smooth_clamp` is:

```julia
function clamp_smooth(x, lo, hi, m)
    d = (hi - lo)
    @assert m <= (0.5 * d)
    x_norm = (x - lo) / d
    a = 1 / (1 - m)
    
    if x_norm < 0
        y = 0
    elseif x_norm < m
        y = (a / (2 * m)) * x_norm^2
    elseif x_norm < (1 - m)
        y = a * x_norm + 0.5 * (1 - a)
    elseif x_norm < 1
        y = 1 - (a / (2 * m)) * (1 - x_norm)^2
    else
        y = 1
    end
    
    return lo + y * d
end
```

Where `m` is the smoothing parameter. This is only C1 continuous which is fine for us now but maybe needs some attention if we need Hessians in the future.

In terms of its behavior, this is with a large `m`:

![Image](https://github.com/user-attachments/assets/8f7a7f8e-5695-4ee2-82d3-b39f350bc961)

Applied to a pump, it means that the pump is a bit "lazy", then "panicky": it pumps slightly less when at half-capacity, and it pumps a little more when over half-capacity. In general, the value for `m` is very small, so deviations are negligible.

evetion:
> it should be helpful for the solver and give some performance benefits.

Lots of maybe here, is there a way to actually measure this (apart from implementing it?). There's always a performance need (Ribasim-NL will only get bigger), but ideally we could order ideas based on their expected benefits.

PR #1932, opened 2024-11-13
Configurable interpolation types

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1919.

Inspired by ongoing discussions with @Huite and @visr on smoothness (see also https://github.com/Deltares/Ribasim/issues/1918, https://github.com/Deltares/Ribasim/issues/1920) I started looking into making the type of interpolation configurable in the several places in the core where interpolation is used. For this PR I am focussing on $Q(h)$ relationships.

We require of the $Q(h)$ interpolation that the flow is $0$ below and at the lowest supplied level. That's trivial to achieve for linear interpolation, but not for non-linear (generally piecewise polynomial of degree >1) interpolation types. This requires a special approach for each method.

The question is which methods to support (and even whether this should be configurable for $Q(h)$ relationships at all). I now experiment with [PCHIP Interpolation](https://docs.sciml.ai/DataInterpolations/stable/methods/#PCHIP-Interpolation) because of its smoothness and non-overshooting properties.

Another point of attention is extrapolation. For linear interpolation extrapolation makes sense, but for higher order interpolation methods extrapolation gets out of hand quick. We could make our own wrapper of interpolation types from `DataInterpolation`, but I also made an issue there to pick it up internally:

https://github.com/SciML/DataInterpolations.jl/issues/355

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -461,9 +461,9 @@ version = "1.7.0"
 
 [[deps.DataInterpolations]]
 deps = ["FindFirstFunctions", "ForwardDiff", "LinearAlgebra", "PrettyTables", "RecipesBase", "Reexport"]
-git-tree-sha1 = "3d81cd1fcba530122a5d6c725aa53521d869816a"
+git-tree-sha1 = "78d06458ec13b53b3b0016daebe53f832d42ff44"
 uuid = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
-version = "6.5.2"
+version = "6.6.0"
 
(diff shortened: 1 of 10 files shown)


SouthEndMusic:
Should the constructors keep the `extrapolate` keyword or can it be replaced by `extrapolation_type::Symbol` which has one of the above supported types (preferably also split in `upper_extrapolation_type` and `lower_extrapolation_type`)?

SouthEndMusic:
This PR has gotten stale. Requires some more refinement before a new attempt.

Issue #1952, opened 2024-11-26
Use mutable structs + `const` instead of `Base.Refvalue` fields


SouthEndMusic:
As suggested by @vsir: https://github.com/Deltares/Ribasim/pull/1927#discussion_r1858172480.

_We use Refs like this now in a few places, but I wonder if this isn't easier to just use mutables with const fields like this:_

```julia
julia> mutable struct Baz
           a::Int
           const b::Float64
       end

julia> baz = Baz(1, 1.5);

julia> baz.a = 2
2

julia> baz.b = 2.0
ERROR: setfield!: const field .b of type Baz cannot be changed
[...]
```

PR #1954, opened 2024-11-26
Get rid of `Base.RefValue` fields

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1952.


visr (approved):
Thanks, this looks more readable.

PR #1971, opened 2024-12-12
Put level as the first column in basin.arrow

visr:
The iMOD timeseries plugin shows the first column by default. That is storage. Most users want to view level first, and have to switch manually. So this simply puts level first, such that we get:

image

### DIFF
--- a/core/src/write.jl
+++ b/core/src/write.jl
@@ -119,8 +119,8 @@ function basin_table(
 )::@NamedTuple{
     time::Vector{DateTime},
     node_id::Vector{Int32},
-    storage::Vector{Float64},
     level::Vector{Float64},
+    storage::Vector{Float64},
     inflow_rate::Vector{Float64},
     outflow_rate::Vector{Float64},
     storage_rate::Vector{Float64},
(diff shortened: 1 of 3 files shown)


evetion:
Changing our Julia structs and output formats so a plotter changes it defaults seems pretty backwards to me. Is changing the behavior of the iMOD plugin that hard?

visr:
No Julia struct has changed, it is only the order of the columns. And since level is used more I think it is actually desirable to have it more to the left, where it is less likely cut off the screen.

Probably changing the iMOD plugin isn't that hard, but there is currently no concept of default variable there, and I think this is actually an improvement.

evetion:
> No Julia struct has changed, it is only the order of the columns. And since level is used more I think it is actually desirable to have it more to the left, where it is less likely cut off the screen.
> 
> Probably changing the iMOD plugin isn't that hard, but there is currently no concept of default variable there, and I think this is actually an improvement.

Well a NamedTuple changed! 😅 That said, this is post-hoc reasoning, the goal is fixing the graph, and this is a bad way to do it. Please make at least an issue to make this configurable in the iMOD/Ribasim plugin before we merge this.

visr:
> Please make at least an issue to make this configurable in the iMOD/Ribasim plugin before we merge this.

Done, see https://github.com/Deltares/imod-qgis/issues/84

evetion (approved):

PR #1974, opened 2024-12-13
Automatically name index

### LINKED ISSUE #1968: Don't require `index.name` to be set to `fid`
This doesn't work unless you add `df.index.name = "fid"`, which feels like it could also be done for us. Ribasim-NL contains many such lines for that reason.

```py
import ribasim_testmodels
model = ribasim_testmodels.trivial_model()
df = model.basin.static.df.reset_index(drop=True)
model.basin.static.df = df
```

```
File a:\repo\ribasim\Ribasim\.pixi\envs\dev\Lib\site-packages\pydantic\main.py:923, in BaseModel.__setattr__(self, name, value)
    [921](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:921)     self.__dict__[name] = value
    [922](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:922) elif self.model_config.get('validate_assignment', None):
--> [923](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:923)     self.__pydantic_validator__.validate_assignment(self, name, value)
    [924](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:924) elif self.model_config.get('extra') != 'allow' and name not in self.__pydantic_fields__:
    [925](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:925)     # TODO - matching error
    [926](file:///A:/repo/ribasim/Ribasim/.pixi/envs/dev/Lib/site-packages/pydantic/main.py:926)     raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')

ValidationError: 1 validation error for TableModel[BasinStaticSchema]
df
  Value error, Expected <class 'pandas.core.series.Series'> to have name 'fid', found 'None' [type=value_error, input_value=   node_id  drainage  pre...       0.0          <NA>, input_type=DataFrame]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

Interestingly `model.basin.state` does not suffer from this issue.

This is controlled by `fid: Index[Int32] = pa.Field(default=1, check_name=True, coerce=True)` in the jinja template. Setting `check_name=False` unfortunately doesn't set the name for us, but perhaps we can find another way for it to do that.

visr:
Fixes #1968

We let pandera check the index name with `check_name = True`. This adds a pandera parser called `_name_index` to set the index to the desired name before validating. This still keeps the index names, but offers more user convenience, since the index name is lost with many pandas operations.

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -217,7 +217,6 @@ class TableModel(FileModel, Generic[TableT]):
 
         # Enable initialization with a DataFrame.
         if isinstance(value, pd.DataFrame | gpd.GeoDataFrame):
-            value.index.rename("fid", inplace=True)
             value = {"df": value}
 
         return value
@@ -386,7 +385,6 @@ class SpatialTableModel(TableModel[TableT], Generic[TableT]):
                     # tell pyarrow to map to pd.ArrowDtype rather than NumPy
                     arrow_to_pandas_kwargs={"types_mapper": pd.ArrowDtype},
                 )
-                df.index.rename(cls.tableschema()._index_name(), inplace=True)
             else:
                 df = None
 
--- a/python/ribasim/ribasim/schemas.py
+++ b/python/ribasim/ribasim/schemas.py
@@ -16,11 +16,13 @@ class _BaseSchema(pa.DataFrameModel):
         add_missing_columns = True
         coerce = True
 
+    @classmethod
+    def _index_name(self) -> str:
+        return "fid"
+
     @pa.dataframe_parser
     def _name_index(cls, df):
-        # Node and Edge have different index names, avoid running both parsers
-        if cls.__name__ not in ("NodeSchema", "EdgeSchema"):
-            df.index.name = "fid"
+        df.index.name = cls._index_name()
         return df
 
     @classmethod
--- a/python/ribasim/ribasim/schemas.py
+++ b/python/ribasim/ribasim/schemas.py
@@ -16,9 +16,12 @@ class _BaseSchema(pa.DataFrameModel):
         add_missing_columns = True
         coerce = True
 
-    @classmethod
-    def _index_name(self) -> str:
-        return "fid"
+    @pa.dataframe_parser
+    def _name_index(cls, df):
+        # Node and Edge have different index names, avoid running both parsers
+        if cls.__name__ not in ("NodeSchema", "EdgeSchema"):
+            df.index.name = "fid"
+        return df
 
     @classmethod
     def migrate(cls, df: Any, schema_version: int) -> Any:
--- a/python/ribasim/ribasim/geometry/edge.py
+++ b/python/ribasim/ribasim/geometry/edge.py
@@ -49,9 +49,10 @@ class EdgeSchema(_GeoBaseSchema):
     edge_type: Series[str] = pa.Field(default="flow")
     geometry: GeoSeries[LineString] = pa.Field(default=None, nullable=True)
 
-    @classmethod
-    def _index_name(self) -> str:
-        return "edge_id"
+    @pa.dataframe_parser
+    def _name_index(cls, df):
+        df.index.name = "edge_id"
+        return df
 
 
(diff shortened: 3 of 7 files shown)


python/ribasim/ribasim/input_base.py
visr, line 1: These are no longer needed because the parser will rename it for us.

python/ribasim/ribasim/schemas.py
```diff
@@ -16,9 +16,12 @@ class Config:
...
 
-    @classmethod
-    def _index_name(self) -> str:
-        return "fid"
+    @pa.dataframe_parser
+    def _name_index(cls, df):
+        # Node and Edge have different index names, avoid running both parsers
+        if cls.__name__ not in ("NodeSchema", "EdgeSchema"):
+            df.index.name = "fid"
+        return df
```
visr, line 24: I was confused for a while since I expected the methods in the subclasses in edge.py and node.py to override this. But when debugging a test failure I saw that first the node parser was applied, naming it "node_id", and then this one, making it end up with the wrong name. Hence the check for those class names.

python/ribasim/ribasim/schemas.py
```diff
@@ -30,595 +33,755 @@ def migrate(cls, df: Any, schema_version: int) -> Any:
 
 class BasinConcentrationExternalSchema(_BaseSchema):
     fid: Index[Int32] = pa.Field(default=1, check_name=True, coerce=True)
+
```
visr, line 36: `gen_python.jl` creates even more whitespace, of which ruff now seems to leave one. I tried to fix this with OteraEngine settings but didn't manage. Not so important though.

visr: Hmm the codegen CI doesn't generate the whitespace, so I removed it again. Not sure why I get different results.

python/ribasim/ribasim/schemas.py
evetion: How about `df.index.name = cls._index_name()`? A mix of both old and new.

evetion (commented):
Are there any Ribasim operations changing the DataFrame index? This is more lenient I guess. Maybe add a debug statement if the index name doesn't match?

python/ribasim/ribasim/schemas.py
visr: Yeah we could go with that as well.

visr:
> Are there any Ribasim operations changing the DataFrame index? This is more lenient I guess. Maybe add a debug statement if the index name doesn't match?

Not that I'm aware. I don't understand exactly what kind of debug statement you had in mind.

evetion:
> > Are there any Ribasim operations changing the DataFrame index? This is more lenient I guess. Maybe add a debug statement if the index name doesn't match?
> 
> Not that I'm aware. I don't understand exactly what kind of debug statement you had in mind.

Then this PR is really for external users that are playing fast and loose with their dataframes right (not using the add API)?

pseudo for debug statement:
```python
if df.index.name != cls.index_name():
  logging.debug("How dare you")
  df.index.name = cls.index_name()
```

visr:
Hmm in this PR I actually removed two explicit index namings because we can now rely on this. I don't think a debug statement then makes sense if we say we'll take care of it.

Here you can see a bunch of uses: https://github.com/search?q=repo%3ADeltares%2FRibasim-NL+index.name+%3D&type=code
True that you wouldn't need it with just add API, but you currently can't do everything we need to do like that, so making an effort to make the direct df API usable IMO isn't so bad.

python/ribasim/ribasim/schemas.py
visr: Done in 007ee67bc71cf3491dea67d94078be0bf2d9a15d

evetion (approved):
Neat!

PR #1977, opened 2024-12-17
Speed up initialization

visr:
For lhm_2024_12_3 this speeds up initialization (`Ribasim.Model(toml_path)`), from 234 seconds to 4 seconds. Large models were especially slow to initialize since in `create_graph` we were constructing NodeID structs for each node with a separate call to the database.

This changes `get_node_ids` to return a `Vector{NodeID}` rather than a `Vector{Int32}`. This means that the type and index of each ID is now directly included, leading to simpler code.

### DIFF
--- a/core/src/graph.jl
+++ b/core/src/graph.jl
@@ -6,6 +6,7 @@ and data of edges (EdgeMetadata):
 [`EdgeMetadata`](@ref)
 """
 function create_graph(db::DB, config::Config)::MetaGraph
+    node_table = get_node_ids(db)
     node_rows = execute(
         db,
         "SELECT node_id, node_type, subnetwork_id FROM Node ORDER BY node_type, node_id",
@@ -40,7 +41,7 @@ function create_graph(db::DB, config::Config)::MetaGraph
         graph_data = nothing,
     )
     for row in node_rows
-        node_id = NodeID(row.node_type, row.node_id, db)
+        node_id = NodeID(row.node_type, row.node_id, node_table)
         # Process allocation network ID
         if ismissing(row.subnetwork_id)
             subnetwork_id = 0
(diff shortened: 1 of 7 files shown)


core/src/graph.jl
visr, line 1: This is responsible for the speedup, going from n database queries to 1.

core/src/parameter.jl
```diff
@@ -52,42 +58,28 @@ This index can be passed directly, or calculated from the database or parameters
...
-            execute(
-                columntable,
-                db,
-                "SELECT COUNT(*) FROM Node WHERE node_type == $(esc_id(node_type_string)) AND node_id <= $value",
-            ),
-        ),
-    )
-    if idx <= 0
-        error("Node ID #$value of type $type is not in the Node table.")
+function NodeID(node_type, value::Integer, node_ids::Vector{NodeID})::NodeID
```
visr, line 60: Often you know the `node_type`, and since `node_ids` has it as well we can check if it is correct. Especially in test code I find `NodeID(:Pump, 5, v)` easier to read than `NodeID(5, v)`. Both is possible however, we call the latter for e.g. listen_node_type handling where we don't know the type.

core/src/read.jl
```diff
@@ -1395,11 +1397,52 @@ function Parameters(db::DB, config::Config)::Parameters
     return p
 end
 
-function get_ids(db::DB, nodetype)::Vector{Int32}
-    sql = "SELECT node_id FROM Node WHERE node_type = $(esc_id(nodetype)) ORDER BY node_id"
+function get_node_ints(db::DB, node_type)::Vector{Int32}
```
evetion, line 1400: Nit: This signature is very similar to get_node_ids and I almost thought you made a typo. `get_node_ids_int32` or similar?

evetion (approved):
LGTM, good to bring back the queries from N to 1. I remember having a discussion on this a while (year?) back, and we were opinion that even though it was suboptimal, it wasn't hurting is yet. Nice to see we have reached that stage/growth now.

Also, TIL about counter (I knew it existed, hadn't seen an example yet).

core/src/read.jl
visr: Ha yeah good point, done.

PR #1980, opened 2024-12-20
Don't update concentrations if not opting in to feature

visr:
I noticed these callbacks take 50% of computational time, for which I will create a separate issue. Currently we always compute, but only write out the results if the experimental feature flag was enabled. Since it is slowing down the model, let's turn off the computation as well.

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -34,25 +34,30 @@ function create_callbacks(
         FunctionCallingCallback(update_cumulative_flows!; func_start = false)
     push!(callbacks, cumulative_flows_cb)
 
-    # Update concentrations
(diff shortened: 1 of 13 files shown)


evetion (approved):
This should've been done when I implemented the experimental section. That said, can you find a way to enable it for all testmodels? Otherwise we test this codepath only for a single testmodel, instead of always/everywhere, which is a big step back.

visr:
> can you find a way to enable it for all testmodels

I can add this to all testmodels:

https://github.com/Deltares/Ribasim/blob/ac0920dee65438ccdefd1c56a2382c65e13120b4/python/ribasim_testmodels/ribasim_testmodels/basic.py#L29

Anything else is probably too magical.

PR #1982, opened 2024-12-20
Add generated, performant snake_case for NodeID.

### LINKED ISSUE #1981: Speed up snake_case for concentrations
When looking at a runtime profile I noticed `update_concentrations!` taking about 50% of the time. Most of that came from `snake_case(Symbol(id.type))`, called here:
https://github.com/Deltares/Ribasim/blob/81a7b7c4a5962a21acd76632ce20448f3a4dd1b5/core/src/util.jl#L1099

We can probably work around that, @evetion suggested either rewriting the method for symbols, fully writing it out or using something like https://github.com/JuliaCollections/Memoize.jl.

#1980 made this performance issue only relate to experimental concentration runs.

evetion:
Fixes #1981

This precalculates all snakecases for NodeType and compiles it. @visr could you benchmark this?

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -190,15 +190,15 @@ function update_concentrations!(u, t, integrator)::Nothing
     # of the basins after processing inflows only
     cumulative_in .= 0.0
 
-    mass .+= concentration[1, :, :] .* vertical_flux.drainage * dt
+    @views mass .+= concentration[1, :, :] .* vertical_flux.drainage * dt
     basin.concentration_data.cumulative_in .= vertical_flux.drainage * dt
 
     # Precipitation depends on fixed area
     for node_id in basin.node_id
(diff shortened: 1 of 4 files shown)


visr (approved):
Ah yes with this I don't see this line anymore in the profile of the LHM model run. I did see a bunch of allocations, so I pushed some views onto this branch to remove them.

With this the concentration callback takes around 20% of runtime. Most of that remaining time is spent iterating over a namedtuple here:

https://github.com/Deltares/Ribasim/blob/ae2589cb2718e1dd6b7babb851857a2bff0bbcb5/core/src/util.jl#L1101-L1105

evetion:
I assumed this code was also used in non-concentration parts? So we might see a speedup in normal runs too?

Good catch with those views. Unless you have ideas on how to further optimize that namedtuple iteration, let's merge this.

PR #1993, opened 2025-01-06
refactor: change to `depends-on`

tdejager:
This makes the `pixi.toml` ready for the next release of pixi which will deprecate `depends_on` in favor of `depends-on`. This has been available for a while now, so it should not cause issues. We found this failure in our downstream tests :)


visr (approved):
Thanks Tim! I'll update our other repos as well.

PR #2006, opened 2025-01-14
Use interpolations to update forcing in Basin

visr:
Fixes another bullet point from #601, namely:

- use interpolation objects instead of `update_basin`, which still relies on time being sorted first

Now we can also use `parse_static_and_time` for Basin, bringing our node types more in line with each other and more code reuse.  For this I had to extend it a bit to support other interpolation types, and support IDs not being in time nor static. We still add time of new data as tstops to ensure the solver knows about net data as it happens.

Labeling this as breaking since before this we allowed a node ID to be in both the static and time tables. In `parse_static_and_time` (the other node types) we specifically check for this with a clear error message. I had to update `basic_transient`, to remove the static table there.

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -10,14 +10,7 @@ function create_callbacks(
     u0::ComponentVector,
     saveat,
 )::Tuple{CallbackSet, SavedResults}
-    (;
-        starttime,
-        basin,
-        flow_boundary,
-        level_boundary,
-        user_demand,
-        tabulated_rating_curve,
-    ) = parameters
+    (; starttime, basin, flow_boundary, level_boundary, user_demand) = parameters
     callbacks = SciMLBase.DECallback[]
 
     # Check for negative storage
@@ -40,6 +33,17 @@ function create_callbacks(
             FunctionCallingCallback(update_concentrations!; func_start = false)
         push!(callbacks, concentrations_cb)
     end
+
+    # Update Basin forcings
+    # All variables are given at the same time, so just precipitation works
+    times = [itp.t for itp in basin.precipitation]
+    tstops = Float64[]
+    for t in times
+        append!(tstops, t)
+    end
+    unique!(sort!(tstops))
+    basin_cb = PresetTimeCallback(tstops, update_basin!; save_positions = (false, false))
+    push!(callbacks, basin_cb)
 
     if config.experimental.concentration
         # Update boundary concentrations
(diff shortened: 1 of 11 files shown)


core/src/callback.jl
```diff
@@ -699,27 +698,45 @@ function save_subgrid_level(u, t, integrator)
...
-"Load updates from 'Basin / time' into the parameters"
+"Update one current vertical flux from an interpolation at time t."
+function set_flux!(
+    fluxes::AbstractVector{Float64},
+    interpolations::Vector{ScalarConstantInterpolation},
+    i::Int,
+    t,
+)::Nothing
+    val = interpolations[i](t)
+    # keep old value if new value is NaN
```
SouthEndMusic, line 709: This only works for constant interpolation, but that is not a concern for this PR

core/src/callback.jl
```diff
@@ -699,27 +698,45 @@ function save_subgrid_level(u, t, integrator)
...
+    return nothing
+end
+
+"""
+Update all current vertical fluxes from an interpolation at time t.
+
+This runs in a callback rather than the RHS since that gives issues with the discontinuities
+in the ConstantInterpolations we use, failing the vertical_flux_means test.
+"""
 function update_basin!(integrator)::Nothing
```
SouthEndMusic, line 722: This wrapper is rather thin, is the method `update_basin!(basin, t)` only used here or also somewhere else?

core/src/parameter.jl
```diff
@@ -421,7 +425,11 @@ end
     demand::Vector{Float64} = zeros(length(node_id))
     allocated::Vector{Float64} = zeros(length(node_id))
     # Data source for parameter updates
-    time::StructVector{BasinTimeV1, C, Int}
+    precipitation::Vector{ScalarConstantInterpolation} = ScalarConstantInterpolation[]
```
SouthEndMusic, line 428: Consider putting these in a substruct to reduce the number of fields of `Basin`

SouthEndMusic (approved):
Few nits, but otherwise it looks good

core/src/callback.jl
```diff
@@ -42,7 +35,13 @@ function create_callbacks(
     end
 
     # Update Basin forcings
-    tstops = get_tstops(basin.time.time, starttime)
+    # All variables are given at the same time, so just precipitation works
+    times = [itp.t for itp in basin.precipitation]
+    tstops = Float64[]
+    for t in times
+        append!(tstops, t)
+    end
```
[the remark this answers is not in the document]
visr: Ha I wrote it first like this but wanted to avoid the splat. The length of `times` is unknown to the compiler and can be long (number of Basin nodes), which the compiler doesn't like.

core/src/callback.jl
visr: Yeah in the Basin constructor I initialize the values. There the integrator doesn't exist yet, hence the wrapper.

core/src/parameter.jl
visr: Yeah looks better, done in e456500e6819db78b202f73cbab74dc03c02d874.

PR #2019, opened 2025-01-27
Upgrade to DataInterpolations v7

### LINKED ISSUE #1621: Setup of Interpolation in `qh_interpolation` contains an "arbitrary" offset of -1.0
I was taking a look at this with @Jingru923, since the inserted offset for the first vertex makes the subsequent validation a bit hacky.

The current code is this:

```julia
"""
From a table with columns node_id, flow_rate (Q) and level (h),
create a ScalarInterpolation from level to flow rate for a given node_id.
"""
function qh_interpolation(node_id::NodeID, table::StructVector)::ScalarInterpolation
    rowrange = findlastgroup(node_id, NodeID.(node_id.type, table.node_id, Ref(0)))
    level = table.level[rowrange]
    flow_rate = table.flow_rate[rowrange]

    # Ensure that that Q stays 0 below the first level
    pushfirst!(level, first(level) - 1)
    pushfirst!(flow_rate, first(flow_rate))

    itp = SmoothedLinearInterpolation(flow_rate, level; extrapolate = true)
    return LinearInterpolation(itp)
end
```

The level is preprended by `first(level) - 1`. This is just a trick, to make sure we don't extrapolate a Q < 0.
However, if I understand correctly, the SmoothedLinearInterpolation creates additional vertices for smoothing. Taking a quick glance at some of the source code, these are computed with splines, right?

In that case, the value of -1.0 places a new vertex, but it will affect the resulting smoothing vertices. That feels somewhat undesirable: i.e. in my opinion the following vertices specify exactly the same Q-h curve:

* [(0.0, 0.0), (1.0, 10.0)]
* [(0.0, 0.0), (0.1, 1.0)]

But does this actually result in exactly the same interpolation object? My intuition is no.

E.g. a different strategy could be to not take -1 as an offset, but to reflect the second vertex around the first:
`pushfirst!(level, level[1] - (level[2] - level[1]))`

Ideally, we could specify directly that interpolation should only occur for Q > 0.

Another solution is that we set `extrapolate=false`, but add another vertex "at infinity" for Q > 0. But this likely results in the same issue with smoothing?

visr:
Instead of adding extra dummy data points to the arrays, we can now directly specify the extrapolation behavior in the DataInterpolations API:

https://docs.sciml.ai/DataInterpolations/stable/extrapolation_methods/

Fixes #1621.


visr:
Oh this still fails a few tests, might need some work in DataInterpolations:

```
ERROR: LoadError: UndefVarError: `extrapolate_integral_left` not defined in `DataInterpolations`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
  [1] integral(A::DataInterpolations.LinearInterpolation{Vector{Float64}, Vector{Float64}, Vector{Float64}, Vector{Float64}, Float64, (1,)}, t1::Float64, t2::Float64)
    @ DataInterpolations a:\.julia\packages\DataInterpolations\5Uosd\src\integrals.jl:27
  [2] formulate_storage!(current_storage::Vector{Float64}, tprev::Float64, t::Float64, flow_boundary::Ribasim.FlowBoundary{@NamedTuple{node_id::Vector{Int32}, time::Vector{Dates.DateTime}, substance::Vector{String}, concentration::Vector{Float64}}})
    @ Ribasim A:\repo\ribasim\Ribasim\core\src\solve.jl:182
```

visr:
The `Int()` calls in this PR can be removed if we update to v7.1, https://github.com/SciML/DataInterpolations.jl/pull/379.

SouthEndMusic:
> Oh this still fails a few tests, might need some work in DataInterpolations:
https://github.com/SciML/DataInterpolations.jl/pull/386

SouthEndMusic:
Yep, with [this fix](https://github.com/SciML/DataInterpolations.jl/pull/391) nothings breaks anymore, the only failing tests are due to some slightly different simulation results.

visr:
@SouthEndMusic could you look into the test failures?

SouthEndMusic (approved):
Looks good, especially my own fix

PR #2020, opened 2025-01-28
More pixi environments

visr:
So far we used the dev environment for almost everything. This meant the dev environment had some large dependencies that have very little to do with one another. The three major ones are QGIS, Quarto and Rust. When we needed to download the dev environment we usually had to wait mainly for these, and these were also usually the ones causing issues not being able to upgrade part of the dev environment. This branch is called py313 because I wanted to get this but for instance QGIS blocks it (and xugrid via numba).

This PR then splits out these major dependencies. The Rust environment doesn't rely on anything else so we don't even need the default environment. QGIS and Docs (Quarto) both extend the dev environment. The vast majority of our CI and tasks need the dev environment, not one of these:

```toml
qgis = { features = ["qgis", "dev"], solve-group = "py312" }
docs = { features = ["docs", "dev"], solve-group = "py312" }
rust = { features = ["rust"], no-default-feature = true }
```

I'm not sure yet this is worth it though.

- Sometimes we just need a little more patience and don't really need the latest Python version that came out 113 days ago
- I wonder how much we will really save in time and bandwidth due to with caching
- It is simpler to have less environments
- Right now I switched `open-vscode.bat` to use the QGIS environment. It would be a pain if QGIS and dev were not together in our VS Code environment, because we want the latest Python hence we detach them.
- I dislike now having to specify the environment almost everywhere because of [ambiguous-environment-selection](https://pixi.sh/latest/features/multi_environment/#ambiguous-environment-selection). Since the qgis and docs environments also include the dev feature, dev tasks are deemed ambiguous. In our case I don't really care which one is picked. I understand the need to be explicit in such cases, but to me that is a usability downside of these superset environments.
- Most of the slowness from managing the large dev environment with pixi comes from ESET real time protection running in the way of pixi, sometimes also causing failures. Moving the repo and pixi cache to a Windows Dev Drive improved performance around 5x, but with Windows Defender in performance mode on the Dev Drive it could probably be a lot faster still. Now it sometimes often takes 15 minutes. Only updating it once a month helps. Here is a screenshot: image
This command finished after 16 minutes, after only minor modifications from #2021. It hardly seems to matter if all the files are already in the cache or not, it is not downloading but validating the cache and installing packages that takes time. This seems mostly an ESET issue, but quite a frustrating one, often I also get failures or slowness from other package managers.

So I think closing this is better. The edits to the dependencies can stay after recombining into the dev feature. I removed a few unused ones.

@Hofer-Julian you may find this interesting, or perhaps have better ideas :)


Hofer-Julian:
> Most of the slowness from managing the large dev environment with pixi comes from ESET real time protection running in the way of pixi, sometimes also causing failures. Moving the repo and pixi cache to a Windows Dev Drive improved performance around 5x, but with Windows Defender in performance mode on the Dev Drive it could probably be a lot faster still. Now it sometimes often takes 15 minutes. Only updating it once a month helps.

Damn 15 minutes is the improved version? That's terrible. I remember that you mentioned slowness, but that's ridiculous. I don't remember experiencing that when I still worked at Deltares, did ICT change something in the meantime?

 > Right now I switched open-vscode.bat to use the QGIS environment. It would be a pain if QGIS and dev were not together in our VS Code environment, because we want the latest Python hence we detach them.
 
Sounds to me like you'll want QGIS part of your default environment then :)
That should also help with ambiguous tasks. If feature defines a task "start" and that feature is added to multiple environments "default" and "test", `pixi run start` will run in the default environment.

visr:
> did ICT change something in the meantime

Yeah migration to Windows 11 with a bunch of new policies. Not sure what changed about ESET. Thanks for checking the timing, I cannot pause it to do such tests, that's good new numbers for my discussion with ICT.

> Sounds to me like you'll want QGIS part of your default environment then :)
That should also help with ambiguous tasks. If feature defines a task "start" and that feature is added to multiple environments "default" and "test", pixi run start will run in the default environment.

I assume you really mean default here, not dev which we use as our development environment. And that the start is a task in default. We get so many ambiguities because we mainly use dev, and use default as a more minimal set to run pytest. If anything, this suggests an approach like in this PR would work better if our "dev" becomes "default", and we accept that this environment is a bit oversized for some things.

Because indeed this is not ambiguous:

```toml
[tasks]
start = "echo howdy"

[environments]
b = []
c = []
```

But this is, which is more the scenario of this PR:

```toml
[feature.dev.tasks]
start = "echo howdy"

[environments]
a = ["dev"]
b = ["dev"]
```

visr:
Closing this in favor of using default as our dev environment: #2024.

PR #2021, opened 2025-01-28
Edit some pixi dependencies

visr:
Useful scraps from #2020.

- add-ribasim-icon is only for Windows
- Jupyter is a metapackage that contains jupyterlab ipykernel ipywidgets.
- Pip is not needed anymore, pixi uses uv.
- libgdal-arrow-parquet not sure we ever used that? qgis in conda-forge has arrow support without this
- plotly we use in our docs, not sure how this wasn't a problem earlier


evetion (approved):
Your confusion might indicate we don't need to add plotly here?

visr:
Probably not, but I rather we explicitly add direct dependencies. Just wonder how it came indirectly. It started giving an error in #2020.

Issue #1505, opened 2024-05-28
Handling default and non-default source priorities in allocation


SouthEndMusic:
The following was decided in a design session with @SouthEndMusic @gijsber @visr @FaridAlavi @Fati-Mon and John:

In the TOML (or in the `Model` constructor when building a model from Ribasim python) a source priority `default_source_priority_*` must be specified for every source node type that is used in the model. These source node types are:
- `Userdemand` return flow (put on top as requested. Can be sub-optimal because at the start the user demand does not have any return flow in its buffer yet)
- `FlowBoundary` and `LevelBoundary` together as `Boundary`
- `Basin` storage (only when it has a `LevelDemand`?)
- `FlowDemand` troughflow
- Main network to subnetwork inlet

Specifying the source priority should be possible when adding a node with the add API.

Multiple sources can have the same priority, and by default all sources nodes of the same type have the same priority. This must be taken into account when looping over the sources in the allocation algorithm.

There were some differences in opinion about how the source priority data should be stored in the GeoPackage file. My conclusion is that only the deviations from the default source priorities should be stored, and that they should be stored in their own dedicated table `AllocationSourceOrder`. This way the table is relatively small as is preferred by @Fati-Mon as a modeler. So expansion to the full explicit 'table' only happens in the model initialization in the core.

Since we have 2 types of priority now (for demands and for sources), as much as possible when a priority is mentioned it should be specified which type of priority it is. This also means a breaking change for column names that are now just `priority` (@evetion should we ease this transition somehow?).

Validation:

- A node addition with the add API should error when a source node is added with a subnetwork ID but no explicit source priority is given nor a default source priority is specified for the given node type. 

Documentation:

- It should be made clear in the documentation that the allocation algorithm first loops over demand priorities, and within those over the source priorities.

SouthEndMusic:
https://github.com/Deltares/Ribasim/pull/1927 introduced a hard coded source ordering per subnetwork:

- User return flow
- Source edges
- Basin storage
- Main network to subnetwork connections
- Buffers (from nodes with a flow demand)

within each category the order is now determined by the input. From this we can deduce 2 categories of sources:
- Sources introduced by a demand node;
- Sources introduced by source edges.

PR #2022, opened 2025-01-29
Source priority input

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1505.

To do:
- [x] Update QGIS plugin based on table changes
- [x] Allow for setting source priority with add API and/or separate function
- [x] Process the source priority data in the core
- [x] Make column renaming `priority -> demand_priority` a non-breaking migration step
- [x] Fix tests
- [x] Update docs


python/ribasim/ribasim/input_base.py
```diff
@@ -197,16 +197,13 @@ def tablename(cls) -> str:
...
-                    table_name = "_".join(names_lowered[n:])
-                    return node_name + delimiter + table_name
-            raise ValueError(f"Found no known node name in {cls_string}")
+        for n in range(1, len(names_lowered) + 1):
+            node_name_snake_case = "_".join(names_lowered[:n])
+            if node_name_snake_case in node_names_snake_case:
+                node_name = "".join(names[:n])
+                table_name = "_".join(names_lowered[n:])
+                return node_name + delimiter + table_name
+        return names[0]
```
SouthEndMusic, line 206: This doesn't work properly yet, but the python tests failed on this becase my new table name neither contains a node name nor is it one word long

core/test/docs.toml
```diff
@@ -21,8 +21,13 @@ ribasim_version = "2025.1.0" # required
...
 [allocation]
-timestep = 86400                   # optional (required if use_allocation = true), default 86400
-use_allocation = false             # optional, default false
+timestep = 86400                                 # optional (required if use_allocation = true), default 86400
+use_allocation = false                           # optional, default false
+default_source_priority_user_demand = 1000       # optional, default 1000
+default_source_priority_boundary = 2000          # optional, default 2000
+default_source_priority_level_demand = 3000      # optional, default 3000
+default_source_priority_flow_demand = 4000       # optional, default 4000
+default_source_priority_subnetwork_inlet = 5000  # optional, default 5000
```
evetion, line 30: I would make it a substructure here, called `source_priority` or similar, and removing `default` from the name.

evetion (commented):
Can you describe the design in this PR? From what I remember we need a configurable priority per nodetype only? And if we needed to override that (not sure), there can be a column in the Node table.

Given that context, why do we need a `python/ribasim/ribasim/allocation/source_priorities.py`?

SouthEndMusic:
> Can you describe the design in this PR?

@evetion It's in the issue: https://github.com/Deltares/Ribasim/issues/1505

evetion:
> > Can you describe the design in this PR?
> 
> @evetion It's in the issue: #1505

It's still nice to summarize what the changes are in a PR description (and later commit message). Can be very short, as in rename `field` to `demand_field`, adds table X, adds config option, especially given the number of files changed here.

SouthEndMusic:
A remark on tests that were changed: In some cases the allocated amounts actually changed because a change in order of sources and how they are clustered together. But more generally, it is a bad idea to test based on obtaining the values from the variables (e.g. `allocation_model.problem[:F]`) directly, since these are the values from the last optimization only, and in particular only for one source priority. It is better to look at `allocation_model.flow`, which is cumulative over all flows for a certain demand priority.

docs/reference/usage.qmd
```diff
@@ -60,7 +60,8 @@ While physically incorrect, it is useful for a first correctness check on a mode
 ## Allocation settings
 Currently there are the following allocation settings:
 - `use_allocation`: A boolean which says whether allocation should be used or not;
-- `timestep`: a float value in seconds which dictates the update interval for allocations.
+- `timestep`: A float value in seconds which dictates the update interval for allocations;
+- `default_source_priority`: An integer per source type for the allocation algorithm: `user_demand`, `boundary`, `level_demand`, `flow_demand`, `subnetwork_inlet`. Flow boundaries and level boundaries are combined in the single category `boundary`. By default all nodes of the same type have the same source priority, so that all source nodes of the same node type within a subnetwork are optimized for at once.
```
evetion, line 64: Why do we share the flow and level boundaries? Can't we distinguish between them?

python/ribasim/ribasim/config.py
```diff
@@ -173,12 +188,15 @@ class Node(pydantic.BaseModel):
         An optional name of the node.
     subnetwork_id : int
         Optionally adds this node to a subnetwork, which is input for the allocation algorithm.
+    source_priority : int
+        Optionally adds a source priority to this node, which is input for the allocation algorithm.
```
evetion, line 192:
```suggestion
        Optionally overrides the source priority for this node, which is used in the allocation algorithm.
```

core/src/util.jl
```diff
@@ -358,35 +358,40 @@ function is_main_network(subnetwork_id::Int32)::Bool
...
-        priority_col = Int32.(coalesce.(priority_col, Int32(0)))
-        if valid_priorities(priority_col, config.allocation.use_allocation)
-            union!(priorities, priority_col)
+
+    for name in names(Ribasim; all = true)
+        type = getfield(Ribasim, name)
+        if !(
+            (type isa DataType) &&
+            type <: Legolas.AbstractRecord &&
+            hasfield(type, :demand_priority)
```
evetion, line 370: While I appreciate this refactor, I'd say the previous code was more readable. I now have to look up all tables with `demand_priority` to get the same understanding. Just a general remark, it's a fine line between preventing duplication bugs and readability.

core/src/read.jl
```diff
@@ -246,7 +246,102 @@ const conservative_nodetypes = Set{NodeType.T}([
...
+    for source_type in source_types
+        if source_type == :boundary
+            default_priority = default_source_priority.boundary
+            default_source_priority_dict[:flow_boundary] = default_priority
+            default_source_priority_dict[:level_boundary] = default_priority
+        else
+            default_source_priority_dict[source_type] =
+                getproperty(default_source_priority, source_type)
+        end
+    end
```
evetion, line 274: This seems a bit convoluted, how about?
```julia
default_source_priority_dict = Dict{Symbol, Int32}(key => getfield(s, key) for key in propertynames(s))
default_source_priority_dict[:flow_boundary] = default_source_priority.boundary
default_source_priority_dict[:level_boundary] = default_source_priority.boundary
```
(if needed, followed by a delete of :boundary)

core/src/read.jl
```diff
@@ -246,7 +246,102 @@ const conservative_nodetypes = Set{NodeType.T}([
...
+                is_source = true
+            else
+                if row.subnetwork_id != 1 # Not in the main network
+                    for id in inflow_ids(p.graph, only(node_ids))
+                        if graph[id].subnetwork_id == 1 # Connects to the main network
+                            is_source = true
+                            source_priority =
+                                default_source_priority_dict[:subnetwork_inlet]
+                            source_type = AllocationSourceType.subnetwork_inlet
+                            break
```
evetion, line 316: This might be the deepest nested code in our codebase! 🎉 

Can we filter the table earlier (maybe even in SQL?) and fill in the defaults in a separate loop?

evetion (approved):
Nice work! 💯 Could you comment on your approach in general?

I'm having a hard time to understand the refactor, mostly because of the rename with AllocationSourceType (going from the old boundary_node basin main_to_sub user_return buffer to user_demand, boundary, level_demand, flow_demand and subnetwork_inlet?). It would be good to document that. Same for the AllocationSource functions using the Val.

The changes in the tests slightly worry me, do you/we fully understand the changes in the source order? Are there other tests that assert correct behavior (if any)?

SouthEndMusic:
@evetion a comment on the approach in general:

In Ribasim Python, source priorities can now be added via the `source_priority` field of the `Node` class passed to the add API. This optional information is then added to the new `source_priority` column of the `Node` table. To void confusion, the old `priority` was renamed to `demand_priority`.

In the core the default or overridden source priorities are assigned to the source nodes in each subnetwork. Currently the sources within a subnetwork with the same priority are solved for at the same time which yield (more) situations with non-uniqueness of an optimal solution.

PR #2023, opened 2025-01-29
Rename Edge to Link

evetion:
*everywhere*

This is automatic rename, and then slowly staging the changes. Main exceptions found so far:
- UGrid uses edge_dimension and similar names
- Networkx has a Graph edges attribute and methods
- Graphs.jl has an edges method and keywords

@visr The only breaking change are that output arrow tables now have `link_id` instead of `edge_id`. I don't think there's a way around that.


visr:
This is pushing the link. An edge to the past.

visr (approved):

PR #2024, opened 2025-01-30
Pixi: use default as our dev environment

visr:
Going the opposite way of #2020. This merges our dev environment into the default environment. The only cost is that our Python unit test CI will be running in a larger environment, but caching helps and this was already one of the quickest CI runs.


evetion (approved):
This simplifies the (number of) environment(s), saving you time right?

visr:
Not really saving time, but mostly saving us from having to think about environments and remembering to select dev when adding packages or loading vscode.

There was just not so much of a benefit of using dev instead of default.

PR #2025, opened 2025-01-31
Fix link_type migration

visr:
I couldn't write migrated models to disk because the validation failed because the Link table had `link_type` twice. One version was filled in with defaults on load since it was missing, and the second one came from the migration code.

This removes link_type before renaming, just like was done for min_crest_level, and uses a new method to capture this pattern and test it.

### DIFF
--- a/python/ribasim/ribasim/migrations.py
+++ b/python/ribasim/ribasim/migrations.py
@@ -7,6 +7,16 @@ from pandas import DataFrame
 # Do the same for write_schema_version in ribasim_qgis/core/geopackage.py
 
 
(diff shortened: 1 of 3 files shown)


evetion:
Hmm, I'm surprised none of the existing migration tests picked up this pattern? Would be good to add that.

visr:
The migration didn't fail, it only failed validation when writing it out. I added decbc872904000dbb9f5c945846ae354667ab847 which would've failed and now shouldn't.

evetion (approved):
Nice, thanks for fixing the last bugs in my rename.

PR #2047, opened 2025-02-09
Allow ruff-vscode to fix notebooks again

visr:


pixi.toml
```diff
@@ -79,7 +79,7 @@ test-ribasim-core = { cmd = "julia --project=core --eval 'using Pkg; Pkg.test()'
 test-ribasim-migration = { cmd = "pytest --numprocesses=4 -m regression python/ribasim/tests" }
 test-ribasim-core-cov = { cmd = "julia --project=core --eval 'using Pkg; Pkg.test(coverage=true, julia_args=[\"--check-bounds=yes\"])'", depends-on = [
     "generate-testmodels",
-] }
+], env = { SSL_CERT_DIR = "", JULIA_SSL_CA_ROOTS_PATH = "" } }
```
visr, line 82: Unrelated, but this is needed to be able to use branches in our Manifest.toml, found in #2046.

.vscode/settings.json
```diff
@@ -2,17 +2,17 @@
...
         "editor.formatOnSave": true
     },
-    "notebook.formatOnSave.enabled": false,
+    "notebook.formatOnSave.enabled": true,
     "notebook.codeActionsOnSave": {
-        // https://github.com/astral-sh/ruff-vscode/issues/593
-        "source.fixAll.ruff": "never",
-        "source.organizeImports.ruff": "never"
+        "notebook.source.fixAll": "explicit",
+        "notebook.source.organizeImports": "explicit"
```
visr, line 8: This was disabled due to the linked issue. Prefixing `notebook` as suggested there fixes the problem for me. The postfix ruff is removed, It is a way for [ruff-vscode](https://github.com/astral-sh/ruff-vscode) to avoid allowing other extensions to run on-save. That is not the default way to do it and I don't think there is an issue with that.

.vscode/settings.json
```diff
@@ -2,17 +2,17 @@
...
     },
     "[python]": {
         "editor.defaultFormatter": "charliermarsh.ruff",
         "editor.formatOnSave": true,
         "editor.codeActionsOnSave": {
-            "source.fixAll": "explicit"
+            "source.fixAll": "explicit",
+            "source.organizeImports": "explicit"
```
visr, line 15: Lets do this as well, rather than wait for pre-commit to do it.

ruff.toml
```diff
@@ -1,5 +1,4 @@
 exclude = ["ribasim_qgis/tomllib/*"]
-extend-include = ["*.ipynb"]
```
visr, line 2: This has been the default since ruff 0.6: https://astral.sh/blog/ruff-v0.6.0#jupyter-notebooks-are-now-linted-and-formatted-by-default

pixi.toml
evetion: Could this break other uses cases (TC?) and if needed, for branches, it would probably be required for all Julia related tasks, not just core-cov?

evetion (approved):

pixi.toml
visr: No this is just when Pkg uses LibGit2 to checkout a branch. The other tasks that may need this already have this set. We can remove this again when https://github.com/JuliaLang/julia/pull/56924 is merged.

Issue #1936, opened 2024-11-14
Place and improve allocation test model WiBo in testbank


gijsber:
Create model Josje (Wibo) and build towards testmodel in testbank. Testmodels in this link: https://publicwiki.deltares.nl/download/attachments/232326570/Bommelerwaard_modellen_gebruikt_voor_berekeningen.zip?version=1&modificationDate=1720040275749&api=v2

PR #2056, opened 2025-02-12
Bommelerwaard test model

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1936.

Notes:
- I had to remove the truth states with "U" and "D" as we no longer support these (also removed this from the docs)
- We no longer support ManningResistance connecting to LevelBoundary so I converted one ManningResistance to LinearResistance (node 29)
- Also updated the Link table in the docs (`usage.md`) as this still mentioned node types
- I had to refactor handling sources for allocation a bit because this model hits an edge (link) case I hadn't considered yet


SouthEndMusic:
The model looks like this:

image

visr (approved):

PR #2057, opened 2025-02-13
Implement model equality

### LINKED ISSUE #928: Testing model equality
**What**
Testing model equality in ribasim Python should work, e.g.
```python
from ribasim_testmodels import basic_model, basic_transient_model

model_1 = basic_model()
model_2 = basic_transient_model()

assert model_1 != model_2
```

**Why**
This is probably never relevant for modelers, but it is useful in tests. I attempted to use this in the test `test_model_merging` (https://github.com/Deltares/Ribasim/pull/914) but could not get it to work.

**How**
The most naive implementation of this is to add a method like this:

```python
def __eq__(self, other):
   for field in self.fields():
      if getattr(self, field) != getattr(self, other):
         return False

   return True
```

to the `Model` and `Node` classes (equality testing of tables is already implemented in https://github.com/Deltares/Ribasim/pull/914 and seems to work well). However, this leads to unbounded recursion errors.

evetion:
Fixes #928
Partially solves https://github.com/Deltares/Ribasim-NL/issues/229

Differences, and such will follow in a separate PR.

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -128,13 +128,13 @@ class BaseModel(PydanticBaseModel):
             getter = (
                 operator.itemgetter(*model_fields)
                 if model_fields
-                else lambda _: pydantic._utils._SENTINEL
+                else lambda _: pydantic._utils._SENTINEL  # type: ignore
             )
             try:
                 return getter(self.__dict__) == getter(other.__dict__)
             except KeyError:
-                self_fields_proxy = pydantic._utils.SafeGetItemProxy(self.__dict__)
-                other_fields_proxy = pydantic._utils.SafeGetItemProxy(other.__dict__)
+                self_fields_proxy = pydantic._utils.SafeGetItemProxy(self.__dict__)  # type: ignore
+                other_fields_proxy = pydantic._utils.SafeGetItemProxy(other.__dict__)  # type: ignore
                 return getter(self_fields_proxy) == getter(other_fields_proxy)
 
         else:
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/input_base.py
```diff
@@ -90,6 +93,53 @@ def _fields(cls) -> list[str]:
...
+        if isinstance(other, BaseModel):
+            self_type = self.__pydantic_generic_metadata__["origin"] or self.__class__
+            other_type = (
+                other.__pydantic_generic_metadata__["origin"] or other.__class__
+            )
+
+            # Perform common checks first
+            if not (
+                self_type == other_type
+                # This comparison has been removed, otherwise we recuse because
```
visr, line 110:
```suggestion
                # This comparison has been removed, otherwise we recurse because
```

python/ribasim/ribasim/input_base.py
```diff
@@ -90,6 +93,53 @@ def _fields(cls) -> list[str]:
...
+                operator.itemgetter(*model_fields)
+                if model_fields
+                else lambda _: pydantic._utils._SENTINEL  # type: ignore
+            )
+            try:
+                return getter(self.__dict__) == getter(other.__dict__)
+            except KeyError:
+                self_fields_proxy = pydantic._utils.SafeGetItemProxy(self.__dict__)  # type: ignore
+                other_fields_proxy = pydantic._utils.SafeGetItemProxy(other.__dict__)  # type: ignore
+                return getter(self_fields_proxy) == getter(other_fields_proxy)
```
visr, line 138: Could you add some comments here, because I can't follow.
If their `__dict__` is not equal, they can still be equal?

visr (approved):
Great! Left a few comments, but looks good otherwise.

python/ribasim/ribasim/input_base.py
evetion: You're reviewing Pydantic code, let's not do that. I will clarify this with a better (license) header.

PR #2063, opened 2025-02-13
Use PAT instead of default token to trigger automatic PRs.

### LINKED ISSUE #1996: Trigger Actions on update/pixi-lock
I missed #1994 in #1992 since I now see only TeamCity was triggered but not GitHub Actions. Any ideas what goes wrong? We already select the branch name `update/pixi-lock`:

https://github.com/Deltares/Ribasim/blob/245c7cafd251bc1c63b33afb54d852a778f2752d/.github/workflows/python_tests.yml#L2-L8

evetion:
Fixes #1996 again. See https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

This PAT is valid for 1 year (!)

### DIFF
--- a/.github/workflows/core_testmodels.yml
+++ b/.github/workflows/core_testmodels.yml
@@ -1,7 +1,7 @@
 name: Julia Run Testmodels
 on:
   push:
-    branches: [main, "update/pixi-lock", "update/julia-manifest"]
+    branches: [main, update/pixi-lock, update/julia-manifest]
     paths-ignore: [".teamcity/**"]
     tags: ["*"]
   pull_request:
--- a/.github/workflows/core_tests.yml
+++ b/.github/workflows/core_tests.yml
@@ -1,7 +1,7 @@
 name: Julia Tests
 on:
   push:
-    branches: [main, "update/pixi-lock", "update/julia-manifest"]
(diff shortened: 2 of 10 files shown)


visr (approved):
This works, thanks. I added two commits.
Also see https://github.com/Deltares/Ribasim/pull/2064#issuecomment-2656962601 in case you want to reconfigure that.

evetion:
> This works, thanks. I added two commits. 

Given that the Github documentation uses quotes for branches with slashes, let's not revert that commit? Or at least I don't see the need for it.

> Also see [#2064 (comment)](https://github.com/Deltares/Ribasim/pull/2064#issuecomment-2656962601) in case you want to reconfigure that.

PAT are the default by Github, we can find another machine account that configures a PAT (and have it as author). There are workarounds, like Github apps, or manually approving a (draft) PR to trigger the subsequent actions, but these all feel convoluted, and this just works as is.

PR #2065, opened 2025-02-13
Update to OteraEngine v1.1

visr:
This removes the newline option, either CRLF or LF works.

This leaves out the other default options and shortens the paths.
Better fix than #2054.


evetion (approved):

visr:
Yes indeed, I tested locally and it produces no diff. This mostly serves to verify to the OteraEngine developer that the fix for our bug report helped, and clean it up a little.

PR #2066, opened 2025-02-13
Format all the YAMLs

visr:
And keep them formatted with a pre-commit hook.
In the default hooks there was already a check-yaml but that only verifies the syntax.
Uses https://github.com/google/yamlfmt.


evetion (approved):
More auto more good

PR #2069, opened 2025-02-14
Make it easier to review Julia manifest updates

visr:
This should put the output of `] update` and `status --outdated` in the PR body, so we can easily see what changed and what is stuck.

Without this it looks like #1990.
With this it looks like [#2070](https://github.com/Deltares/Ribasim/pull/2070).

### DIFF
--- a/.github/workflows/julia_auto_update.yml
+++ b/.github/workflows/julia_auto_update.yml
@@ -27,5 +27,5 @@ jobs:
           branch: update/julia-manifest
           title: Update Julia manifest
           commit-message: "Update Julia manifest"
-          body-path: .pixi/update-manifest-julia.md.md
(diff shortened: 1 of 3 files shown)


evetion (approved):
Does it make sense to have a Pkg.status first? Now it's a huge dump of text, and a the updates that can't be done.

visr:
Yeah the status should be computed after the update, but we can print it first. Ideally we filter the update output more to show only the changes, and any possible errors, but not the installation and precompilation.

https://github.com/Deltares/Ribasim/pull/2070

visr:
This looks pretty ok now, see the updated text in #2070. Not as nice as the Pixi update, but don't have time for that now.

PR #2076, opened 2025-02-17
Convert autodiff Boolean to ADTypes

### LINKED ISSUE #2029: Start using ADTypes rather than Bool for autodiff
I noticed in the test logs a new deprecation warning:

```
┌ Warning: Using a `Bool` for keyword argument `autodiff` is deprecated. Please use an `ADType` specifier.
│   caller = _process_AD_choice(ad_alg::Bool, chunksize::Val{0}, diff_type::Val{:forward}) at misc_utils.jl:147
└ @ OrdinaryDiffEqCore ~/.julia/packages/OrdinaryDiffEqCore/33WQj/src/misc_utils.jl:147
```

See https://github.com/SciML/ADTypes.jl.
In our config we have `autodiff = true/false`. I guess we'd keep that for now and just map `true` to `AutoForwardDiff`.

visr:
This should fix the deprecation warning, that does this as well:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/5e78c0ac9403e5c9110905ac87c66670832abfd4/lib/OrdinaryDiffEqCore/src/misc_utils.jl#L134-L148

Fixes #2029.

### DIFF
--- a/core/Project.toml
+++ b/core/Project.toml
@@ -5,6 +5,7 @@ manifest = "../Manifest.toml"
 version = "2025.1.0"
 
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
 BasicModelInterface = "59605e27-edc0-445a-b93d-c09a3a50b330"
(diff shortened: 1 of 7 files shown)


core/Project.toml
```diff
@@ -55,6 +56,7 @@ Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 RibasimMakieExt = ["Makie", "DataFrames"]
 
 [compat]
+ADTypes = "1.13"
```
SouthEndMusic, line 59: alphabeticly ordered?

SouthEndMusic (approved):
Looks good (y)

core/Project.toml
visr: Upper case is sorted before lowercase -_-
This is where Pkg wants it.

PR #2077, opened 2025-02-18
Use StateVector instead of ComponentVector

visr:
This gets rid of our main use of ComponentArrays in favor of a minimal Ribasim state specific replacement called StateVector. ComponentArrays is a convenient package, but after recent breakage I wondered if it isn't better to make a more minimal and more static alternative for the state vector. IMO this leads to more readable code, and better stacktraces, at the cost of defining StateVector ourselves. Other ComponentArrays uses have mostly been replaced already by normal struct-of-arrays. The state needs to be an `AbstractVector{T}` and has clear components that all should be contiguous.

There are still a few other uses that need replacing.

This is a draft because I haven't managed to get SparseConnectivityTracer working with this properly. It already helped to support `similar(u, Int)`, but during tracing `du` stays a StateVector, but `u` becomes a Vector.

```jl
ERROR: MethodError: no method matching water_balance!(::StateVector{…}, ::Vector{…}, ::Ribasim.Parameters{…}, ::Float64)
The function `water_balance!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  water_balance!(::StateVector, ::StateVector, ::Ribasim.Parameters, ::Number)
   @ Ribasim a:\repo\ribasim\Ribasim\core\src\solve.jl:4

Stacktrace:
 [1] (::Ribasim.var"#251#252"{…})(du::StateVector{…}, u::Vector{…})
   @ Ribasim a:\repo\ribasim\Ribasim\core\src\util.jl:795
 [2] trace_function(::Type{…}, f!::Ribasim.var"#251#252"{…}, y::StateVector{…}, x::StateVector{…})
   @ SparseConnectivityTracer a:\.julia\packages\SparseConnectivityTracer\aHOLr\src\trace_functions.jl:56
 [3] _jacobian_sparsity(f!::Function, y::StateVector{…}, x::StateVector{…}, ::Type{…})
   @ SparseConnectivityTracer a:\.julia\packages\SparseConnectivityTracer\aHOLr\src\trace_functions.jl:87
 [4] jacobian_sparsity
   @ a:\.julia\packages\SparseConnectivityTracer\aHOLr\src\adtypes_interface.jl:60 [inlined]
 [5] get_jac_prototype(du0::StateVector{…}, u0::StateVector{…}, p::Ribasim.Parameters{…}, t0::Float64)
   @ Ribasim a:\repo\ribasim\Ribasim\core\src\util.jl:794
 [6] Ribasim.Model(config::Ribasim.config.Config)
   @ Ribasim a:\repo\ribasim\Ribasim\core\src\model.jl:135
```

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -99,7 +99,7 @@ using ComponentArrays: ComponentVector, ComponentArray, Axis, getaxes
 using Dates: Dates, DateTime, Millisecond, @dateformat_str
 
 # We customize broadcasting for our StateVector.
-using Base.Broadcast: Broadcasted, ArrayStyle
+using Base.Broadcast: Broadcasted, ArrayStyle, Extruded
 
 # Callbacks are used to trigger function calls at specific points in the similation.
 # E.g. after each timestep for discrete control,
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -98,6 +98,9 @@ using ComponentArrays: ComponentVector, ComponentArray, Axis, getaxes
 # internally we use a Float64; seconds since the start of the simulation.
 using Dates: Dates, DateTime, Millisecond, @dateformat_str
 
+# We customize broadcasting for our StateVector.
+using Base.Broadcast: Broadcasted, ArrayStyle
+
 # Callbacks are used to trigger function calls at specific points in the similation.
 # E.g. after each timestep for discrete control,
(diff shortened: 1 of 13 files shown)


visr:
I'll close this. It was working but this needed more overloads for StructVector than I was comfortable maintaining going forward. I think it's better to stick with ComponentArrays for the state, which is widely used and tested in SciML packages. I'm making another small PR to bring some of the smaller refactors here into main.

Issue #744, opened 2023-11-07
DiscreteControl with time


harm-nomden-sweco:
We were discussing the implementation of pumping stations (pumps). Is it possible to change the "greater than" levels for the DiscreteControl over time (e.g. summer/winter? 
Otherwise we might need to go for PidControl at several locations, PidControl does have a time element.

evetion:
- [ ] Check whether dynamic control states are needed (instead of just >)
- [ ] Check interpolation behaviour (forward seems more likely than linear)

The control callback can interpolate a timeseries of greater thans without problem. We propose adding an optional column time to Control / Condition. If left empty, it's set to the starttime, (dummy timeseries) and extrapolated.

PR #2079, opened 2025-02-19
Time dependent thresholds

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/744

Notable changes:
- A mandatory `condition_id` and an optional `time` column were added to the `DiscreteControl / condition` table. Sorting of this table is no longer based on `node_id, compound_variable_id, greater_than` but on `node_id, compound_variable_id, condition_id`
- The time points of the `greater_than` timeseries were added to the `tstops`


python/ribasim/ribasim/schemas.py
```diff
@@ -225,6 +225,12 @@ class DiscreteControlConditionSchema(_BaseSchema):
     greater_than: Series[Annotated[pd.ArrowDtype, pyarrow.float64()]] = pa.Field(
         nullable=False
     )
+    condition_id: Series[Annotated[pd.ArrowDtype, pyarrow.int32()]] = pa.Field(
```
evetion, line 228: Maybe this should be the first field in this schema? Or second, after node_id? Not sure here, maybe sort order?

python/ribasim/ribasim/migrations.py
```diff
@@ -118,3 +118,16 @@ def migration_func(df: DataFrame, schema_version: int) -> DataFrame:
...
+def discretecontrolconditionschema_migration(
+    df: DataFrame, schema_version: int
+) -> DataFrame:
+    if schema_version < 5:
+        warnings.warn(
+            "Migrating outdated DiscreteControl / condition table.", UserWarning
+        )
+        n_rows = len(df)
+        df["time"] = Series([None] * n_rows, dtype="timestamp[ms][pyarrow]")
+        df["time"] = Series(range(n_rows), dtype="int32[pyarrow]")
```
evetion, line 132:
```suggestion
        df["condition_id"] = Series(range(1,n_rows+1), dtype="int32[pyarrow]")
```

core/test/run_models_test.jl
```diff
@@ -366,17 +366,19 @@ end
...
-    (; u, p, t, sol) = integrator
+    (; du, u, p, t, sol) = integrator
     current_storage = p.basin.current_properties.current_storage[Float64[]]
 
     day = 86400.0
 
     @test only(current_storage) ≈ 1000.0
     # constant UserDemand withdraws to 0.9m or 900m3 due to min level = 0.9
     BMI.update_until(model, 150day)
+    formulate_storages!(current_storage, u, u, p, t)
```
evetion, line 377: Can you explain why these changes are now needed?

core/src/callback.jl
```diff
@@ -469,75 +469,54 @@ end
...
 - The nodes that are controlled by this DiscreteControl node must have the same control state, for which they have
     parameter values associated with that control state defined in their control_mapping
 """
 function apply_discrete_control!(u, t, integrator)::Nothing
     (; p) = integrator
     (; discrete_control) = p
-    (; node_id) = discrete_control
+    (; node_id, truth_state, compound_variables) = discrete_control
     du = get_du(integrator)
-    water_balance!(du, u, p, t)
```
evetion, line 488: Can you explain why this is not needed anymore?

core/src/read.jl
```diff
@@ -1805,10 +1846,15 @@ function load_structvector(
         nt = merge(
             nt,
             (;
-                time = DateTime.(
-                    replace.(nt.time, r"(\.\d{3})\d+$" => s"\1"),  # remove sub ms precision
-                    dateformat"yyyy-mm-dd HH:MM:SS.s",
-                )
+                time = map(
+                    val ->
+                        ismissing(val) ? DateTime(0) :
```
evetion, line 1851: Do we want to replace 0 with the starttime as defined in the config?

core/src/read.jl
```diff
@@ -883,81 +887,123 @@ function Basin(db::DB, config::Config, graph::MetaGraph)::Basin
...
 
+function get_greater_than!(greater_than, conditions_compound_variable, starttime)::Nothing
+    (; node_id) = first(conditions_compound_variable)
+    errors = false
+
+    for condition_group in
+        IterTools.groupby(row -> row.condition_id, conditions_compound_variable)
+        condition_group = StructVector(condition_group)
+
+        if any(ismissing, condition_group.time)
```
evetion, line 898: Can this happen, as we replace missings in the load_structvector? I think the check would be for multiple identical timesteps?

evetion (approved):
Nice work, also in terms of reducing/simplifying code used.

core/test/run_models_test.jl
SouthEndMusic: `current_storage` is a state derived variable. It's latest value depends on with which `u` `formulate_storages!` was called last. This depends on solver internals as wel as callbacks, so calling `formulate_storages!` here with the latest state means we also get the latest `current_storage`. This magically went well before without these explicit calls here, but not anymore.

PR #2080, opened 2025-02-20
Implement `diff` on Models.

evetion:
Fixes https://github.com/Deltares/Ribasim-NL/issues/229

Implements `diff` on Models and its children. Either gives back `None` when models are equal, or a (nested) dict with the differences. When the differences are not a DataFrame (like the toml config), the dict has `self` and `other` as key. For DataFrames we return a dict with `diff` as key, and a [datacompy](https://github.com/capitalone/datacompy) Comparison object.

```python
>>> nbasic == basic
False
>>> x = nbasic.diff(basic)
{'basin': {'node': {'diff': <datacompy.core.Compare object at 0x16e5a45c0>},
           'static': {'diff': <datacompy.core.Compare object at 0x16eb90080>}},
 'solver': {'saveat': {'other': 86400.0, 'self': 0.0}}}
>>> x["basin"]["static"]["diff"].report()
DataComPy Comparison
--------------------

DataFrame Summary
-----------------

  DataFrame  Columns  Rows
0      self        5     5
1     other        5     4

Column Summary
--------------

Number of columns in common: 5
Number of columns in self but not in other: 0 []
Number of columns in other but not in self: 0 []

Row Summary
-----------

Matched on: index
Any duplicates on match values: No
Absolute Tolerance: 0
Relative Tolerance: 0
Number of rows in common: 4
Number of rows in self but not in other: 1
Number of rows in other but not in self: 0

Number of rows with some compared columns unequal: 0
Number of rows with all compared columns equal: 4

Column Comparison
-----------------

Number of columns compared with some values unequal: 0
Number of columns compared with all values equal: 5
Total number of values which compare unequal: 0

Sample Rows Only in self (First 10 Columns)
-------------------------------------------

     node_id  drainage  potential_evaporation  infiltration  precipitation
fid
4         18      <NA>                   <NA>          <NA>            4.0

```

I've tried deepdiff and other diff plugins, but none seem to combine nested Pydantic models *and* DataFrames. Comparing DataFrames using the built-in `compare` only works for identical shapes (limiting it's use).

- [x] Ignore meta_ keys in DataFrames

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -20909,7 +20909,7 @@ packages:
 - pypi: python/ribasim
   name: ribasim
   version: 2025.1.0
-  sha256: bbdb286b50f624bd7d9b68d8c43e60b009e188e1d7dcbdd624f21698857a8635
+  sha256: 1beeb16859cc9084dbf4bdc93f5ff7da3e8c032bd82d0702db9218a396ee578e
   requires_dist:
   - datacompy>=0.16
   - geopandas>=1.0


visr (approved):
The diff looks good!

Issue #153, opened 2023-04-06
Cyclic forcing


visr:
Right now forcing tables look like this:

note_id | time | precipitation | potential_evaporation
---------             | ------- | ------------ | -----
1 | 2020-01-01 | 2.0 | 1.0
1 | 2020-02-01 | 2.1 | 1.1

Many parameters are highly cyclical, typically yearly, but possibly with a different period. It is worth considering if and how we should support this. @Huite suggested using the same forcing timeseries schema, but adding an option like `cycle = "year"` (or month/day/...).

If we support this, it will be easier for users with cyclic data to prepare the right input, and input file sizes will be smaller. It does mean some extra complexity in the interpretation of timeseries, e.g. we'd need a TOML to be able to interpret it.

For parameters that have 2 values per year, one for summer and one for winter, the reduction of filesize is probably not worth it since we support nonequidistant timeseries input, so only 2 rows per year are needed. We can also consider adding convenience methods in the preprocessing tooling instead, and stick to regular timeseries in the core.

visr:
This ideally would make use of DataInterpolations: https://github.com/SciML/DataInterpolations.jl/issues/360

evetion:
- [ ] Would be good to check all current extrapolation methods in core

Given that the complexity of extrapolation is done in DataInterpolations, it adds relatively little complexity to do this in the Julia (and doing so in Python would be more complex).

Given that this needs to be set on individual nodes (different cycles for different userdemands), the only table that fits is the Node table, where we suggest a `cyclic_time` as a optional boolean column. A prerequisite of implementing this is watching the movie Tennet.

visr:
This needs #2019.

PR #2081, opened 2025-02-20
Add test model with cyclic forcing

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/153.

### DIFF
--- a/python/ribasim/ribasim/config.py
+++ b/python/ribasim/ribasim/config.py
@@ -190,6 +190,8 @@ class Node(pydantic.BaseModel):
         Optionally adds this node to a subnetwork, which is input for the allocation algorithm.
     source_priority : int
         Optionally overrides the source priority for this node, which is used in the allocation algorithm.
+    cyclic_forcing : bool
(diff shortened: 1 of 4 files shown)


visr:
Looks good, but there are still some tests to fix it seems:

```
FAILED python\ribasim\tests\test_io.py::test_sort - AssertionError: assert ['node_id', '...condition_id'] == ['node_id', '...greater_than']
  
  At index 2 diff: 'condition_id' != 'greater_than'
  
  Full diff:
    [
        'node_id',
        'compound_variable_id',
  -     'greater_than',
  +     'condition_id',
    ]
= 1 failed, 53 passed, 1 skipped, 1 xfailed, 1 xpassed, 30 warnings in 58.63s =
```

Does this need to merge master?

SouthEndMusic:
@visr this is because I did a small follow-up of https://github.com/Deltares/Ribasim/pull/2079 where the sorting of the `DiscreteControl / condition` table was changed, and apparently we have 2 sources of truth for those (one in Python and one in Julia). I wasn't aware this was part of a test, will quickly fix.

Issue #1942, opened 2024-11-18
Reduce CLI latency


visr:
In a REPL the `trivial` test model runs in about 0.03 seconds on the second run, but 20 seconds with the CLI. We put a lot of effort into making the solver fast to run. Little effort has been spent to quickly start a simulation. Users running the `trivial` test model from the CLI (v2024.11) need to wait 20 seconds each time. For short simulations or quick tests this hinders the iteration speed of modelers. Here is the output of [hyperfine](https://github.com/sharkdp/hyperfine) on `trivial`:

```
❯ hyperfine 'ribasim ribasim.toml'
Benchmark 1: ribasim ribasim.toml
  Time (mean ± σ):     20.835 s ±  0.280 s    [User: 10.951 s, System: 0.364 s]
  Range (min … max):   20.606 s … 21.525 s    10 runs
```

Especially on a warm filesystem cache the loading of the binary is quick, `Info: Starting a Ribasim simulation` appears quickly. It then takes about 10 seconds for `Simulating   0%` to appear, and another 10 seconds for the simulation to finish.

That indicates that it still needs to compile a lot of code during a CLI run. Ten seconds worth of initialization / validation code and ten seconds of simulation code. I hoped that https://github.com/Deltares/Ribasim/blob/v2024.11.0/build/precompile.jl would precompile more of what we need, but apparently it doesn't. We can consider looking into this with [SnoopCompile.jl](https://github.com/timholy/SnoopCompile.jl). Right now we don't yet use [PrecompileTools.jl](https://github.com/JuliaLang/PrecompileTools.jl) in Ribasim.jl. We could start doing this as well.

I assume in time having small binaries with [`juliac`](https://github.com/Deltares/Ribasim/issues/1632#issuecomment-2296107713) will also help here.

One way to completely avoid this issue is to run `Ribasim.main("ribasim.toml")` from Julia instead. If we'd support some way in the CLI to run multiple model (separate TOMLs) with one command this would similarly avoid the latency when doing multiple runs each time. It may be good to support separate TOMLs also for #47.

visr:
If we don't use the precompile file in PackageCompiler, the build is 6 minutes faster and only 6 MB smaller (than 313 MB) but the latency is about twice as high.

Testing a simple PrecompileTools workload like

```julia
using PrecompileTools: @compile_workload
@compile_workload begin
    main(normpath(@__DIR__, "../../generated_testmodels/basic/ribasim.toml"))
end
```

[...]

PR #2086, opened 2025-02-21
Store state node IDs in parameters

visr:
So we can remove `id_from_state_index`.

### DIFF
--- a/core/src/parameter.jl
+++ b/core/src/parameter.jl
@@ -1003,6 +1003,8 @@ const ModelGraph = MetaGraph{
     const water_balance_reltol::Float64
     # State at previous saveat
     const u_prev_saveat::C11 = ComponentVector()
+    # Node ID associated with each state
+    const node_id::Vector{NodeID} = NodeID[]
 end
 
 # To opt-out of type checking for ForwardDiff
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -1700,19 +1700,17 @@ function Parameters(db::DB, config::Config)::Parameters
 
     subgrid = Subgrid(db, config, basin)
 
-    # aligned with the state
-    node_id = vcat(
-        tabulated_rating_curve.node_id,
-        pump.node_id,
-        outlet.node_id,
-        user_demand.node_id,
-        user_demand.node_id,
-        linear_resistance.node_id,
-        manning_resistance.node_id,
-        basin.node_id,
-        basin.node_id,
-        pid_control.node_id,
-    )
+    u_ids = state_node_ids((;
+        tabulated_rating_curve,
+        pump,
+        outlet,
+        user_demand,
+        linear_resistance,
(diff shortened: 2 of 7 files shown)


visr:
These are scraps from #2077. Thinking about this, and having to deal a bit more with huge stacktraces from the large ComponentArrays types, made me consider another approach. We could store the `UnitRange{Int64}` of each state component in the parameters, and use `view(u, p.is_pump)` instead of `u.pump`. Storing the resulting SubArray like in StateVector from #2077 can maybe also be done, but is more complex because you'd have to support Duals and GradientTracer as well as Float64.

I'm hoping this can also help with latency (#1942). Since ComponentArrays puts the full Axes into its type signature, it will specialize differently on each model. Because if the state length or component lengths differ, it needs to compile separate specializations.

```jl
# uses MethodAnalysis, f(x) can be anything
julia> specializations = methodinstances(f)
2-element Vector{Core.MethodInstance}:
 MethodInstance for f(::ComponentVector{Int64, Vector{Int64}, Tuple{Axis{(a = ViewAxis(1:3, Shaped1DAxis((3,))), b = ViewAxis(4:5, Shaped1DAxis((2,))))}}})
 MethodInstance for f(::ComponentVector{Int64, Vector{Int64}, Tuple{Axis{(a = ViewAxis(1:2, Shaped1DAxis((2,))), b = ViewAxis(3:5, Shaped1DAxis((3,))))}}})
```

Here the two ComponentVectors only differ slightly:
```jl
julia> using ComponentArrays
julia> a = ComponentVector(; a = [1,2,3], b = [4,5])
julia> b = ComponentVector(; a = [1,2], b = [3,4,5])
julia> typeof(a)==typeof(b)
false
```

Many methods that we call are likely to already be precompiled for simple `Vector{Float64}`, but none will be for our ComponentVector state (or StateVector).

core/src/parameter.jl
```diff
@@ -1003,6 +1003,8 @@ const ModelGraph = MetaGraph{
     const water_balance_reltol::Float64
     # State at previous saveat
     const u_prev_saveat::C11 = ComponentVector()
+    # Node ID associated with each state
```
SouthEndMusic, line 1006: Can you explain why you put this in the parameters instead of in the state object?

core/src/read.jl
```diff
@@ -1700,6 +1700,20 @@ function Parameters(db::DB, config::Config)::Parameters
 
     subgrid = Subgrid(db, config, basin)
 
+    # aligned with the state
+    node_id = vcat(
```
SouthEndMusic, line 1704: Is there a way to have a single source of truth for this order?

SouthEndMusic (commented):
Getting rid of `id_from_state_index` is nice. I left some questions.

SouthEndMusic (approved):

PR #2100, opened 2025-02-25
Remove PR template

visr:
This isn't really helping. I usually write a commit message with what it fixes and then I just need to remove this.
Plus there isn't always an issue that it fixes.


evetion (approved):
I give it half a year before we introduce a template again.

Issue #2094, opened 2025-02-24
Add cyclic timeseries support for demand nodes


SouthEndMusic:
As a follow-up of https://github.com/Deltares/Ribasim/pull/2081.

Add cyclic timeseries support for:
- `UserDemand`
- `FlowDemand`
- `LevelDemand`

PR #2102, opened 2025-02-25
Support and test cyclic demands

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2094


SouthEndMusic:
Cyclic `PidControl` is automatically supported via `parse_static_and_time`.

core/src/model.jl
```diff
@@ -97,27 +101,31 @@ function Model(config::Config)::Model
...
+            level_demand.max_level,
+            level_demand.min_level,
+            user_demand.demand_itp...,
+            user_demand.return_factor,
+            vcat(
+                [
+                    [cv.greater_than for cv in cvs] for
+                    cvs in discrete_control.compound_variables
+                ]...,
+            )...,
```
visr, line 128: Would be good to avoid this splatting. From the vcat docs:
```
To concatenate a large vector of arrays, reduce(vcat, A) calls an efficient method when A isa
  AbstractVector{<:AbstractVecOrMat}, rather than working pairwise.
```

visr (approved):

PR #2104, opened 2025-02-26
Run DataFrame validators ASAP.

### LINKED ISSUE #2101: Fix broken `DiscreteControl / condition` migration
Currently loading older models in Ribasim Python yield a Pydantic validation error in the `DiscreteControl / condition` table because of the missing mandatory `condition_id` column:

```
C:\Users\konin_bt\Repositories\Deltares\Ribasim\python\ribasim\ribasim\migrations.py:85: UserWarning: Migrating outdated DiscreteControl / variable table.
  warnings.warn(
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[18], line 1
----> 1 model = ribasim.Model.read(toml_path)

File ~\Repositories\Deltares\Ribasim\python\ribasim\ribasim\model.py:294, in Model.read(cls, filepath)
    292 if not Path(filepath).is_file():
    293     raise FileNotFoundError(f"File '{filepath}' does not exist.")
--> 294 return cls(filepath=filepath)

    [... skipping hidden 2 frame]

File ~\Repositories\Deltares\Ribasim\.pixi\envs\default\Lib\site-packages\pydantic\main.py:214, in BaseModel.__init__(self, **data)
    212 # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
    213 __tracebackhide__ = True
--> 214 validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
    215 if self is not validated_self:
    216     warnings.warn(
    217         'A custom validator is returning a value other than `self`.\n'
    218         "Returning anything other than `self` from a top level model validator isn't supported when validating via `__init__`.\n"
    219         'See the `model_validator` docs (https://docs.pydantic.dev/latest/concepts/validators/#model-validators) for more details.',
    220         stacklevel=2,
    221     )

ValidationError: 1 validation error for TableModel[DiscreteControlConditionSchema]
df
  Value error, column 'condition_id' in DataFrameSchema {'node_id': <Schema Column(name=node_id, type=DataType(int32[pyarrow]))>, 'compound_variable_id': <Schema Column(name=compound_variable_id, type=DataType(int32[pyarrow]))>, 'condition_id': <Schema Column(name=condition_id, type=DataType(int32[pyarrow]))>, 'greater_than': <Schema Column(name=greater_than, type=DataType(double[pyarrow]))>, 'time': <Schema Column(name=time, type=DataType(timestamp[ms][pyarrow]))>} requires a default value when non-nullable add_missing_columns is enabled [type=value_error, input_value=     node_id  compound_va...

[703 rows x 3 columns], input_type=DataFrame]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

This should be fixed by the migration function, but for some reason, the method where the migration should be called:

https://github.com/Deltares/Ribasim/blob/bfcb554f5dcdf4dfbf0f95c7e99fbe5032e68e08/python/ribasim/ribasim/input_base.py#L312-L321

is never called for this particular table.

evetion:
Fixes #2101

Also, the migration Series will match the index of the DataFrame, thus dropping the last id (range is zero based, whereas fid starts with 1).

I saw more errors locally, might be because the hws_model has been "fixed" but the version has not been set, throwing off our migration logic?


python/ribasim/ribasim/migrations.py
```diff
@@ -109,15 +114,21 @@ def outletstaticschema_migration(df: DataFrame, schema_version: int) -> DataFram
 for node_type in ["UserDemand", "LevelDemand", "FlowDemand"]:
     for table_type in ["static", "time"]:
 
-        def migration_func(df: DataFrame, schema_version: int) -> DataFrame:
+        def migration_func(
+            df: DataFrame,
+            schema_version: int,
+            node_type: str = node_type,
+            table_type: str = table_type,
```
evetion, line 121: @SouthEndMusic this fixes a late binding bug

visr (approved):
This looks good, thanks for the various fixes here.
So `_check_schema` wasn't called for some tables because the field_validator wasn't marked as mode "before"? Why did it do some but not all?

visr:
I used this branch to update the HWS integration model, and triggered the integration and regression tests, and both pass!

evetion:
> This looks good, thanks for the various fixes here. So `_check_schema` wasn't called for some tables because the field_validator wasn't marked as mode "before"? Why did it do some but not all?

I thought I answered this, but apparently not. For what it's worth, I believe the `_check_schema` wasn't called *yet*. Without a set order, I guess pandera validators came first (and errored). I will investigate this later, seems good to understand this better.

Issue #2060, opened 2025-02-13
Linear Basin implementation for allocation


SnippenE:
Introduce the piecewise linear approximation of the level - storage relationship
Introduce the storage conservation constraints

first step #2041

SouthEndMusic:
This has been implemented, but a proper tests of the 'phantom storage' and application of the Douglas Peucker algorithm must be added

Issue #1494, opened 2024-05-22
A linear objective function that achieves equal ratio allocation


SouthEndMusic:
## The objective function

@jarsarasty suggested the following linear objective function that also achieves equal ratio allocation.

Say we have demands $d_i$, allocated flows $F_i$, and the allocated fraction $f$. We want to achieve $F_i = f d_i$, and so we introduce the constraints

$$
u_i \ge F_i - f d_i, \quad l_i \ge f d_i -F_i
$$

where $u_i, l_i \ge 0$ are the upper error and the lower error respectively. Then to minimize the error we minimize

$$
\sum_{i \in I} u_i + l_i.
$$

This however penalizes absolute flow amounts, so if we want to penalize relative errors we either minimize

$$
\sum_{i \in I} \frac{u_i + l_i}{d_i},
$$

or use in stead of the above the constraints

$$
u_i \ge \frac{F_i}{d_i} - f, \quad l_i \ge f-\frac{F_i}{d_i}
$$

(which I think is nicest).

## My thoughts on this objective function

I like this concept a lot, because to me it is very explainable, more so than [the first objective function that achieves equal ratio allocation](https://github.com/Deltares/Ribasim/pull/1366). An added bonus that Jesús mentioned is that in general linear problems can be solved faster than quadratic problems.

The error behavior in this method is different from in the quadratic case, i.e. each incremental deviation from the target allocation is weighed equally. I think this makes the results also more explainable though, and we haven't had much feedback yet about how the optimization plays out when equal ratio allocation cannot be achieved.

The role of $f$ needs some more thought. In the current implementation this $f$ is not a priori known. We could compute it as (the minimum of $1$ and) the sum of the available source capacity over the sum of demands (within a priority). This however doesn't take the graph topology and constraining nodes into account. So ideally $0 \le f \le 1$ itself is also maximized, but then the question is how do we combine that with the objective above.

SnippenE:
Paused until need of lineair function is asked for by users

PR #2108, opened 2025-02-26
Piecewise linear basin profile approx in allocation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2060.

An additional complexity here is that because we change the way we represent basins in the allocation problems, level demand has to be reformulated as well. The previous approach was to have a 'buffer' attached to basins with a level demand which acted as a source when the level in the basin was too high, and as a `UserDemand` when the level in the basin was too low. This source or demand was computed with data from the physical layer.

This PR introduces the level of each basin explicitly as a variable in the optimization problems. Therefore we can optimize for the level directly. The error for other demand nodes are still formulated in terms of flow though, so the error for level demand should be formulated that way too, so all terms are treated the same way in the loss function. Therefore we formulate the error term for the level demand as

$$
\left| \frac{L_\text{end} - l_\text{target}}{\Delta t_\text{allocation}} \right|,
$$

where $L_\text{end}$ is the level of a particular basin at the end of the allocation timestep.

One thing to properly think trough here is what the behavior of a basin with a level demand should be when optimizing for a different priority than the priority of the level demand. I would say intuitively: for priorities higher than the level demand, all storage of the basin should be available as a source, and for the priority of the level demand and lower priorities only the disk of water above the maximum tolerated level should be available, if applicable.


SouthEndMusic:
Here's a bit of theoretical background to the goal of this PR.

For each basin, a table is given with levels and associated areas: $h_i, A_i \in \mathbb{R}, i = 1, \ldots n$. This data is sorted by level and the areas are strictly positive and non-decreasing. To get a continuous profile we interpolate this data linearly:

$$
A(h) = A_i + (A_{i+1} - A_i)\frac{h - h_i}{h_{i+1} - h_i}, \quad h \in [h_i, h_{i+1}].
$$

Then to get the volume (storage) of water at some level $\ell$ we integrate from the bottom, given by the first and lowest level in the table:

$$
   V(\ell) = \int_{h_1}^\ell A(h)\text{d}h,
$$

giving a piecewise quadratic storage - level relationship. This is the relationship we want to approximate linearly. In the code we actually use the inverse of this relationship (`storage_to_level`) which is more useful in the physical layer, which I implemented upstream and is described [here](https://docs.sciml.ai/DataInterpolations/stable/inverting_integrals/). In this PR I evaluate this relationship in an arbitrary `n_samples_per_interval = 5 equi-spaced points in each interval between data points.

The level can also increase above the last level $h_n$, for which we extend the linear area and quadratic storage increase from the last data interval in the physical layer. It seems however to do this in the allocation problem that we have to make some educated guess for the maximum level that is going to be attained at model initialization, but @visr says that in practice this is probably doable.

jarsarasty:
Hi @SouthEndMusic, 

Originally, the objective function of this problem is quadratic, and this pull request introduces **binary** decision variables, so your problem becomes a mixed-integer quadratic one. Unfortunately, there are no very good open source solvers for this type of discrete nonlinear optimization problem. In particular, _HiGHS cannot solve QP models where some of the variables must take integer values. You could try other free solvers like bonmin, but the solution will probably not be very fast.

Other alternatives, as discussed in our meeting last week, would be 1) to convert the objective function into a **linear** one (this is probably a no regret action), 2) to try to exploit the problem structure to **avoid the use of binary variables** (this will require a bit more investigation, and is not guaranteed to be achievable), or 3) try to solve the problem in stages, for example by first solving a simplified version with linear storage level tables, and then using that initial solution to solve a more detailed model of the storage curve (will require some effort to set up).

I would suggest trying first action 1), implemented as an option, so that the user can choose between the quadratic or linear objective. We could discuss this approach further and refine what has been suggested [here](https://github.com/Deltares/Ribasim/issues/1494).

visr:
It makes sense to try to use a linear objective function. I'd prefer to not support multiple objective functions unless there is a good reason for it, to keep the maintenance lower.

SouthEndMusic:
@jarsarasty thanks for your reply. I'll start implementing the linear objective function you suggested.

jarsarasty:
Hi @SouthEndMusic,

Indeed, the product of decision variables in the optimization model should be avoided because it introduces nonlinearity, which can significantly complicate the problem and make it harder to solve.

SouthEndMusic:
@jarsarasty this package looks very interesting for constructing linear approximations:

https://github.com/LICO-labs/PiecewiseLinApprox.jl

jarsarasty (commented):
Hi @SouthEndMusic,

These are my comments for now.

SouthEndMusic:
Superseded by https://github.com/Deltares/Ribasim/pull/2266.

Issue #2015, opened 2025-01-20
Remove bundled root project artifacts


visr:
From [v2024.11.0](https://github.com/Deltares/Ribasim/releases/tag/v2024.11.0) to [v2025.1.0](https://github.com/Deltares/Ribasim/releases/tag/v2025.1.0) the `ribasim_windows.zip` size grew from 230 MB to 313 MB.

We should investigate why and see if we can get it back to the old size. Presumably one of the Manifest.toml updates is responsible for this. The difference is entirely in the number of artifacts that get bundled, `share\julia\artifacts\` used to be 10 directories but is now 55. I assume we don't need all those.

PR #2113, opened 2025-02-27
Remove RibasimMakieExt

visr:
Everytime Ribasim needed to be recompiled and Makie was available we had to spend considerable time recompiling this extension. We avoid having to pay this price by moving it to `utils/plot.jl` instead.

```julia
using Ribasim

m = Ribasim.run("ribasim.toml")
include("../utils/plot.jl")
using CairoMakie
plot_basin_data(m)
```

Helps latency #1942 and we should check if this helps with #2015 (edit: it doesn't). CairoMakie is still in our root Manifest.toml since we use it in our docs.


visr:
I'll go ahead and merge since this makes it easier to investigate the latency. Already discussed this with @SouthEndMusic.

PR #2115, opened 2025-02-28
Replace ComponentVector with Vector

### LINKED ISSUE #2091: Use Vector instead of ComponentVector
For the state. Alternative to #2077, described in https://github.com/Deltares/Ribasim/pull/2086#issuecomment-2675500741.

visr:
Fixes #2091.
Not only does the state but also the state_inflow_link and state_outflow_link such that the dependency can be dropped altogether.

It is a bit more verbose to replace `du.evaporation` with

```julia
du_evaporation = view(du, state_ranges.evaporation)
```

We can potentially organize this a bit more still. I already did some refactoring in solve.jl because instead of passing state_range and other parts of p everywhere I found it easier and more consistent to just pass p.

This also moves some of the parameter fields that were set after p creation into the p constructor.


visr:
Promising, running all testmodels on Linux went from 17 minutes to 7.

core/src/solve.jl
```diff
@@ -194,11 +184,14 @@ end
 Smoothly let the evaporation flux go to 0 when at small water depths
 Currently at less than 0.1 m.
 """
-function update_vertical_flux!(basin::Basin, du::AbstractVector)::Nothing
+function update_vertical_flux!(du::AbstractVector, p::Parameters)::Nothing
```
SouthEndMusic, line 187:
```suggestion
function update_vertical_flux!(du::Vector, p::Parameters)::Nothing
```

core/src/solve.jl
```diff
@@ -754,21 +752,38 @@ function limit_flow!(
         tabulated_rating_curve,
         basin,
         allocation,
+        state_ranges,
     ) = p
 
+    u_tabulated_rating_curve = view(u, state_ranges.tabulated_rating_curve)
```
SouthEndMusic, line 758: This repetition is a bit sad

SouthEndMusic (approved):
Nice that:

- We can now use `u` instead of `parent(u)` for `ForwardDiff` cache 'indexing', at least for as long as we still support that
- We can get rid of some type parameters

core/src/solve.jl
visr: Indeed. We could make it a bit shorter, but I think that would hurt readability more.

Issue #2122, opened 2025-03-03
Fix integration test


visr:
The results have changed too much, and the speed as well. Running again to check if it was noise.

https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_IntegrationTest_Linux

PR #2125, opened 2025-03-03
Revert "Make min_upstream_level and max_downstream_level controllable…

visr:
… (#2088)"

This reverts commit c87c70f9674ad7332510c839f7677b79a95a9b29.

This caused a considerable slowdown of the HWS model. I think it is just because parameter updates are slow, and this does 3 instead of 1.

Related to https://github.com/Deltares/Ribasim/issues/2122


visr:
This wasn't just slow, but also changed the results. Good to revert this quickly.

Issue #1960, opened 2024-12-05
Document representing Basins


visr:
Before Ribasim becomes more widely used, it is good to consider feedback from new users. One such feedback is that some names we use may cause some confusion.

**Edge**
Why: The name Edge is commonly used in graph theory, but is not so common in hydrological modeling.
Alternatives: Link.

**Basin** 
Why: Basin's don't necessarily need to align with basins or watersheds. Basins always need some explanation of them representing (a part of a) water body with a certain storage and level.
Alternatives:
- Waterbody: We should try to think ahead if this would still make sense if we will add other node types representing water bodies, like https://github.com/Deltares/Ribasim/issues/1609 or reservoirs.
- LevelPool was mentioned by Joe Hughes. I see this used in Surface-Water Routing (SWR) docs, and see references like in [HEC-RAS docs](https://www.hec.usace.army.mil/confluence/rasdocs/ras1dtechref/6.0/performing-a-dam-break-study-with-hec-ras/inflow-flood-routing-a-through-reservoir/level-pool-routing).
- Reach, used in SWR and other software, makes it clear that it is a part of a river or stream.
- LevelStorage
- **Bucket** ==> preference (based on discussion team 16-1-2025)

## Implementation
These renames can be very breaking for users, which is a reason not to wait too long. Though doing them right now isn't great for https://github.com/Deltares/Ribasim-NL progress. Ribasim Python migration can do the renames, and potentially offer the old names as API aliases temporarily. The core could have a period in which it looks for the old table names if the new ones don't exist.

Note: Implementation can also have an impact on the BMI and thus the coupling with MFMS

Decided on 13-2-2025 to implement the renaming in better explenaition in documentation how the concept basin works and some example names that could be used as aliases (reservoir, lake, river reach, canal)

gijsber:
Proposal 16/1/2024

Basin --> Bucket
Edge-->Link

gijsber:
Since it might be challenging to get a proper name that works under all circumstances, I have the following suggestion (already discussed and approved by Edwin).

1. we introduce a NodeTypeAlias (which in the case of a Basin could then be Lake, Reservoir, Channel, Canal, Tunnel etc)
2. the available NodeTypeAliases for a specific NodeType are defined  in the toml file
4. the QGIS plugin recognizes NodeTypeAliases and shows them as a different icons on the map
5. the Nodes table get an extra field NodeTypeAlias to register how the user wanst the node to be represented

[...]

PR #2135, opened 2025-03-06
Update basin.qmd

Fati-Mon:
#1960

### DIFF
--- a/docs/reference/node/basin.qmd
+++ b/docs/reference/node/basin.qmd
@@ -2,9 +2,33 @@
 title: "Basin"
 ---
 
-The Basin is the central node in each schematization, since it is the only one that stores water.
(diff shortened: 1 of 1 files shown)


visr:
I feel like the paragraphs for reservoirs, lakes, river reaches and canals essentially all say the same. It makes sense, since they all require the same tables. So perhaps we should just list them to avoid duplicated texts?

For confluence, I don't think we generally want to add a Basin just to represent a confluence. The Basin in which multiple links flow acts as a confluence.

All water (and land) must be assigned to a single Basin. The size of a Basin can vary a lot. Typically we only need to split up Basins as needed to represent the network more accurately.

visr (approved):
Thanks!

Issue #1743, opened 2024-08-20
`ManningResistance` causes unexpected oscillations


SouthEndMusic:
In the `basic` model, the levels of the basins around the `ManningResistance` look like this with `saveat = 0`:

![plot](https://github.com/user-attachments/assets/f6f8c13e-852f-458a-90bc-602f3384f405)

These oscillations are damped by setting `k = 100.0` in `formulate_flow!` for the `ManningResistance`: 

![plot](https://github.com/user-attachments/assets/8be5b8e1-2ac6-44e4-9720-232ef75fbed8)

@Huite do you think that's a valid change? It relaxes the approximation of the absolute value function.

visr:
Oh nice find. To add some context, the basic model has only static parameters, so we expect just smoothly going to a dynamic equilibrium like in the lower figure. There is a constant inflow on the upstream side, so that level should be slightly higher the whole time.

SouthEndMusic:
It would be nice if we had a more structural solution than parameter fiddling, though. I think the problem (still) is that the flow is too sensitive to head difference changes around a head difference of 0.

Huite:
> @Huite do you think that's a valid change? It relaxes the approximation of the absolute value function

It's currently k=1000.0 right? I haven't looked at atan enough to know in at which $\Delta h$ it starts to make a difference.

Don't we use a quadratic sigmoid somewhere else?
That has the upside of being cheaper to compute and is more explicit in when it's "turned" on.

SouthEndMusic:
Not sure what a quadratic signmoid is or what this means:
> ... and is more explicit in when it's "turned" on.

What we want is to limit $\frac{\text{d}Q}{\text{d}\Delta h}$ which is essentially the inverse of the resistance used in `LinearResistance`. So we could make that an optional parameter which we enforce in the `ManningResistance` formulation (I would have to look into what's the best way to do that).

Huite:
I don't think it's a proper term, I mean something like this, more in general for reduction factors:

![image](https://github.com/user-attachments/assets/d1da4a57-d725-4864-8af3-3ea394af4800)

In our case, the Sf isn't naturally bounded (0, 1), so you specify when the (linear reduction) starts. But to get smooth gradients you adjust the function a bit, $\Omega$ control when you start adjusting. For linear relationship you'd only use $\Omega$.

And what I mean with "turning on/off", basically this (for arctan):

[...]

PR #2136, opened 2025-03-06
Decrease relaxed root coefficient to 1e-5

visr:
@ngoorden is looking at this: https://github.com/Deltares/Ribasim-NL/issues/258#issuecomment-2694895342

### DIFF
--- a/core/integration_test/hws_integration_test.jl
+++ b/core/integration_test/hws_integration_test.jl
@@ -22,7 +22,7 @@
 
     @testset "Results values" begin
         @test basin.node_id == basin_bench.node_id
-        @test all(q -> abs(q) < 0.2, basin.level - basin_bench.level)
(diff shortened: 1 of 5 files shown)


ngoorden:
We can't commit to this branch. We recommend to use 1e-5 so the effect of relaxed_root is minimal on outcomes.

visr:
Lowered it to 1e-5.

visr:
The reason @ngoorden recommends 1e-5 meter/meter is because these low slopes are not common in flowing water. With 1e-3 many situations in the Netherlands were effectively not in our Manning but relaxed root domain.

The oscillations from https://github.com/Deltares/Ribasim/issues/1743 are not reintroduced by this lower threshold, I made similar plots to confirm. Some changes in the state due to this change are to be expected.

SouthEndMusic (approved):

Issue #714, opened 2023-10-31
Allow `Pump` and `Outlet` nodes to be controlled by allocation


SouthEndMusic:
**What**
After a refinement session, we (@SouthEndMusic, @gijsber, @visr) came to the conclusion that it would be nice to have a special control state "allocation", which is not defined via the `static` tables but means that the flow of a `Pump` or `Outlet` is set by allocation. However, when working out this issue so that @Jingru923 can collaborate on it, I came to some slightly different ideas on how to mark a node as being controlled by allocation, see below.

**How**
- [ ] Add a new optional string column to the `static` tables of `Pump` and `Outlet` named `controlled_by`, with allowed values "nothing" (default), "allocation" and "PID". We could set PID controllable nodes to PID by default to avoid a breaking change.
- [ ] _Add a flag to `Pump` and `Outlet` nodes to indicate if and how they are controlled_: The allocation algorithm must know which nodes are allocation controlled so that their `flow_rate` can be set. These nodes already have a boolean flag `is_pid_controlled`, so instead of adding a new flag, let's modify this one to an enumerator `ControlType` with values of the `controlled_by` options above.
- [ ] _Compute and assign the `flow_rate`_: after a run of the allocation algorithm, the flow through nodes that are allocation controlled must be computed and assigned. The assigned flow must be the sum over the priorities of the flow through the inflow edge of allocation controlled nodes. This sum can be computed during the allocation algorithm in pre-allocated memory, or be computed afterwards using the data in the `flow_record` (@visr let me know whether you think the latter is a good idea).
- [ ] _Don't do PID control if not in that control type_: Currently we assume that when PID control is available for a node, this node is always PID controlled (except when the PID control node is not active). If we want to incorporate PID control in the above (this could also be a follow-up issue), we need to check whether nodes are actually PID controlled when doing the PID part of `water_balance!`.

**Validation**
- Which conflicts between the different control types do we want to validate for?
- Allocation controlled nodes should at least be part of a subnetwork (and allocation should be active).
- If validation whether PID is available (see below).

**Reasons why I prefer this over the original idea**
- We don't have to implement and maintain special control states;
- Nodes can be assigned one of the control types at initialization without `DiscreteControl` getting involved;
- If desired,`DiscreteControl` can change the control type during simulation (for PID control it must be checked whether it is available for the node, preferably during initialization). Note that this means that `DiscreteControl` overarches the other control types, which is also why it is not one of the `ControlType` options.

[...]

PR #2139, opened 2025-03-10
Control from allocation

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/714

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -651,7 +651,7 @@ end
 function apply_parameter_update!(parameter_update)::Nothing
     (; name, value, ref) = parameter_update
 
-    # Ignore this parameter update of the associated node does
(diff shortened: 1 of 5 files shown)


SouthEndMusic:
@gijsber @Fati-Mon I made the model as discussed:

allocation_control

Currently there is no flow at all in this model, and here's why: allocation calculates the amount of flow available as the flow from the level boundary in the physical layer. However, at the beginning of the simulation there is no flow there yet because the outlet does nothing (the demand is 0 at the start). So there's an impass: the outlet doesn't flow because allocation doesn't allocate anything, allocation doesn't allocate anything because there is no flow.

Of course this is a flawed mechanism, which is again solved by allocation having a better model of the outlet.

PR #2144, opened 2025-03-11
Write changelog for release

### LINKED ISSUE #2141: [core] When failing on an invalid link, give the link id instead of the source id
As mentioned by @DanielTollenaar. Can you give a traceback, so we know which error message to fix exactly?

visr:
Already adds a note on #2136 assuming something like that will be merged.

- Includes the `snake_case` part of #2114.
- Adds an upper bound on Pandera in pyproject.toml as well (pixi.toml already has it)
- Fixes #2141 by adding link_id to the error message.


docs/changelog.qmd
```diff
@@ -5,6 +5,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
+- Support time dependent Pump and Outlet tables. [#2110](https://github.com/Deltares/Ribasim/pull/2110)
+
+### Changed
+- Edges are now called links. [#2023](https://github.com/Deltares/Ribasim/pull/2023)
+- All geometries are coerced to 2D. [#2111](https://github.com/Deltares/Ribasim/pull/2111)
+
+### Fixed
+- Faster initialization of large tables. [#2026](https://github.com/Deltares/Ribasim/issues/2026)
+- Declare incompatibility with Pandera v0.23+. [#2120](https://github.com/Deltares/Ribasim/issues/2120)
+- Decrease the usage of a reduction factor in ManningResistance. [#2136](https://github.com/Deltares/Ribasim/issues/2136)
```
SouthEndMusic, line 32: I was a bit confused by this, because strictly speaking this is about something different from what we call the reduction factor

SouthEndMusic (approved):

docs/changelog.qmd
visr: Good point, renamed to Decrease the smoothing around `Δh = 0`

PR #2152, opened 2025-03-17
CI: Force rebuilding on tags

### LINKED ISSUE #2040: `ribasim --version` is wrong for releases
Previously we had the issue #1655 that failed the release tests. The tests we fixed, but the issue described in https://github.com/Deltares/Ribasim/issues/1655#issuecomment-2400784264 remains.

The README.md for the v2025.1.0 build states:

```toml
version = "2024.11.0-74-ge2b86cd"
commit = "https://github.com/Deltares/Ribasim/tree/e2b86cdb0d"
branch = "https://github.com/Deltares/Ribasim/tree/main"
```

That commit is the right one: https://github.com/Deltares/Ribasim/tree/e2b86cdb0d. So why is it not recognizing the tag? The release workflow is supposed to trigger on the tag being pushed, as mentioned here: https://ribasim.org/dev/release.html#create-a-new-release. Perhaps that is not the case? Or if it already built the same commit that is later tagged it doesn't build new artifacts or something?

### LINKED ISSUE #2128: Make [skipci] work on TC
Or [TC], as each push to main will now trigger everything. It would be good to be able to skip the CI based on the commit message. See https://www.jetbrains.com/help/teamcity/2024.12/configuring-vcs-triggers.html#trigger-rules-examples-1 on how to configure this.

evetion:
Teamcity re-uses builds in a build chain when possible, so once the tag triggered the release chain, it re-used the previously build executable(s) for the commit of the tag. We can disable this by `reuseBuilds = ReuseBuilds.NO` in the specific snapshot builds.

This also enables us to skip builds on Teamcity (and already on Github) with a commit message that starts with `[skip ci]`. I've documented this as well.

Finally, I noted we create duplicate builds, once for a(ny) branch, and once for a created PR. This is bad (especially because I disabled some re-use above), so I've added some branchfilter to the default branch only (as [advised when using PR integration](https://www.jetbrains.com/help/teamcity/pull-requests.html#Interaction+with+VCS+Roots)). I also added this PR integration to the testmodel build which didn't have it yet.

Fixes #2040 
Fixes #2128

### DIFF
--- a/docs/dev/ci.qmd
+++ b/docs/dev/ci.qmd
@@ -30,6 +30,10 @@ They include the following checks:
 * Ribasim Python tests: Runs Ribasim Python tests on multiple platforms and multiple Python versions
 * QGIS Tests: Runs QGIS unit tests
 
+The Github CI configuration is stored in version control,
+and can be found in the `.github` folder.
+
+One can [skip running the CI](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs) using `[skip ci]` in the git commit message.
 
 # TeamCity
 Ribasim has another cloud-based CI service based on [TeamCity](https://www.jetbrains.com/teamcity/).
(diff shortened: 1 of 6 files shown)


docs/dev/ci.qmd
```diff
@@ -41,6 +45,13 @@ graph LR
     A-->C(Release)
 ```
 
+The Teamcity CI configuration itself is also stored in version control, and can be found in the `.teamcity` folder.
+Note that changes are only applied once merged to the `main` branch.
+One can test the validity of their local configuration with `mvn teamcity-configs:generate`
+(requiring a local installation of [maven](https://maven.apache.org/)).
+
+Like Github, one can skip running the Teamcity CI by starting the git commit message with `[skip ci]`.
```
visr, line 53: Above it says

> using `[skip ci]` in the git commit message

which is it, at the start or anywhere?

PR #2153, opened 2025-03-17
VSCode parallel tests

visr:
This adds `"julia.numTestProcesses": 6` and sorts the settings.
This makes VSCode run the tests in parallel. I did some quick checks before and around 6 seemed optimal for my hardware.
Probably few people will have hardware that would cause this to be slower.


evetion (approved):
That sort makes the diff terrible 😅

visr:
That's why I put it in a separate commit ;)

PR #2155, opened 2025-03-18
[skip ci] Trigger TC builds on PRs.

evetion:
#2152 introduced a branch filter for the builds that used the PR feature, but they need to include PR branches as well. Currently we build only for main.


evetion:
I do like how easy it is to get the CI to be green 😈

visr (approved):

Issue #2159, opened 2025-03-18
Configure FEWS visualization: stacked bar plot


visr:
Based on `basin.arrow`

https://user-images.githubusercontent.com/4471859/179259174-0caccd4a-c51b-449e-873c-17d48cfc8870.png

PR #2161, opened 2025-03-18
Add `model.to_fews(region_dir)`

visr:
This adds a function `model.to_fews(region_dir)` that converts the network and results to files that Delft-FEWS can directly handle. It is marked as experimental for now.

@gijsber is working on a Delft-FEWS configuration that can be used to visualize model results, to complement our existing tools. We'll likely add this configuration to this monorepo since it is generic. #2159 also pertains to this work.

What is especially nice is the spatio-temporal support of Delft-FEWS, so we can make visualizations like this:

image

In theory we can support similar functionality with QGIS, but looking at the plots in https://github.com/Deltares/Ribasim/pull/1369 this would likely need work in QGIS itself. So this is really a quick win to be able to inspect models better.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -780,7 +780,7 @@ class Model(FileModel):
         ** Warning: This method is experimental and is likely to change. **
 
         To run this method, the model needs to be written to disk, and have results.
-        The Node and Link tables are written to shapefiles in the REGION_HOME/Config directory.
+        The Node, Link and Basin / area tables are written to shapefiles in the REGION_HOME/Config directory.
         The results are written to NetCDF files in the REGION_HOME/Modules directory.
         The netCDF files are NetCDF4 with CF-conventions.
 
@@ -806,10 +806,16 @@ class Model(FileModel):
         assert df_link is not None
         assert df_node is not None
 
+        df_basin_area = self.basin.area.df
+        if df_basin_area is None:
+            # Fall back to the Basin points if the area polygons are not set
+            df_basin_area = df_node[df_node["node_type"] == "Basin"]
+
         network_dir = region_home / "Config/MapLayerFiles/{ModelId}"
         network_dir.mkdir(parents=True, exist_ok=True)
         link_path = network_dir / "{ModelId}Links.shp"
         node_path = network_dir / "{ModelId}Nodes.shp"
+        basin_area_path = network_dir / "{ModelId}Areas.shp"
 
         with warnings.catch_warnings():
             warnings.filterwarnings(
@@ -822,6 +828,7 @@ class Model(FileModel):
             )
             df_link.to_file(link_path)
             df_node.to_file(node_path)
+            df_basin_area.to_file(basin_area_path)
 
     def _results_to_fews(self, region_home: DirectoryPath) -> None:
         """Convert the model results to NetCDF with CF-conventions for importing into Delft-FEWS."""
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import shutil
+import warnings
 from collections.abc import Generator
 from os import PathLike
 from pathlib import Path
@@ -462,6 +463,55 @@ class Model(FileModel):
         context_file_loading.set({})
         return self
 
+    @property
+    def toml_path(self) -> FilePath:
+        """
+        Get the path to the TOML file if it exists.
+
+        Raises
+        ------
+        FileNotFoundError
+            If the model has not been written to disk.
+
+        Returns
+        -------
+        FilePath
+            The path to the TOML file.
+        """
+        if self.filepath is None:
+            raise FileNotFoundError("Model must be written to disk.")
+        return FilePath(self.filepath)
+
+    @property
+    def results_path(self) -> DirectoryPath:
+        """
+        Get the path to the results directory if it exists.
+
+        This checks for the presence of required result files in the directory.
+
+        Raises
+        ------
+        FileNotFoundError
+            If any of the required result files are missing.
+
+        Returns
+        -------
+        DirectoryPath
+            The path to the results directory.
+        """
+        toml_path = self.toml_path
+        results_dir = DirectoryPath(toml_path.parent / self.results_dir)
+        # This only checks results that are always written.
+        # Some results like allocation_flow.arrow are optional.
+        filenames = ["basin_state.arrow", "basin.arrow", "flow.arrow"]
+        for filename in filenames:
+            if not (results_dir / filename).is_file():
+                raise FileNotFoundError(
+                    f"Cannot find {filename} in '{results_dir}', "
+                    "perhaps the model needs to be run first."
+                )
+        return results_dir
+
     def plot_control_listen(self, ax):
         """Plot the implicit listen links of the model."""
         df_listen_link = pd.DataFrame(
@@ -585,7 +635,7 @@ class Model(FileModel):
 
         return ax
 
-    def to_xugrid(self, flow: bool = False, allocation: bool = False):
+    def to_xugrid(self, add_flow: bool = False, add_allocation: bool = False):
         """Convert the network to a `xugrid.UgridDataset`.
 
         Either the flow or the allocation data can be added, but not both simultaneously.
@@ -593,12 +643,12 @@ class Model(FileModel):
 
         Parameters
         ----------
-        flow : bool
+        add_flow : bool
             add flow results (Optional, defaults to False)
-        allocation : bool
+        add_allocation : bool
             add allocation results (Optional, defaults to False)
         """
-        if flow and allocation:
+        if add_flow and add_allocation:
             raise ValueError("Cannot add both allocation and flow results.")
 
         node_df = self.node_table().df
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -584,20 +585,20 @@ def plot(
 
         return ax
 
-    def to_xugrid(self, add_flow: bool = False, add_allocation: bool = False):
+    def to_xugrid(self, flow: bool = False, allocation: bool = False):
```
visr, line 588: This change is breaking but I don't think anyone uses it yet. This brings it more in line with the `to_fews` argument names.

python/ribasim/ribasim/utils.py
```diff
@@ -66,12 +66,6 @@ def _link_lookup(uds) -> Series[Int32]:
     )
 
 
-def _time_in_ns(df) -> None:
-    """Convert the time column to datetime64[ns] dtype."""
-    # datetime64[ms] gives trouble; https://github.com/pydata/xarray/issues/6318
-    df["time"] = df["time"].astype("datetime64[ns]")
```
visr, line 72: The linked issue is now fixed in xarray. I don't think we need to keep it to support old versions of optional dependencies.

python/ribasim/ribasim/model.py
```diff
@@ -638,36 +639,41 @@ def to_xugrid(self, add_flow: bool = False, add_allocation: bool = False):
...
+    def _toml_path(self) -> FilePath:
+        """Get the path to the TOML file, raising an error if it doesn't exist."""
         toml_path = self.filepath
         if toml_path is None:
-            raise FileNotFoundError("Model must be written to disk to add results.")
-        return toml_path
+            raise FileNotFoundError("Model must be written to disk.")
+        return FilePath(toml_path)
+
+    def _results_dir(self) -> DirectoryPath:
```
evetion, line 656: This could be a public method with @property? And another property with `has_run` or `has_results`?

python/ribasim/ribasim/model.py
evetion: Why can't `to_fews` take an `add_network` argument? I think `add_x` is much clearer than `x`.

python/ribasim/ribasim/model.py
```diff
@@ -738,3 +741,67 @@ def _add_allocation(self, uds):
...
+
+        To run this method, the model needs to be written to disk, and have results.
+        The Node and Link tables are written to shapefiles in the Config directory.
+        The results are written to NetCDF files in the Modules directory.
+        The netCDF files are NetCDF4 with CF-conventions.
+
+        Parameters
+        ----------
+        region_home: str | PathLike[str]
+            Path to the Delft-FEWS REGION_HOME directory.
```
evetion, line 764: Is REGION_HOME an environment variable that we can use as default here?

python/ribasim/ribasim/model.py
```diff
@@ -738,3 +741,67 @@ def _add_allocation(self, uds):
...
+        network: bool = True,
+        results: bool = True,
+    ) -> None:
+        """
+        Write the model network and results into files used by Delft-FEWS.
+
+        ** Warning: This method is experimental and is likely to change. **
+
+        To run this method, the model needs to be written to disk, and have results.
+        The Node and Link tables are written to shapefiles in the Config directory.
```
evetion, line 757:
```suggestion
        The Node and Link tables are written to shapefiles in the REGION_HOME/Config directory.
```

python/ribasim/ribasim/model.py
```diff
@@ -738,3 +741,67 @@ def _add_allocation(self, uds):
...
+        The results are written to NetCDF files in the Modules directory.
```
evetion, line 758:
```suggestion
        The results are written to NetCDF files in the REGION_HOME/Modules directory.
```

python/ribasim/ribasim/model.py
```diff
@@ -738,3 +741,67 @@ def _add_allocation(self, uds):
...
+    def _network_to_fews(self, region_home: DirectoryPath) -> None:
+        """Write the Node and Link tables to shapefiles for use in Delft-FEWS."""
+        df_link = self.link.df
+        df_node = self.node_table().df
+        assert df_link is not None
+        assert df_node is not None
+
+        network_dir = region_home / "Config/MapLayerFiles/{ModelId}"
+        link_path = network_dir / "{ModelId}Links.shp"
+        node_path = network_dir / "{ModelId}Nodes.shp"
```
evetion, line 781: Did you mean these to be f-strings? If so, you miss the f, and the argument. edit: Maybe not, given the rest of the code, what a weird way of naming files.

python/ribasim/tests/test_model.py
```diff
@@ -203,15 +203,39 @@ def test_xugrid(model, tmp_path):
...
+
+
+@pytest.mark.parametrize(
+    "model",
+    [basic_model(), outlet_model(), pid_control_equation_model(), trivial_model()],
+)
+def test_to_fews(model, tmp_path):
+    region_home = tmp_path
+    network_dir = region_home / "Config/MapLayerFiles/{ModelId}"
+    network_dir.mkdir(parents=True, exist_ok=True)
```
evetion, line 224: This should only create region_home if it not exists, because the code above doesn't test for the existence of the folders below it.

python/ribasim/ribasim/model.py
```diff
@@ -738,3 +741,67 @@ def _add_allocation(self, uds):
...
+
+    def _network_to_fews(self, region_home: DirectoryPath) -> None:
+        """Write the Node and Link tables to shapefiles for use in Delft-FEWS."""
+        df_link = self.link.df
+        df_node = self.node_table().df
+        assert df_link is not None
+        assert df_node is not None
+
+        network_dir = region_home / "Config/MapLayerFiles/{ModelId}"
+        link_path = network_dir / "{ModelId}Links.shp"
```
evetion, line 780: Shapefiles truncate our column names:

> Normalized/laundered field name: 'from_node_id' to 'from_node_'

evetion (approved):
Looks good, some minor comments. Note that your test requires a model that has been run (as it needs arrow files), so it fails on our default Python tests.

python/ribasim/ribasim/model.py
visr: Can do!

python/ribasim/ribasim/model.py
visr: It is a variable that can be used inside FEWS configurations, but I don't think it is commonly set as an environment variable. @gijsber can confirm.

python/ribasim/ribasim/model.py
visr: Yeah it is a bit odd, the idea was that this way it can easily be replaced. We want to support importing multiple different models in the future.

python/ribasim/tests/test_model.py
visr: I don't understand your comment. Are you saying that `exist_ok=True` is not needed. That may be true usually since tmp_path is generally not reused. This felt a bit more robust for debugging.

gijsber:
REGION_HOME is indeed and interval variable which is the root folder of the Fews-application. everything Fews is relative to this root. At what path exactly REGION_HOME resides is up to the user (could be c:\fews or burried deep insde your One Drive folder, but it should not be on a network drive given the high frequent IO

python/ribasim/ribasim/model.py
visr: Indeed. The FEWS Config already uses the truncated column names. FEWS also accepts GeoJSON but that is probably not ideal either given the lack of CRS support (assuming they use the latest standard). Hopefully they'll support GeoPackage or another format without these issues. Obligatory http://switchfromshapefile.org/.

Pyogrio does warn about this truncation by the way. We should check if we can silence that for this call, since we know.

gijsber:
Hi @visr , I my latest update (posted in my Bulletin) I now also expect a separate shape file for the Areas:
{ModelId}Areas.shp next to the Nodes and Links

gijsber:
I guess it gives an error at that moment. 
Easiest workaround is ..in case no area polygons are defined, write the Areas.shp with the content of Nodes.shp or even better only the Basin nodes.

python/ribasim/tests/test_model.py
evetion: Sorry, I meant that the normal code doesn't check whether the folders exist, it just tries to write. I'm not sure if pandas creates intermediate folders.

The fact that you create all intermediate folders for a test implies that it doesn't, and that's something we should catch/fix?

visr: Oh yeah I didn't make it create all paths on purpose since if it wouldn't exist it probably wouldn't be a correct destination.

Though we do create the whole path on model.write, and it may be convenient to be able to write to a region_home that doesn't exist in case you want to inspect the output or need to send it to somewhere. So I'll change that and remove this.

python/ribasim/ribasim/model.py
visr: We now have `model.toml_path` and `model.results_path`, which both error if they don't exist. `model.results_dir` returns the `results_dir` config entry, which is "results" by default.

I didn't add `has_run` or `has_results` yet as I don't think there's much of a need for it yet, but we could always consider it in another PR?

visr:
@evetion maybe a quick check since I changed quite a lot?

PR #2174, opened 2025-03-26
Linear allocation objective

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1494.

The objective function I have implemented now is of the form

$$
\sum_{i \in I} U_i + L_i,
$$

as described in the issue. I have reformulated the associated constraints as 

$$
d_i U_i \ge F_i - fd_i, \quad d_i L_i \ge d_i f- F_i
$$

to avoid division problems when $d_i = 0$. The 'target fraction' $f$ is given by the minimum of 1 and the total capacity divided by the total demand.

As mentioned by @jarsarasty, it could be that we have to add another optimization in goal programming fashion to allocate the capacity that is left after this first optimization. That is because the a priori computed fraction $f$ as described above is only ideal if there are no constraints between sources and demands.


SouthEndMusic:
I still have to fix some things, but its pretty neat that the "equal_fraction_allocation" test still passes at first try 👍

SouthEndMusic:
It's also immediately clear that the performance is much better with this objective function than with the quadratic one

SouthEndMusic:
The remaining few failing tests are because of the changed outcomes of the allocation algorithm.

SouthEndMusic:
@jarsarasty I had the following brainwave: In stead of just penalize any deviation from the target fraction of the demand, have the steepest decline in loss from 0 to the target fraction, and then a less steep decline until the full demand is allocated, and then increasing the loss again afterwards. This is nicely convex. 

The slopes of the first and last section can be the same as before, but the slope of the middle section is debatable. My first thought was to take half of the slope of the first section.

Elaborated_loss

jarsarasty:
Hi @SouthEndMusic ,

As discussed, the failed test is only happening on Linux. This test indicates that some values for saveat and dt are incorrect. Could you please confirm if there is indeed an error in those values or if the test should be declared as broken for now?

SouthEndMusic:
Hi @jarsarasty,

I ran the tests on CI again, and this time they don't fail on Linux, so that still suggests non-deterministic behavior (or maybe hardware-specific behavior). 

I understand the confusion, but these error messages

```
┌ Error: Invalid saveat
│   saveat = -Inf
└ @ Ribasim.config D:\a\Ribasim\Ribasim\core\src\config.jl:341
┌ Error: Invalid saveat
│   saveat = NaN
└ @ Ribasim.config D:\a\Ribasim\Ribasim\core\src\config.jl:341
┌ Error: A finite saveat must be an integer number of seconds.
│   saveat = 3.1415
└ @ Ribasim.config D:\a\Ribasim\Ribasim\core\src\config.jl:337
┌ Error: Invalid dt
│   dt = 0.0
└ @ Ribasim.config D:\a\Ribasim\Ribasim\core\src\config.jl:361
┌ Error: Negative storage detected in Basin #1
└ @ Ribasim D:\a\Ribasim\Ribasim\core\src\callback.jl:470
```

do not indicate failing tests; these are error messages which are deliberately triggered during testing.

docs/concept/allocation.qmd
```diff
@@ -144,22 +144,37 @@ There are several types of variable whose value has to be determined to solve th
...
     \min E_{\text{user demand}} + E_{\text{level demand}} + E_{\text{flow demand}}
 $$
 
 The error between the flows and user demands is denoted by $E_{\text{user demand}}$, where
 $$
-    E_{\text{user demand}} = \sum_{(i,j)\in E_S\;:\; i\in U_S} d_j^p(t)\left(1 - \frac{F_{ij}}{d_j^p(t)}\right)^2
+    E_{\text{user demand}} = \sum_{(i,j)\in E_S\;:\; i\in U_S} \left|F_{ij} - fd_j^p(t)\right|,
 $$
 
+which is implemented using auxiliary variables.
```
jarsarasty, line 171: You could explain how this part was implemented (similar to what you wrote in the description of the pull request). If this information is not very relevant to end users, it could just be included in a footnote.

jarsarasty (changes requested):
Hi @SouthEndMusic 

I have made a few suggestions for implementation. A short meeting may help clarify some of my questions.

jarsarasty (approved):
Hi @SouthEndMusic 

Thanks for addressing the comments.

PR #2175, opened 2025-03-27
Implement Junction Node

### LINKED ISSUE #1609: Add `Junction` node
It came out of discussions with @Fati-Mon that it would be nice to have a confluence node, which I think exists in Ribasim 7-8. It comes down to a node where flows from multiple edges can come in, which is then summed on one outgoing edge. Currently we only support something like this by bringing these multiple flows together into a basin, but the downside of this is that you then need profile information for this basin.

evetion:
Fixes #1609 

- [x] Implement tests

### DIFF
--- a/core/src/graph.jl
+++ b/core/src/graph.jl
@@ -137,23 +137,24 @@ function simplify_graph!(
 
     # Remove junctions by iteratively simplifying from IN--J--OUT to IN--OUT
     for junction_id in get_node_ids(db, NodeType.Junction)
-        for in_nb in collect(inneighbor_labels(graph, junction_id)),
-            out_nb in collect(outneighbor_labels(graph, junction_id))
+        for in_neighbor in collect(inneighbor_labels(graph, junction_id)),
+            out_neighbor in collect(outneighbor_labels(graph, junction_id))
 
-            link_id = graph[in_nb, junction_id].id
+            link_id = graph[in_neighbor, junction_id].id
             external_link_ids = get(link_mapping, link_id, [link_id])
 
-            link_id = graph[junction_id, out_nb].id
+            link_id = graph[junction_id, out_neighbor].id
             append!(external_link_ids, get(link_mapping, link_id, [link_id]))
 
             link_metadata = LinkMetadata(;
                 id = new_link_id,
                 type = LinkType.flow,
-                link = (in_nb, out_nb),
+                link = (in_neighbor, out_neighbor),
             )
 
             # Create new flow link when no Junctions remain
-            if in_nb.type != NodeType.Junction && out_nb.type != NodeType.Junction
+            if in_neighbor.type != NodeType.Junction &&
+               out_neighbor.type != NodeType.Junction
                 for external_link_id in external_link_ids
                     push!(I, external_link_id)
                     push!(J, new_link_id)
@@ -163,11 +164,11 @@ function simplify_graph!(
 
             link_mapping[new_link_id] = external_link_ids
             new_link_id += 1
-            if haskey(graph, in_nb, out_nb)
+            if haskey(graph, in_neighbor, out_neighbor)
                 errors = true
-                @error "Duplicate link: Junction links form cycle." in_nb out_nb external_link_ids
+                @error "Duplicate link: Junction links form cycle." in_neighbor out_neighbor external_link_ids
             else
-                graph[in_nb, out_nb] = link_metadata
+                graph[in_neighbor, out_neighbor] = link_metadata
             end
         end
         # Will also remove the edges
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -1060,8 +1060,8 @@ function CompoundVariable(
     for row in variables_compound_variable
         listen_node_id = NodeID(row.listen_node_id, node_ids_all)
         if listen_node_id.type == NodeType.Junction
-            @error "Cannot listen to Junction #$(listen_node_id.value)"
-            error("Invalid `listen_node_id` for CompoundVariable #$(node_id.value).")
+            @error "Cannot listen to Junction node" listen_node_id node_id
+            error("Invalid `listen_node_id`.")
         end
         # Placeholder until actual ref is known
         diff_cache_ref = DiffCacheRef()


core/src/graph.jl
```diff
@@ -82,6 +92,50 @@ function create_graph(db::DB, config::Config)::MetaGraph
...
+            if in_nb.type != NodeType.Junction && out_nb.type != NodeType.Junction
+                # TODO Validate this new connection
+                for edge_id in edge_ids
+                    push!(I, current_link_id)
+                    push!(J, edge_id)
+                end
+                push!(flow_links, link_metadata)
+            end
+
+            mapping[current_link_id] = edge_ids
```
visr, line 129: Going back to edge ID are we?

core/src/graph.jl
```diff
@@ -82,6 +92,50 @@ function create_graph(db::DB, config::Config)::MetaGraph
         end
         graph[id_src, id_dst] = link_metadata
     end
+
+    # Setup sparse matrix for junctions mapping
+    flow_link_ids = [flow_link.id for flow_link in flow_links]
+    I, J = flow_link_ids, copy(flow_link_ids)
+    mapping = Dict{Int32, Vector{Int32}}()
```
visr, line 99: Good to add a comment, it maps x to y.

core/src/validation.jl
```diff
@@ -12,12 +12,13 @@ neighbortypes(::Val{:basin}) = Set((
     :outlet,
     :user_demand,
 ))
-neighbortypes(::Val{:terminal}) = Set{Symbol}() # only endnode
+neighbortypes(::Val{:terminal}) = Set{Symbol}()
+neighbortypes(::Val{:junction}) = Set((:basin, :junction))
```
visr, line 16: This looks like the original directed Confluence validation of https://github.com/Deltares/Ribasim/issues/1609#issuecomment-2730054537.

But Junction should also be able to go to connector nodes right?

core/src/parameter.jl
```diff
@@ -14,7 +13,7 @@ const SolverStats = @NamedTuple{
 }
 
 # LinkType.flow and NodeType.FlowBoundary
-@enumx LinkType flow control none
+@enumx LinkType flow control junction none
```
visr, line 16: What is a junction link and why do we need it?

core/src/graph.jl
```diff
@@ -80,8 +92,59 @@ function create_graph(db::DB, config::Config)::MetaGraph
             errors = true
             @error "Invalid link: the opposite link already exists (this is only allowed for UserDemand)." link_id id_src id_dst
         end
+        new_link_id = max(link_id, new_link_id)
         graph[id_src, id_dst] = link_metadata
     end
+
+    # Setup sparse matrix for junctions mapping
```
SouthEndMusic, line 99: I suggest making the `Junction` node processing a separate function completely disjoint from the above for clarity. The `Junction` node IDs can easily be retrieved with

```julia
get_node_ids(db, NodeType.Junction)
```

and get the flow links that connect to a Junction node using `inoutflow_ids`. 

Also, shouldn't `new_link_id += 1` be in the above loop?

core/src/read.jl
```diff
@@ -825,11 +825,16 @@ function Outlet(db::DB, config::Config, graph::MetaGraph)::Outlet
...
+function Terminal(db::DB, _::Config)::Terminal
     node_id = get_node_ids(db, NodeType.Terminal)
     return Terminal(node_id)
 end
 
+function Junction(db::DB, _::Config)::Junction
+    node_id = get_node_ids(db, NodeType.Junction)
+    return Junction(; node_id)
+end
+
```
SouthEndMusic, line 837: Since it's not necessary for dispatch, why not entirely remove the config argument?

python/ribasim_testmodels/ribasim_testmodels/allocation.py
```diff
@@ -1065,9 +1065,9 @@ def allocation_training_model():
...
     )
 
-    # Confluence
+    # Junction
     conf = model.basin.add(
-        Node(3, Point(-1.5, -1), subnetwork_id=1, name="confluence"),
+        Node(3, Point(-1.5, -1), subnetwork_id=1, name="junction"),
         [
             basin.Profile(area=[672000, 5600000], level=[0, 6]),
             basin.State(level=[4]),
```
visr, line 1073: We should probably change this test model to use an actual junction rather than a Basin that is used as a confluence.

docs/_quarto.yml
```diff
@@ -78,6 +78,7 @@ website:
             - reference/node/linear-resistance.qmd
             - reference/node/manning-resistance.qmd
             - reference/node/terminal.qmd
+            - reference/node/junction.qmd
```
visr, line 81: The tutorials should also be switched over to use a Junction node. Right now they use a Basin called confluence, see https://ribasim.org/tutorial/natural-flow.html#basin-node-confluence.

We could keep the name "confluence" but make the Basin a Junction instead.

core/src/write.jl
```diff
@@ -218,30 +218,36 @@ function flow_table(
...
         for (j, cvec) in enumerate(saveval)
             (; flow, flow_boundary) = cvec
-            flow_rate[i + (j - 1) * nflow] =
-                get_flow(flow, p_non_diff, 0.0, link; boundary_flow = flow_boundary)
+            for link in links
+                f = get_flow(flow, p_non_diff, 0.0, link; boundary_flow = flow_boundary)
+                flow_rate[i + (j - 1) * nflow] += f
+            end
         end
     end
```
SouthEndMusic, line 252: Would be nice if you can just get the output flows as the flow_link_map multiplied by the internal model flows in place using `mul!`.

docs/reference/node/junction.qmd
```diff
@@ -0,0 +1,65 @@
+---
+title: "Junction"
+---
+
+A Junction is a node that does nothing, but can be used to simplify networks that would otherwise result in many overlapping edges.
```
SouthEndMusic, line 5: I don't find this the greatest first sentence of the description of the shiny new node type you implemented

python/ribasim/ribasim/validation.py
```diff
@@ -14,16 +24,19 @@
...
+        "Junction",
         "LevelBoundary",
         "Basin",
         "Terminal",
     ],
     "TabulatedRatingCurve": [
+        "Junction",
         "LevelBoundary",
         "Basin",
         "Terminal",
```
visr, line 42: What can't FlowBoundary flow into a Junction? I think that'd be useful.

SouthEndMusic (commented):
It looks good, I left some refactoring suggestions.

docs/reference/node/junction.qmd
evetion: I'm happy to take suggestions.

visr: A Junction node allows explicitly representing confluences and bifurcations in the network. It doesn't introduce new behavior but makes it easier to make the network layout recognizable.

core/src/graph.jl
evetion: I've moved the logic to a separate function. The new_link_id is the max of the existing + 1, so it should not be in the loop. I've clarified that as well.

core/src/write.jl
evetion: After some investigation, the reason I did not do this yet is because the sparse matrix map uses the actual link_ids, not the logical indices, so it can't directly be used. I'm hesitant to make the change to logical indices, as everything else is id based (easy to make a mistake).

python/ribasim_testmodels/ribasim_testmodels/allocation.py
evetion: I've reverted this change.

docs/_quarto.yml
evetion: I've made an issue for this, as we determined that we cannot easily change the existing models (they need basins).

SouthEndMusic (approved):
Looks good, left a few nits

Issue #2189, opened 2025-03-28
Limit backtracking iterations


Huite:
In discussion with @visr 

The number of backtracking iterations is currently set at its default value of 1000:

https://github.com/JuliaNLSolvers/LineSearches.jl/blob/3259cd240144b96a5a3a309ea96dfb19181058b2/src/backtracking.jl#L15

This likely far too many. Note that every bactracking iteration requires an evaluation of the Newton-Raphson residual --- basically the `rhs!` -- which isn't that cheap. In my opinion, it makes more sense to put the maximum backtracking iterations at a value of 10 or so (the values I see "in the wild" for other nonlinear solves vary between 5-20). If you need more than 10 iterations are so, you're probably better off decreasing the time step, reformulating, and trying again.

In testing this with Martijn, the models crashes with a\ LineSearchException with a value of `10`. This shouldn't happen, this should be to signal convergence failure and a reason to apply the `failfactor` (by default 2, i.e. cut the step size in half).

To address this, we can catch the exception and reset the residual so that it'll be rejected. See:
https://github.com/Deltares/Ribasim/compare/main...catch-monitor?expand=1

For the particular model, this doesn't give any performance benefit, but:

* It should give performance benefits for very poorly converging models (since we spend less time on a difficult problem and move to an easier problem earlier)
* It is more robust: with adaptive time stepping, there should always be an option to decrease the time step (until dtmin). The exception basically blocks this, because the exception isn't caught / shouldn't be raised in this context.

Note that LineSearch.jl doesn't raise this exception, but communicates a ReturnCode instead.

https://github.com/SciML/LineSearch.jl/blob/dca4be86debb63fde9d1d4b4555f27c5139afe3a/src/backtracking.jl#L131

So this would be another reason to move to LineSearch.jl.

PR #2190, opened 2025-03-28
Catch too many iterations in backtracking

visr:
Trying out ideas from #2189.
I ran some tests using a combined waterboard model that runs to 50% in 10s, but hangs longer than I can wait at 62%.
I hoped that this would help it get over the bump more quickly, but it doesn't seem to so far, it has been running for 15 minutes. It might still be working, but we probably need an example with a less severe hang.
Besides it looks like the try-catch does have an effect on allocations, these are the timings until 50%:

```
before: 10.881255 seconds (7.48 M allocations: 1.813 GiB, 2.79% gc time, 1 lock conflict, 0.14% compilation time)
after: 10.332818 seconds (15.52 M allocations: 2.424 GiB, 4.52% gc time, 3 lock conflicts, 0.14% compilation time)
```

Both show the same `model.integrator.sol.stats`:

```
SciMLBase.DEStats
Number of function 1 evaluations:                  588
Number of function 2 evaluations:                  0
Number of W matrix evaluations:                    67
Number of linear solves:                           585
Number of Jacobians created:                       1
Number of nonlinear solver iterations:             585
Number of nonlinear solver convergence failures:   0
Number of fixed-point solver iterations:                     0
Number of fixed-point solver convergence failures:           0
Number of rootfind condition calls:                2380
Number of accepted steps:                          340
Number of rejected steps:                          2
```

### DIFF
--- a/core/src/util.jl
+++ b/core/src/util.jl
@@ -773,7 +773,7 @@ relaxed_root(x::GradientTracer, threshold::Real) = x
 get_level_from_storage(basin::Basin, state_idx::Int, storage::GradientTracer) = storage
 
 @kwdef struct MonitoredBackTracking{B, V}
-    linesearch::B = BackTracking()
+    linesearch::B = BackTracking(; iterations = 10)
     dz_tmp::V = []
     z_tmp::V = []
 end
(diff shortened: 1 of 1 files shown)


visr:
Closing this. I think the idea is good but it hasn't shown increased performance in practice. The issue may be that a try-catch at such a deep level in the code is simply too costly to reap any benefit. It would be better if there was no LineSearchException thrown, but rather a return code indicating failure. This is exactly what is done in the new approach that https://github.com/Deltares/Ribasim/pull/1784 tries to switch to, however that approach isn't done for another few months.

PR #2192, opened 2025-03-31
Allow max 1 outflow neighbour for FlowBoundary

### LINKED ISSUE #2072: Only allow having exactly one outflow link per `FlowBoundary` node

visr:
Fixes #2072. It was possible to let one FlowBoundary flow into multiple Basins. This is no longer allowed, so this is a breaking change. As far as I know it wasn't used though. It is quite confusing to have multiple outflow neighbours, because a Q at the FlowBoundary that goes to n Basins then becomes a nQ total inflow. And usually a FlowBoundary represents a single physical inflow point, that should always go to one point only.

The LevelBoundary can still be connected via resistance nodes to multiple Basins, but this makes sense when representing e.g. the sea.

### DIFF
--- a/core/src/parameter.jl
+++ b/core/src/parameter.jl
@@ -584,7 +584,7 @@ end
 
 """
 node_id: node ID of the FlowBoundary node
-outflow_links: The outgoing flow link metadata
+outflow_link: The outgoing flow link metadata
 active: whether this node is active and thus contributes flow
 cumulative_flow: The exactly integrated cumulative boundary flow since the start of the simulation
 cumulative_flow_saveat: The exactly integrated cumulative boundary flow since the last saveat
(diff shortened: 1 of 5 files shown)


python/ribasim/tests/test_validation.py
```diff
@@ -14,18 +14,16 @@
...
-            "Links have to be unique, but link with from_node_id 16 to_node_id 1 already exists."
-        ),
+        match=re.escape("Node 16 can have at most 1 flow link outneighbor(s) (got 1)"),
     ):
         model.link.add(
             model.flow_boundary[16],
             model.basin[1],
-            name="duplicate",
+            name="multiple-outflows",
         )
```
visr, line 27: I tried to come up with a `link.add` that triggers the unique edges validation first, but I don't think we can trigger it anymore because we'd need a from_node that supports multiple outputs and to_node that supports multiple inputs, that are allowed to be connected. I don't think that exists anymore.

PR #2200, opened 2025-04-01
Implement success

visr:
This is a follow-up of #2191. An issue with that is that it logged a warning about crashing simulations also when it succeeded. Also it didn't log the bottlenecks when you interrupt a (hanging) simulation.

Now it does both, this is what is shown when interrupting a simulation, stacktrace is clipped off:

```
│   toml_path = c:\Users\visser_mn\Downloads\lhm-sel\lhm-sel.toml
│   cli.ribasim_version = 2025.2.0
│   starttime = 2020-01-01T00:00:00
│   endtime = 2021-01-01T00:00:00
└ @ Ribasim C:\ProgramData\DevDrives\repo\ribasim\Ribasim\core\src\main.jl:40
┌ Warning: Simulation crashed or interrupted.
└ @ Ribasim C:\ProgramData\DevDrives\repo\ribasim\Ribasim\core\src\main.jl:51
┌ Warning: Convergence bottlenecks in descending order of severity:
│   Outlet #591225 = 48782.889544290454
│   Outlet #591226 = 48782.889544290454
│   ManningResistance #340784 = 0.04819449424793999
│   Pump #340303 = 0.045638255628477994
│   Pump #340194 = 0.04563825461806383
└ @ Ribasim C:\ProgramData\DevDrives\repo\ribasim\Ribasim\core\src\logging.jl:55
ERROR: InterruptException:
Stacktrace:
  [1] bracketstrictlymontonic(v::Vector{Float64}, x::Float64, guess::Int64, o::Base.Order.ForwardOrdering)
```

This implements `Base.success(::Model)::Bool`. Before we checked the SciML retcode. I noticed that the retcode of an ongoing simulation that gets interrupted is also successfull. So now we check both the retcode and if it `is_finished`.

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -16,7 +16,7 @@ function BMI.finalize(model::Model)::Nothing
 end
 
 function BMI.update(model::Model)::Nothing
-    step!(model.integrator)
+    SciMLBase.step!(model.integrator)
     return nothing
 end
 
--- a/core/src/main.jl
+++ b/core/src/main.jl
@@ -46,12 +46,18 @@ function main(toml_path::AbstractString)::Cint
                     model = Model(config)
                     try
                         solve!(model)
-                    finally
-                        @warn "Simulation crashed or interrupted. Writing results."
+                    catch
+                        # Catch errors thrown during simulation.
+                        @warn "Simulation crashed or interrupted."
+                        log_bottlenecks(model; converged = false)
                         write_results(model)
+                        display_error(io)
+                        return 1
                     end
 
-                    if successful_retcode(model)
+                    write_results(model)
+
+                    if success(model)
                         log_bottlenecks(model; converged = true)
                         @info "The model finished successfully."
                         return 0
@@ -67,19 +73,27 @@ function main(toml_path::AbstractString)::Cint
                     end
 
                 catch
+                    # Catch errors thrown before the model is initialized.
                     # Both validation errors that we throw and unhandled exceptions are caught here.
-                    # To make it easier to find the cause, log the stacktrace to the terminal and log file.
-                    stack = current_exceptions()
-                    Base.invokelatest(Base.display_error, stack)
-                    Base.invokelatest(Base.display_error, io, stack)
+                    display_error(io)
                     return 1
                 end
             end
         end
     catch
-        # If it fails before we get to setup the logger, we can't log to a file.
+        # Catch errors thrown before the logger is initialized.
         # This happens if e.g. the config is invalid.
-        Base.invokelatest(Base.display_error, current_exceptions())
+        display_error()
         return 1
     end
+end
+
+"Print a stacktrace to the terminal and optionally a log file."
+function display_error(io::Union{IOStream, Nothing} = nothing)::Nothing
+    stack = current_exceptions()
+    Base.invokelatest(Base.display_error, stack)
+    if io !== nothing
+        Base.invokelatest(Base.display_error, io, stack)
+    end
+    return nothing
 end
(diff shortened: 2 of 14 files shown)


visr:
It was good that this adds `is_finished`, which now checks if the `t` at the end is equal to `endtime`, because it highlighted an issue that I wasn't aware of. Two models had `allocation.timestep = 100_000`, which doesn't fit exactly in days. So we had a remainder timestep that we didn't take. This is now fixed:

https://github.com/Deltares/Ribasim/blob/e007c98bf7b105c56c56428076cba9d4bb75819f/core/src/model.jl#L350-L359

At first I thought `SciMLBase.solve!` perhaps didn't mean it would end at the endtime, but it does, it was our own custom allocation timestep code that was wrong.

core/src/bmi.jl
```diff
@@ -16,7 +16,7 @@ function BMI.finalize(model::Model)::Nothing
 end
 
 function BMI.update(model::Model)::Nothing
-    step!(model.integrator)
+    SciMLBase.step!(model.integrator)
```
SouthEndMusic, line 19: Why? Was there a name conflict?

core/src/main.jl
```diff
@@ -67,19 +73,27 @@ function main(toml_path::AbstractString)::Cint
...
+        display_error()
         return 1
     end
 end
+
+"Print a stacktrace to the terminal and optionally a log file."
+function display_error(io::Union{IOStream, Nothing} = nothing)::Nothing
+    stack = current_exceptions()
+    Base.invokelatest(Base.display_error, stack)
+    if io !== nothing
```
SouthEndMusic, line 95:
```suggestion
    if !isnothing(io)
```

core/src/model.jl
```diff
@@ -324,20 +340,26 @@ end
...
 Solve a Model until the configured `endtime`.
 """
-function SciMLBase.solve!(model::Model)::Model
+function solve!(model::Model)::Model
     (; config, integrator) = model
+    (; tspan) = integrator.sol.prob
     if config.allocation.use_allocation
-        (; tspan) = integrator.sol.prob
         (; timestep) = config.allocation
         allocation_times = 0:timestep:(tspan[end] - timestep)
```
SouthEndMusic, line 348: It looks like `allocation_times` is only used to determine `n_allocation_times`, so maybe this can be simply

```julia
n_allocation_times = int(floor(tspan[end]/timestep))
```

core/test/bmi_test.jl
```diff
@@ -8,15 +9,21 @@
     @test BMI.get_time_step(model) ≈ dt0 atol = 5e-3
     @test BMI.get_start_time(model) === 0.0
     @test BMI.get_current_time(model) === 0.0
-    @test BMI.get_end_time(model) ≈ 3.16224e7
+    endtime = BMI.get_end_time(model)
+    @test endtime ≈ 3.16224e7
```
SouthEndMusic, line 13: revert?

SouthEndMusic (approved):
Looks good, added some nits

core/src/bmi.jl
visr: Before this PR we overloaded `SciMLBase.solve!` and `SciMLBase.step!`, and now they are just `Ribasim.solve!` and `Ribasim.solve!`, so internally `solve!` is `Ribasim.solve` and we qualify the SciMLBase methods.

The overloading was not really useful and potentially confusing. And with some bad luck they can lead to more invalidations.

core/test/bmi_test.jl
visr: Why? The value is the same. I just need the variable for an extra test of the time at the end through BMI.

core/src/model.jl
visr: Done, good call.

Issue #2196, opened 2025-03-31
Support allocation results in Delft-FEWS


visr:
This probably has similarities to #2166. To be able to debug allocation results over time and space we probably need support in FEWS.

These are the two result files for allocation:

https://ribasim.org/reference/usage.html#allocation---allocation.arrow
https://ribasim.org/reference/usage.html#allocation-flow---allocation_flow.arrow

Both have the discrete variable `priority`. Perhaps this can be implemented like `substance` for concentration, using realizations (ensembles)?

The `allocation_flow` file also has optimization_type, see #2195. Though those are fixed so perhaps can be split into separate variables.

gijsber:
Best to implement the priority label items as a priority realization (similar to substance realization).
non-priority based items (e.g. the total demand and total realized) should not be in the realization but can be  included as a normal scalars (similar to basin.nc or flow.nc)

JohnRushKucharski-Deltares:
This could potentially be moved out of the 31 July deadline.

PR #2208, opened 2025-04-04
Add temporal Flow and Basin results to QGIS.

### LINKED ISSUE #2171: Visualize flows in QGIS
As we now try to do in FEWS (#1511). Non ugrid follow up of #969.

Simplest form, show the flows of the latest timestep as Link colors/width. Ideally there's some integration with the time manager in the future.

@DanielTollenaar

### LINKED ISSUE #2178: Show topology in QGIS
instead of the linestrings. Might be nice to have it as a toggle, like selection of a single node to show direct links. Helps with debugging.

evetion:
And adds "show topology" button on the layer context menu.

Fixes #2178
Fixes #2171 

This creates a new Results group in QGIS for each loaded model, and adds a BasinNode and FlowLink layer to it. These are duplicates of the Node and Link layer for the result data (and thus only keeping `node_type=Basin`, and `link_type=flow`).

Result arrow files (basin, flow and concentration) are read with pandas and indexed on time (concentration is widened first) and saved on the widget. Fields on the BasinNode and FlowLink are created to store the result variables. A specific time slice, by default the last, or by a signal from the temporal manager is then used to update these fields. The temporal manager is initialized based on the start/endtime and timestep from .toml file.

In effect you now have Node and Link tables with all result variables for a given timestep, meaning you're free to add symbology/labels as you would for any (static) layer.

The layer context menu is expanded to show a "Show topology" checkmark, which sets a layer variable, which in turn is used by the layer symbology to either show `@geometry` or `make_line(start_point(@geometry), end_point( @geometry))`.

---

For @huite and @JoerivanEngelen, updating the (memory) layers on the fly, and not using QGIS to read arrow files is essential for performance. Further performance was gained by not iterating over the layer features and updating them, but by setting them directly based on the fid (order). Note that the memory copy does not use the same fids as our geopackage layers, and you can't control them. The last factor of performance was disabling the automatic undo functionality (or rather creating 1 undo, instead of one for each feature) when updating the layer.


visr:
This is very very nice. I like how it dynamically fills in the tables and things I tried like adding a label with the flow rate just seem to work:
image

Here I used `format_number("flow_rate", 2)` to avoid getting the full Float64 printed. Perhaps we should add something like this in the default style?

If I remove a model and load it in in the same QGIS session it seems the show topology context is getting added another time:

image

Also it seemed to only work for me on the results but not the Link layer?

Further the BasinNode symbology uses `relative_error` by default(last column); can that be `level` (first column)?

These other things are perhaps better as follow ups:
- Extend https://ribasim.org/guide/qgis.html#sec-results
- Extend https://ribasim.org/dev/qgis_test_plan.html#result-inspection-tests
- iMOD Time Series Plot only links to Link and Node, not the new results layers. (not sure if this would make sense, but seems potentially confusing for people switching between timeseries and spatial results)

visr:
If I do "remove group" on the HWS model and then open it again I consistently get:

```
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 620, in _update_arrow_layers
    self._update_arrow_layer(
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 593, in _update_arrow_layer
    layer.startEditing()
RuntimeError: wrapped C/C++ object of type QgsVectorLayer has been deleted
```

evetion:
> Traceback (most recent call last):
>   File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 620, in _update_arrow_layers
>     self._update_arrow_layer(
>   File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 593, in _update_arrow_layer
>     layer.startEditing()
> RuntimeError: wrapped C/C++ object of type QgsVectorLayer has been deleted
> ```

I've seen this before and refactored a large part of the signals because of it, but I think it only triggers once you got another error first?


> Here I used `format_number("flow_rate", 2)` to avoid getting the full Float64 printed. Perhaps we should add something like this in the default style?

Sure, will add it. What do you suggest for Basin labels?



> Also it seemed to only work for me on the results but not the Link layer?

Yes, it only sets a variable on the layer, the style has to use that variable. Will update the default style to do so as well.

> If I remove a model and load it in in the same QGIS session it seems the show topology context is getting added another time:

Ouch, nice find. It's hijacking of the contextmenu signal, and it does so for the DatasetWidget (repeated for each dataset), that should probably be the Ribasim widget, or check for an existing widget.

> Further the BasinNode symbology uses `relative_error` by default(last column); can that be `level` (first column)?

Sure, but that's hard to style consistently, would have to make an update symbology after arrow load for it, based on min/max or so.

Huite:
> Traceback (most recent call last):
File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 620, in _update_arrow_layers
self._update_arrow_layer(
File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 593, in _update_arrow_layer
layer.startEditing()
RuntimeError: wrapped C/C++ object of type QgsVectorLayer has been deleted

> I've seen this before and refactored a large part of the signals because of it, but I think it only triggers once you got another error first?

This is because you're dealing with basically parallel garbage collection. If you delete the group, the Qt memory manager will delete the group, but Python won't. So if you then to try access the Python object, it'll try to look at something that's been deleted.
You could try connecting to the group deleted signal (or something). But this signal might not exist, or there might be a bunch of ways in which the group can get up deleted.

Alternatively, you just catch the exception, which will be fairly robust -- and update the Python state accordingly.

visr:
> What do you suggest for Basin labels?

Hmm by default nothing I guess so it doesn't get confused with flow rates? Otherwise `level`, would be nice if that is already prepared and folks just have to enable it.

> would have to make an update symbology after arrow load for it, based on min/max or so

You wanted that for `flow_rate` as well right? Also fine with me as a follow up.

visr:
Another follow up, since I see you already add concentration data, would be to add allocation results, see also https://github.com/Deltares/Ribasim/issues/2196.

evetion:
Ok, this is good for review/merge.

Fixed:
- Remove our Python reference (used in the update arrow layers) to our layers when layers are about to be removed. Should fix the C++ Wrapper errors.
- Hide the Results group by default
- Link layer now also supports topology view
- FlowLink layer has labels by default
- BasinNode layer now visualizes level (range will be useless for most cases though) and has labels by default.
- Show topology button will only show up once now
- Layers are now temporal enabled, enabling resetting of temporal ranges after manual changes:
Screenshot 2025-04-04 at 19 15 17

visr (approved):

PR #2209, opened 2025-04-04
Add computation time

### LINKED ISSUE #2207: Record the time to simulate
@rbruijnshkv and @DanielTollenaar both indicated that they would like it if the log file contained the time it took to simulate a model. We can add this, perhaps we can fish this out of the progress logger in the integrator, which is also used for the progress bar on stdout. This wouldn't include latency and model initialization, but that's probably fine. They can use a command line timing command of choice for that. Also they mentioned wanting to read this time programmatically from the log file, which is probably fragile because the log file format is not fixed and meant for humans not computers.

Additionally we could add a column to this file: https://ribasim.org/reference/usage.html#solver-statistics---solver_stats.arrow
Which would be the number of seconds it took to solve that period. Summing the column gives the total duration.

visr:
Fixes #2207. This logs the computation time:

```
│   toml_path = "models\\hws\\hws_2025_3_0\\hws.toml"
│   cli.ribasim_version = "2025.2.0"
│   starttime = 2023-01-20T00:00:00
└   endtime = 2024-04-27T00:00:00
Simulating 100%|███████████████████████████████████████████████████████████████████████████████| Time: 0:00:50
[ Info: Computation time: 49 seconds, 955 milliseconds
```

The time is also at the end of the progress bar, but this bar is not logged to `ribasim.log` so there was no record of it there.

There is also a new column `computation_time` in `solver_stats.arrow`, with the time in milliseconds for that period. I went for millisecond since the order of seconds per day is only really applicable for very large models.

```python
In [1]: import pandas as pd

In [2]: df = pd.read_feather("solver_stats.arrow")

In [3]: df
Out[3]:
          time  computation_time  rhs_calls  linear_solves  accepted_timesteps  rejected_timesteps
0   2023-01-20         2244.7489       7712           7709                2850                 458
1   2023-01-21          632.4497       2513           2513                 737                 128
2   2023-01-22          164.0670        662            662                 195                  36
3   2023-01-23          387.5130       1428           1428                 402                  80
4   2023-01-24          190.5293        751            751                 214                  37
..         ...               ...        ...            ...                 ...                 ...
458 2024-04-22          132.2348        542            542                  67                   6
459 2024-04-23          181.5635        703            703                  64                  10
460 2024-04-24           67.9207        292            292                  32                   2
461 2024-04-25          117.8029        451            451                  30                   8
462 2024-04-26            5.1752         16             16                   7                   0

[463 rows x 6 columns]

In [4]: df["computation_time"].sum() / 1000
Out[4]: np.float64(50.0540202)
```

### DIFF
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -343,7 +343,8 @@ Solve a Model until the configured `endtime`.
 function solve!(model::Model)::Model
     (; config, integrator) = model
     (; tspan) = integrator.sol.prob
-    if config.allocation.use_allocation
+
+    comptime_s = @elapsed if config.allocation.use_allocation
         (; timestep) = config.allocation
         n_allocation_times = floor(Int, tspan[end] / timestep)
         for _ in 1:n_allocation_times
@@ -360,5 +361,7 @@ function solve!(model::Model)::Model
         SciMLBase.solve!(integrator)
     end
     check_error!(integrator)
+    comptime = canonicalize(Millisecond(round(Int, comptime_s * 1000)))
+    @info "Computation time: $comptime"
     return model
 end
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -94,7 +94,7 @@ using TerminalLoggers: TerminalLogger
 
 # Date and time handling; externally we use the proleptic Gregorian calendar,
 # internally we use a Float64; seconds since the start of the simulation.
-using Dates: Dates, DateTime, Millisecond, @dateformat_str
+using Dates: Dates, DateTime, Millisecond, @dateformat_str, canonicalize
 
 # Callbacks are used to trigger function calls at specific points in the similation.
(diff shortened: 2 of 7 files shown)


core/src/model.jl
```diff
@@ -360,5 +361,7 @@ function solve!(model::Model)::Model
         SciMLBase.solve!(integrator)
     end
     check_error!(integrator)
+    comptime = canonicalize(Millisecond(round(Int, comptime_s * 1000)))
+    @info "Computation time: $comptime"
```
evetion, line 365: It this captured by our logging system (and thus ends up in the log file)?

evetion (approved):
A nice and small (efficient!) PR

core/src/model.jl
visr: Yes, I checked. And there are tests for this for other log messages.

PR #2217, opened 2025-04-07
Add `label_flow_rate` function to QGIS

visr:
Will be hooked up after #2215.
Also adds concentration to Bommelerwaard, we run it in all other test models, this is good to test the combined output.

### DIFF
--- a/ribasim_qgis/core/model.py
+++ b/ribasim_qgis/core/model.py
@@ -43,7 +43,7 @@ def get_database_path_from_model_file(model_path: Path) -> Path:
     )
 
 
-@qgsfunction(args="auto", group="Custom", referenced_columns=[])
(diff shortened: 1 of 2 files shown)


ribasim_qgis/core/model.py
```diff
@@ -39,3 +41,21 @@ def get_database_path_from_model_file(model_path: Path) -> Path:
...
+
+
+@qgsfunction(args="auto", group="Custom", referenced_columns=[])  # type: ignore
+def label_flow_rate(value: float) -> str:
+    """
+    Format the label for `flow_rate`.
+
+    Above 1, show 2 decimals.
+    Show 0 as 0.
+    Below 1, show 3 significant digits and scientific notation.
```
evetion, line 53:
```suggestion
    Between 0 and 1, and below 1, show 3 significant digits and scientific notation.
```

evetion (approved):

PR #2238, opened 2025-04-14
Add cumulative flow resetting

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1897.

This resets the cumulative flows when the water balance error gets above 1% of the tolerated value.

This works, but I still think that resetting the state like this can mess with the solver. Maybe it affects the adaptive time stepping, maybe it causes a very wrong initial guess in the Newton solve of the next timestep.

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -390,6 +390,7 @@ function check_water_balance_error!(
     (; basin, water_balance_abstol, water_balance_reltol, state_ranges, starttime) =
         p_non_diff
     errors = false
+    reset = false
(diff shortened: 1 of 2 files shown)


evetion:
Have we ever seen the water balance error dropping below our current thresholds? While I've seen it growing, the cumulative flow state needs to start exceeding 1e13 for any floating point errors to trigger it `eps(1e13) = 0.001953125`.

SouthEndMusic:
Closing this as not needed

PR #2242, opened 2025-04-15
Reduce residual calls

SouthEndMusic:
A follow-up to https://github.com/Deltares/Ribasim/pull/2224. I managed to reduce the number of residual calls in the 'managing' of the backtracking.

I cannot find any type instabilities in the relaxation code with Cthulhu anymore. The profile with `hws_2025_4_0` now looks like this:

image

What I also find interesting, is that time in the backtracking algorithm is completely dominated by the first call to the residual, which I think tells us that most of the time the residual reduction criterion is immediately satisfied. Maybe we can come up with a cheap criterion in the monitoring to skip the backtracking all together.

ping @visr @Huite

### DIFF
--- a/core/src/util.jl
+++ b/core/src/util.jl
@@ -813,22 +813,40 @@ function OrdinaryDiffEqNonlinearSolve.relax!(
     linesearch::MonitoredBackTracking,
 )
     (; linesearch, dz_tmp, z_tmp) = linesearch
+    atmp::Vector{Float64} = nlsolver.cache.atmp
 
-    # Store step before relaxation
-    @. dz_tmp = dz
+    let dz = dz,
+        integrator = integrator,
+        nlsolver = nlsolver,
+        f = f,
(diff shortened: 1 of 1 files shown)


Huite:
> which I think tells us that most of the time the residual reduction criterion is immediately satisfied.

What does "immediate" mean here? The backtracking starts with a relaxation step of I think of 0.5. If that works, then in this case a fixed relaxation of 0.5 should work too.

Or is immediate: after it's managed to fit a quadratic interpolant for the merit function?

visr (approved):
This looks like a clear performance gain that will help us until we can switch to https://github.com/Deltares/Ribasim/pull/1784.

Note that these benchmarks are with the current HWS model, which is a really easy model for the solver, so it is not a good stress test for the solver. When we introduced relaxation we did use more problematic models as well I believe which showed that there was a major benefit to using this type of monitored backtracking.

Issue #2133, opened 2025-03-05
Refactor parameter initialization in core


SouthEndMusic:
The function `parse_static_and_time` is very type unstable, because it does a lot of different things for a lot of different types. I propose the following stages in parameter initializations for nodes:

- Retrieve the node ids with `get_node_ids(db, node_type_string)`
- Initialize the node object making as much use of `@kwdef` and the `zeros(T, length(node_ids))` and otherwise `Vector{T}(undef, length(node_ids))`  patterns as possible, not caring yet about the actual parameter values
- Load the relevant tables with `load_structvector(db, config, MyTableV1)`
- Per parameter, call 

```julia
parse_parameter!(
   parameter::Vector{T}, 
   node_ids::Vector{NodeID}; 
   static::Union{StructVector, Nothing} = nothing,
   time::Union{StructVector, Nothing} = nothing,
   default::Union{Nothing, T},
   is_complete::Bool = true
)::Nothing where {T}
```

which is a trimmed down version of `parse_static_and_time`. Here the old `time_interpolatables` and `interpolation_type` are inferred from `T`. This for instance allows different fields of the same node type to have different interpolation types.

- Lastly, call a separate function `build_control_mapping!` to initialize control states, also split out from `parse_static_and_time`.

This approach makes the parameter parsing less opaque, and gets rid of the `parse_static_and_time` versus 'completely custom implementation' divide, for instance in the case of `TabulatedRatingCurve`. If it makes much more sense to parse certain parameters together, specific methods for that can be written. Another advantage of this approach is that the constructors of the node objects are less likely to bloat up when new features and/or fields are added.

PR #2251, opened 2025-04-22
Refactor parameter initialization

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2133.

This refactor makes parameter initialization more explicit and I think therefore easier to understand. It is also more type stable, because the type of each parameter is obtained from the fields in the `AbstractParameterNode` objects.

It is possible that this refactor changes untested behavior. One place where I noticed that the order of parsing matters is with `DiscreteControl`: the control state simulation starts out with depends on it. I had to change the initial control state to the last control state parsed to make all tests pass (although the failing test was a test model where the controlled flow depends on itself which I think is rare in practice).

### DIFF
--- a/core/src/allocation_optim.jl
+++ b/core/src/allocation_optim.jl
@@ -635,7 +635,7 @@ function save_demands_and_allocations!(
 
         if node_id.type == NodeType.UserDemand
             # UserDemand nodes
-            if user_demand.has_priority[node_id.idx, demand_priority_idx]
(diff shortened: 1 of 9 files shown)


visr (approved):
Looks like a clear improvement. Will hopefully also make it a bit easier to work on this part of the code.

visr:
Hmm this slowed down the HWS benchmark from 1 minute to 2-3 minutes. Not sure why CI didn't catch this. We'd have to look into it and can hopefully fix it without having to revert this.

PR #2256, opened 2025-04-24
Fix QGIS Docker CI; add pandas and return code

evetion:
in CI. Before we would actually ignore the exit code, and the CI has been silently failing on the pandas import since (at least) the flow vis results.

### DIFF
--- a/.docker/start.sh
+++ b/.docker/start.sh
@@ -1,7 +1,0 @@
-#!/usr/bin/env bash
-set -eux
-
-docker compose -f compose.yml up -d --force-recreate --remove-orphans
-echo "Installation of the plugin Ribasim"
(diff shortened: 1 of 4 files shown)


visr (approved):
This just fixes the issue rather than allowing failure right?
See #2257 for the Delwaq failure.

evetion:
> This just fixes the issue rather than allowing failure right? See #2257 for the Delwaq failure.

It does both. Before we would ignore any exit code, so the CI was green, even though there were failures: https://github.com/Deltares/Ribasim/actions/runs/14618848867/job/41013357755#step:4:1014

visr:
Yeah ok, "allowing failure" sounded to me like making it green despite errors, but that was the case before, but not anymore.
Thanks for the fix!

PR #2258, opened 2025-04-25
Use CVector for our state

### LINKED ISSUE #2231: Investigate adding a feature to `ComponentArrays.jl` to use it for the state without causing latency
That would probably be a `ComponentVector` where the partitioning of the underlying array is not part of the type information

visr:
Fixes #2231.

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -147,6 +147,8 @@ using DataStructures: OrderedSet, OrderedDict, counter, inc!
 
 export libribasim
 
+include("carrays.jl")
+using .CArrays: CVector
 include("schema.jl")
 include("config.jl")
 using .config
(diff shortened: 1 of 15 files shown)


SouthEndMusic (approved):
Looking good

visr:
Before merging we should look a bit more into this regression in the number of allocations:

```
hws_2024_7_0 (before, after)
29.171531 seconds (1.60 M allocations: 2.987 GiB, 1.07% gc time)
32.488307 seconds (11.39 M allocations: 3.779 GiB, 1.48% gc time)
```

For instance `DiffEqBase.ODE_DEFAULT_NORM(u, 0.0)` has (before, after):

```
24.303 ns
491.228 ns (5 allocs: 240 bytes)
```

visr:
I created #2262 to look into the allocation regression. Since I won't have time to look into it right now but we don't want this getting stale, let's merge.

Issue #2274, opened 2025-05-08
Instructions on Updating Ribasim


peter-n-jansson:
**Issue** I had trouble updating my Ribasim from 2024.11 to 2025.2. I had the 2024 version from the DSD but run into dependency issues (I heard the operative term was ‘dependency hell’) when I tried to add the 2025 ribasim to that environment. Instead, I needed to create a new pixi environment and set up everything from scratch, only bringing over the old scripts and data from the older version.  

**Proposal** Add instructions on updating versions to the documentation (new issue). I would recommend adding line-by-line commands to show how this can be done through pixi. Also, highlight which command lines allow you to add other python packages to the pixi environment.  

 

Here are the commands I used. Would recommend something like this to be pasted on the documentation: 

![Image](https://github.com/user-attachments/assets/073a9a03-5cef-4056-a912-9630b2fb366e)

visr:
Thanks for writing up your experience. Would be good to consider this together with #2248.

PR #2275, opened 2025-05-09
Add alternative installation method using pixi for Ribasim in the docs

simulutions:
While following the tutorial, I had trouble with installation. These steps should fix it.
Similar issue was already there: https://github.com/Deltares/Ribasim/issues/2274


docs/install.qmd
```diff
@@ -110,7 +110,14 @@ The Ribasim Python package is [registered in PyPI](https://pypi.org/project/riba
 ```sh
 pip install ribasim
 ```
-
+or create a pixi environment in your modelling project with:
+```sh
+pixi init && pixi add ribasim
```
evetion, line 115: Maybe best to remove the &&, not sure if it works in Windows, and you do the add ribasim below as well.

Huite: > not sure if it works in Windows

It'll work in batch, not in powershell; and we indeed don't know what people are using...

simulutions: Yes ofcourse, forgot to remove it. It's done now.

PR #2278, opened 2025-05-15
Support storage column in basin profile

### LINKED ISSUE #2254: Support `storage` column in `Basin / profile`
Since #280 we define profiles with $A(h)$, integrating in the core to get the $S(h)$. For many reservoirs only the latter is wrong, and asking users to convert to $A(h)$ is error-prone.

We can add a column `storage` such that users can fill this in instead of `area`.

simulutions:
Fixes #2254

Notes:

- supports now 3 cases of initialisation of Basins with either level-area, level-storage or both relations as input
- we interpolate level-area piecewise linear and storage-level is always done using the inverse of the integral of area-level, no matter how the basin was initialized


core/test/run_models_test.jl
```diff
@@ -676,3 +676,34 @@ end
...
+    @test success(model)
+end
+
+@testitem "init_basin_both_area_and_storage" begin
+    toml_path = normpath(
+        @__DIR__,
+        "../../generated_testmodels/basic_basin_both_area_and_storage/ribasim.toml",
+    )
+    @test ispath(toml_path)
+    model = Ribasim.run(toml_path)
```
SouthEndMusic, line 706: It's good to test that the models run with this new initialization (integration tests), but it would be nice to also test the output of your code (unit tests). For instance that the profiles are the same with the different initialization methods. You could also try to trigger some of the validation errors as we do in `validation_test.jl`.

python/ribasim_testmodels/ribasim_testmodels/basic.py
```diff
@@ -458,3 +458,219 @@ def cyclic_time_model() -> Model:
     model.edge.add(fb, bsn)
 
     return model
+
+
+def basic_basin_only_area_model() -> Model:
```
SouthEndMusic, line 463: It would be nice if you could prevent some code duplication here. Maybe split out a function with the schematization that these new test models have in common.

SouthEndMusic (commented):
Overall this looks good! I left a few comments. I'm also missing an update of the Basin documentation in `basin.qmd`.

core/src/read.jl
```diff
@@ -1790,22 +1773,70 @@ function load_structvector(
...
+"""trapezoidal integration"""
+function trapezoidal_int_area_to_storage(
+    dfdx::Vector{Float64},
+    x::Vector{Float64},
+)::Vector{Float64}
+    n = length(dfdx)
+    f = Vector{Float64}(undef, n)
+    f[1] = 0
+    for i in 1:(n - 1)
+        Δx = x[i + 1] - x[i]
```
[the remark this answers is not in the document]
simulutions: done

python/ribasim_testmodels/ribasim_testmodels/basic.py
simulutions: done

core/test/run_models_test.jl
simulutions: Done, both unit and validation test

core/src/read.jl
```diff
@@ -1847,3 +1878,45 @@ function set_concentrations!(
         end
     end
 end
+
+function interpolate_basin_profile_relations!(
+    basin::Basin,
+    profiles::StructVector{BasinProfileV1},
+)::Nothing
+    i = 0
+    for group in IterTools.groupby(row -> row.node_id, profiles)
```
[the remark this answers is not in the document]
simulutions: done

SouthEndMusic (approved):
I added some final comments, nice work!

PR #2292, opened 2025-05-22
Only run CI on relevant commits.

evetion:
I feel bad accepting a suggestion, or pushing too many small doc commits because I trigger so much CI. And [skipci] is not always applicable.


visr (approved):
Looks good to me. Lets try to pay attention to what runs on PRs for a while to make sure it is working as intended.

PR #2301, opened 2025-05-26
Use tasks for s3 download and upload

visr:
So we can use them as depends-on tasks in other workflows.

To be able to run the integration and regression tests locally with these tasks:

```
pixi run test-ribasim-regression
pixi run model-integration-test
```

Developers now need to go to https://s3-console.deltares.nl/ to set up personal access keys.
Make a copy of `.env.default` called `.env` and fill in the access key and secret key there. That will allow us to download from MinIO the reference data we need.


pixi.toml
```diff
@@ -83,6 +89,8 @@ test-ribasim-core-cov = { cmd = "julia --project=core --eval 'using Pkg; Pkg.tes
 test-ribasim-regression = { cmd = "julia --project=core --eval 'using Pkg; Pkg.test(julia_args=[\"--check-bounds=yes\"], test_args=[\"regression\"])'", depends-on = [
     "generate-testmodels",
     "test-ribasim-migration",
+    { task = "s3-download", args = ["benchmark/", "benchmark/"] },
+    { task = "s3-download", args = ["hws_migration_test/", "hws_migration_test/"] },
```
visr, line 93:
```suggestion
    { task = "s3-download", args = ["benchmark/", "benchmark/"] },
    { task = "s3-download", args = ["hws_migration_test/", "hws_migration_test/"] },
    "generate-testmodels",
    "test-ribasim-migration",
```

This was too late

PR #2308, opened 2025-05-29
Add debugging guide to the docs

evetion:
This can be expanded later, but this is already useful in the current stage of development.


visr (commented):
Rather than introducing a FAQ section, I feel like this should just be a new guide; https://diataxis.fr/how-to-guides/. Guides don't have to be long.

Also I learned something. I do use `saveat=0`, but was surprised that this can trigger a water balance error. The reason is this check is done during `save_flow`. Perhaps they should be decoupled, but that is another topic.

Issue #2262, opened 2025-04-29
CArray issues


visr:
~~Before merging~~ we should look a bit more into this regression in the number of allocations:

```
hws_2024_7_0 (before, after)
29.171531 seconds (1.60 M allocations: 2.987 GiB, 1.07% gc time)
32.488307 seconds (11.39 M allocations: 3.779 GiB, 1.48% gc time)
```

For instance `DiffEqBase.ODE_DEFAULT_NORM(u, 0.0)` has (before, after):

```
24.303 ns
491.228 ns (5 allocs: 240 bytes)
```

_Originally posted by @visr in https://github.com/Deltares/Ribasim/issues/2258#issuecomment-2835441709_

evetion:
This feature also broke our regression tests, some solvers produce a centimeter of difference, not really affected by tolerances at all.

visr:
A big part of this is fixed with `Base.iterate(x::CArray, state...) = iterate(getdata(x), state...)`. You can see if that is enough to fix the regression tests. But I still see mapreduce being slower.

evetion:
Make sure to implement iterate fully, correctly and test it like crazy. Iterate is the reason I'm like, told you so.

evetion:
> A big part of this is fixed with `Base.iterate(x::CArray, state...) = iterate(getdata(x), state...)`. You can see if that is enough to fix the regression tests. But I still see mapreduce being slower.

I can't reproduce OP regarding `iterate`. In both cases I have
```
julia> @time DiffEqBase.ODE_DEFAULT_NORM(u, 0.0);
  0.000003 seconds (1 allocation: 16 bytes)
```

visr:
Hmm this is what I saw benchmarking on 1.12, but probably better to do it on 1.11

```
using Ribasim, DiffEqBase, Chairmarks
@time m = Ribasim.run(raw"hws_2025_4_0\hws.toml")

u = m.integrator.u
v = Ribasim.getdata(u)

@b DiffEqBase.ODE_DEFAULT_NORM($u, 0.0)
@b DiffEqBase.ODE_DEFAULT_NORM($v, 0.0)
# u - 1.12 - 248 ns (after iterate 237 ns)
# v - 1.12 - 32 ns
```

After adding `iterate` `UNITLESS_ABS2` was equally fast. It should be noted that since DiffEqBase has a separate method for Array and AbstractArray that the code is different. Perhaps we can even add another method for CArray that looks more like the Array method.

[...]

PR #2309, opened 2025-05-30
Return of the ComponentArray

visr:
This reintroduces ComponentArrays to quickly address #2262. We should still get back to something like CArrays because it helps a lot with latency, but this should not come at the cost of some correctness issues that caused the regression tests to fail.


visr:
Luckily we don't need to merge this.

PR #2313, opened 2025-06-02
Transient basin forcing

### LINKED ISSUE #2291: Add reduction factor to flow expressions for LinearResistance and Manningresistance in docs

### LINKED ISSUE #2298: Cyclic_time model is not cyclic
When testing #2277 (and increasing the model time to a thousand years), I saw something odd in the cyclic time model, the precipitation is not cyclic, even though it should be:

Image

Not sure whether I missed something in the model that explains the lack of bumps in the second year, but otherwise this seems to be a bug in applying the extrapolation. I've investigated by checking the interpolation object (opening the model with `Model("path")`, and `itp = model.integrator.p.p_non_diff.basin.forcing.precipitation[1]`) but it seems to work fine. So it might not be called at the right timesteps?

@SouthEndMusic

SouthEndMusic:
Fixes #2298, fixes #2291

Notes:
- Refactored getting `tstops` from timeseries interpolations in type stable manner
- Refactored callbacks so that now always all of them are created
- Running 1000 year `cyclic_time` model now takes 15 seconds
- Concentrations cannot yet be periodic (silent problem)
- Fixed some unit formatting and added reduction factor for resistance nodes in docs

### DIFF
--- a/core/src/callback.jl
+++ b/core/src/callback.jl
@@ -26,12 +26,9 @@ function create_callbacks(
         FunctionCallingCallback(update_cumulative_flows!; func_start = false)
     push!(callbacks, cumulative_flows_cb)
 
-    if config.experimental.concentration
-        # Update concentrations
-        concentrations_cb =
-            FunctionCallingCallback(update_concentrations!; func_start = false)
-        push!(callbacks, concentrations_cb)
-    end
+    # Update concentrations
+    concentrations_cb = FunctionCallingCallback(update_concentrations!; func_start = false)
+    push!(callbacks, concentrations_cb)
 
     # Update Basin forcings
     # All variables are given at the same time, so just precipitation works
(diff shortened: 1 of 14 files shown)


core/src/callback.jl
```diff
@@ -26,36 +26,29 @@ function create_callbacks(
...
-    times = [itp.t for itp in basin.forcing.precipitation]
-    tstops = Float64[]
-    for t in times
-        append!(tstops, t)
-    end
-    unique!(sort!(tstops))
+    tstops = Vector{Float64}[]
+    t_end = seconds_since(config.endtime, config.starttime)
+    get_timeseries_tstops!(tstops, t_end, basin.forcing.precipitation)
+    tstops = sort(unique(vcat(tstops...)))
```
visr, line 38:
```suggestion
    tstops = sort(unique(reduce(vcat, tstops)))
```

visr (approved):

PR #2324, opened 2025-06-02
Don't reresolve test environment

visr:
See https://pkgdocs.julialang.org/v1/api/#Pkg.test Since we check in our Manifest, those are the versions we want to use.
Might help a bit with test duration.


visr:
Wow compared to #2318, Windows run testmodels goes from 31 to 23 minutes, and Windows Julia tests goes from 35 to 27 minutes.

evetion (approved):

Issue #2329, opened 2025-06-02
Regression test failure


visr:
As commented in https://github.com/Deltares/Ribasim/issues/2262#issuecomment-2912786112, our regression tests are currently broken. It seemed that the CArray PR caused it, but when we revert it locally the issue doesn't go away. This is possibly because 8b8768b2f27ee0eeaf0928b22aec59da11124b26 has been merged since. Also #2277 is relevant, since higher tolerances give the different algorithms more freedom to diverge. Still it is not clear why the CArray PR would cause a difference here. The only thing we can think of is due to creating copies where mutations are expected inside the solvers, leading to differences.

The original failing test after merging #2258 can be found at https://dpcbuild.deltares.nl/buildConfiguration/Ribasim_RegressionTest_Linux/5731166 and started reporting 8 errors:

```
regression_ode_solvers_trivial
	QNDF sparse=true/false autodiff=true/false (4 failures)
	Rodas5P sparse=false autodiff=true/false (2 failures)
	ImplicitEuler sparse=false autidiff=true/false (2 failures)

Expression: basin.storage[end] ≈ 16.530443f0
Evaluated: 16.54382866633898 ≈ 16.530443f0

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18189103978862112 ≈ 0.181817438

Expression: basin.storage[end] ≈ 16.530443f0
Evaluated: 16.54382866634262 ≈ 16.530443f0

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18189103978864113 ≈ 0.181817438

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18181745391330392 ≈ 0.181817438

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18181745491110854 ≈ 0.181817438

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18183812096947194 ≈ 0.181817438

Expression: basin.level[end] ≈ 0.181817438
Evaluated: 0.18183447857871027 ≈ 0.181817438
```

With #2277 the number of failures went up from 8 to 14, which is understandable due to higher tolerances.
8b8768b2f27ee0eeaf0928b22aec59da11124b26 should be reverted (done in 72b3b51abb5be76d0e4e03efec5c88f8dba75807), it further raised the number of errors from 14 to 52.

PR #2331, opened 2025-06-02
Revert "Use CVector for our state (#2258)"

visr:
This reverts commit c16280eb0f0b4d47bb144fc3a2382a5c5c8667bb (#2258).
To look more into #2329.


evetion:
@visr Your allow_reresolve=false might've fixed it when downgrading to the old manifest. I swear I have diffed the test output in terms of package hashes, but not good enough it seems.

evetion:
🎉 With the old tolerances we're back to normal. That it works on mac with the new could be mac specific.

I've seen this pass *with* the cvector too (!), so now we need to narrow down the package changes that triggered it.

PR #2339, opened 2025-06-03
support storage column in basin profile 2

simulutions:


python/ribasim/ribasim/__init__.py
```diff
@@ -1,6 +1,6 @@
-__version__ = "2025.3.0"
+__version__ = "2025.4.0"
 # Keep synced write_schema_version in ribasim_qgis/core/geopackage.py
```
evetion, line 2: This still needs to be done

evetion (approved):
Could you rebase your latest three commits on main?

python/ribasim/ribasim/__init__.py
```diff
@@ -1,6 +1,6 @@
-__version__ = "2025.3.0"
+__version__ = "2025.4.0"
```
evetion, line 1: And let's revert this, it was an example change I made. Sorry for that.

python/ribasim/ribasim/__init__.py
simulutions: done

PR #2341, opened 2025-06-03
BMI: return 1 on update if solve not successful

visr:
This changes the return code of update and update_until to only return 0 if the solve was successful.
@HendrikKok does this seem ok to you? Or should other methods be changed as well? For other methods like `get_current_time` it is a bit unclear to me what a failure would mean. I assume that it failed to get the current time, and not that the solve of the last update was unsuccesful.


evetion (approved):
It's a bit ugly that the macro returns in failures, but you have to pass success yourself. Can't you add `return update_retcode(model)` to the macro as well?

PR #2342, opened 2025-06-04
Generate a Ribasim-python SBOM

deltamarnix:
Similar to the julia SBOM generation, I've generated an SBOM via sbom4python. The only thing I've done is pass on my module `-m ribasim`, and it's still a bit magic to me which packages are then picked up. I will have to learn more from the documentation, but at least there is an SBOM with information.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -36,13 +36,6 @@ initialize-julia = { depends-on = [
     "update-registry-julia",
     "instantiate-julia",
 ] }
-generate-sbom-ribasim-core = { cmd = "julia --project --check-bounds=yes -- utils/generate-sbom.jl", depends-on = [
-    "instantiate-julia",
-], inputs = [
-    "core/Project.toml",
-], outputs = [
-    "Ribasim.spdx.json",
-] }
 # Docs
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -170,6 +174,18 @@ ribasim-core-testmodels = { cmd = "julia --project --threads=4 utils/testmodelru
...
 
+[tasks.generate-sbom-ribasim-core]
+cmd = "julia --project -- utils/generate-sbom.jl"
+depends-on = ["instantiate-julia"]
+inputs = ["core/Project.toml"]
+outputs = ["Ribasim.spdx.json"]
+
+[tasks.generate-sbom-ribasim-python]
+cmd = "sbom4python -m ribasim --system --sbom spdx --format json --output Ribasim-python.spdx.json"
+env = { PATH = "$PATH;$CONDA_PREFIX\\Lib\\site-packages\\magic\\libmagic" }
```
deltamarnix, line 185: Likely not going to work on linux. What is a smart way to make this work for both?

deltamarnix: Forward slashes work, but I need a semi-colon, while linux wants colons.

PR #2343, opened 2025-06-04
migration for new storage column

simulutions:
fixes the migration for models that were build before we supported storage profiles in basins


evetion (approved):
Looking good, I think you only miss a line in `ribasim_qgis/core/nodes.py` to add the field storage. You can follow https://ribasim.org/dev/addnode.html (applies to new nodes, but useful for new fields to nodes as well) to check the usual code locations.

Issue #2336, opened 2025-06-03
CI check-bounds settings


visr:
This can possibly improve CI runtime.
See https://github.com/Deltares/Ribasim/pull/2332#issuecomment-2932330170.

PR #2345, opened 2025-06-04
Use `check-bounds=yes` when starting Julia

evetion:
Per suggestions in #2336. We choose to run check-bounds=yes everywhere. For longer runner models this decreases performance (I see a 20% drop), but we should precompile less overall, skipping the several minute hit on CI.

Other CI fixes:
- Timeout to 60 minutes (as we have 6 hours per #2338).
- Move the TC cache generation to the upper project, as regression/integration couldn't make find it now.

### DIFF
--- a/.github/workflows/core_testmodels.yml
+++ b/.github/workflows/core_testmodels.yml
@@ -19,6 +19,7 @@ jobs:
   test:
     name: Julia ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


visr:
Ah yes so you're suggesting to use it for everything, including integration/regression tests? That is probably fine since we care mostly about relative differences.

Should we still add it to the rest of the julia commands as well, like `update-manifest-julia`, `generate-sbom-ribasim-core` etcetera?

And we can probably remove `julia_args=[\"--check-bounds=yes\"]` from Pkg.test calls right, since that is the default?

Even the build task we can start with `--check-bounds=yes`, given this won't be used for the actual compilation:

https://github.com/JuliaLang/PackageCompiler.jl/blob/57505caad062b5b1985bcbbb2fb15b9b6e9e6d99/src/PackageCompiler.jl#L1031C4-L1032

And I think we should probably add this to the vscode settings.json:

```
    "julia.additionalArgs": [
        "--check-bounds=yes"
    ],
```

Issue #2300, opened 2025-05-22
How are "tops" of tables defined?


peter-n-jansson:
When defining stage-discharge relations for tabulated rating curves, there is some unclarity on how the tops of tables are defined. For example, here, I would like the max gate capacity to be 2076.5 but was told to add an incremental increase 2017.51 so the value can be extrapolated properly. Is this necessary? 

 **Proposal:** in the documentation, provide explicit instructions on how to define different types of "tops" for such tables, typically for gates/trc:
- stable top: is the increment necessary?
- finite top: what if the gate closes after a certain level is reached?
- extrapolating top 

Another question: in RIBASIM 7, the final entry in the stage-discharge tables for gates tend to be -1. This was apparently common practice, that users interpreted to signify the end data entry. But what does the -1 entry actually do, and is that functionality worth bringing forth to the new ribasim?

![Image](https://github.com/user-attachments/assets/4aefbc13-5716-4128-aaed-f740dd823097)

SouthEndMusic:
@visr we could also as a follow-up of https://github.com/Deltares/Ribasim/issues/2281 add some extrapolation configuration. That wouldn't add any latency because the extrapolation type is not part of the type parameters.

visr:
@SouthEndMusic let's keep that one in the pocket for now unless we see more people finding it confusing.

@peter-n-jansson I forgot to mention about the -1, that is just a way for users to tell it is the end of the relation, because the relation was a fixed length array. We don't use fixed length relations.

peter-n-jansson:
> By the way also feel free to just click the pencil in https://github.com/Deltares/Ribasim/blob/main/docs/reference/node/tabulated-rating-curve.qmd and start adding or modifying things that you find confusing.


Just made some amendments to clarify

peter-n-jansson:
Strange, I made two merge requests with separate edits using the pencil tool. But I don't see it on the history. I'll make them again on a fork and send a pull request

peter-n-jansson:
![Image](https://github.com/user-attachments/assets/25444d2e-3e04-4b90-8b51-ee980f38919c)

Now in pull request #2346. Apparently pending review. If there's a (rookie) mistake somewhere, let me know!

PR #2346, opened 2025-06-04
Adjustments to the tabulated rating curve

peter-n-jansson:
As discussed on issue #2300


visr (approved):
Thanks for your contribution. It agree it is a good idea to add the example of capping the flow rate and the hint on circumventing decreasing flow rates.

I looked at the whole section a bit and decided it needed some restructuring (already before this PR). So I added that here, please have a look if you agree with this. We had some discussion of the interpolation spread over two sections with two different example Q(h) relations. I now moved this all to the interpolation section and switched to one example.

PR #2347, opened 2025-06-04
Simplify exact integration

SouthEndMusic:
I committed these changes directly to main by mistake. Please have a look at the first 2 commits shown below.

These changes make sure that there is a single source of truth for exact integration, before that code was found both in `set_current_basin_properties!` to compute the exact integrations up to the `t` at which the RHS is called and in `update_cumulative_flows!` to compute the exact integration up to the end of the latest successful timestep.


visr (approved):
Looks good to me! Can be closed :)

PR #2359, opened 2025-06-05
Remove backtracking, add more caching

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2189
Fixes https://github.com/Deltares/Ribasim/issues/1778


visr:
Unless I messed up something in my benchmarking, it seems that the HWS time profile still looks about the same as https://github.com/Deltares/Ribasim/pull/2330#issuecomment-2939557331, but the allocation situation has improved, going from ~90% to ~70% in relaxation.

The issue with the profiler saying there is GC and dynamic dispatch around the closure remains. Still I would've expected a speedup if this does less iterations. Instead I see a 2x regression that I don't understand. At least a part of that is due to now running with `check-bounds=yes`, so not completely fair. This is still a bit of a mystery to me. Great to have progress on this front though.

SouthEndMusic:
I've been fiddling around quite a bit in this branch, but in the end:
- Backtracking was removed completely, as it didn't do anything other than take time
- Most time interpolations are now cached. I introduced a `TimeDependentCache` for that, and renamed the existing `DiffCache` `StateTimeDependentCache`, to make more clear what they are. Lastly I renamed `ParametersNonDIff` to `ParametersIndependent` for consistency.

SouthEndMusic:
I don't understand what went wrong in CI, I cannot reproduce the failing `minimal_subnetwork` model or the `sparse and AD/FDM jac solver options` tests locally

visr:
Restarting CI fixed one of them, the other timed out. It isn't clear to me why CI takes so much longer on this branch, and why there are spurious failures. That makes me a bit hesitant to merge this as one. Perhaps we should first remove relaxation separately?

visr (approved):
Hmm I see the CI time regression also on main, so this branch doesn't seem responsible for it. Great, then we can merge it at once.

PR #2360, opened 2025-06-06
Error message when no basins

simulutions:


core/src/model.jl
```diff
@@ -172,6 +172,10 @@ function Model(config::Config)::Model
     @debug "Read database into memory."
 
     u0 = build_state_vector(parameters.p_non_diff)
+    if length(u0) == 0
+        error("Models without states are unsupported, please add a Basin node.")
+    end
```
visr, line 177:
```suggestion
    if isempty(u0)
        @error "Models without states are unsupported, please add a Basin node."
        error("Model has no state.")
    end
```

This is a bit of a weird pattern but we tend to use this. The error macro throws a clear user facing error message that is not fatal, execution continues after. During validation we tend to log all the errors first, and the throw an error saying validation failed.

In this case we want to stop immediately so an error log is followed by an error.

Perhaps in a separate PR we can look into improving this, with e.g. a fatal log that throws, or deciding that the error message on top of the stacktrace is readable enough in the log file and stdout that we can only error.

core/test/run_models_test.jl
```diff
@@ -742,3 +742,14 @@ end
...
+
+@testitem "model without basin" begin
+    toml_path = normpath(@__DIR__, "../../generated_testmodels/no_basin/ribasim.toml")
+    @test ispath(toml_path)
+    model = Ribasim.run(toml_path)
+    @test model isa Ribasim.Model
+    @test success(model)
+
+    (; p_non_diff) = model.integrator.p
+    @test p_non_diff.basin === nothing
```
visr, line 754: I guess this model name should start with "invalid_" so it is not expected to run in the run test models CI. You can have a look at the other invalid models to see how we tend to test for error logs etc.

visr (approved):

PR #2361, opened 2025-06-10
Switch to latest `SparseConnectivityTracer` release

SouthEndMusic:
It took some digging, but it turns out we needed a `Base.map` overload for `CArrays` to make it work.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2337,11 +2337,9 @@ version = "1.11.0"
 
 [[deps.SparseConnectivityTracer]]
 deps = ["ADTypes", "DocStringExtensions", "FillArrays", "LinearAlgebra", "Random", "SparseArrays"]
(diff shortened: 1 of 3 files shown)


SouthEndMusic:
Ah I found what the problem is, interesting that this didn't show up earlier. The problem is that for the `basic` model the (initial) infiltration is 0 (and the initial potential evaporation is not). Because of this 0, the effect of of the infiltration is not propagated. 

This is worth investigating more broadly, because I've noticed before that when there are false negatives in the Jacobian sparsity, models do run but much slower.

SouthEndMusic:
Turns out that things are fixed on the new SCT release from today, see also https://julialang.slack.com/archives/C6G240ENA/p1749564441760959

visr (approved):
Great, thanks. I added tests and removed the map! overload which didn't seem to be fully correct or necessary.

PR #2366, opened 2025-06-11
Use the Kahan algorithm for summing flows

SouthEndMusic:
I wondered whether this could help with convergence or decrease water balance errors. I have found no indication that it does significantly, but it might be interesting for future unstable models.


SouthEndMusic:
Closing this as not needed

PR #2370, opened 2025-06-12
Write log file via BMI

### LINKED ISSUE #1351: Write ribasim.log also when using BMI
Currently the logger is created in `Ribasim.main`, which is the CLI entrypoint. When using BMI this function is not used, so no logfile is created.

visr:
Fixes #1351

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -25,10 +25,7 @@ function BMI.finalize(model::Model)::Nothing
     log_finalize(model)
     logger = global_logger()
     try
-        # Pick the IOStream out of our composed LoggingExtras.jl logger,
-        # the FileLogger contains the file handle.
-        # This is internal API, but our unit tests cover it.
-        io = logger.logger.loggers[1].logger.logger.stream
+        io = logger_stream(logger)
         close(io)
     catch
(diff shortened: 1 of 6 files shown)


core/src/bmi.jl
```diff
@@ -12,6 +22,14 @@ Write all results to the configured files.
 """
 function BMI.finalize(model::Model)::Nothing
     write_results(model)
+    log_finalize(model)
+    logger = global_logger()
+    try
+        io = logger.logger.loggers[1].logger.logger.stream
```
evetion, line 28: 😱

evetion, line 28: Please rewrite this somehow, or explain why the rabbithole is so deep. Maybe also have a specific catch here?

evetion (approved):
LGTM, happy to see log support for BMI.

core/src/bmi.jl
visr, line 28:
```suggestion
        # Pick the IOStream out of our composed LoggingExtras.jl logger,
        # the FileLogger contains the file handle.
        # This is internal API, but our unit tests cover it.
        io = logger.logger.loggers[1].logger.logger.stream
```

PR #2371, opened 2025-06-12
Go back to an older SparseConnectivityTracer

visr:
Effectively reverting #2361. Since that was merged some CI jobs have seen a large slowdown. I want to check if this was responsible.

For instance the docs used to build in 34m, but now time out at 60m.

Before the SCT update at 28965e41f1a2c52fafbf41e730001ad5d8ab1dcb we had
Run Testmodels Ubuntu was 24m (of which 22m the run models step, including that since it is less dependent on caching)
Run Testmodels Windows was 32m (24m)

After the SCT update at 93ebc169fca620e4f6cee6f6f01bfdb2e05cbf3e we had
Run Testmodels Ubuntu was 36m (35m)
Run Testmodels Windows was 56m (38m)


visr:
I think the testmodels slowdown must have been something else, perhaps even a GitHub Actions disruption, because main is currently showing roughly the same times again.

On this PR the docs still time out, though they have been timing out every time since updating SCT, taking up all time in the Render Quarto Project step. What is suspicious is that the last thing it is doing is always the discrete control reference page, let's see if I can reproduce that.

```
[48/57] reference/node/discrete-control.qmd

Starting julia-1.11 kernel...Done

Executing 'discrete-control.quarto_ipynb'
Error: The operation was canceled.
```

visr:
Hmm running again (515127fad2d43253f74f29e46bf093c5257075ae) shows the timeout happening earlier. So something has made quarto over 2x slower. But nothing we did, because the quarto version stayed the same.

PR #2372, opened 2025-06-13
Replace docs code block with markdown

visr:
Seeing if this helps with https://github.com/Deltares/Ribasim/pull/2371#issuecomment-2969343086. I can run the snippet fine locally.

```
df |> markdown_table(String) |> clipboard
```

Also make quarto more verbose.


visr:
This isn't it, see https://github.com/Deltares/Ribasim/pull/2371#issuecomment-2969512114

PR #2373, opened 2025-06-16
Test `LinearResistance` and `TabulatedRatingCurve` in allocation

### LINKED ISSUE #2148: Linearize `TabulatedRatingCurve` curve for allocation

SouthEndMusic:
Fixes the `LinearResistance` part of #2149, fixes #2148.

Here's a comparison of `LinearResistance` flow in a simple model between the physical layer and allocation:

image

It even takes `max_flow_rate` into account properly!

Here's a similar comparison of `TabulatedRatingCurve` flow:

image

### DIFF
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -592,12 +592,13 @@ function add_linear_resistance!(
 
                 Δlevel_min = min_inflow_level - max_outflow_level
                 Δlevel_max = max_inflow_level - min_outflow_level
+                Δlevel_max_flow = resistance[node_id.idx] * max_flow
 
                 input = [-Δlevel_max_flow, Δlevel_max_flow]
                 output = [-max_flow, max_flow]
 
                 if Δlevel_min < -Δlevel_max_flow
-                    pushfirst!(input, Δllevel_min)
+                    pushfirst!(input, Δlevel_min)
                     pushfirst!(output, -max_flow)
                 end
 
--- a/core/src/allocation_init.jl
+++ b/core/src/allocation_init.jl
@@ -103,8 +103,9 @@ function add_basin!(
     values_storage = Dict{NodeID, Vector{Float64}}()
     values_level = Dict{NodeID, Vector{Float64}}()
 
-    # TODO: Not just split in segments given by input data, but look for changes in slope
+    # TODO: Use DouglasPeucker algorithm here and for TabulatedRatingCurve
 
+    # TODO: Find lowest level within subnetwork of basins and level boundaries
     lowest_level = minimum(itp -> itp.t[1], level_to_area)
 
     for node_id in basin_ids_subnetwork
@@ -607,7 +608,6 @@ function add_linear_resistance!(
                     push!(output, max_flow)
                 end
 
-                Δlevel_max_flow = resistance[node_id.idx] * max_flow
                 piecewiselinear(problem, Δlevel, input, output)
             end
         end,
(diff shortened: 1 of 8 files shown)


SouthEndMusic:
I've implemented a version of the [Douglas-Peucker algogirthm](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm) that looks at the relative error w.r.t. the original piecewise linear relation:

image

This can reduce the number of points while stil having a good approximation of the relation (we care about this number of points as emphasized by @jarsarasty as it can have a large performance aspect in optimization). Although the level-storage relationship is piecewise quadratic we can also use this there, because we can just sample that at many points and then apply this algorithm.

This may be of interest to @evetion, as I found when looking into these algorithms that they have a large application to geological data.

evetion:
> This may be of interest to @evetion, as I found when looking into these algorithms that they have a large application to geological data.

Geospatial you mean? For sure, they're famous and widely used. One would often go for the Visvalingam-Whyatt algorithm, which can produce nicer geometries, not sure if that's of interest here. Also note that these algorithms have been implemented in Julia over at https://juliageo.org/GeometryOps.jl/stable/source/transformations/simplify

SouthEndMusic:
Yeah I saw that those algorithms are implemented there, but I figured that would be an unnecessary dependency and it looks like this only supports an absolute error while @gijsber prefers a relative error

core/src/allocation_init.jl
```diff
@@ -103,8 +103,9 @@ function add_basin!(
     values_storage = Dict{NodeID, Vector{Float64}}()
     values_level = Dict{NodeID, Vector{Float64}}()
 
-    # TODO: Not just split in segments given by input data, but look for changes in slope
+    # TODO: Use DouglasPeucker algorithm here and for TabulatedRatingCurve
```
jarsarasty, line 106: I want to suggest an algorithm that does not require an epsilon parameter for excluding points.

core/src/allocation_optim.jl
```diff
@@ -512,55 +503,73 @@ function save_allocation_flows!(
...
     objective::AllocationObjective,
-    optimization_type::AllocationOptimizationType.T,
 )::Nothing
-    args = (allocation_model, integrator, objective, optimization_type)
     if objective.type == AllocationObjectiveType.demand
-        optimize_for_demand!(args...)
+        set_demands!(problem, p_independent, objective)
     elseif objective.type == AllocationObjectiveType.source_priorities
-        optimize_for_sources!(args...)
+        nothing
```
jarsarasty, line 524: Should we log a message indicating that source priority is not yet supported?

core/src/allocation_optim.jl
```diff
@@ -512,55 +503,73 @@ function save_allocation_flows!(
...
+        )
+    elseif objective.type == AllocationObjectiveType.source_priorities
+        nothing
+    end
     return nothing
 end
 
-function optimize_for_demand!(
+function optimize_for_objective!(
     allocation_model::AllocationModel,
```
jarsarasty, line 561: Shall we include docstrings?

core/src/allocation_util.jl
```diff
@@ -134,3 +134,53 @@ function get_minmax_level(p_independent::ParametersIndependent, node_id::NodeID)
...
+    u::Vector{T}
+    t::Vector{T}
+    selection::Vector{Bool} = zeros(Bool, length(u))
+    rel_tol::T
+end
+
+"""
+Perform a modified Douglas-Peucker algorithm to down sample the piecewise linear interpolation given
+by t (input) and u (output) such that the relative difference between the new and old interpolation is
+smaller than ε_rel on the entire domain when possible
```
jarsarasty, line 148: (rel_tol)

PR #2381, opened 2025-06-16
Switch to QuartoNotebookRunner

### LINKED ISSUE #2078: Try QuartoNotebookRunner.jl to speed up docs build
Currently our docs build take 26 minutes. Our example notebooks take up a lot of that time. They are executed in Python, and then run the core with subprocess starting julia for each model. So it needs to load and compile each time. I suspect that if we run from Julia instead it will be faster since compiled code remains available in the session. With QuartoNotebookRunner v0.13 it should be possible to combine both Python and Julia.

https://discourse.julialang.org/t/ann-native-julia-engine-for-quarto-using-quartonotebookrunner-jl/112753/56

visr:
Fix #2078. Continued from #2376.

Updates the old branch https://github.com/Deltares/Ribasim/compare/main...qmd

https://quarto.org/docs/computations/julia.html

### DIFF
--- a/docs/guide/delwaq.qmd
+++ b/docs/guide/delwaq.qmd
@@ -151,13 +151,18 @@ We can use this table to plot the results of the Delwaq model, both spatially as
 from ribasim.delwaq import plot_fraction
 
 plot_fraction(nmodel, 1)  # default tracers, should add up to 1
+plt.draw()
 plot_fraction(nmodel, 9, ["Foo", "Bar"])  # custom tracers
+plt.draw()
 plot_fraction(nmodel, 9, ["Continuity"])  # mass balance check
+plt.draw()
 ```
 
 ```{python}
 from ribasim.delwaq import plot_spatial
 
 plot_spatial(nmodel, "Bar")
+plt.draw()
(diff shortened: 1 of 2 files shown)


visr:
At first this was hanging on drawing plots, but now it finishes after 22 minutes, quite a nice speedup.

The downside are
- this crashes on Windows, cause unknown, hard to debug due to https://github.com/PumasAI/QuartoNotebookRunner.jl/issues/277
- I didn't manage to get the plots to show inline yet

It seems to work better from WSL. Though on Windows I was able to get a seaborn plot inline just like the example in https://github.com/PumasAI/QuartoNotebookRunner.jl/pull/255, though a simple matplotlib example fails, see also https://github.com/visr/qnr-plot-mwe.

The showing of inline results is also different because this is not jupyter, where we can use ipython magic and such, but just PythonCall. Perhaps there is more we can do configuring matplotlib and plotly backends to work well in this scenario, and can also  engage upstream.

https://plotly.com/python/renderers/
https://matplotlib.org/stable/users/explain/figure/backends.html

evetion:
Have you tried displaying plots using https://quarto.org/docs/authoring/figures.html#layout?

Issue #2319, opened 2025-06-02
LevelDemand in allocation


SouthEndMusic:
For the LevelDemand node in the new allocation implementation:
- we must support demands of multiple priorities. The input should look like a sequence of levels defining intervals where priorities decrease with level
- we must be tested: some existing tests must be adapted, possibly new tests have to be added 
- we must make a final decision on whether maximum levels will also be supported. If so, it might be easiest do handle minimum levels and maximum levels in separate goals

SouthEndMusic:
This has been implemented, not tested

PR #2386, opened 2025-06-19
Expand and test `LevelDemand` functionality

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2319.
Fixes https://github.com/Deltares/Ribasim/issues/2418.


SouthEndMusic:
Here's how I now implement the level demand in allocation:

For a given Basin we have the the storage $S$ relative lower error $E_\text{rel, lower}$ and the relative upper error $E_\text{rel, lower}$. For a certain priority $p$ we have a minimum target level $h^p_\text{min}$ and a maximum target level $h^p_\text{max}$, and the level to storage function $S(h)$. Lastly we have the fair fractions $f_\text{lower}$ and $f_\text{upper}$. Then we have this constraint for the lower error:

$$
   S(h_\text{min}^p) \cdot E_\text{rel, lower} \ge f_\text{lower} \cdot  S(h_\text{min}^p) - S
$$

and this constraint for the upper error:

$$
    S(h_\text{max}^p) \cdot E_\text{rel, upper} \ge S - f_\text{upper} \cdot  S(h_\text{max}^p) 
$$

For now I keep the fair fractions $1$, I'll compute them properly in a follow-up PR. Note that this formulation is simpler than what I wrote in https://github.com/Deltares/Ribasim/issues/2418; I decided for simplicity that I do not compute the error with respect the result of the previous optimization (from the same timestep), but just the deviation from the target storage.

@simulutions @jarsarasty

SouthEndMusic:
I updated the Julia Tests workflow so that now when allocation fails, the failing problem together with analysis data is written as an artifact that can be downloaded from Github 👍

simulutions (commented):
Looks good, big one though. Lets have a discussion tomorrow about how everything is implemented. The code thouches a lot of aspects I can learn about

simulutions (approved):

PR #2387, opened 2025-06-19
Use mirrors-typos to avoid tag warnings

### LINKED ISSUE #2351: Typos v1 mutable reference warning
I saw this already since #2186, and pre-commit gave a warning, but it is getting noisier and more annoying, now also with popups in VSCode, saying:

```
[WARNING] The 'rev' field of repo 'https://github.com/crate-ci/typos' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.
```

The problem is that "Update pre-commit hooks" just takes the latest tag, and that is now almost always this v1 tag. Ideally it would always take the latest normal tag.

https://github.com/crate-ci/typos/releases/tag/v1
https://github.com/crate-ci/typos/releases/

Upstream issue and possible workaround: https://github.com/crate-ci/typos/issues/390#issuecomment-2764488396

visr:
Fixes #2351

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
     rev: 0.8.1
     hooks:
       - id: nbstripout
-  - repo: https://github.com/crate-ci/typos
+  - repo: https://github.com/adhtruong/mirrors-typos
     rev: v1.33.1
     hooks:
       - id: typos


evetion (approved):
Ok for fixing the warnings, but bit nasty to use a unknown mirror of it, seems like a good candidate for a supply chain attack.

PR #2391, opened 2025-06-20
Switch from UMFPACK to KLU sparse linear solver

visr:
I was talking to @Huite about linear solvers, and the last time I compared some options was in https://github.com/Deltares/Ribasim/pull/2043#issuecomment-2642537825.

I did benchmarks using `AmstelGooienVecht_parameterized_2025_6_13` since it is quite large and uses daily varying meteo forcing.

```
UMFPACKFactorization()
18.932428 seconds (20.04 M allocations: 6.932 GiB, 2.68% gc time)
PardisoJL(; vendor = :MKL)
47.083817 seconds (19.95 M allocations: 3.339 GiB, 0.60% gc time)
MKLPardisoIterate()
43.992427 seconds (19.95 M allocations: 3.339 GiB, 0.70% gc time)
MKLPardisoFactorize()
48.077815 seconds (19.95 M allocations: 3.339 GiB, 0.67% gc time)
KLUFactorization()
10.374421 seconds (19.93 M allocations: 1.061 GiB, 1.13% gc time)
```

So Pardiso is 2x slower than UMFPACK, but KLU is 2x faster, with a lot less allocated memory as well. This is probably better than MKL since that doesn't really support Apple anymore. As mentioned in [the LinearSolve docs](https://docs.sciml.ai/LinearSolve/stable/tutorials/accelerating_choices/#Understanding-Performance-of-Sparse-Linear-Solves), the best choice is highly problem dependent.

### DIFF
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -21,6 +21,7 @@ using OrdinaryDiffEqTsit5: Tsit5
 using OrdinaryDiffEqSDIRK: ImplicitEuler, KenCarp4, TRBDF2
 using OrdinaryDiffEqBDF: FBDF, QNDF
 using OrdinaryDiffEqRosenbrock: Rosenbrock23, Rodas4P, Rodas5P
+using LinearSolve: KLUFactorization
 
 export Config, Solver, Results, Logging, Toml
 export algorithm,
@@ -313,6 +314,9 @@ function algorithm(solver::Solver; u0 = [])::OrdinaryDiffEqAlgorithm
 
     if algotype <: OrdinaryDiffEqNewtonAdaptiveAlgorithm
         kwargs[:nlsolve] = NLNewton()
+        if solver.sparse
+            kwargs[:linsolve] = KLUFactorization()
+        end
     end
 
     if function_accepts_kwarg(algotype, :step_limiter!)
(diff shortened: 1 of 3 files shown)


visr:
> Rijkswaterstaat/modellen/lhm_peil_2025_6_0

How long does this run for you? It is very slow and not finished. Indeed good to check, perhaps for a short time period. Possibly we need to switch to an iterative method as suggested in the docs:

> A Krylov subspace method with proper preconditioning will be better than direct solvers when the matrices get large enough.

Though for that it would be good to see how large of a fraction the linear solve takes. In HWS it was 30% on main, on AGV (bigger model) 50%. This PR cut the 50% down to 20%. Would be good to know that fraction in the model you linked, since it gives an indication for speedup.

For iterative solvers we'd need to look more into preconditioners and such. Probably better to merge this first as a new baseline, in case it scales well to LHM.

visr:
After merging #2114 it is about 1.5x slower in my no-specialize dev setup, which is to be expected. Interestingly the amount of allocations went down a ton in no-specialize as well.

KLU pre and post #2114
```
10.374421 seconds (19.93 M allocations: 1.061 GiB, 1.13% gc time)
15.553228 seconds (79.08 k allocations: 782.321 MiB, 0.54% gc time)
```

I tested a 5 minute run of `lhm_peil_2025_6_0`, and there I didn't observe any significant difference with or without KLU.

evetion:
I've ran lhm_peil_2025_6_0 and this PR is twice as fast (53 minutes vs 1h 52 minutes).

SouthEndMusic (approved):
Good stuff!

PR #2395, opened 2025-06-23
Test piecewise linear basin profile

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2060


simulutions (approved):
test makes sense! Maybe a docstring or comment here and there be helpfull for me but I approve. Also dont forget the failing test

PR #2401, opened 2025-06-24
control by allocation

simulutions:
Fixes https://github.com/Deltares/Ribasim/issues/714.
Fixes https://github.com/Deltares/Ribasim/issues/2420 (see also https://github.com/Deltares/Ribasim/pull/2421).


core/src/util.jl
```diff
@@ -1152,3 +1152,28 @@ function eval_time_interp(
...
+function should_skip_update_q(
+    active::Bool,
+    control_type::ControlType.T,
+    control_type_::ControlType.T,
+    p::Parameters,
+)::Bool
+    (; p_mutable) = p
+    (; all_nodes_active) = p_mutable
+
+    # Update is not needed if inactive and all nodes are not active
```
SouthEndMusic, line 1165: I think it's more clear to write `if active || all_nodes_active`, and describe it as 'either the node is actually active or it is forced to be active for sparsity detection'

core/src/util.jl
```diff
@@ -1152,3 +1152,28 @@ function eval_time_interp(
...
+    if !active && !all_nodes_active
+        return true
+    end
+
+    # If the control_type does not match the control_type_ skip update q, UNLESS the combination is None/Allocation
+    if (control_type != control_type_)
+        if control_type_ == ControlType.None && control_type == ControlType.Allocation
+            return false
+        else
```
SouthEndMusic, line 1174: This isn't the prettiest logic. The problem is that previously `ControlType` was `ContinuousControlType` and nodes that were controlled by DiscreteControl fell under `ControlType.None` (which is still the case). However, now in the non continuously controlled case we have to differentiate between allocation and non-allocation controlled nodes. 

I think the nicest way to refactor this is to go back to the `ContinuousControlType` situation where allocation controlled nodes have `ContinuousControlType.None`, and add a separate Boolean vector `allocation_controlled` to `Pump` and `Outlet` which is only used read in the allocation context (all handled in `set_control_type!`). We might in the future allow DiscreteControl to alter this vector.

SouthEndMusic (approved):
Looks good! As a last demonstration, can you simulate and plot the situation where the Outlet capacity is constraining when the demand is above the Outlet capacity?

core/src/util.jl
simulutions: agree

core/src/util.jl
simulutions: fixed it

core/src/allocation_init.jl
```diff
@@ -683,52 +681,57 @@ function add_manning_resistance!(
...
+        [node_id = node_ids_non_alloc_controlled],
+        flow[node_data.inflow_link[node_id.idx].link] ==
+        q * get_low_storage_factor(problem, node_data.inflow_link[node_id.idx].link[1]),
+        base_name = "$(constraint_name)_constraint"
     )
     return nothing
 end
 
-function add_outlet!(
+function add_pump!(
```
SouthEndMusic, line 713: This and `add_outlet` are now rather trivial wrappers of `add_node_with_flow_control`, consider calling this directly in the `AllocationModel` constructor.

core/src/allocation_init.jl
```diff
@@ -683,52 +681,57 @@ function add_manning_resistance!(
     return nothing
 end
 
-function add_pump!(
+# Extract common functionality for pumps and outlets
```
SouthEndMusic, line 684: This reads more as a PR description than a description of this function

core/src/parameter.jl
```diff
@@ -671,7 +671,9 @@ control_type: one of None, ContinuousControl, PidControl, Allocation
         Vector{ScalarLinearInterpolation}(undef, length(node_id))
     control_mapping::Dict{Tuple{NodeID, String}, ControlStateUpdate} =
         Dict{Tuple{NodeID, String}, ControlStateUpdate}()
-    control_type::Vector{ControlType.T} = fill(ControlType.None, length(node_id))
+    control_type::Vector{ContinuousControlType.T} =
+        fill(ContinuousControlType.None, length(node_id))
+    allocation_controlled::Vector{Bool} = fill(false, length(node_id))
```
SouthEndMusic, line 676: For both Pump and Outlet, also add a description of this field to the docstring

core/src/parameter.jl
```diff
@@ -705,7 +707,9 @@ control_type: one of None, ContinuousControl, PidControl, Allocation
     max_downstream_level::Vector{ScalarLinearInterpolation} =
         Vector{ScalarLinearInterpolation}(undef, length(node_id))
     control_mapping::Dict{Tuple{NodeID, String}, ControlStateUpdate} = Dict()
-    control_type::Vector{ControlType.T} = fill(ControlType.None, length(node_id))
+    control_type::Vector{ContinuousControlType.T} =
+        fill(ContinuousControlType.None, length(node_id))
+    allocation_controlled::Vector{Bool} = fill(false, length(node_id))
```
SouthEndMusic, line 712: This is fine, `zeros(Bool, length(node_id))` would also work

core/src/solve.jl
```diff
@@ -558,51 +558,59 @@ function formulate_flow!(
     return nothing
 end
 
-function formulate_flow!(
-    du::CVector,
-    pump::Pump,
+# Common code for both pump and outlet flow formulation
+function formulate_controlled_flow!(
+    du_component::SubArray{<:Number},
+    component::Union{Pump, Outlet},
```
SouthEndMusic, line 564: `node` in stead of `component` is more in line with the rest of the code

core/src/solve.jl
```diff
@@ -558,51 +558,59 @@ function formulate_flow!(
     return nothing
 end
 
-function formulate_flow!(
-    du::CVector,
-    pump::Pump,
+# Common code for both pump and outlet flow formulation
+function formulate_controlled_flow!(
```
SouthEndMusic, line 562: This name is somewhat misleading; pumps and outlets can also have just a static or time dependent flow rate (from `flow_rate_itp`)

core/src/solve.jl
```diff
@@ -558,51 +558,59 @@ function formulate_flow!(
...
+    du_component::SubArray{<:Number},
+    component::Union{Pump, Outlet},
     p::Parameters,
     t::Number,
-    control_type_::ControlType.T,
+    control_type_::ContinuousControlType.T,
+    current_flow_rate::Vector{<:Number},
+    component_cache::NamedTuple,
+    reduce_Δ_level::Bool = false,
```
SouthEndMusic, line 570:
```suggestion
    reduce_Δlevel::Bool = false,
```

core/src/solve.jl
SouthEndMusic: Nice code reduction (y)

simulutions: Renamed it to formulate_pump_or_outlet_flow

PR #2406, opened 2025-06-26
allocation_init: Unique flow conservation constraint names based on node type

jarsarasty:
Having access to constraint sets per node type could be useful for debugging since the constraint names would be more informative.


core/src/allocation_init.jl
```diff
@@ -182,7 +182,10 @@ function add_flow!(
 end
 
 """
-Equate the inflow and outflow of conservative nodes
+Add flow conservation constraints for conservative nodes.
+Ensures that inflow equals outflow for nodes that conserve mass (pumps, outlets,
```
visr, line 186:
```suggestion
Ensures that inflow equals outflow for nodes that conserve mass (pumps, outlets,
```

The trailing whitespace here fails our tests. We use pre-commit to automatically strip these, which can be configured once with `pixi run install`.

jarsarasty: Thanks, @visr

jarsarasty:
Thanks for this clarification, @SouthEndMusic. 

Having access to constraint sets per node type could be useful for debugging since the constraint names would be more informative. However, we can wait for your upcoming pull request, and we can keep this one as a draft or close it if necessary.

SouthEndMusic (approved):
Good stuff!

PR #2407, opened 2025-06-26
Scale storage and flow in allocation

SouthEndMusic:


core/src/allocation_optim.jl
```diff
@@ -217,27 +237,31 @@ function set_simulation_data!(
...
 
 function reset_goal_programming!(allocation_model::AllocationModel)::Nothing
-    (; problem, Δt_allocation) = allocation_model
+    (; problem, Δt_allocation, scaling) = allocation_model
 
     # From demand objectives
     JuMP.fix.(problem[:user_demand_allocated], 0.0; force = true)
-    JuMP.fix.(problem[:flow_demand_allocated], -MAX_ABS_FLOW; force = true)
-    JuMP.fix.(problem[:basin_allocated], -MAX_ABS_FLOW * Δt_allocation; force = true)
+    JuMP.fix.(problem[:flow_demand_allocated], -MAX_ABS_FLOW / scaling.flow; force = true)
```
simulutions, line 244: Why would you also scale the initial value of allocated demand? If its goal is to just hold an arbitrary large value that probably is never reached, it is better to keep it as high as possible.

core/test/allocation_test.jl
```diff
@@ -22,7 +22,7 @@
 
     flow = allocation_model.problem[:flow]
 
-    flow_value(id_1, id_2) = JuMP.value(flow[(id_1, id_2)])
+    flow_value(id_1, id_2) = get_flow_value(allocation_model, (id_1, id_2))
 
     @test flow_value(NodeID(:Basin, 2, p_independent), NodeID(:Pump, 5, p_independent)) ≈
           pump.flow_rate[1](t)
```
simulutions, line 28: Do you have a test that shows this feature solves a problem or adds a performance increase? I think we need proof it improves something otherwise it only clutters code

simulutions (changes requested):

core/src/allocation_optim.jl
SouthEndMusic: Will make a separate issue for this

simulutions (approved):
Nice that the training model runs! proves usefullness of this feature

core/test/run_models_test.jl
```diff
@@ -785,3 +785,10 @@ end
     @test area_col ≈ [areas[1]; areas[2]]
     @test storage_col ≈ [storages[1]; storages[2]]
 end
+
+@testitem "Allocation training model" begin
+    toml_path =
+        joinpath(@__DIR__, "../../generated_testmodels/allocation_training/ribasim.toml")
+    model = Ribasim.run(toml_path)
+    @test Ribasim.success(model)
+end
```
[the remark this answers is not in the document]
SouthEndMusic: Nice catch ahah. But we currently don't run the allocation models in `run_testmodels` but I want to keep track of this one. I'll add this to the reverting issue

PR #2408, opened 2025-06-26
fix: resolve pixi task parsing errors and Windows symlink issues

jarsarasty:
I found a couple of issues when trying to run pixi locally. I'm not sure if this is relevant to other users.

Here are some fixes I implemented:

The 'args' field is not supported in pixi task configuration, causing parsing failures. What was done:
- Remove invalid 'args' field from pixi task definitions (s3-download, s3-upload, migrate-model)
- Remove 'args' field from task dependencies that caused "expected a string, found table" errors

On Windows, creating symlinks requires administrator privileges.
- Remove problematic target-specific task dependency sections with incorrect syntax
- Add platform-specific generate-sbom-ribasim-python tasks with proper PATH environment handling
- Fix Windows symlink privilege error in QGIS plugin installation by using file copying instead of symlinks on Windows. The install script now detects the platform and uses shutil.copytree() instead of symlink_to() to avoid permission errors while maintaining the same functionality.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -73,23 +73,14 @@ build = { "cmd" = "julia --project --check-bounds=yes build.jl", cwd = "build", 
     "initialize-julia",
 ] }
 # Tests
-s3-download = { cmd = "python utils/s3_download.py {{ remote }} {{ local }}", args = [
-    "remote",
-    "local",
-] }
-s3-upload = { cmd = "python utils/s3_upload.py {{ source }} {{ destination }}", args = [
-    "source",
-    "destination",
-] }
+s3-download = { cmd = "python utils/s3_download.py {{ remote }} {{ local }}" }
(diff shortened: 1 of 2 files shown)


visr:
Great that you're trying this out. I think the task parsing issues are possibly because your pixi version is outdated, this was added recently. Can you try again after `pixi self-update` to update to v0.48.2?

For the symlinks, I think what is needed is to turn on Developer Mode in your Windows settings under System / For developers. Copying the files has the downside that you have to run a pixi task each time to copy the plugin while you are developing it, whereas I prefer for it to use the repo content directly. Note that Developer Model is only needed to develop Ribasim, indeed we want to avoid this requirement for end users.

Both the Developer Mode and minimum pixi version needed should ideally be mentioned in https://ribasim.org/dev/, though the pixi version may get outdated as we start using newer versions.

PR #2410, opened 2025-06-26
Warn Python users of version incompatibilities (adds `packaging` dep)

### LINKED ISSUE #2403: Warning on `Model.read` when Ribasim version mismatches.
We already do this in the core (since #1181) if the version in the TOML is different than the CLI. We can also do something similar in Python on `Model.read`. Now users with older ribasim versions try to read newer models and get `ValidationError: 1 validation error for TableModel[BasinProfileSchema]`.

visr:
Fixes #2403. Adds the packaging dependency to parse and compare versions.
No warning is given on lower versions in the TOML, since we can migrate.

### DIFF
--- a/python/ribasim/ribasim/model.py
+++ b/python/ribasim/ribasim/model.py
@@ -180,11 +180,6 @@ class Model(FileModel):
     def _serialize_path(self, path: Path) -> str:
         return str(path)
 
-    @field_serializer("ribasim_version")
-    def _serialize_ribasim_version(self, value: str) -> str:
-        """Serialize ribasim_version as the current package version when writing."""
-        return ribasim.__version__
-
     def model_post_init(self, __context: Any) -> None:
         # When serializing we exclude fields that are set to their default values
         # However, we always want to write `input_dir`, `results_dir`, and `ribasim_version`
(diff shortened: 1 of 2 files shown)


python/ribasim/tests/test_model.py
```diff
@@ -362,3 +364,79 @@ def test_model_diff(basic):
...
+
+    # Set a newer version
+    config["ribasim_version"] = "3030.1.0"
+
+    # Write the modified TOML back
+    with open(toml_path, "wb") as f:
+        tomli_w.dump(config, f)
+
+    # Test that a warning is issued when reading the model
+    with warnings.catch_warnings(record=True) as w:
```
evetion, line 387: You could also use https://docs.pytest.org/en/stable/how-to/capture-warnings.html#warns

evetion (approved):

python/ribasim/ribasim/model.py
```diff
@@ -150,12 +180,17 @@ def _update_used_ids(self) -> "Model":
     def _serialize_path(self, path: Path) -> str:
         return str(path)
 
+    @field_serializer("ribasim_version")
```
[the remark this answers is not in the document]
visr: It isn't, thanks, removed.

python/ribasim/tests/test_model.py
visr: Ah yes much shorter.

Issue #2248, opened 2025-04-18
Installation guide


visr:
Right now our [installation docs](https://ribasim.org/install.html) is not bad, but it is a bit laborious to install all 3 components and have them work together nicely.

The install Ribasim Python instructions are too brief, we only give `pip install ribasim` as a command. There are still plenty of Anaconda users out there, that may not be aware of the different channels and how it becomes almost impossible to solve when they interact. See for instance the note in https://conda-forge.org/docs/user/transitioning_from_defaults/

> In 2024, the workflow that has packages co-installed from Anaconda's channel and conda-forge is no longer supported.

We should still recommend Pixi, but we can try a little harder to document how to integrate it in the users existing workflows.

Regarding the Pixi installation, we currently only link to the Pixi docs, but having a guide that walks users through setting up a project with `pixi init`, `pixi add ribasim` with some extra hints in case things go wrong

- cannot install pixi; use archive https://github.com/prefix-dev/pixi/issues/3639
- find versions installed `pixi list`
- other unknown issue: `pixi clean`

If we have such a guide, we could even make it easier to install the core and QGIS plugin also via Pixi:
- core: add a task or a Ribasim Python CLI to download the correct core release into the Pixi project
- QGIS: make it easy to install the plugin to the global QGIS installation (or local pixi QGIS but most wouldn't use this)

Ted052025:
We tried to summarize the steps that we followed, maybe that will help writing up the instructions: 
Setting up RIBASIM with Pixi:
1. Set up the working directory.
2. Download ribasim.exe.
3. Use the prompt to set up the Pixi environment and import Ribasim and related packages.
4. Activate the Python environment and import Ribasim Pixi.
5. Open a code platform (e.g., VSCode, Spyder) and connect to the environment. Find python.exe in the working directory (C:\ribasim\v3\pixi\envs\default\python.exe).

For QGIS:
1. Install QGIS.
2. Download the Ribasim plugin and import it.
3. Import the iMOD plugin.

PR #2415, opened 2025-06-27
Add instructions to install using Pixi

### LINKED ISSUE #2274: Instructions on Updating Ribasim

visr:
Fixes #2274.
Fixes parts of #2248.

### DIFF
--- a/docs/install.qmd
+++ b/docs/install.qmd
@@ -62,10 +62,58 @@ Deltares also provides the iMOD QGIS plugin, which can be used to inspect the re
 For larger edits using Python is recommended.
 One can also use Ribasim Python to build entire models from base data, such that your model setup is fully reproducible.
 
-Users can choose to use Ribasim Python or QGIS plugin or a combination of them to build a Ribasim model.
-And then use the Ribasim core to run the simulation.
+# Install all components into a Pixi workspace (recommended) {#sec-install-all}
 
-# Install Ribasim core
+We recommend most users to install all components into one [Pixi](https://pixi.sh/) workspace.
+With it we can provide quick and reproducible steps to get started with modeling.
+Note that if you only wish to visualize models, you can [just install the QGIS plugin](@sec-install-qgis-plugin).
+
+First install Pixi following the [Pixi installation documentation](https://pixi.sh/latest/installation/).
+Note that if the recommended installation methods don't work due to restriction on your PC, the [zipped executable](https://pixi.sh/latest/installation/#download-from-github-releases) is likely to still work.
+
+```sh
+# Open your terminal in a new empty project folder.
+# Create a new Pixi workspace.
+pixi init
+
+# Add Ribasim, QGIS and download & unpack tools.
+# Remove qgis if you prefer to use an existing QGIS installation.
+pixi add ribasim==2025.4.0 gh 7zip qgis
+
+# Download latest Ribasim QGIS plugin.
+pixi run gh release download v2025.4.0 --repo Deltares/Ribasim --clobber --pattern ribasim_qgis.zip
+
+# Start QGIS and install the Ribasim and iMOD plugins manually.
+# If you did not `pixi add qgis`, start QGIS yourself.
+pixi run qgis
+
+# After installation you can remove the plugin ZIP file
+pixi run rm ribasim_qgis.zip
+
+# Download latest Ribasim core for Windows, unpack and remove ZIP file.
+# Takes some time due to download size.
+# Replace `windows` with `linux` if you are using Linux.
+pixi run gh release download v2025.4.0 --repo Deltares/Ribasim --clobber --pattern ribasim_windows.zip
+pixi run 7z x ribasim_windows.zip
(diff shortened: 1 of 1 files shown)


docs/install.qmd
```diff
@@ -62,10 +62,58 @@ Deltares also provides the iMOD QGIS plugin, which can be used to inspect the re
...
+Note that if the recommended installation methods don't work due to restriction on your PC, the [zipped executable](https://pixi.sh/latest/installation/#download-from-github-releases) is likely to still work.
+
+```sh
+# Open your terminal in a new empty project folder.
+# Create a new Pixi workspace.
+pixi init
+
+# Add Ribasim, QGIS and download & unpack tools.
+# Remove qgis if you prefer to use an existing QGIS installation.
+pixi add ribasim==2025.4.0 gh 7zip qgis
```
visr, line 81: Perhaps when the `pixi task` CLI gets more complete, we can define more of these commands as tasks that depend on each other, and take the Ribasim version as arguments. Then users can use those tasks later on to update Ribasim as well.

We could possibly also use some of the QGIS profile tricks we do in our dev setup to install plugins automatically, and make `pixi run qgis` use an isolated profile. But that quickly becomes a lot more complicated because you have to run Python code, and get that on the system and such, and users might prefer to use the default profile anyway.

visr:
@xldeltares do you mind having a look if this addresses your concerns raised in https://github.com/Deltares/Ribasim/issues/2248#issuecomment-2935391052?

docs/install.qmd
```diff
@@ -62,10 +62,58 @@ Deltares also provides the iMOD QGIS plugin, which can be used to inspect the re
...
+pixi run gh release download v2025.4.0 --repo Deltares/Ribasim --clobber --pattern ribasim_qgis.zip
+
+# Start QGIS and install the Ribasim and iMOD plugins manually.
+# If you did not `pixi add qgis`, start QGIS yourself.
+pixi run qgis
+
+# After installation you can remove the plugin ZIP file
+pixi run rm ribasim_qgis.zip
+
+# Download latest Ribasim core for Windows, unpack and remove ZIP file.
```
jarsarasty, line 93: Thank you, @visr !

Perhaps this note could be added: "If this is your first time using the GitHub CLI, run `pixi run gh auth login` to authenticate."

visr: Thanks. Since the assets are publicly available, I assumed this wasn't needed. I'll update this PR to just download using curl instead.

docs/install.qmd
```diff
@@ -62,10 +62,58 @@ Deltares also provides the iMOD QGIS plugin, which can be used to inspect the re
...
+Note only that QGIS plugins are by default installed into the default user profile, which is shared.
+You need to create separate user profiles if you wish to install and use different versions of the QGIS plugin at the same time.
+
+If you wish to check what is installed in your workspace, run `pixi list`.
+When you run into what seem like installation issues, try `pixi clean`, followed by `pixi install`.
+
+If your editor does not automatically detect the right Python environment, point it to `.pixi/envs/default/python.exe`.
+Starting your editor via Pixi will also help it find the environment since it will already be active.
+For Visual Studio Code, you can run `pixi run code .` to open your workspace.
+
```
Fati-Mon, line 124: Thanks, Martijn, this step-by-step explanation is really helpful, especially for users who aren’t that familiar with the process yet. I assume this workflow also works for those using Spyder or other IDEs, right? And does it also make it easier to update to newer Ribasim versions going forward?

docs/install.qmd
```diff
@@ -62,10 +62,58 @@ Deltares also provides the iMOD QGIS plugin, which can be used to inspect the re
...
+# Install Ribasim core {#sec-install-core}
```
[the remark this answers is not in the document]
visr: Yes indeed they still apply. And they may be helpful to read for some more details as well. Perhaps I should mention that at the start, because I want to avoid duplicating too much info. Though I want to avoid users having to go back and forth between sections as well...

docs/install.qmd
Fati-Mon: Yes, that would be helpful, thanks!

docs/install.qmd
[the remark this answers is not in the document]
Fati-Mon: Yes, I agree that it would be better to place that part in the beginning, maybe under the first install section. But then again also agree you don't want them to go back and forth. Perhaps this is fine then.

PR #2416, opened 2025-06-30
Migrate surface_runoff in Ribasim Python.

evetion:
After bashing @simulutions for not updating migrations after his `storage` addition, I make the same mistakes with `surface_runoff`. 😅


visr:
QGIS test failure is real: `self.assertTrue(int(cursor.fetchone()[0]) == 6, "schema_version is wrong")`

After fixing that, can you also trigger the integration tests on this branch?

evetion:
> QGIS test failure is real: `self.assertTrue(int(cursor.fetchone()[0]) == 6, "schema_version is wrong")`
> 
> After fixing that, can you also trigger the integration tests on this branch?

Integration tests have been fixed already :) (by uploading a migrated model).

PR #2419, opened 2025-06-30
make_user_demand_allocated_bound

SouthEndMusic:
I saw that I used constraints of the form "the inflow to a UserDemand is at least what is allocated to the UserDemand" but this is better expressed as a bound.


simulutions (approved):
Looks good!

Issue #2388, opened 2025-06-19
Refactor concentration to drop StructVector


evetion:
This should also make the type instability from #2114 go away.

visr:
A bit related to a bullet in #601: replace PresetTimeCallback for concentration with interpolation objects

Issue #2105, opened 2025-02-26
Support cyclic `Basin / concentration`


SouthEndMusic:
As a follow-up of https://github.com/Deltares/Ribasim/issues/2095 and https://github.com/Deltares/Ribasim/pull/2102.

This requires some refactoring, because we handle cyclic timeseries via the  extrapolation functionality of `DataInterpolations.jl`, but currently the transient concentration data is handled via callbacks in stead of interpolation objects + `tstops`.

SouthEndMusic:
To make sure that cyclic concentrations work, for each node type which can have concentration data make sure that `cyclic_times` is considered in parameter initialization, and make sure that the concentration `tstops` are taken into account in `get_timeseries_tstops`.

Issue #1894, opened 2024-10-09
User demand should return concentrations instead of removing it


evetion:
Because the UserDemand in and outflows are on separate edges, #1849 sees them as different boundaries altogether, ignoring that the return flow is actually a part of the inflow.

visr:
In time, we may want to support some process or mass injection or removal at the UserDemand node. Until then, indeed it is best to assume the outflow concentrations are equal to the inflow concentrations.

visr:
That setting the concentration, not injection, right? Injection adds mass to the inflow.

evetion:
Each outflow from the UserDemand is multiplied times the set concentration of the UserDemandConcentration table, so it's a mass injection?

gijsber:
The result is a mass injection and I assume that the User demand return flow has only one concentration, either as specified in this table or (when this table is not populated by the user) the same concentration as the abstraction.

evetion:
Note that the offline Delwaq coupling correctly returns concentrations for cyclic UserDemands (from/to the same Basin, as the flows are merged), but like Ribasim not for the other User Demands (seperate boundaries). I'll investigate with Erwin how to support this.

evetion:
@visr Should we remove any (only built-in) UserDemand tracer input when outputing from a UserDemand, so we keep the concentration at 1. Otherwise, depending on the capacity versus the basin storage, we can get very high concentrations, which indicate the number of cycles more than just tracing a source.

I don't think such a high number is really useful? The 0-1 number is still good to determine the percentage of water flowing through a userdemand, but any further accumulation is better tracked with a normal substance (which works already).

PR #2426, opened 2025-07-02
Add concentration interpolation objects

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2105
Fixes https://github.com/Deltares/Ribasim/issues/2388
Fixes https://github.com/Deltares/Ribasim/issues/601 (the last 2 bullet points)

 Refactor steps:
- The concentration update callbacks and associated caches were removed
- Concentration interpolations (potentially cyclic) were added which are called in `update_concentrations`
- Substance masses are stored in a vector per Basin

Potential bugs fixed:
- Not taking drainage into account in `cumulative_in`
- UserDemand: counting substances from inflow to outflow twice

Regarding UserDemand: I now use that the flow is one directional. I see that the assumption is made that all the substance that is abstracted ends up back in the return flow Basin, regardless of the return factor. @evetion is this a conscious decision? This is at least weird when the return factor is $0$. _Edit: the return factor is now taken into account, so that UserDemand consumes the substance that is in the difference between the inflow and the outflow._


evetion:
> Regarding UserDemand: I now use that the flow is one directional. I see that the assumption is made that all the substance that is abstracted ends up back in the return flow Basin, regardless of the return factor. @evetion is this a conscious decision? This is at least weird when the return factor is 0 . _Edit: the return factor is now taken into account, so that UserDemand consumes the substance that is in the difference between the inflow and the outflow._

There's https://github.com/Deltares/Ribasim/issues/1894 to discuss UserDemand, and it seems I thought I had fixed it. I remember running checks with the Continuity tracer, but maybe that was only with a return factor of 1? Doesn't `flow_update_on_link` take the return factor into account?

SouthEndMusic:
Essentially what we have is a volume balance computation over a timestep where we do not treat all water the same, but as separate commodities:
- A commodity of the volume that comes into the model from each boundary
- A commodity for the volume that is in each Basin at the start of the timestep

We can build a linear system for this, where:
- We have a conservation equation per basin per commodity
- The sum of all commodities over a link must equal the flow over that link as determined by the physical layer

The number of unknowns is the number of non-boundary links times the number of commodities, and the number of equations is equal to the number of basins plus the number of links. So in general that is an underdetermined problem without further assumptions or constraints.

SouthEndMusic:
Ah, I wondered why you formulated it like that. I think I fixed the NaN concentration problem; when Basins are empty this leads to division by zero for mass/storage. It is not clear to me where you want to add this check on the Python side.

evetion:
> Ah, I wondered why you formulated it like that. I think I fixed the NaN concentration problem; when Basins are empty this leads to division by zero for mass/storage. It is not clear to me where you want to add this check on the Python side.

Hmm, maybe it's not as straightforward as I thought. Maybe just add such a test to `@testitem "results"`, so we at least test 1 model.

Issue #2445, opened 2025-07-08
repro: Very slow waterbed models


evetion:
Some Ribasim-NL waterboard models are really slow, and can be fixed (sped up) by enlarging the storage of the basins, indicating it's the quick change in level that is the cause of the slowdown. We expect that the change in level propagates and has its own downstream effects, like a waterbed. A hunch is that it might involve (inversing) ManningResistance nodes.

I've uploaded the model sent to me by Ron to `repro-models/HollandseDelta_parameterized`. It might be good to focus on just a small area, such as the area around Ridderkerk (`#6415`), as there seem to be slow nodes in that area (indicated by convergence).

PR #2446, opened 2025-07-09
Improve continuous control performance

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2445.

The model `repro-models/HollandseDelta_parameterized` makes heavy use of ContinuousControl. I had the hypothesis that the fact that the functions that define the continuous relations for ContinuousControl are piecewise linear is hard on the solver due to the lack of smoothness. To test this I changed the interpolation type to `PCHIPInterpolation` (see [here](https://docs.sciml.ai/DataInterpolations/stable/methods/#PCHIP-Interpolation)), not as a change to merge, but just to see what a smoother interpolation type does. And indeed, with this smoother interpolation I can locally run the first 4 months of the simulation in 9 min. 

There are several things we can do to improve the ContinuousControl performance structurally:
- Indeed use a smoother interpolation type than LinearInterpolation. We of course have to use an interpolation type that can be understood by modelers, so this could be a good use case for [smoothed linear interpolation](https://southendmusic.github.io/CachedInterpolations.jl/stable/examples/#Smoothed-linear-interpolation)
- Treat ContinuousControl more similar to DiscreteControl, in the sense that the continuous control logic is not part of the ODE system, but the control logic is computed at the start of the timestep and then kept fixed over the timestep. This approach probably has the best performance gain, but also results in a loss of accuracy. Alternatively we could make the continuous-but-not-part-of-the-ODE-system-control a new node type (or add a switch in the TOML between callback computed or ODE computed), but then we need to make the subtle difference very clear in the docs.

@visr @gijsber @Huite let me know whether you think this warrants a design session. Maybe in the particular case of the `HollandseDelta` model the schematization can also be made a bit more solver friendly

As a side node, I think we can expect similar issues with models that have a lot of TabulatedRatingCurve nodes with very sharp transitions in the q(h) relations

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1677,7 +1677,9 @@ version = "1.11.0"
 
 [[deps.SparseConnectivityTracer]]
 deps = ["ADTypes", "DocStringExtensions", "FillArrays", "LinearAlgebra", "Random", "SparseArrays"]
-git-tree-sha1 = "182990067a09adf950274f97f38f68c76f81d2d0"
+git-tree-sha1 = "bb394bcac0275a08e71506eaa23132dfd9d383a5"
+repo-rev = "add_smoothed_linear_interpolation_support"
+repo-url = "https://github.com/SouthEndMusic/SparseConnectivityTracer.jl"
 uuid = "9f842d2f-2579-4b1d-911e-f412cf18a3f5"
 version = "0.6.21"
 
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.5"
 manifest_format = "2.0"
-project_hash = "f6a0770e1fb1a6057f88838586f8e70915108dcb"
+project_hash = "a15a36ab8f1f06e75263ce0939505903c612166d"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "be7ae030256b8ef14a441726c4c37766b90b93a3"
@@ -359,7 +359,9 @@ version = "1.7.0"
 
 [[deps.DataInterpolations]]
 deps = ["EnumX", "FindFirstFunctions", "ForwardDiff", "LinearAlgebra", "PrettyTables", "RecipesBase", "Reexport"]
-git-tree-sha1 = "13c4d72ac01c0a441d72eced702332f8eabc8366"
+git-tree-sha1 = "01dd49bd6a0f61728a0b9b2c3d6de5c95cd2b1e9"
+repo-rev = "add_smoothed_linear_interpolation"
+repo-url = "https://github.com/SciML/DataInterpolations.jl"
 uuid = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
 version = "8.1.0"
 
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -66,6 +66,8 @@ using DataInterpolations:
     LinearInterpolation,
     SmoothedConstantInterpolation,
     LinearInterpolationIntInv,
+    CubicHermiteSpline,
+    PCHIPInterpolation,
     invert_integral,
     get_transition_ts,
     derivative,
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -64,6 +64,7 @@ using LinearAlgebra: mul!
 using DataInterpolations:
     ConstantInterpolation,
     LinearInterpolation,
+    SmoothedLinearInterpolation,
     SmoothedConstantInterpolation,
     LinearInterpolationIntInv,
     CubicHermiteSpline,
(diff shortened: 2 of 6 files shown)


SouthEndMusic:
This model also contains a lot of Manning resistance nodes, so I tried some things there as well without much succes. One thing I tried was relaxing the Manning resistance flow, i.e. computing the flow for a new timestep with a fraction of the flow of the previous timestep.

In general I think that the Manning resistance nodes still cause stability problems because around a head difference of 0 the flow is very sensitive to level fluctuations. I discussed this with Claude and one of the things it suggested that I thought was interesting is introducing an artificial viscosity term to dampen oscillations induced by Manning resistance nodes. I don't yet fully understand how to implement that, @Huite do you have thoughts on this?

visr:
Yeah if we can find a smooth interpolation for ContinuousControl / TabulatedRatingCurve that helps performance and is still quite close to the linear interpolation we could just enable that always.

Huite:
> Indeed use a smoother interpolation type than LinearInterpolation. We of course have to use an interpolation type that can be understood by modelers, so this could be a good use case for [smoothed linear interpolation](https://southendmusic.github.io/CachedInterpolations.jl/stable/examples/#Smoothed-linear-interpolation)

> Yeah if we can find a smooth interpolation for ContinuousControl / TabulatedRatingCurve that helps performance and is still quite close to the linear interpolation we could just enable that always.

You could consider relaxing the "close to linear interpolation" requirement since reality is generally smooth rather than piecewise linear. The reason we picked linear interpolation -- I think -- is because it's conceptually simple to implement and understand for a modeler, it obviously has no basis in reality. If e.g. a spline (or whatever) is numerically advantageous, I think most would be happy with it as long as it matches the control points exactly and doesn't look too bendy.

Splines are a somewhat common approach in unsaturated zone modeling for the constitutive functions. Something like Mualem-Van Genuchten needs to be computed at every formulation to compute conductivity and moisture content from pressure; it's cheaper to approximate with a spline (and it'll support all kind of tabulated/empirical constitutive functions as well).

> In general I think that the Manning resistance nodes still cause stability problems because around a head difference of 0 the flow is very sensitive to level fluctuations. I discussed this with Claude and one of the things it suggested that I thought was interesting is introducing an artificial viscosity term to dampen oscillations induced by Manning resistance nodes. I don't yet fully understand how to implement that, @Huite do you have thoughts on this?

I've heard of artificial viscosity, but our formulation doesn't consider viscosity explicitly. So you have to alter some values/parameters and my guess is that it's effectively a similar approach as the relaxed root.

What's causing the instability exactly? My guess is that it's reversing the direction of the flow, not so much the steep derative?
I'd guess that a one-directional Manning resistance is more stable, but that might be worthwhile to check with an ad hoc experiment.

Huite:
> It is not immediately clear to me how this would be formulated in a way that is both smooth in the flow direction reversal and not a departure from the current formulation in a neighborhood around that point.

What I mean with ad hoc is: hopefully there's a simple model which triggers the instability. Then just "break" the functionality it and allow flow in only one direction. The steep derivative remains, so then it's the obvious culprit. But on reflection, it does seem more obvious that it's the steepness rather than the flow direction switching. My thinking was that that maybe the flow reversal causes "bouncy" behavior and instability through the network thereby causing the Newton iteration to struggle; but that doesn't really sound very parsimonious when there's also a singularity to blame.

On that topic: singularities are unphysical. So maybe we should consider a conceptual model that has less singular behavior.

Huite:
Continuing my last point: Manning and Chezy assume turbulent flow. This is the wrong type of physics for near zero flows; at some point we get laminar flow. If you were to approximate with Poisseuille flow, it's numerically excellent: fully linear!

What we could do is compute Reynolds number, and choose the appropriate approximation. ChatGPT-o3 suggests:

* Laminar: Re < 500 or depth < 1 cm in small channels: Poisseuille / Darcy-Weisbach
* Transitional: 500 <= Re <= 4000: Churchill or Swamee-Jain
* Fully turbulent: Re > 4000: existing Manning (or Colebrook-White)

You can use e.g. a continuous weighting function to keep the function smooth.

In retrospect, I obviously should've done this from the start! This is superior in terms of physical representation and numerical characteristics!

SouthEndMusic:
I now use SmoothedLinearInterpolation, see https://github.com/SciML/DataInterpolations.jl/pull/441. Here's an example of the first ContinuousControl function of the Hollandse Delta model:

https://github.com/SciML/DataInterpolations.jl/pull/441

evetion:
> I now use SmoothedLinearInterpolation, see [SciML/DataInterpolations.jl#441](https://github.com/SciML/DataInterpolations.jl/pull/441). Here's an example of the first ContinuousControl function of the Hollandse Delta model:
> image

I understood this was even faster, now running a significant portion of HD in minutes? I'd say how CC works internally is non-breaking, so I don't mind too much that it doesn't hit it's data points exactly. That said, it would be good to document any large changes in behavior.

PR #2450, opened 2025-07-10
Use pandera v0.25, and revert using pyarrow dtypes

### LINKED ISSUE #2120: Ribasim Python breakage wit SQLite and Pandera
Two new releases of dependencies break the latest release of Ribasim Python.

New error on string literal syntax in sqlite 3.49. Fixed in https://github.com/Deltares/Ribasim/pull/2117/commits/951ff5d74089b6dc68540aeff2b00fd064794b79.
For now users must `pixi add sqlite=3.48`, which is done in Ribasim-NL.

Pydantic 0.23 breaks Ribasim Python, some error related to `to_schema`. For now users must `pixi add pandera=0.22`, which is done in Ribasim-NL, and Ribasim, until we can fix this.

We could modify our upper bounds in the conda-feedstock, but since we'll probably make a release soon let's just document, fix and release. The pandera fix might require at least 0.23.

visr:
This fixes #2120, we can update to the latest Pydantic release (0.25) again. There is still one open bug https://github.com/unionai-oss/pandera/issues/2016. That triggered me to essentially revert #1781 and switch back to the default pandas dtypes, because with them we are less likely to hit various compatibility issues in the ecosystem.

Because we like nullable strings, integers and Booleans, we do use these now:
- Nullable integer data type: https://pandas.pydata.org/docs/user_guide/integer_na.html
- Nullable Boolean data type: https://pandas.pydata.org/docs/user_guide/boolean.html
- StringDtype: https://pandas.pydata.org/docs/user_guide/text.html

Furthermore this gets rid of Series in the type annotations using https://pandera.readthedocs.io/en/stable/dataframe_models.html#using-data-types-directly-for-column-type-annotations. This was needed because otherwise MyPy would complain about not allowing certain dtypes, but it is cleaner anyway.

This also switches back to the default Timestamp with ns precision for the same reason. Possibly we need to support other units in the future if we do very long simulations.


python/ribasim/ribasim/input_base.py
```diff
@@ -515,15 +514,7 @@ def _from_db(cls, path: Path, table: str):
         with closing(connect(path)) as connection:
             if exists(connection, table):
                 # pyogrio hardcodes fid name on reading
-                df = gpd.read_file(
-                    path,
-                    layer=table,
-                    engine="pyogrio",
-                    fid_as_index=True,
-                    use_arrow=True,
```
evetion, line 523: Might be good to check/confirm that use_arrow/engine is now default? I remember using arrow was way faster, and actually using the arrow dtypes was a bit faster on top of that. Would be good to keep the initial performance gain.

python/ribasim/ribasim/delwaq/generate.py
```diff
@@ -73,18 +73,17 @@ def _make_boundary(data, boundary_type):
...
     )
-    # Convert Arrow time to Numpy to avoid needing tzdata somehow
-    piv.time = piv.time.astype("datetime64[ns]").dt.strftime("%Y/%m/%d-%H:%M:%S")
+    piv.time = piv.time.dt.strftime("%Y/%m/%d-%H:%M:%S")
     boundary = {
         "name": bid,
         "substances": list(map(_quote, piv.columns[1:])),
-        "df": piv.to_string(formatters={"time": _quote}, header=False, index=False),
+        "df": piv.to_string(
+            formatters={"time": _quote}, header=False, index=False, na_rep=-999
```
evetion, line 85: Why did you move from the fill_value to na_rep here? Not sure if it matters if the tests run.

visr: Just reverting parts of https://github.com/Deltares/Ribasim/pull/1781#discussion_r1773291247. This is slightly nicer that it doesn't write .0 after -999.

python/ribasim/ribasim/input_base.py
visr: Yeah good point, pyogrio is default if installed, arrow is not: 6e9365ee164bf1f5d5cf404b283bd628ee52e2f8

PR #2458, opened 2025-07-14
Smoother rating curve

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1919


visr:
Nice! Let's switch to PCHIPInterpolation for this and #2446. I think you also need to add Plots to the root environment, though I hope that won't undo the gains from #2441 too much.

visr (approved):

PR #2465, opened 2025-07-17
Support giving `model_dir` to `run_delwaq`

visr:
Right now `run_delwaq` doesn't take arguments and assumes that the model is in a model dir next to the ribasim.delwaq folder. To make it possible to run models in other places this adds a kwarg `model_dir`, with the same default.

Perhaps we should make the kwarg mandatory, but I don't want to break it right now.

### DIFF
--- a/python/ribasim/ribasim/delwaq/util.py
+++ b/python/ribasim/ribasim/delwaq/util.py
@@ -11,6 +11,7 @@ import numpy as np
 import numpy.typing as npt
 import pandas as pd
 
+from ribasim.delwaq.generate import output_path as model_dir
 from ribasim.utils import MissingOptionalModule
 
 try:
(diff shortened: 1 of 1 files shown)


python/ribasim/ribasim/delwaq/util.py
```diff
@@ -153,21 +157,30 @@ def ugrid(G) -> xugrid.UgridDataset:
...
+                  If None, uses the D3D_HOME environment variable.
+    """
     if d3d_home is None:
-        raise ValueError("D3D_HOME is not set.")
-    else:
-        pd3d_home = Path(d3d_home)
-    binfolder = (pd3d_home / "bin").absolute()
-    folder = Path(__file__).parent
-    inp_path = folder / "model" / "delwaq.inp"
+        d3d_home = os.environ.get("D3D_HOME")
```
evetion, line 171: This can be moved to the argument?

evetion (approved):

Issue #2383, opened 2025-06-18
Run all test models with allocation again in CI


SouthEndMusic:
In https://github.com/Deltares/Ribasim/pull/2266 `tesmodelrun.jl` was modified to skip test models using allocation. This should be reverted once the new allocation code is stable enough.

SouthEndMusic:
Similarly, https://github.com/Deltares/Ribasim/pull/2386/commits/63198bf25d61870c0e49f73f5e82f762e25bd5c0 should be reverted

SouthEndMusic:
The test `Allocation example model` looks like it's also covered by `run_models_test`

visr:
Also revert this test skip: 15abb72bf740757d7ca35a80a896988307e8329c

PR #2474, opened 2025-07-18
Use `engine: julia` and `juliacall` in docs

### LINKED ISSUE #2078: Try QuartoNotebookRunner.jl to speed up docs build

visr:
ac079e4f4c10bef1cc517c072b896d2b0ad76f67: use [QuartoNotebookRunner](https://quarto.org/docs/computations/julia.html) for our pages that have mainly Julia code.

fc0e157066d8183c4fc1748e88924bd3ff3f199b: use [JuliaCall](https://juliapy.github.io/PythonCall.jl/stable/juliacall/) in the Jupyter Python QMD files, as an alternative to `subprocess` that avoids restarting Julia every time, speeding up the docs.

dfc2af16e1f222eae04f5b78f760c9e03f862838 is needed since we now fail explicitly if one of the simulations in our docs fail, listed it in #2383.

Fixes #2078
Fixes #2381

Initially I tried to use the Julia engine everywhere, since it also allows Python cells in the same document. However it just uses regular Python, not ipython/jupyter, so I would have had to change a lot of existing code that assumed jupyter, and get less rich output, and in some cases, like interactive Plotly plots, I didn't manage at all to get it to render to an interactive inline HTML.


visr:
The docs job was between 23 and 48 minutes (not sure why so much variation, this is outside of cache misses I think), now it seems around 16 minutes.

PR #2486, opened 2025-07-24
Recreate CI env locally: add devcontainer

### LINKED ISSUE #2498: Recreate CI environment in devcontainer
Sometimes in the CI the unit tests fail on ubuntu but not in windows.
To enable developers to debug such situation, we can create a docker devcontainer to which vscode can attach. This way we can run the failing unit test locally and debug what is going wrong. 

Additionally  we get another feature for free: we can run our entire environment in a container on github using codespaces.

verheem:
fixes #2498


visr:
Quarto currently doesn't have Linux aarch64 support hence it is excluded there. After https://github.com/conda-forge/quarto-feedstock/pull/43 is merged we can make it an unconditional dependency again.

PR #2487, opened 2025-07-24
Remove Legolas, take 2

### LINKED ISSUE #1127: Move away from Legolas
While investigating https://github.com/Deltares/Ribasim/issues/1087, @deltamarnix encountered difficulties with one of our dependencies, Legolas. We don't use much of its functionalities at the moment, and it makes the code quite complex. Let's look into removing it.

visr:
This addresses the parts of #2483 that I wasn't happy with. Now schema.jl contains the Schema module, which has a module for each node type. Each node type module then has a struct for each table schema. So `Schema.Basin.State` is now a struct. We use the type of this thing to load the right table, and the resulting table will be a `StructVector{Schema.Basin.State}`, meaning when you iterate over the rows you get these structs filled with data.

Fixes #1127.


visr:
Yeah it being tied to tables is the reason I went for this now, but I've wanted to do this since https://github.com/Deltares/Ribasim/issues/1127, which explains the reasons. I'll add a few unit tests for the new utility function still.

PR #2488, opened 2025-07-24
Workflows: Include configuration file for Claude

jarsarasty:
Adds a comprehensive CLAUDE.md file to provide AI coding assistants (specifically Claude) with essential context and guidance when working with the Ribasim codebase.
Complements existing .github/copilot-instructions.md with more comprehensive guidance

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -141,9 +141,9 @@ We use the `GitHub` repository https://github.com/Deltares/Ribasim for issues an
 
 ### Primary Data Formats
 - **SQLite/GeoPackage**: Model database storage
-- **Arrow**: High-performance data exchange and large results
+- **Arrow**: Results or tables too large for SQLite
 - **TOML**: Configuration files
-- **NetCDF**: Conversion target for external tools
+- **NetCDF**: Conversion to NetCDF for interop
 
 ### Network Architecture
 Models are represented as directed graphs where:


jarsarasty:
Thank you for the feedback, @evetion and @visr! 

You're right about avoiding duplication and aligning with Deltares' official tooling.

I've updated this PR, expanding the existing copilot-instructions.md with:
- Detailed Pixi commands for all development workflows (building, testing, linting, docs) 
- Testing strategy with single test execution examples
- Network architecture explanation

CLAUDE.md now points to copilot-instructions.md. In this way, a single source of truth is used while maintaining compatibility for different AI assistants. 

Please note that although Claude via Copilot recognizes `copilot-instructions.md`, standalone tools like Claude Code may expect different instruction files by default. For more information see [here](https://www.anthropic.com/engineering/claude-code-best-practices).

.github/CLAUDE.md
visr, line 1: Are you sure this is picked up by Claude Code? I thought CLAUDE.md needed to be in the root. And what about making it a symlink to the copilot-instructions?

.github/copilot-instructions.md
```diff
@@ -61,20 +61,42 @@ See also the developer docs at `docs/dev`.
...
+pixi run quarto-preview           # Preview Quarto documentation
+pixi run quartodoc-build          # Build Python API docs
 
 # Model Generation
-pixi run generate-testmodels     # Generate test models
+pixi run generate-testmodels      # Generate test models
+
+# Running models
+pixi run ribasim-core             # Run Julia core with command line args
+pixi run ribasim-core-testmodels  # Run all testmodels with core
```
visr, line 99: It should be able to figure out many of these by reading pixi.toml itself, which is already linked.

I'm not against adding these, but overall I think we should aim to keep the instructions as short as possible. Rather than trying to make it complete from the beginning, having a short copilot-instructions.md that we expand over time as we run into issues during usage may work better. See also https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#writing-effective-repository-custom-instructions.

.github/copilot-instructions.md
```diff
@@ -147,14 +180,22 @@ We use the `GitHub` repository https://github.com/Deltares/Ribasim for issues an
...
+- **Integration tests** - `core/integration_test/`
+- **Regression tests** - `core/regression_test/`, marked with `@pytest.mark.regression`
 
-### Python Tests
-- Unit tests in `python/*/tests/`
-- Mark regression tests with `@pytest.mark.regression`
+### Running Single Tests
+```bash
+# Julia single test file
+julia --project=core --check-bounds=yes core/test/specific_test.jl
```
visr, line 194: This won't work, because a test environment needs to be used rather than the core environment.

I don't know the right version right now, perhaps just leave it out.

.github/copilot-instructions.md
```diff
@@ -147,14 +180,22 @@ We use the `GitHub` repository https://github.com/Deltares/Ribasim for issues an
...
+
+# Python single test
+pytest python/ribasim/tests/test_specific.py
```
visr, line 197: I think it knows this quite well, I see Sonnet doing versions of this all the time.

visr (changes requested):
Let's see if we can merge this

PR #2489, opened 2025-07-25
Add a guide on updating Ribasim models

### LINKED ISSUE #2382: Documentation/Guide on model migration
**What**
Guide/documentation on what needs to be done to migrate model from an older to a newer version.

**Why**
Since ribasim is under fast improvement and development, models build with older version are foresee to be updated frequently. it would be helpful for a user to have guidance/documentation on what needs to be done to migrate model from an older to a newer version.

**How**
I guess it can be as simple as a model read and write, as indicated in the [changelog](https://ribasim.org/changelog.html).  

**Additional context**
The current tutorial and how-tos give clear instructions on model building and writing, but not yet covering the topic of reading in and updating the model.

visr:
Fixes #2382.

### DIFF
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -41,6 +41,7 @@ website:
       contents:
         - guide/examples.qmd
         - guide/modeling-process.qmd
+        - guide/updating-ribasim.qmd
         - guide/qgis.qmd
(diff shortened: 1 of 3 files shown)


SouthEndMusic (approved):
Looks good, added some small comments

PR #2537, opened 2025-08-21
Add FlowBoundary / Area.

### LINKED ISSUE #2449: Add `FlowBoundary / area`
**What**
For Basin we have an area table: https://ribasim.org/reference/node/basin.html#area
@DanielTollenaar suggest we add an area table to FlowBoundary as well.

**Why**
To show what upstream area is associated with a FlowBoundary. For Ribasim-NL this data is available (foreign input).

**How**
The table should be optional, and individual node areas should also be optional.

**Additional context**
For #2436 this probably also makes sense to have.

evetion:
Fixes #2449 

Cyclic time testmodel with a `FlowBoundary / Area`
Screenshot 2025-08-21 at 20 40 57

### DIFF
--- a/python/ribasim_testmodels/ribasim_testmodels/basic.py
+++ b/python/ribasim_testmodels/ribasim_testmodels/basic.py
@@ -448,7 +448,7 @@ def cyclic_time_model() -> Model:
         ],
     )
 
-    flow_boundary_geometry = Point(0.0, 0.0)
+    flow_boundary_geometry = Point(0.0, 2.0)
     fb = model.flow_boundary.add(
         Node(4, Point(0, 1), cyclic_time=True),
         [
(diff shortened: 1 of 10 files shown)


python/ribasim_testmodels/ribasim_testmodels/basic.py
```diff
@@ -448,13 +448,17 @@ def cyclic_time_model() -> Model:
         ],
     )
 
+    flow_boundary_geometry = Point(0.0, 0.0)
```
visr, line 451:
```suggestion
    flow_boundary_geometry = Point(0.0, 2.0)
```

You put the FlowBoundary / area around the Basin, which I think is a bit confusing. I moved it up, such that the area is upstream of the FlowBoundary node, roughly like so:

Image

visr (approved):
Nice, thanks!

PR #2540, opened 2025-08-25
Improve error message for DiscreteControl with no data

visr:
Otherwise it fails in the next line with "reducing over an empty collection is not allowed"


SouthEndMusic (approved):
LGTM

PR #2541, opened 2025-08-25
double the upper bound

### LINKED ISSUE #2473: Decide if we allow overflow of basins in both physical and allocation or not
According to the docs, we do not allow overflow of allocation in basins but we do allow in physical.
We need to be consistent in both cases since the current design will cause infeasibilities if the physcal layer has over flow of a basin.

simulutions:
fixes #2473


visr (changes requested):
There are a lot of new allocation test failures, probably related.

visr (approved):

PR #2548, opened 2025-08-29
Speed up loading Arrow tables in core

visr:
Loading large Arrow tables was very slow. Profiling showed it was due to a warning log message from Arrow.jl about time having a nanosecond precision, lower than DateTime which has millisecond precision. This warning was harmless, only annoying and apparently super slow. It started appearing after going back to Pandas defaults in #2450.

Upstream issue: https://github.com/apache/arrow-julia/issues/559

This avoids that by passing `convert = false` to Arrow.Table, and doing the conversion with our own function that doesn't have this log.

### DIFF
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -1818,7 +1818,7 @@ function load_structvector(
     end
 
     table = StructVector{T}(nt)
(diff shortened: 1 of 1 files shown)


SouthEndMusic (approved):
LGTM

Issue #2550, opened 2025-08-29
Only load simulation period from timeseries


visr:
When looking into initialization performance in https://github.com/Deltares/Ribasim/pull/2548 I saw that most of the remaining time is spent in Basin / time processing, especially `parse_forcing!` which iterates over the time series groups with `IterTools.groupby`.

Some models are saved with many years worth of forcing, even though we usually only run one year. That should be fine, but it is not very efficient to then load, process and keep all that data in memory. It would be more efficient to filter out the data outside the simulation period during or right after loading.

A naive implementation could be adding this at the end of `load_structvector`:
```jl
    table = StructVector{T}(nt)

    @show length(table)
    # drop all rows that are not in the simulation period, except for padding
    if hasproperty(table, :time) && !isempty(table)
        sort!(table; by = x -> x.time)
        (; time) = table
        i1 = searchsortedlast(time, starttime)
        i2 = searchsortedfirst(time, endtime)
        # adjust if starttime is before the data or endtime is after the data
        i1 = max(i1, 1)
        i2 = min(i2, length(time))
        # this means that there is no data in the requested time range
        table = table[i1:i2]
        # we rely on sorted_table! below to put it back in the correct order
    end
    @show length(table)

    return sorted_table!(table)
```

However this isn't quite correct, since for each node the timeseries can be different. And we also want to retain the timestamps right before and after the simulation period, such that we can still correctly use interpolation.

If we go the structured approach with #2521 this would become a lot easier to implement, so I propose doing that first.

We could also write a utility script that filters all timeseries outside a certain period so we can shrink models and make them initialize faster, for testing.

Here is a profile of the `parse_forcing!` slowness:

<img width="632" height="563" alt="Image" src="https://github.com/user-attachments/assets/ba87b9ca-81c1-43d8-8c20-e27c32bc9be1" />

PR #2551, opened 2025-08-29
Add `pixi run filter-model`

visr:
This utility can be run to avoid some of the pain from https://github.com/Deltares/Ribasim/issues/2550.

It removes all data outside the simulation period. Note that if there is no timestamp on the starttime or endtime this can affect results or break the model. It is just meant as a convenience function for developers, nothing public.

Output looks like this:
```
Filtering time tables to period: 2023-01-20 00:00:00 to 2024-04-27 00:00:00
Basin / time                        removed 6774500 / 7760500 rows (87.3%)
FlowBoundary / time                 removed 45 / 973 rows (4.6%)
LevelBoundary / time                removed 432 / 1360 rows (31.8%)
Total rows removed across all time tables: 6774977
```

I also modified both `filter-model` and `migrate-model` to accept one argument, in which case it will modify the model in place, use with caution.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -178,10 +178,8 @@ write-allocation-problems = { cmd = "julia --project --check-bounds=yes --thread
     "generate-testmodels",
     "initialize-julia",
 ] }
-migrate-model = { cmd = "python utils/migrate_model.py {{ input_toml }} {{ output_toml }}", args = [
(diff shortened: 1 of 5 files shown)


SouthEndMusic (approved):
Looks good, also nice info feedback from the call.

Issue #2051, opened 2025-02-11
Optimize for all sources in a subnetwork with the same source priority at the same time


SouthEndMusic:
In https://github.com/Deltares/Ribasim/pull/2022 source priorities were introduced for allocation. In that PR it was first attempted to solve for all sources in a subnetwork that have the same source priority at the same time. The problem that this situation has is that it becomes more likely that the optimization problems do not have an unique optimal solution, which can lead to non-deterministic behavior. Therefore it was reverted to solving for each source individually.

Solving per source means solving a lot more optimization problems, which affects performance (and potentially also optimization results of lower quality). In the default case where source priorities are not unique, the optimization per source priority is done in the order of the Node IDs, which is somewhat arbitrary. So if we want to bring solving for all sources that have the same priority at once, we need to introduce a way to make the solution of the optimization unique. 

One way to do this is by adding a secondary objective. Then solving an optimization problem has the following steps:
- Solve the optimization problem with the 'fair distribution' objective that we have now
- Fix the allocated amounts 
- Optimize for a second objective, for instance minimizing flow cost by giving each edge a cost. To ensure uniqueness of this problem, each edge could have a different cost by introducing some deterministic noise.

PR #2553, opened 2025-09-01
Test source priorities

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2051.

The test model looks like this:

image

The idea is that the UserDemand gradually increases its demand over a single priority. The level boundaries go from more desirable to less desirable to the right. We see that this is correctly reflected in the flows over the pumps:

image

I did have to change the objective for the source priorities because it did exactly the opposite of what we want. The objective function to minimize is now

$$
\sum w F
$$

where $F$ is the flow out of a source and $w$ is its source priority. This means that the higher the source priority value (which means the lower the desirability of the source) the more the solver is incentivized to lower $F$.


simulutions (approved):
Lekker bezig

PR #2567, opened 2025-09-08
Fix CLI panic for threads arg

visr:
I should've triggered the CLI tests manually on #2562.

It failed with
```
thread 'main' panicked at src/main.rs:50:55:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The problem was that when neither `--threads` nor `JULIA_NUM_THREADS` was set, it tried to unwrap `cli.threads` which was none.

This avoids that case and adds some tests.


visr:
This is lhm_ctwq scaling, x = nthreads, y=seconds, on my Windows machine with 12 physical cores and 24 logical. The current default `auto` uses 24, which isn't a great default.
image

This is the same model on H7 (Linux) with 4 vcpu:
image

@evetion also did some tests and saw that it was faster to only have `@threads` on set current basin properties. So we should probably amend this PR to:
- remove all but one `@threads`
- use `get_physical` from the `num_cpus` crate to avoid the `auto` setting

In follow up PRs we can look into lower overhead threading setups.

PR #2581, opened 2025-09-16
Update typing in Ribasim Python

visr:
I asked Claude to go over the type annotations, and also update to the syntax supported by our minimum Python version 3.11.

It suggested in several places to use `object` instead of `Any`, the latter bails out of typing. Also it could refine the types of our context variables and generator.

### DIFF
--- a/python/ribasim/ribasim/geometry/node.py
+++ b/python/ribasim/ribasim/geometry/node.py
@@ -1,5 +1,3 @@
-from typing import Any
-
 import geopandas as gpd
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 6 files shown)


python/ribasim/ribasim/geometry/node.py
```diff
@@ -142,7 +142,7 @@ def plot(self, ax=None, zorder=None) -> Any:
             "": "k",
         }
         if self.df is None:
-            return
+            return ax
```
evetion, line 145: This now plots an empty ax, we should check whether that's what we want.

python/ribasim/ribasim/input_base.py
```diff
@@ -600,12 +598,12 @@ def _write_geopackage(self, path: Path) -> None:
 
 
 class ChildModel(BaseModel):
-    _parent: Any | None = None
+    _parent: "Model | None" = None
```
evetion, line 601: This is not a model perse, it should be a base model.

evetion (approved):
Some nits. It would be nice to mark Claude/CoPilot uses in our PRs, or use the integration, as that would enable iterative improvements based on my review. It's also good to mark the code that has been AI generated in our codebase.

python/ribasim/ribasim/input_base.py
evetion: Yeah (and then remove the if TYPE_CHECKING part). If that doesn't work, I'll work on this branch.

PR #2586, opened 2025-09-22
Add `smooth_clamp` function 

### LINKED ISSUE #1918: Smooth clamp function

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/1918. 

The function I implemented is somewhat different than the one suggested in the issue. In the clamp function I replaced the sections on `[lo - d, lo + d]` and `[hi - d, hi + d]` C1 smooth with a quadratic polynomial section, where `d = a (hi - lo) / 2`, so ` 0 <= a <= 1` is the relative smoothing parameter. The only edge case this cannot handle smoothly is when `isinf(lo) xor isinf(hi)`.
I haven't noticed any significant performance improvements with this but it's quite little regret.

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -436,7 +436,7 @@ function linear_resistance_flow(
 
     Δh = h_a - h_b
     q_unlimited = Δh / resistance[node_id.idx]
-    q = clamp(q_unlimited, -max_flow_rate[node_id.idx], max_flow_rate[node_id.idx])
+    q = smooth_clamp(q_unlimited, -max_flow_rate[node_id.idx], max_flow_rate[node_id.idx])
     return q * low_storage_factor_resistance_node(p, q_unlimited, inflow_id, outflow_id)
 end
 
@@ -698,7 +698,7 @@ function formulate_pump_or_outlet_flow!(
                 end
             end
         end
-        q = clamp(q, lower_bound, upper_bound)
+        q = smooth_clamp(q, lower_bound, upper_bound)
 
         # Special case for outlet: check level difference
         if reduce_Δlevel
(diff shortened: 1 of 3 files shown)


Huite (changes requested):
This looks like a nice implementation! I'd replace `α` by `δ`, and omit the default value.
Maybe switch the conditionals around so it matches the behavior of the regular clamp function when lo and hi are provided "wrongly".

SouthEndMusic:
Closing this as it doesn't seem needed

PR #2587, opened 2025-09-22
Catch residuals from OrdinaryDiffEq internals

SouthEndMusic:
See also https://github.com/Deltares/Ribasim/pull/2448/files.

The benefit here is that we catch each computation of the residuals, and do not depend on reaching into an internal cache which might break at some point.


SouthEndMusic:
I realize now that this can probably be simplified, because now we compute averages both in `update_cumulative_flows!` and `save flow`.

SouthEndMusic:
Closing as stale

PR #2594, opened 2025-09-23
Use quarto-publish task

visr:
The action in https://github.com/Deltares/Ribasim/pull/2593 cannot find quarto, and this does not seem configurable. So it is failing on main.

Since we run render the step before, I call `--no-render`.
This uses git worktrees and when trying out this task locally pre-commit failed on gh-pages which doesn't have a pre commit yaml, hence PRE_COMMIT_ALLOW_NO_CONFIG.

The `_publish.yml` is created by quarto, and the docs seem to indicate this needs an ID, but I haven't been able to generate that yet.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -58,6 +58,7 @@ quarto-render = { cmd = "PYTHON_JULIAPKG_PROJECT=$PIXI_PROJECT_ROOT quarto rende
     "generate-testmodels",
     "initialize-julia",
 ], env = { "PYTHON_JULIACALL_CHECK_BOUNDS" = "yes", "PYTHON_JULIACALL_HANDLE_SIGNALS" = "yes", "PYTHON_JULIAPKG_OFFLINE" = "yes" } }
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -58,6 +58,7 @@ quarto-render = { cmd = "PYTHON_JULIAPKG_PROJECT=$PIXI_PROJECT_ROOT quarto rende
     "generate-testmodels",
     "initialize-julia",
 ], env = { "PYTHON_JULIACALL_CHECK_BOUNDS" = "yes", "PYTHON_JULIACALL_HANDLE_SIGNALS" = "yes", "PYTHON_JULIAPKG_OFFLINE" = "yes" } }
+quarto-publish = { cmd = "quarto publish gh-pages --no-render --no-prompt --no-browser", cwd = "docs", env = { "PRE_COMMIT_ALLOW_NO_CONFIG" = "1" } }
```
evetion, line 61: Might be good to comment this is for CI only, and wouldn't work out of the box if I run it locally right?

evetion (approved):

PR #2595, opened 2025-09-23
[python] Forbid any extra values in (base)model

### LINKED ISSUE #2577: ribasim.Model accepts and passes arbitrary keys
The right way to configure writing the results to NetCDF is

```toml
[results]
format = "netcdf
```

Which is done in Python like so:

```py
from ribasim import Model
from ribasim.config import Results

model = Model(
    starttime="2020-01-01",
    endtime="2020-01-20",
    crs="EPSG:28992",
    results=Results(format="netcdf"),
)
```

A user instead passed `results_format = "netcdf"` as a keyword argument to Model. This was accepted, and written to the TOML. Then the core crashed saying `results_format` is not a valid key.

I'd expect the Model construction to fail when supplied with invalid keys, to catch errors as early as possible.

evetion:
Fixes #2577.

This might break some Ribasim-NL usage.

I had to hardcode the edge field, as some (test)models still use it directly.

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -266,6 +266,10 @@ class TableModel(FileModel, Generic[TableT]):
     df: DataFrame[TableT] | None = Field(default=None, exclude=True, repr=False)
     _sort_keys: list[str] = PrivateAttr(default=[])
 
+    model_config = ConfigDict(
+        extra="allow",
+    )
+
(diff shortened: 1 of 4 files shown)


evetion:
Ok, made sure extra fields are allowed for initiating a TableModel to fix the mypy errors. I've removed the .edge fallback.

visr (approved):

PR #2605, opened 2025-10-01
Use mixed mode AD for Jac evaluation

### LINKED ISSUE #2176: Investigate performance for duplicated models
Performance drops significantly (i.e. hangs) when creating a combined model (without interconnecting them), even though the models all run individually.

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/2176 (?)

I haven't really looked into scaling yet, but I found an avenue for performance improvement: evaluating the Jacobian with mixed mode AD where the reverse mode is `AutoMooncake()`. Some notes:

- I haven't been able to make `CArrays` compatible with `Mooncake.jl`, so for now the state is a `Vector{Float64}` and wrapped in a `CArray` where needed
- Mooncake doesn't like `@threaded` code but it should be fine for our application, an issue came out of my discussions with a developer: https://github.com/chalk-lab/Mooncake.jl/issues/791. This is probably not going to be solved before merging this PR. There is a workaround with `@mooncake_overlay`.
- How many rhs calls are needed for either the forward or reverse mode depends on the sparse matrix coloring, I added a loop in `get_diff_eval` which tries different colorings to obtain the cheapest way to evaluate the Jacobian
- I found a bug in the `DifferentiationInterface` code for sparse matrix coloring for mixed mode AD which makes certain small models fail, I've been told this is easy to fix so it will probably be fixed shortly: https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/857
- I made the mixed mode Jacobian evaluation opt-in as an experimental feature because it adds a lot of precompile time.


SouthEndMusic:
This isn't needed with https://github.com/Deltares/Ribasim/issues/3113

Issue #1946, opened 2024-11-21
Improve Delwaq netCDF


evetion:
Currently our own netCDF is a valid ugrid, but the converted output from Delwaq isn't, and Delwaq gives errors on reading ours. When fixed, we could probably plot the Delwaq produced netCDF directly in QGIS/Quickplot.

evetion:
```
 Warning: retrieving attribute edge_coordinates failed --          -43
 Warning: retrieving attribute node_id failed --          -43
 Warning: retrieving attribute node_long_name failed --          -43
 One or more mesh attributes missing -
 this may cause problems in some postprocessing programs
```

Could be picked up in our current netcdf push (#2604 and others).

evetion:
@visr The generate code (in `python/ribasim/ribasim/delwaq/generate.py`) uses

```python
    uds = ugrid(G)
    uds.ugrid.to_netcdf(output_path / "ribasim.nc")
```

visr:
This looks like DelWAQ not liking the way https://github.com/Deltares/xugrid/ writes their NetCDF (or just the content we put in), is there some documentation that indicates what DelWAQ expects?

Huite:
I have to catch the train now, but this rings a bell: I think Delwaq expects a number of optional UGRID attributes that xugrid does not write by default. They can be easily added via the respective assign_coords methods: https://deltares.github.io/xugrid/api/xugrid.UgridDatasetAccessor.assign_node_coords.html

PR #2619, opened 2025-10-06
Add edge_coordinates attributes to generate.py

visr:
Fixes a part of #1946.

This option currently only adds edge coordinates:

https://github.com/Deltares/xugrid/blob/4e3851385ba43db89191e923ea19922ad467c531/xugrid/ugrid/ugrid1d.py#L265-L266

The warning about this missing is now gone, but the node_id and node_long_name warnings remain:

```
 Warning: retrieving attribute node_id failed --          -43
 Warning: retrieving attribute node_long_name failed --          -43
 One or more mesh attributes missing -
 this may cause problems in some postprocessing programs
```

### DIFF
--- a/python/ribasim/ribasim/delwaq/generate.py
+++ b/python/ribasim/ribasim/delwaq/generate.py
@@ -356,9 +356,9 @@ def generate(
             )
         )
 
-    # Generate mesh and write to NetCDF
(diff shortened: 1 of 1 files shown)


evetion:
Thanks!

PR #2621, opened 2025-10-07
Fix remaining DelWAQ netCDF warnings

### LINKED ISSUE #1946: Improve Delwaq netCDF

visr:
Follow up of #2619, fixes #1946.

Implements https://github.com/Deltares/Ribasim/issues/1946#issuecomment-3376328165

This removes the remaining two warnings.

Though in the logging it still says:

```
  Number of WARNINGS            :     2
  Number of ERRORS during input :     0
```

But I guess that refers to other warnings that are not logged to the terminal.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -5,5 +5,6 @@
 /tutorial/crystal-basin/
 guide/data/
 *.html
-*.quarto_ipynb
 objects.json
+
+**/*.quarto_ipynb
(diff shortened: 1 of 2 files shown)


docs/.gitignore
visr, line 1: I ran `pixi run quarto-preview` and this entry was added, even though a similar one was there already. So I removed *.quarto_ipynb and commit it to avoid this diff.

evetion (approved):

PR #2624, opened 2025-10-08
Formulate in terms of flows, solve in terms of storages

SouthEndMusic:
TL;DR: Computing the Jacobian with the performance of storage states without losing the water balance accuracy of flow states.

Our state vector $u$  consists mostly of cumulative flows, which we sum in our rhs to obtain the storages. Hence our rhs is of the form $f(Au)$, where $A$ is a highly sparse matrix, and $Au$ gives the sums of all the cumulative flows without other storage contributions like those from flow boundary and initial storage (and the PID integral terms). Using the chain rule, the Jacobian w.r.t. $u$ is

$$
   J_u \[f(Au)](u) = J\[f](Au)A,
$$ 

i.e. the Jacobian of $f$ in $Au$ multiplied by $A$. In the code $Au$ is denoted by `u_reduced`. The Jacobian that is now computed with automatic differentiation is only $J\[f](u_\text{reduced})$, which is a matrix with the same number of rows as the full Jacobian but much fewer colums. Furthermore, the product $Au$ is computed much less frequently because it only happens outside rhs calls.

It turns out that this is very beneficial for the lhm, making the Jacobian evaluation so much cheaper that it went from >80% of the runtime to <10%.

Since the code around AD is becoming somewhat elaborate and complex, I introduced a new file `differentiation.jl`.

(The branch name is not accurate)

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.7"
 manifest_format = "2.0"
-project_hash = "5fc5384346375b45391dc378ebc18a7e72201de3"
+project_hash = "6a3acc77e36731f0a10f5bd4f969b426915cdc16"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "27cecae79e5cc9935255f90c53bb831cc3c870d7"
@@ -2835,7 +2835,7 @@ uuid = "8e850ede-7688-5339-a07c-302acd2aaf8d"
 version = "1.59.0+0"
 
 [[deps.oneTBB_jll]]
-deps = ["Artifacts", "JLLWrappers", "Libdl"]
+deps = ["Artifacts", "JLLWrappers", "LazyArtifacts", "Libdl"]
 git-tree-sha1 = "d5a767a3bb77135a99e433afe0eb14cd7f6914c3"
 uuid = "1317d2d5-d96f-522e-a858-c73665f53c3e"
 version = "2022.0.0+0"


visr:
Ooh very interesting indeed.

core/src/differentiation.jl
```diff
@@ -0,0 +1,477 @@
...
+
+In the RHS of the ODE systems these states are summed to obtain their contribution to
+the Basin storages, and this is the only way these states are used in the RHS. The RHS is thus of the form
+`f(u) = g(A*u)` where `A` is a highly sparse matrix with the following structure:
+
+    ⎡ max one 1 and ⎢         ⎢         ⎢  ⠀⎤
+    ⎢ one -1 per    ⎢    I    ⎢    I    ⎢ 0 ⎢      Basin rows
+A = ⎢ col by graph  ⎢  evap.  ⎢  infl.  ⎢   ⎢
+    ⎢---------------⎢---------⎢---------⎢---⎢
+    ⎣       0       ⎢    0    ⎢    0    ⎢ I⠀⎦      PID integral rows
```
simulutions: Is I here the identity matrix? so ones on the diagonal of the "sub matrix" but then what is I evap and I infl? Could you clarify

core/src/differentiation.jl
```diff
@@ -0,0 +1,477 @@
...
+
+`Jf(u) = Jg(A*u)*A.`
+
+Note that the Jacobian of `g` has `length(u)` rows and `length(u_reduced)` columns. It turns out that taking this multiplication
+with `A` out of the AD Jacobian computation is very advantageous in terms of computation time. In the code `f` and `g` are both methods
+of `water_balance!`. We denote the Jacobian of `g` as `J_intermediate`. The custom Jacobian object that wraps `J_intermediate` is called
+`HalfLazyJacobian`, which contains the parameters to implicitly define `A` and other parameters needed for AD Jacobian evaluation.
+
+It turns out that we can make use of this structure of the Jacobian in the linear solve as well. The linear system is of the form
+`Wa = b`, where `W = -γ⁻¹I - J_intermediate*A`. It turns out this special form of the Jacobian can be utilized to solve the linear system in 2
```
simulutions: What is gamma or where can I find its definition?

core/src/config.jl
```diff
@@ -366,15 +366,21 @@ function get_ad_type(solver::Solver; specialize = true)
     end
 end
 
+struct RibasimLinearSolve{A} <: SciMLLinearSolveAlgorithm
+    algorithm::A
+end
```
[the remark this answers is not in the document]
SouthEndMusic: I added some documentation to clarify. In general variables and type parameters live in different places, so I tend not to worry about name clashes between them.

core/src/differentiation.jl
SouthEndMusic: Gamma is a solver algorithm dependent scalar, see the link I added.

SouthEndMusic:
> I reckon a cleaner implementation results from u = [S; Q] then eliminating Q in the linear solve

@Huite based on my experience with this PR, that goes beyond the hackability of the Newton solve from OrdinaryDiffEq.jl. It's not designed to deal with that

Huite:
Okay, but if it's that the case, it's probably worth documenting!

I should most likely study your changes more, but here's my half-baked view / assumptions:

* DifferentialEquations just wants to call `linsolve(W, rhs)`
* I guess we cannot control how `W` is formed, but we can control `J` and we can avoid formulating the terms we do not care about. J is backed by a sparse array and its shape has to concur with `u` so they might be structural or non-structural zeros?
* We may provide our own linsolve. Inside that linsolve we take the relevant part of W, solve for the update in S, then cheaply construct the update in Q.
* DifferentialEquations is none the wiser, it just gets the update for `u`.

(You can reasonably ask me to check this with a toy example...)

simulutions (approved):

visr:
Thanks for the comments @Huite. Since we're doing a release and we need the performance gain for LHM we'll go ahead with this. Agree that it would be nice if we can simplify this more in the future, in the interest of maintainability.

Huite:
I had a try at making a toy example with a very simple system of ODEs. The principle should be simple, the `jac` function can be provided; a linsolve can be provided as well and requires overloading/wrapping the `dolinsolve` as is done here. Unfortunately, that requires a number of additional overloads (as can be seen in the `differentiation.jl` file).

Of course Bart has already done the hard work of figuring this out. However, even if we can simplify some things, providing a custom linear solver that matches the DiffEq API requirements isn't trivial and should probably have its own section in the developer documentation.

PR #2646, opened 2025-10-13
Use JULIA_CPU_TARGET

visr:
Rather than the kwarg in JuliaC, let's try if using the environment variable JULIA_CPU_TARGET works better.

Related to https://github.com/Deltares/Ribasim/pull/2644

The native binaries are for `skylake-avx512`. Most people including me don't have `avx512`, hence I get:

```
❯ ribasim.exe .\ribasim.toml
ERROR: Unable to find compatible target in cached code image.
Target 0 (skylake-avx512): Rejecting this target due to use of runtime-disabled features
 ```

### DIFF
--- a/build/cli/src/main.rs
+++ b/build/cli/src/main.rs
@@ -54,19 +54,23 @@ fn main() -> ExitCode {
     }
 
     let shared_lib_path = match OS {
-        "windows" => exe_dir.join("bin/libribasim.dll"),
-        "linux" => exe_dir.join("lib/libribasim.so"),
-        "macos" => exe_dir.join("lib/libribasim.dylib"),
+        "windows" => "bin/libribasim.dll",
+        "linux" => "lib/libribasim.so",
+        "macos" => "lib/libribasim.dylib",
         _ => unimplemented!("Your OS is not supported yet."),
     };
+    let full_shared_lib_path = exe_dir.join(shared_lib_path);
     unsafe {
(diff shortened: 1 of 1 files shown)


evetion:
On Teamcity you should
> enable the Show all personal builds option under [User Profile](https://www.jetbrains.com/help/teamcity/configuring-your-user-profile.html) | General | UI settings.
to see my builds and download their artefacts.

Some patterns:
- Building for each new architecture increases the build time. Some experiments with the default flags and adding skylake-avx512 exceeded 2 hours on Windows.
- Linux and Windows differ significantly, Windows is easily twice as slow, and testing also differs.

On Teamcity Linux, this PR builds and all tests are green(!). On main, with setting `JULIA_CPU_TARGET` to `generic` or `generic;native` it builds and only the basic or basic_transient test fail with exitcode -11.

On Teamcity Windows, this PR fails horribly, Rust can't find the procedure, and neither can Python in the BMI tests (note that my last commit never touched the Python code, so that could be fair). On main, my personal `JULIA_CPU_TARGET` builds are green, and tests pass.

evetion:
> We can wait a little, but may need to start thinking about reverting stuff. Ideally only JuliaC -> PackageCompiler would be enough, but perhaps we need to go back to Julia 1.11 as well.

If we wait, we should use the old release/executable for the DSD. So no new features since then.

PR #2653, opened 2025-10-17
Add `run_ribasim(toml_path)` function

visr:
This adds a script entrypoint to the pyproject.toml that forwards to the ribasim.exe core, but adding automatic installation:

```
❯ pixi run ribasim --version
Ribasim CLI v2025.5.0 not found. Downloading...
Downloading: [=====-----------------------------------] 13.4% (44.1/327.8 MB)
```

It is stored in a cache dir that will be reused. It first looks on the path if it can find a compatible ribasim.exe

This feature can be used from Python as `run_ribasim`, like `run_delwaq`.

EDIT: the script entrypoint and automatic downloading is now taken out, see https://github.com/Deltares/Ribasim/issues/2662. Now we have:

image

### DIFF
--- a/build/tests/test_cli.py
+++ b/build/tests/test_cli.py
@@ -201,17 +201,38 @@ def test_run_ribasim_invalid_model(tmp_path):
 
 
 def test_run_ribasim_in_notebook(tmp_path, monkeypatch):
-    """Test run_ribasim() in notebook mode by mocking _is_notebook."""
+    """Test run_ribasim() in notebook mode by mocking _subprocess_handling."""
     from ribasim import cli
+    from ribasim.cli import SubprocessHandling
 
     model = ribasim_testmodels.basic_model()
     toml_path = tmp_path / "ribasim.toml"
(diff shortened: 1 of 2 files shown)


simulutions (approved):
Nice feature, well commented and tested so no comments  👍

evetion (commented):
I appreciate the effort in getting the different shells to work, but it's a lot of untested code.

I'm sad to see the auto downloader go, do we need some documentation on how users can download Ribasim and edit their PATH?

visr:
> I appreciate the effort in getting the different shells to work, but it's a lot of untested code.

True. This uses mocking to at least exercises the code paths. Other than that I manually tested this on all the shells, which comes down to running `run_ribasim(toml_path)` on a simulation that is long enough to show a progress bar and making sure you see similar output as ribasim.exe, with a progress bar updating in place.

Regarding the PATH documentation we already have this: https://ribasim.org/install.html#adding-ribasim-to-path-on-windows

PR #2667, opened 2025-10-25
Revert "Formulate in terms of flows, solve in terms of storages (#2624)"

visr:
This reverts commit e179af04030003ee2330d6950db55c49fd948f1b.

#2624 provides a great performance speedup for very large models, but I now see this is too much at the expense of small models.

For instance after https://github.com/Deltares/Ribasim/pull/2652#issuecomment-3418828189 we got down to an acceptable 10s wall time for `tabulated_rating_curve_control` from ribasim.exe. It seems #2624 brings that up to 1m11s. This can be reduced by setting the number of threads to 1, though only partially.

```
tabulated_rating_curve_control NoSpecialize: 10s total, Computation time: 21 milliseconds
tabulated_rating_curve_control NoSpecialize+2624: 1m11s total, Computation time: 10 seconds
tabulated_rating_curve_control NoSpecialize+2624-t1: 46s total, Computation time: 10 seconds
```

Since for the upcoming courses no-one is going to be running the LHM I feel we should revert this now, giving us time to find ways to make #2624 quicker to start.


SouthEndMusic:
@visr I did some digging with Claude, and I identified 2 probable causes of the latency:
- The length of the state vector is one of the type parameters of the `LinearSolution` object:
https://github.com/Deltares/Ribasim/blob/543e0269c856cbf3d4c652af8021c2f40c197b7c/core/src/differentiation.jl#L313-L321
There are now two of these objects (for the full state and the reduced state)
- I think by far the biggest culprit is this part:
https://github.com/Deltares/Ribasim/blob/543e0269c856cbf3d4c652af8021c2f40c197b7c/core/src/solve.jl#L785-L791
I tried to be clever here by avoiding threading if there aren't any connector node types with more than one node. The problem is that the non-threaded branch is not hit in `precompile.jl`. Speaking of which: maybe we should add some more models, involving transient forcing, allocation, control?

PR #2691, opened 2025-10-31
Fix the last Heisenbug

### LINKED ISSUE #2670: Undeterministic unstable solutions in physical layer due to allocation
In allocation_physics_test.jl ,the test "Primary Secondary Network Model" has a TODO: sometimes the model does not converge the first time step, rerun up to 5 times.

We have a model with allocation, and at t=0.0, every once in a while the physical layer reports the timestep is forced below floating point accuracy. The cause might be that the solution is unstable.

We need to investigate where this comes from, why this happens undeterministically and how to prevent this.

Possibly OrderedSet will help: https://github.com/Deltares/Ribasim/pull/2669

verheem:
Fixes #2670  and #2639

Summary: 
The last allocation heisenbug was caused due to reading uninitialized memory during the jacobian generation. 

Uninitialized memory cause:
The problem is that we would store a constant scalar flow_rate for pump or outlet as a time dependent interpolations in the 'StateAndTimeDependentCache'. For optimization purpose, we minimize the amount of interpolation evaluations since these are expensive. 

What we do is check if the flow rate has changed, or if we are at a new time step, since only then would we need to evaluate the time-interpolation to get the new flow rate.

However, based on the  constant flow_rate, this is never the case, which gave us problems at t = 0.

The solution was as simple as to just have different caches for time independent flow rates (constant scalars) and time dependent flow rates (interpolation objects).

### DIFF
--- a/core/src/solve.jl
+++ b/core/src/solve.jl
@@ -678,12 +678,13 @@ function formulate_pump_or_outlet_flow!(
             current_flow_rate[id.idx]
         elseif isassigned(node.time_dependent_flow_rate, node_idx)
             # get the time dependent flow rate from interpolation or cached value
-
-            # eval_time_interp is not used here because current_flow_rate
-            # lives in state_time_dependent_cache (for ContinuousControl support),
-            # and thus also has to be updated if t is not new but the last evaluation
-            # was with the other version of the cache (normal versus the one for AD)
-            node.time_dependent_flow_rate[node_idx](t)
+            eval_time_interpolation(
+                node.time_dependent_flow_rate[node_idx],
+                current_flow_rate,
+                id.idx,
+                p,
+                t,
+            )
         else
             # get the scalar flow rate from  (for DiscreteControl, Control by allocation or flows from the Static table)
             node.flow_rate[id.idx]


SouthEndMusic:
Let's review; the flow of a Pump or Outlet can come from:

- The corresponding `Static` table
- The corresponding `Time` table
- `DiscreteControl`
- `PidControl`
- `ContinuousControl`
- Control by allocation

These values are handled as follows:
- The values from the `Static` and `Time` table both end up in time interpolation objects (trivial ones for the static values) in the `flow_rate::Vector{ScalarLinearInteprolation}` field of the corresponding struct. This is part of the `ParametersIndependent` section of the parameters, which are considered constant by `DifferentiationInterface.jl`. When the a Pump or Outlet has `ContinuousControlType.None`, these are used.
- `PidControl` and `ContinuousControl` write their flows to `StateTimeDependentCache` for differentiability, which is obscured by the `CacheRef` system (i.e. the hoops mentioned in the writeup). This part is implemented quite well: The control type of Pumps and Outlets is set in `set_control_type!`, and in `formulate_pump_or_outlet_flow!` overwriting the flows in `state_time_dependent_cache` with the `flow_rate` interpolations is skipped for these control types.
- Control by allocation updates the trivial interpolation objects in in `flow_rate` in place, discrete control replaces them by trivial interpolation objects constructed during initialization. I'd say that these should both suffer from the Dual version of the `current_flow_rate_*` not being updated properly bug.

My conclusion is that `DiscreteControl`, Control by allocation and flows from the `Static` table should get their own (shared) cache in `ParametersIndependent` (preferably in the corresponding node struct). Then the logic in the part below should be:

```julia
flow_rate = if node_from_time_table
    <get flow rate from interpolation or cached value in `current_flow_rate_*`, now from `TimeDependentCache`>
elseif control_type != ContinuousControlType.None
   <get flow rate from cached value in `current_flow_rate_*` from `StateTimeDependentCache`>
else
   <get flow rate from new cache for `DiscreteControl`, Control by allocation and flows from the `Static` table>
end
```

https://github.com/Deltares/Ribasim/blob/c233e567f5a752d241d0f909489c3accd004c829/core/src/solve.jl#L660-L664

To avoid an indexing mess, it's easiest to just initialize each of these caches for each node, even if they aren't going to use it.

core/test/allocation_physics_test.jl
```diff
@@ -210,31 +210,18 @@ end
...
-        # Commented out to avoid DimensionMismatch: arrays could not be broadcast to a common size: a has axes Base.OneTo(19) and b has axes Base.OneTo(20)
-        # if !all(isapprox.(multiple_subs, single_sub; atol = 1e-8))
-        #     println(
-        #         "The flows over link $link_id differ by ",
-        #         maximum(single_sub .- multiple_subs),
-        #     )
-        #     @test false broken = true
-        # end
+        if !all(isapprox.(multiple_subs, single_sub; atol = 1e-8))
+            println(
```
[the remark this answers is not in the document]
simulutions: yup, otherwise we have too little informartion about the failing test

core/src/parameter.jl
```diff
@@ -776,7 +778,7 @@ A cache for intermediate results in `water_balance!` which can depend on both th
 this cache is required for automatic differentiation, where e.g. ForwardDiff requires these vectors to
 be of `ForwardDiff.Dual` type. This second version of the cache is created by DifferentiationInterface.
 """
-const StateTimeDependentCache{T} = @NamedTuple{
+const StateAndTimeDependentCache{T} = @NamedTuple{
```
SouthEndMusic: An interesting way to inflate your diff :P

core/src/solve.jl
```diff
@@ -5,7 +5,7 @@ water_balance!(du::CVector, u::CVector, p::Parameters, t::Number)::Nothing = wat
     du,
     u,
     p.p_independent,
-    p.state_time_dependent_cache::StateTimeDependentCache,
+    p.state_and_time_dependent_cache::StateAndTimeDependentCache,
```
SouthEndMusic: I'm not sure why this type annotation is actually here

simulutions: Hmm weird. Deleted

core/src/parameter.jl
simulutions: yeah sorry for that one xD. Next time I will do it in a separate PR

Issue #2438, opened 2025-07-07
Improve substance propagation using subtimesteps


SouthEndMusic:
See the description of the numerical scheme [here](https://github.com/Deltares/Ribasim/pull/2426#issuecomment-3035404726).

TL;DR: We can set up an ODE system which describes the spread of water over the last physical layer timestep, and solve that with a single timestep of an implicit numerical method. The result can then be used to compute the amount of substance that came along with those volumes.

Huite:
> The parallel ODE system idea is interesting, but doesn't that create much more overhead than putting everything in a single problem?

That depends! The parallel ODE would be worthwhile if the flow vs solute transport time scales are different, or if the solution method is different.

> In the current implementation we already solve for the concentrations every dt.

Yes, yet DELWAQ doesn't use the instantaneous flows (and the results may be satisfactory).

> The discussion of this issue and in the linked PR are about how many basins a substance can travel within a single timestep. Currently the the concentration computations do an Euler forward step of the same length as the physical layer timestep, which means that over these timesteps can only travel one Basin further.

And to compare, Euler backward requires a linear solve, but atleast only a single one (for a linear problem). Unlike Euler forward this is at least stable, but this has a different opposite problem. Where Euler forward can "only move one basin per step", Euler backward can move too many for a large time step and we get numerical dispersion. But you could look at higher order methods like second order [upwind](https://en.wikipedia.org/wiki/Upwind_scheme) which should be less diffusive spatially; but then you might also prefer second order in time (e.g. BDF2)...

Hence my general intuition: if you're solving system of ODEs, you might as well bring the appropriate tools (DifferentialEquations).

SouthEndMusic:
Closing in favor of https://github.com/Deltares/Ribasim/issues/3125

PR #2701, opened 2025-11-09
Add ResidenceTime tracer

evetion:
One of my shorter PRs, quite effective I'd say.

This adds a ResidenceTime tracer, which tracks the average *age* of the water. Since we assume instantaneous perfect mixing (we're tracking water after all), this equals the average residence time. You could also track it using `storage/outflow_rate`, but that uses average flows, and can't handle no (Inf) or negative outflow.

The PR is short because age is easy to calculate; all inflows have zero concentration (age) by default, and we add a simple aging method (+dt) for all mass already in the basin. In contrast, tracking age for a tracer is much more involved.

HWS with the tracer visualised in days (output is seconds) at the end of the simulation, nicely showing dead zones in the model.
Screenshot 2025-11-09 at 21 21 23

In preparation of #2438. Given the `dt`s of hours in the HWS model above, one can see that tracers along the main rivers will be delayed, as their residence times have a similar order. If you would divide this output by the `dt` you have the Courant number (or as close to it in a 0d model like this).

### DIFF
--- a/core/test/run_models_test.jl
+++ b/core/test/run_models_test.jl
@@ -310,6 +310,21 @@ end
     table = Ribasim.concentration_data(model)
     @test "Continuity" in table.substance
     @test all(isapprox.(table.concentration[table.substance .== "Continuity"], 1.0))
+    summed_source_concentrations = reduce(
+        +,
+        [
+            table.concentration[table.substance .== substance] for substance in [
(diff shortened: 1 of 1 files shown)


visr (commented):
Nice feature! Some tests and docs would be nice though.

Issue #2686, opened 2025-10-28
Installer for core


visr:
Following discussion in https://github.com/Deltares/Ribasim/issues/2662 we decided to work on a WiX installer for the Ribasim core, to simplify the installation process. This would replace the installation script added in https://github.com/Deltares/Ribasim/pull/2682. We keep the zips as-is.

We can check if [cargo-wix](https://github.com/volks73/cargo-wix) would make it easier to create a WiX installer, given that we build a Rust CLI wrapper.

It should also offer an uninstall feature. We can still consider `ribasim self-update` as well as long as it doesn't interfere with this. If we do self-update we probably want to standardize the zip naming to allow for more platflorms and architectures in the future, following rust target conventions also used for e.g. uv.

PR #2710, opened 2025-11-14
Create MSIX app package

visr:
Fixes https://github.com/Deltares/Ribasim/issues/2686.
This follows the Microsoft guide from https://learn.microsoft.com/en-us/windows/msix/package/manual-packaging-root

image

It automatically puts `ribasim.exe` on the path at `C:\Users\visser_mn\AppData\Local\Microsoft\WindowsApps\ribasim.exe` via the recommended `AppExecutionAlias` mechanism.

The [capabilities](https://learn.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-manual-conversion#capabilities) are on `runFullTrust`. I looked a bit into this but makeappx.exe states that for this package manifest this is required.

The "Launch when ready" toggle cannot be controlled by the manifest. If left checked you'll just see a quick flash of a terminal which then closes, as you'd expect when running ribasim.exe without arguments.

Uninstallation is automatically available:

image

This still needs to be hooked up to TeamCity's code signing. The MSIX can only be used when signed. To test I signed it with a self-signed certificate, but that doesn't work on other computers. I provide some PowerShell commands here that can be used to create the `ribasim_windows.msix` under the `/build/` directory.

```ps1
# Add Windows SDK to PATH for this session
$env:Path += ";C:\Program Files (x86)\Windows Kits\10\App Certification Kit"
# Versioned path: c:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x64\

# copy files needed for MSIX to the build directory
# we could also check them in there, but I'd rather not have these files in the .zip artifacts
# we could also supply a mapping file with `/f` instead of the directory argument `/d` to `makeappx`
# but that would require us to maintain the mapping file with currently over 1k files
Copy-Item AppxManifest.xml ribasim/
Copy-Item logo-150.png ribasim/
Copy-Item logo-44.png ribasim/

# pack the directory ribasim to ribasim_windows.msix, overwriting existing msix files
makeappx pack /o /d ribasim /p ribasim_windows.msix

# probably best to use existing signing TeamCity scripts as a template
SignTool sign /fd SHA256 /sha1 <cert_sha> ribasim_windows.msix
```

This puts ribasim.exe on the path, but not the bin dir with all DLLs and libribasim.dll. For libribasim users it is probably a better idea to keep using the zip install.

EDIT: whilst `ribasim --version` works fine, running a model doesn't, it cannot find libribasim:

```
❯ ribasim ribasim.toml

thread 'main' panicked at src\main.rs:63:49:
called `Result::unwrap()` on an `Err` value: LoadLibraryExW { source: Os { code: 126, kind: Uncategorized, message: "The specified module could not be found." } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is related to there being limited file permissions in the installation directory under WindowsApps.
One way to avoid this is to put ribasim.exe in the bin dir with all the libraries.
We still need to see if we can avoid  TimeZones.jl writing to scratchspaces in this restricted dir:

<details><summary>Scratchspaces</summary>
<p>

```
InitError(mod=:TZData, error=Base.IOError(msg="mkdir(\"C:\\\\Program Files\\\\WindowsApps\\\\Deltares.Ribasim_2025.6.0.0_x64__m018s8shtq18w\\\\share\\\\julia\\\\scratchspaces\"; mode=0o777): permission denied (EACCES)", code=Int32(-4092)))
uv_error at .\libuv.jl:106 [inlined]
#mkdir#8 at .\file.jl:185
mkdir at .\file.jl:177 [inlined]
#mkpath#9 at .\file.jl:241
mkpath at .\file.jl:235
#mkpath#9 at .\file.jl:239
mkpath at .\file.jl:235
#mkpath#9 at .\file.jl:239
mkpath at .\file.jl:235 [inlined]
#get_scratch!#10 at D:\buildAgent\work\ecd2b8f9b25b1609\.julia\packages\Scratch\tDtSY\src\Scratch.jl:245
get_scratch! at D:\buildAgent\work\ecd2b8f9b25b1609\.julia\packages\Scratch\tDtSY\src\Scratch.jl:230 [inlined]
macro expansion at D:\buildAgent\work\ecd2b8f9b25b1609\.julia\packages\Scratch\tDtSY\src\Scratch.jl:315 [inlined]
_scratch_dir at D:\buildAgent\work\ecd2b8f9b25b1609\.julia\packages\TimeZones\dt27j\src\TimeZones.jl:35
__init__ at D:\buildAgent\work\ecd2b8f9b25b1609\.julia\packages\TimeZones\dt27j\src\tzdata\TZData.jl:19
jfptr___init___195813.1 at C:\Program Files\WindowsApps\Deltares.Ribasim_2025.6.0.0_x64__m018s8shtq18w\bin\libribasim.dll (unknown line)
jl_apply at C:/workdir/src\julia.h:2157 [inlined]
jl_module_run_initializer at C:/workdir/src\toplevel.c:76
_finish_julia_init at C:/workdir/src\init.c:902
```

</p>
</details> 

We can look into using Preferences.jl to avoid writing here. That doesn't seem directly supported though. It seems that I have `scratchspaces\f269a46b-ccf7-5d73-abea-4c690281aa53\build\`, so perhaps it is only writing an empty directory? Alternatively, we can look into making the scratchspaces folder writable, see https://learn.microsoft.com/en-us/windows/msix/psf/psf-filesystem-writepermission
EDIT: fixed by https://github.com/Deltares/Ribasim/pull/2738

If that works, we can also look if it is possible to specify `libribasim.dll` in the AppxManifest.xml as a product, and see if other apps (BMI) can then use that. EDIT: no this is not possible


build/AppxManifest.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
+  xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities">
+  <Identity Name="Deltares.Ribasim" Version="2025.6.0.0"
+    Publisher="CN=Stichting Deltares, O=Stichting Deltares, L=Delft, C=NL"
```
visr: This is identical to the publisher of the TeamCity signed ribasim.exe, which is a requirement.

build/AppxManifest.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
+  xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities">
+  <Identity Name="Deltares.Ribasim" Version="2025.6.0.0"
```
visr: The version consists of four components. I suggest using the fourth digit as a way to create new builds for the same code. You cannot update the installed version without increasing the version number.

build/AppxManifest.xml
```diff
@@ -0,0 +1,39 @@
...
+    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.14393.0"
+      MaxVersionTested="10.0.26200.0" />
+  </Dependencies>
+  <Capabilities>
+    <rescap:Capability Name="runFullTrust" />
+  </Capabilities>
+  <Applications>
+    <Application Id="Ribasim" Executable="ribasim.exe" EntryPoint="Windows.FullTrustApplication">
+      <uap:VisualElements DisplayName="Ribasim" Description="Water resources modeling"
+        BackgroundColor="transparent" Square150x150Logo="logo-150.png" Square44x44Logo="logo-44.png" />
```
visr: I created these from the original `lo_ribasim__beeldmerk.jpg` from the RIBASIM sharepoint. This is 1109 x 1110 pixels. I first converted to PNG and then resampled to the required sizes.

We should probably check with the visualization department if they don't have an original PNG or SVG rather than JPG. It would probably also be better to have the corners transparent rather than white.

https://deltares.sharepoint.com/:f:/r/teams/RIBASIM/Shared%20Documents/Website/Logo/final?csf=1&web=1&e=bpoUdF

build/AppxManifest.xml
visr: The `Identity Name="Deltares.Ribasim"` part is important to make sure we don't have to change that later on.
I'm not aware of other MSIX packages at Deltares (none in the Microsoft Store). But it could be that e.g. people want us to use "Stichting Deltares" rather than "Deltares". If we have to change that, I think they essentially become two separate Ribasims, which we want to avoid.

deltamarnix: Yes, let's look at that last number like the `build` argument in conda forge. So, creating a new build with the same underlying build tag

build/AppxManifest.xml
deltamarnix: Is it possible to automate the version number? Maybe template it in a way? Examples on StackOverflow show how to load dthe AppxManifest.xml in powershell and do some xml transformations.

https://stackoverflow.com/questions/57094081/how-can-i-automate-supplying-a-version-in-appxmanifest

visr: Yes should be possible, made a separate issue for it: https://github.com/Deltares/Ribasim/issues/2757

PR #2732, opened 2025-11-25
Recommend Pixi VSCode Extension

visr:
And tell the python environment extension to use it.

https://github.com/renan-r-santos/pixi-code/
https://github.com/microsoft/vscode-python-environments

There is a lot changing related to python environments in VSCode. Yesterday it suddently wouldn't find my environment, despite starting with `pixi run code .` But this helped. I suggest keeping `open-vscode.bat` as well for the time being. I don't think this does any harm even if you don't use the environments extension. I do have it enabled now, with:

```json
"python.useEnvironmentsExtension": true
```

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,15 +1,16 @@
 {
     "recommendations": [
-        "julialang.language-julia",
-        "ms-python.python",
-        "ms-python.mypy-type-checker",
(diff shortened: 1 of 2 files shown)


evetion (approved):
LGTM. Why didn't you also put `"python.useEnvironmentsExtension": true` into the json?

PR #2733, opened 2025-11-25
Allow replacing/deleting links.

evetion:
Also exposes these methods (`remove_node` and `remove_link`) on the model instance.


visr (approved):
Looks good to me, added a comment on the added Python API.

PR #2762, opened 2025-12-08
[core] Use SQLite fork for faster read performance.

### LINKED ISSUE #2752: Investigate sqlite performance for large tables.
Since we're dropping arrow, and netcdf might not be optimal for some tables. Sqlite is used for pretty large databases, so our geopackage/python code could be optimized by some pragmas further still.

https://github.com/Deltares/Ribasim/issues/2521

evetion:
This uses the strict mode of SQLite.jl (parsing using the declared type, not detecting it for each row/column) using a fork of mine that fixed the strict usage (and some type instabilities).

The previous manual written loop was much slower than just using columntable again.

Fixes #2752

### DIFF
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -1951,6 +1951,7 @@ function load_data(
     end
 end
 
+"Faster alternative to Tables.columntable that preallocates based on the schema."
 function sqlite_columntable(
     table::Query,
     db::DB,


core/src/read.jl
```diff
@@ -1966,24 +1965,16 @@ function sqlite_columntable(
...
-                        dateformat"yyyy-mm-dd HH:MM:SS.s",
-                    )
-                end
-            end
-            nt[name][i] = val
+    for (name, column) in pairs(columntable(table))
+        if name == :time
+            df = dateformat"yyyy-mm-dd HH:MM:SS.s"
+            # remove sub ms precision
+            @. nt[name] = [ismissing(val) ? DateTime(config.starttime) : DateTime(replace(val, r"(\.\d{3})\d+$" => s"\1"), df) for val in column]
```
visr: This is a bit hard to read, could you split this up?

PR #2763, opened 2025-12-09
Add indexes to geopackage.

### LINKED ISSUE #1973: Database disk image is malformed
Locally I cannot run `pixi run generate-testmodels`. It tends to fail on `basic` or `basic_arrow`. I have no idea why this started happening now. It looks like `self.filepath = filepath` during `model.write` is triggering the reading of the database, which doesn't look desirable.

Sometimes can be reproduced by

```py
import ribasim_testmodels
model = ribasim_testmodels.basic_model()
model.write("d:/temp/model/trivial.toml")
```

```
Pixi task (generate-testmodels in dev): python utils/generate-testmodels.py
Generating allocation_training
Generating allocation_example
Generating backwater
Generating basic_arrow
Generating basic
  File "A:\repo\ribasim\Ribasim\utils\generate-testmodels.py", line 49, in <module>
    model.write(datadir / model_name / "ribasim.toml")
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\model.py", line 307, in write
    self.filepath = filepath
    ^^^^^^^^^^^^^
  File "A:\repo\ribasim\Ribasim\.pixi\envs\dev\Lib\site-packages\pydantic\main.py", line 923, in __setattr__
    self.__pydantic_validator__.validate_assignment(self, name, value)
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\model.py", line 129, in _ensure_edge_table_is_present
    self.edge.df = self.edge.df.set_geometry("geometry", crs=self.crs)
    ^^^^^^^^^^^^
  File "A:\repo\ribasim\Ribasim\.pixi\envs\dev\Lib\site-packages\pydantic\main.py", line 923, in __setattr__
    self.__pydantic_validator__.validate_assignment(self, name, value)
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\input_base.py", line 121, in _check_filepath
    data = cls._load(filepath)
           ^^^^^^^^^^^^^^^^^^^
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\input_base.py", line 240, in _load
    ddf = cls._from_db(db, cls.tablename())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\input_base.py", line 378, in _from_db
    if exists(connection, table):
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "A:\repo\ribasim\Ribasim\python\ribasim\ribasim\db_utils.py", line 15, in exists
    cursor.execute(
sqlite3.DatabaseError: database disk image is malformed
```

evetion:
This adds an index to each node_id (and link_id) column in the geopackage. This makes identification in QGIS much faster, and even the LHM now feels snappy (before it took a few seconds for a node to be identified).

The root cause is that the (QGIS specific) relations between node tables required a lot of lookups, and without an index, that's a full search in tables. That becomes noticeable at the scale of the LHM. This PR does make geopackages 10% larger.

Ideally we would also make the relations explicit in the geopackage, but (geo)pandas doesn't support that. Creating tables ourselves is doable, but not sure if it's worth it.

Also fixes #1973, by not opening the database (triggered by setting the filepath) while writing it.

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -596,7 +596,7 @@ class SpatialTableModel[TableT: _BaseSchema](TableModel[TableT]):
                 self.tableschema()._index_name(),
                 unique=True,
             )
-        _add_styles_to_geopackage(path, self.tablename())
+            _add_styles_to_geopackage(connection, self.tablename())
 
 
 class ChildModel(BaseModel):
--- a/python/ribasim/ribasim/styles.py
+++ b/python/ribasim/ribasim/styles.py
@@ -1,8 +1,7 @@
 import logging
-from contextlib import closing
 from datetime import datetime
 from pathlib import Path
-from sqlite3 import connect
+from sqlite3 import Connection
 
 STYLES_DIR = Path(__file__).parent / "styles"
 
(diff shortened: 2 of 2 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -387,7 +387,7 @@ def write(self, filepath: str | PathLike[str]) -> Path:
             self._validate_model()
 
         filepath = Path(filepath)
-        self.filepath = filepath
+        self.set_filepath(filepath)
```
visr: Nice find!

PR #2764, opened 2025-12-09
Remove activity

### LINKED ISSUE #2623: Remove (in)activity for now
As it makes the code more complex (sparsity patterns can change over time, among others) and is hardly used.

We should document new ways of quickly enabling/disabling nodes (maybe just removing them with Python?).

A in between is to remove the temporal functionality, and only support active in the Node table?

evetion:
Fixes #2623

~Draft for now, as I'd have to inspect the tests.~ An inactive resistance is probably infinite instead of zero, and I guess some tests can be removed altogether.

### DIFF
--- a/python/ribasim/ribasim/migrations.py
+++ b/python/ribasim/ribasim/migrations.py
@@ -28,7 +28,7 @@ def check_inactive(df: DataFrame, name: str):
     nodes = df["node_id"][~df["active"].isin([True, None])].tolist()
     if len(nodes) > 0:
         raise ValueError(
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/migrations.py
visr: By just dropping the column we can convert inactive nodes to active ones, changing results is perhaps unexpected ways. What do you think about aborting the migration if any inactive nodes are found, with a message that this is no longer supported, and they adapt the parameters to get the same effect? Setting flow_rate to 0 and resistance to Inf probably goes a bit too far, since I don't expect much/any usage.

PR #2765, opened 2025-12-10
Generate consistent binary Geopackage

evetion:
By removing all dynamic timestamps from the geopackage tables. This is required for DVC trains in Ribasim-NL, as otherwise rerunning a generate script with identical input, would yield different binary outputs (and thus md5 hashes).

The ability to do so is explicitly mentioned for an OGR config setting: https://gdal.org/en/stable/drivers/vector/gpkg.html#configuration-options.

### DIFF
--- a/python/ribasim/ribasim/db_utils.py
+++ b/python/ribasim/ribasim/db_utils.py
@@ -2,6 +2,10 @@ from contextlib import closing
 from pathlib import Path
 from sqlite3 import Connection, connect
 
+# A fixed date for `last_change` fields in the metadata tables in a geopackage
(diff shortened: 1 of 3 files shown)


python/ribasim/ribasim/db_utils.py
```diff
@@ -20,10 +20,12 @@ def exists(connection: Connection, name: str) -> bool:
 
 
 def _set_gpkg_attribute_table(connection: Connection, table: str) -> None:
+    from ribasim import fake_date
```
visr: Is is really preferred style to put these internal imports inside functions? Or is there another reason to do it like this?

PR #2767, opened 2025-12-12
add skip_fair_fraction to Allocation struct

verheem:
To solve directional flow in allocation, for example polder level management in the most general case, where water may be allocated from boezem to polder but not the other way around, we need the possibility to skip the fair fraction objective. This may be relevant in the LHM


verheem:
> @verheem can this be closed?

This would be relevant for turning on allocation in the LHM, but requires more work. I added a small description.

PR #2775, opened 2025-12-16
Fix UserDemand migration

evetion:
As #2772 was incorrect (it overwrote an older migration)


visr (approved):
Looks good, I resolved the merge conflict by having both is schema < 4 and if schema < 9.

PR #2808, opened 2026-01-12
Switch between control state and ribasim.allocation

### LINKED ISSUE #2703: Support switching between control and allocation
> It is intended that we can have control switch between `Ribasim.allocation` or other control states. An example is a reservoir that can release water based on downstream demand, but then switch to flood operation without any allocation when needed. I'm not sure this currently works, but it would be great to work out and document such an example.
> 
> You are right that it is a bit magic, but at the same time it aligns well in the existing concept of `control_state`. Having to introduce a separate concept can possibly also make things more difficult to grasp. We add the `Ribasim.` prefix to signify that this value has special meaning to Ribasim. If we don't do it already, we should error on any control states with that prefix that we don't know, to reserve it for our use. 

 _Originally posted by @visr in [#2527](https://github.com/Deltares/Ribasim/issues/2527#issuecomment-3497844894)_

https://ribasim.org/concept/allocation.html#control-by-allocation

verheem:
fixes #2703

### DIFF
--- a/core/src/bmi.jl
+++ b/core/src/bmi.jl
@@ -3,7 +3,7 @@
 
 Initialize a [`Model`](@ref) from the path to the TOML configuration file.
 """
-function BMI.initialize(T::Type{Model}, config_path::AbstractString)::Model
+function BMI.initialize(::Type{Model}, config_path::AbstractString)::Model
     config = Config(config_path)
     mkpath(results_path(config))
     io = open(results_path(config, "ribasim.log"), "w")
@@ -89,11 +89,11 @@ function BMI.get_value_ptr(model::Model, name::String)::Vector{Float64}
 end
 
 BMI.get_current_time(model::Model)::Float64 = model.integrator.t
-BMI.get_start_time(model::Model)::Float64 = 0.0
+BMI.get_start_time(::Model)::Float64 = 0.0
 BMI.get_time_step(model::Model)::Float64 = get_proposed_dt(model.integrator)
 
 function BMI.get_end_time(model::Model)::Float64
     return seconds_since(model.config.endtime, model.config.starttime)
 end
 
-BMI.get_time_units(model::Model)::String = "s"
+BMI.get_time_units(::Model)::String = "s"


core/src/callback.jl
```diff
@@ -672,6 +670,9 @@ function set_new_control_state!(
     # If there is a change, update parameters and the discrete control record
     control_state_now = discrete_control.control_state[discrete_control_id.idx]
     if control_state_now != control_state_new
+        print(
+            "New control state $control_state_now for $discrete_control_id at t = $(integrator.t): \n",
+        )
```
visr: Should we turn this into debug logging? Or remove.

python/ribasim_testmodels/ribasim_testmodels/allocation.py
```diff
@@ -2196,3 +2196,112 @@ def polder_management_model() -> Model:
...
+    level_demand7 = model.level_demand.add(
+        Node(7, Point(0.0, 1.0)),
+        [level_demand.Static(min_level=[0.9], max_level=[0.9], demand_priority=1)],
+    )
+
+    # Give all nodes subnetwork_id 1
+    for node in model._nodes():
+        # check if node is None
+        if node.node.df is not None:
+            node.node.df.subnetwork_id = 1
```
visr: Is this still needed given #2766?

I'd prefer not to use internal functions (leading underscore) in `ribasim_testmodels`, as people might use them as examples.

core/src/allocation_optim.jl
```diff
@@ -1040,6 +1040,35 @@ function reset_cumulative!(allocation_model::AllocationModel)::Nothing
...
+    (; problem) = allocation_model
+
+    try
+        constraints = problem[constraint_key]
+        for constraint in constraints
+            JuMP.delete(problem, constraint)
+        end
+    catch err
+        err isa KeyError || rethrow(err)
+    end
```
visr: Usually try-catch should be avoided in inner code like this, as part of the [style guide](https://docs.julialang.org/en/v1/manual/style-guide/#Don't-overuse-try-catch) but also for performance reasons. Could you rewrite this to use `if haskey(problem, constrain_key)`? That also makes it clear where we expect this code to not apply.

Probably not something to prematurely optimize, but it this the first time we modify the JuMP problem structure (as opposed to values) at runtime? I wonder about the performance implications, if this means JuMP/HiGHS has to throw away caches and such.

core/src/callback.jl
simulutions: lol no forgot to remove this one

visr:
This needs some work to pass CI it seems. On main `polder_management_subnetwork_id_1` fails, and this adds more failures. I tried to `pixi run write-allocation-problems` locally, but that failed for me:

```
Wrote allocation problem(s) for multiple_route_priorities
┌ Error: Failed to process model multiple_source_priorities
│   exception = SQLite.SQLiteException("no such column: route_priority")
└ @ Main C:\ProgramData\DevDrives\repo\ribasim\Ribasim\utils\write_allocation_problems.jl:54
```

PR #2811, opened 2026-01-15
Replace cascading polder example

### LINKED ISSUE #2736: Docs: Replace cascading polder example
Given that https://github.com/Deltares/Ribasim/issues/2723 is a very good example, we need to update https://ribasim.org/guide/examples.html#guidance-of-modeling-a-cascade-of-polder-basins

verheem:
Fixes #2736

### DIFF
--- a/python/ribasim_testmodels/ribasim_testmodels/allocation.py
+++ b/python/ribasim_testmodels/ribasim_testmodels/allocation.py
@@ -2057,7 +2057,7 @@ def polder_management_model() -> Model:
     evaporation[180:270] = 0
     evaporation[270:366] = 1e-6
 
-    basin_data = [
+    basin_data: list[TableModel[Any]] = [
         basin.Profile(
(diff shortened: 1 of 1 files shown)


docs/getting-started/examples.qmd
```diff
@@ -557,271 +557,563 @@ ax.plot(times, level_demands, color="k", ls=":", label="target level")
...
+
+- You need explicit control over switching behavior
+- You want to implement custom control logic
+
+The key differences:
+
+- **Discrete control** instead of allocation for pumps and outlets
+- **Hysteresis logic** to prevent rapid switching (different on/off thresholds)
+- **No level demand nodes** - target levels are embedded in discrete control logic
+- **Flow demands** still work but with regular discrete control
```
visr: This reads a bit like it is equivalent, or at least I don't see clear downsides to this approach. Should we add this? I guess a downside is that you cannot work with priorities, and that the intention might not be as clear?

I'm even wondering a bit if we should add this approach to the docs at all, since we probably want people to use allocation for this.

Issue #1813, opened 2024-09-09
Refactor node table in Python


evetion:
Given node ids are unique again #1717, we can move away from the multinodetable approach, which should simplify our code (and validation) considerably. This would not be breaking, so low priority for now.

visr:
Right now we have `mode.node_table()` which creates the Node table on demand, and the actual data is stored per node type under e.g. `model.terminal.node`.

It would probably indeed simplify things to have the real table in one place. It would probably be breaking to remove `model.terminal.node`, and getting the node table for one node type is quite convenient as well. So we should see it we can keep that.

PR #2878, opened 2026-02-06
[Python] Lazy reading and writing.

evetion:
This refactors the read/write (and _load/_save) methods considerably and introduces:

- `Model.read(toml_path, internal=False, external=False)` only reads the model config, and nothing else. It means no tables, or empty ones in the case of Node/Link, which can be written to disk as well. It's much like the `Model(startime, endtime, crs)` constructor result in that sense.
- `Node.read(external=True, internal=True)` to load all node tables from disk (overwriting/resetting them) if any changes were done.
- `TableModel.read()` for individual tables, same caveats as above.

- `Model.write(toml_path, toml=True, internal=True, external=True)` to control writing all geopackage tables (internal) and/or all arrow/netcdf files (external).
- `Node.write(internal=True, external=True)` for writing nodes (e.g. all Basin tables)
- `TableModel.write()` to write individual tables

One needs to set the Model (toml) filepath first, so the input folder can be determined.

This allows for more granular control of writing for the HydroMT plugin. However, note that this API introduces some edge cases that need careful handling. It becomes much easier to create invalid/inconsistent models. For example, reading a single table and afterwards adding new nodes might clash (non-unique node ids) with another yet unread table. That case will be solved by #1813. Another one is writing a single non-spatial table, yielding a sqlite database, not a valid geopackage.

This also refactors ChildModel and introduces a ParentModel (both now using composition instead of inheritance) so an empty lazy table can find the toml path of its (grandparent) Model.

@hboisgon see the `test_hydromt.py` for API examples that might kickstart the HydroMT plugin.

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -239,6 +239,7 @@ class FileModel(BaseModel, ABC):
 
     @classmethod
     def allows_lazy(cls) -> bool:
+        """Whether this FileModel allows lazy loading."""
         return True
 
     @classmethod
@@ -308,6 +309,11 @@ class FileModel(BaseModel, ABC):
 
 
 class ChildModel(BaseModel):
+    """Base class to represent models that are contained in a ParentModel.
+
+    These include NodeModels and all Tables, but also the Config groups.
+    """
+
     _parent: BaseModel | None = None
     _parent_field: str | None = None
 
@@ -319,7 +325,10 @@ class ChildModel(BaseModel):
 
     @property
     def root(self) -> BaseModel | None:
-        """Return the upmost parent model of this ChildModel."""
+        """Return the *uppermost* parent model of this ChildModel.
+
+        Useful to get the Model directly from a Table.
+        """
         if self._parent is None:
             return self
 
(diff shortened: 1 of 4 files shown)


visr:
Shouldn't we strive for hydromt to only use public API though? Do you see issues with adding the new `_save` methods/kwargs to `write`?

evetion:
We can easily make new write methods to make it public yes. But maybe we should put it into an advanced documentation section? It becomes easy to write incomplete and thus invalid models on read.

The lazy load is actually even more difficult, and sorta requires the refactor of the multinodetable. Now you can create or load tables that have a nodeid clash with a table that's not loaded.

That's why I'm ok with the hydromt plugin handling that state in a consistant state internally.

python/ribasim/ribasim/input_base.py
```diff
@@ -282,19 +299,44 @@ class ChildModel(BaseModel):
...
     @model_validator(mode="after")
-    def _check_parent(self) -> "ChildModel":
+    def check_parent(self) -> "ChildModel":
         if self._parent is not None and self._parent_field is not None:
             self._parent.model_fields_set.update({self._parent_field})
         return self
 
+    @property
+    def root(self) -> BaseModel | None:
+        """Return the parent model of this ChildModel."""
```
visr: Would parent then not be a better name?

visr (approved):

PR #2880, opened 2026-02-06
Load NetCDF input in QGIS

### LINKED ISSUE #318: QGIS support NetCDF input files
In addition to #170, we should support loading tables from Arrow files into QGIS as well. This would be a read-only table due to the lack of editing support of Arrow in QGIS.

Right now QGIS loads a model from a GeoPackage, but really we should load the TOML and find the GeoPackage and Arrow tables from there.

EDIT: renamed from Arrow to NetCDF. I think just loading the NetCDF as a layer without the relations set up like we have for GeoPackage would be enough.

verheem:
fixes #318

### DIFF
--- a/ribasim_qgis/core/nodes.py
+++ b/ribasim_qgis/core/nodes.py
@@ -29,6 +29,22 @@ from ribasim_qgis.core import geopackage
 STYLE_DIR = Path(__file__).parent / "styles"
 
 
+def pascal_to_snake_case(name: str) -> str:
+    """Convert PascalCase to snake_case.
+
+    Parameters
+    ----------
+    name : str
+        PascalCase string to convert.
+
+    Returns
+    -------
+    str
+        snake_case version of the input string.
+    """
+    return "".join(["_" + c.lower() if c.isupper() else c for c in name]).lstrip("_")
+
+
 class Table:
     """Base class for Ribasim input layers."""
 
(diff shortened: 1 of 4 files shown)


ribasim_qgis/core/nodes.py
```diff
@@ -120,3 +120,125 @@ def load_nodes_from_geopackage(path: Path) -> dict[str, Table]:
...
+    return external_tables  # type: ignore[return-value]
+
+
+class ExternalTable(Table):
+    """Represents a table stored in an external NetCDF file."""
+
+    def __init__(self, input_type: str, path: Path):
+        super().__init__(input_type, path)
+
+    def layer_from_geopackage(self) -> QgsVectorLayer:
```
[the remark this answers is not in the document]
simulutions: layer_from_file seems cleanest to me

ribasim_qgis/core/nodes.py
```diff
@@ -120,3 +120,128 @@ def load_nodes_from_geopackage(path: Path) -> dict[str, Table]:
...
+    external_files = {}
+
+    # Derive node types from existing tables set
+    # Extract unique node types (the part before ' / ')
+    node_types = {table.split(" / ")[0] for table in tables if " / " in table}
+
+    for node_type in node_types:
+        # Convert PascalCase to snake_case for TOML lookup
+        snake_case_type = "".join(
+            ["_" + c.lower() if c.isupper() else c for c in node_type]
```
evetion: nit: This could live in an utility function (maybe it exists already?).

ribasim_qgis/core/nodes.py
```diff
@@ -120,3 +120,128 @@ def load_nodes_from_geopackage(path: Path) -> dict[str, Table]:
...
+        ).lstrip("_")
+
+        if snake_case_type not in toml_data:
+            continue
+
+        node_config = toml_data[snake_case_type]
```
evetion: nit: You can write these patterns more easily with `toml_data.get(snake_case_type, {})`. You get an empty dict back if if the key doesn't exist.

evetion (approved):

PR #2882, opened 2026-02-09
Switch from pre-commit to prek

visr:
> prek is a reimagined version of pre-commit, built in Rust. It is designed to be a faster, dependency-free and drop-in alternative for it, while also providing some additional long-requested features.

Prek v0.3.2 added Julia support (it was added to pre-commit a year ago, but it wasn't working reliably for me, see https://github.com/Deltares/Ribasim/pull/2825). So this adds the Runic formatting check back.

https://github.com/j178/prek/releases/tag/v0.3.2
https://prek.j178.dev/

- [x] Needs `j178/prek-action` to be whitelisted first, I made an issue for that. EDIT: now uses prek from pixi
- [x] Needs 1ce27a90e2fc1c0cc61167e3da1ed921c836ab3a to be reverted, but I want to use it to verify that the runic hook works first.

### DIFF
--- a/.github/workflows/prek_auto_update.yml
+++ b/.github/workflows/prek_auto_update.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install prek hooks
         run: pixi run install-prek
       - name: Run prek on all files
-        run: pixi run prek
+        run: pixi run prek-run
         continue-on-error: true
       - uses: peter-evans/create-pull-request@v8
         with:
--- a/.github/workflows/prek_check.yml
+++ b/.github/workflows/prek_check.yml
@@ -16,4 +16,4 @@ jobs:
       - name: Install prek hooks
         run: pixi run install-prek
       - name: Run prek on all files
-        run: pixi run prek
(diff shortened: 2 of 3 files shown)


visr:
One thing that always made me a little uneasy was how we e.g. have a dev environment with a locked versions, but in pre-commit the version is generally unlocked. Sometimes the hook revision controls the tool version, but not always, and we have to keep this synchronized to avoid possible edit wars.

I extended this PR to use as much as possible the [builtin](https://prek.j178.dev/builtin/) hooks, adding some in the process. For all repo hooks that have CLIs in conda-forge, I switched to the local repo, with `language: system`, and added them (typos, nbstripout, yamlfmt) to our Pixi environment. The `language: system` tells prek/pre-commit to assume the tool is already installed. So these local hooks are essentially the hooks as they were defined in the repo, but changing the language to system.

Since Runic is not in conda-forge we keep that as-is.

With this we also don't need the prek GitHub Action anymore, since we use the one in our Pixi environment.

PR #2886, opened 2026-02-10
Introduce Node table at Model again.

### LINKED ISSUE #1813: Refactor node table in Python

evetion:
Partially fixes #1813. Has some conflicts with #2878 (also uses `ParentModel`).

### DIFF
--- a/python/ribasim_testmodels/ribasim_testmodels/allocation.py
+++ b/python/ribasim_testmodels/ribasim_testmodels/allocation.py
@@ -466,7 +466,7 @@ def medium_primary_secondary_network_verification_model() -> Model:
     model = medium_primary_secondary_network_model()
     # set all subnetwork ids to 2 for verification purposes
     assert model.node.df is not None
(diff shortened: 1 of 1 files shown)


python/ribasim/ribasim/model.py
```diff
@@ -154,34 +151,19 @@ def _validate_ribasim_version(cls, v: str) -> str:
...
-            k,
-            v,
-        ) in self._children().items():
-            v._parent = self
-            v._parent_field = k
-        return self
-
-    @model_validator(mode="after")
-    def _ensure_link_table_is_present(self) -> "Model":
+    def _ensure_spatial_table_is_present(self) -> "Model":
```
visr: We have also `Basin / area` as a spatial table. Should we rename spatial to network here? Or topology, or node_and_link?

visr (approved):

PR #2896, opened 2026-02-11
default interpolation types from linear to constant

### LINKED ISSUE #1249: Transient forcing: forward fill versus interpolation and principle of least surprise
I was expecting forward fill on my user_demand, but I got linear interpolated values:

image

image

I guess there are reasons for this, but it's not quite Principle of Least Surprise to me.
I see that it's mentioned in the docs: https://deltares.github.io/Ribasim/core/usage.html#userdemand-time

Precipitation uses forward fill. This is desirable, because precipitation is generally provided as e.g. daily sums. If you start interpolating between sum values, you don't actually meet the sum, because the vertices of the interpolation are always the most extreme values.

Anyway, what is the thinking behind this? Why is demand interpolated rather than forward filled?

verheem:
Fixes #1249 

timeseries for which the default interpolation type changed from linear to constant:

- ConcentrationData
- LevelBoundary
- Pump
- Outlet
- PidControl

### DIFF
--- a/python/ribasim/ribasim/geometry/link.py
+++ b/python/ribasim/ribasim/geometry/link.py
@@ -159,15 +159,20 @@ class LinkTable(SpatialTableModel[LinkSchema]):
         if self.df is not None and link_id in self.df.index:
             # Remove from node table
             self.df = self.df.drop(link_id)
+            if link_id in self._used_link_ids:
+                self._used_link_ids.node_ids.remove(link_id)
+
             if self.df.empty:
                 self.df = None
 
     def _remove_node_id(self, node_id: NonNegativeInt):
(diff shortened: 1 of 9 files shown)


core/test/time_test.jl
```diff
@@ -131,15 +131,13 @@ end
 
     model = Ribasim.run(toml_path)
     storage = Ribasim.get_storages_and_levels(model).storage
+    flow = DataFrame(Ribasim.flow_data(model))
+    flow_basin = filter(:to_node_id => ==(3), flow).flow_rate
```
visr: Doesn't seem to be used?
```suggestion

```

Issue #790, opened 2023-11-15
A QGIS Ribasim plotting window


Huite:
#788 got my noggin' joggin'.

All the Ribasim input is essentially in long table form. This should make it relatively easy to support a general plotting window to help with inspecting model input. 

Essentially:

* Allow spatial selection on the nodes. See e.g. imod plugin timeseries logic.
* Use a dropdown to select from one of the properties tables. These can be filtered -- at the least -- on NullGeometry tables.
* Fetch all the associated nodes from the selected table.
* Group by the node id.
* General x-y plots can be supported, set a selector for the x-axis variable, and one for the y-axis variable.
* Some specific logic is required to support e.g. a datetime axis, but this isn't too hard to setup with pyqtgraph.
* To reduce tedium, a default x and y column could be set for each table. E.g. for tabulated rating curve, it automatically draws the Q-h relationship. For the profile, the area (x) versus level (y). For timeseries, it automatically puts the time on the x-axis.

Performance might be a bit of an issue. Pandas is available, and arrow data can be read efficiently (although it will load all data into memory). Data from the geopackage can be ingested via the PyQGIS API, which would result a list of dicts.

Although if we can get QGIS installed via pixi, we would be able to use pyarrow and pyogrio instead.

Huite:
Something that came up today: there's many cases where you probably want to look at more than one state, but which should not be plotted in the same plotting window, such as basin levels / storages and edge flows.

An easy solution is to allow the spawning of separate plotting windows. This allows us to spawn two plotting widgets, one for the levels and one for the flows.

visr:
As @deltamarnix mentioned in https://github.com/Deltares/Ribasim/issues/342#issuecomment-2560980191, we looked a bit at https://github.com/ghtmtt/DataPlotly. It looks like it can already do many of the things listed above. Some are still a bit cumbersome, but perhaps we can work around that or contribute upstream.

This is plotting a `Basin / profile` relation for the currently selected Basin:
![Image](https://github.com/user-attachments/assets/6f7c47ac-46e1-46fb-966d-f1afec88a2ba)
This works due to #1756. However if you select multiple nodes ideally you'd want to be able to group by node_id, like described in this feature request: https://github.com/ghtmtt/DataPlotly/issues/83#issuecomment-482407075. Without grouping it will think it is one line: ![Image](https://github.com/user-attachments/assets/d9121eb6-71b7-473f-a052-008364dbda29)

[...]

PR #2903, opened 2026-02-16
QGIS: Support NetCDF results, add Plotly widget

visr:
Fixes https://github.com/Deltares/Ribasim/pull/2618

This updates the QGIS temporal controller / spatial view from https://github.com/Deltares/Ribasim/pull/2208 to netCDF, and **replaces our use of the iMOD timeseries widget with a new widget with a Plotly plot**.

Plotly comes with QGIS. I considered using DataPlotly at first, but it wouldn't help us much versus using plotly directly, and as noted in https://github.com/Deltares/Ribasim/issues/790#issuecomment-3602598630 wflow-qgis follows a similar approach. This is just line plots, but if we want we can also consider supporting fancier stacked histogram water balance plots for Basins, and fraction plots for water quality for instance.

- [x] Docs (update guide, also update old screenshots)
- [x] Tests (at least in the manual test plan)
- [x] Trying it out on large models (seems reasonable, takes a few secs to plot timeseries of a hundred nodes, but you don't want to plot so many, spatial results feel similar as before)
- [x] Go over the code (used AI)

I didn't want to have more than one Ribasim button, so I made the current one a dropdown, and moved the other options that were previously in the context menu there.

Screenshot 2026-02-16 212052

### DIFF
--- a/docs/dev/qgis_test_plan.qmd
+++ b/docs/dev/qgis_test_plan.qmd
@@ -74,6 +74,39 @@ Intended behavior: The same model is loaded twice, but there is only a connectio
 - While running, right-click the same model and click "Run Ribasim model": _Warning message "Simulation already running for hws/ribasim"_.
 - Load a different model and run it: _Second simulation starts normally_.
 
+# Plot widget tests
+
+## Open the plot widget
+
+- Run `models/hws` or another model that produces results.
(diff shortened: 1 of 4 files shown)


visr:
Yep, thanks for the lift! QGIS 3.40 and 3.44 are bundled with Plotly 5. Plotly 6 should give some free performance boosts as well for sending e.g. numpy arrays to JS. I didn't use it here since I'm worried about it not working everywhere, but Plotly also has WebGL support that you might want to try out.

https://plotly.com/python/performance/

I didn't try very large plots yet, but 188cdedf1eb03e5050d6a54fbaf02d09a42dde6e made it snappier to load, because including plotly.js in every HTML was a bit slow.

PR #2912, opened 2026-02-19
A Boolean for allocation controlled

### LINKED ISSUE #2889: A Boolean for control by allocation
Implement a Boolean for control by allocation as opposed to the magic "Ribasim.allocation", as suggested in https://github.com/Deltares/Ribasim/issues/2527#issuecomment-3497426147.

When we do this, we should directly validate that it is only used for Outlet or Pump, and not other unsupported node types, otherwise users may expect it to do something. Came up in https://github.com/Deltares/Ribasim/issues/2883.

verheem:
fixes #2889

example of an outlet that switches between distinct allocation states

Screenshot 2026-02-26 at 09 50 06

### DIFF
--- a/core/src/allocation_optim.jl
+++ b/core/src/allocation_optim.jl
@@ -1108,6 +1108,7 @@ function update_flow_variable_bounds!(
     flow = problem[:flow]
     for flow_link in flow_links_subnetwork
         flow_var = flow[flow_link]
+        JuMP.is_fixed(flow_var) && continue
         JuMP.set_lower_bound(
             flow_var,
             flow_capacity_lower_bound(flow_link, p_independent) / scaling.flow,


python/ribasim/ribasim/migrations.py
```diff
@@ -95,12 +95,36 @@ def pidcontrolstaticschema_migration(df: DataFrame, schema_version: int) -> Data
...
+
+
 def outletstaticschema_migration(df: DataFrame, schema_version: int) -> DataFrame:
     if schema_version < 2:
         _rename_column(df, "min_crest_level", "min_upstream_level")
     if schema_version < 9:
         check_inactive(df, name="Outlet / static")
         df.drop(columns="active", inplace=True, errors="ignore")
+    if schema_version < 11:
+        df = _migrate_allocation_controlled(df)
```
visr:
```suggestion
        _migrate_allocation_controlled(df)
```

for consistency

python/ribasim/ribasim/migrations.py
```diff
@@ -95,12 +95,36 @@ def pidcontrolstaticschema_migration(df: DataFrame, schema_version: int) -> Data
...
+        # clear the control_state; for rows that had it alongside other states, rename it
+        has_other_states = df.groupby("node_id")["control_state"].transform(
+            lambda x: (
+                (x != "Ribasim.allocation").any() & (x == "Ribasim.allocation").any()
+            )
+        )
+        # Rename "Ribasim.allocation" to "allocation" where there are multiple control states
+        df.loc[mask & has_other_states, "control_state"] = "allocation"
+        # Clear control_state where it was the only state
+        df.loc[mask & ~has_other_states, "control_state"] = None
```
visr: Seems like a good solution. Did you run it manually with `pixi run migrate-model` on a model with allocation control? Because our migration test model doesn't hit most of these lines.

python/ribasim/ribasim/schemas.py
```diff
@@ -301,6 +302,7 @@ class PumpStaticSchema(_BaseSchema):
     min_upstream_level: float = pa.Field(nullable=True)
     max_downstream_level: float = pa.Field(nullable=True)
     control_state: pd.StringDtype = pa.Field(nullable=True)
+    allocation_controlled: pd.BooleanDtype = pa.Field(nullable=True)
```
visr: Reminder to add this field to TRC once https://github.com/Deltares/Ribasim/pull/2907 is merged.

python/ribasim/ribasim/migrations.py
verheem: worked

visr (approved):

Issue #2518, opened 2025-08-05
Support compression for NetCDF results


visr:
From https://github.com/Deltares/Ribasim/pull/2497#issuecomment-3143977156:

> For NetCDF I did not yet implement compression, compression settings are ignored, though this is documented.

The current compression settings look like this:

```toml
[results]
format = "arrow"      # optional, default "arrow", can also be "netcdf"
compression = true    # optional, default true, using zstd compression
compression_level = 6 # optional, default 6
```

The NCDatasets options are `deflatelevel::Int` and `shuffle::Bool`: https://juliageo.org/NCDatasets.jl/stable/variables/#CommonDataModel.defVar

Wflow exposes `deflatelevel` as `compressionlevel` to users, and defaults to no compression but recommends no higher than 1. This is from their docs:

To limit the size of the resulting netCDF file, file compression can be enabled. Compression
increases computational time but can significantly reduce the size of the netCDF file. Set
the `compressionlevel` variable to a value between `0` and `9`. A setting of `0` means no
compression, while values between 1 and 9 indicate increasing levels of compression (1:
least compression, minimal run-time impact, 9: highest compression, maximum run-time
impact). If file size is a concern, we recommend using a value of `1`, as higher compression
levels generally have a limited effect on file size.

```toml
[output.netcdf_grid]
path = "output_moselle.nc"          # Location of the output file
compressionlevel = 1                # Compression level (default 0)
```

I believe HDF5 and so NetCDF support many more compression options than just `deflate`, but I guess this just hasn't been set up in NCDatasets the way they have in HDF5.jl: https://juliaio.github.io/HDF5.jl/stable/#Chunking-and-compression. So if we want to default to e.g. Blosc we'd need to add support to NCDatasets and verify that it is sufficiently well-supported in e.g. Delft-FEWS, QGIS for it to be the default.

We could start with making the `compression_level` default 1 for NetCDF and keep it at 6 for Arrow, and then also consider to always apply shuffle.

Issue #2933, opened 2026-02-25
Align concentration output tables


evetion:
Currently the experimental functionality writes to `concentration.nc`, whereas the Delwaq parse writes to the `concentration_external` table (in memory). The latter cannot be visualised. I've been advising some code snippets to set the external filename of that table to `concentration.nc`, so on writing it behaves similar to the built-in functionality. We should come up with a design before v1 and document it (better).

EDIT: change was made in https://github.com/Deltares/Ribasim/pull/2937#discussion_r2859788706, but needs to be tested if the sturcture is equal to the concentration.nc written by Ribasim.

PR #2937, opened 2026-02-26
remove arrow support

### LINKED ISSUE #2520: Drop support for Arrow results
https://github.com/Deltares/Ribasim/pull/2497 added support for NetCDF. Having both tabular and N dimensional array formats is not great for the simplicity of this codebase.

See https://github.com/Deltares/Ribasim/pull/2497#issuecomment-3143977156 for the reasoning.

https://github.com/Deltares/Ribasim/issues/2519 would need to be done first before doing this, and possibly https://github.com/Deltares/Ribasim/issues/2518.

If we do this we should also make sure `allocation_flow` results are added.

verheem:
Fixes #2520
Fixes https://github.com/Deltares/Ribasim/issues/2518


core/Project.toml
visr: Why did you have to add this back? Note that they are also mentioned in the Project.toml in the repo root. And after removing you need to `Pkg.resolve()`.

python/ribasim/ribasim/delwaq/parse.py
```diff
@@ -56,6 +56,7 @@ def parse(
     df.sort_values(["time", "node_id"], inplace=True)
 
     model.basin.concentration_external = df
-    df.to_feather(model.results_path / "basin_concentration_external.arrow")
+    ds = df.set_index(["time", "node_id", "substance"]).to_xarray()
+    ds.to_netcdf(model.results_path / "basin_concentration_external.nc")
```
visr: @evetion would this order need to be time, substance, node_id, the inverse of the Julia order for the results? https://github.com/Deltares/Ribasim/commit/ca97265dca3f512517ccc1ea63aa261f1b2ed061

And the filename `concentration.nc` as discussed in https://github.com/Deltares/Ribasim/issues/2933?

Issue #2915, opened 2026-02-20
Can we include the listener links in the network definition?


gijsber:
Currently, the listener links are a bit hidden in the datamodel, possibly beause you need to state for each listening node what kinfd of variable you are listening to.

Maybe we can still accommodate this in the network model by introducing a level_listen_link and a flow_listen_link as two different link types.

It would be good if this is addressed in the Links Tables such that other tools can also use this information
related to #341

Issue #341, opened 2023-06-20
Show implicit Control edges in QGIS


evetion:
There's no straightforward solution, we either:

- Add a geometry field to the Control / logic table, but this violates the special cases of Node/Edge as the only spatial tables.
- Add these edges to the Edge table, but since we support multiple listen_node_ids, and require the listen_variable we duplicate this connection data.
- Visualize the control network with a special action button, making a (hidden?) layer that is separate from the input file.

Thinking about this, I have a slight preference for the first.

SouthEndMusic:
Can someone give a priority to this issue?

PR #2949, opened 2026-02-27
Add implicit listen edges to Link table on write

visr:
Fixes https://github.com/Deltares/Ribasim/issues/2915
Fixes https://github.com/Deltares/Ribasim/issues/341

Screenshot from QGIS `discrete_control_of_pid_control`, which now matches the `model.plot()` style:

Screenshot 2026-02-27 171223


evetion:
In HydroMT we wont touch control yet, so it should be fine I guess.

Issue #2688, opened 2025-10-28
No QGIS Basin results with one Basin


visr:
If you run and load `two_basin` into QGIS, you can see the Results / Basin layer with data.
If you do the same for `bucket`, which has only one Basin, nothing is shown on the map.
Looking at the attribute table of the scratch layer the node_id and node_type are filled in, but all the numeric results are NULL. If you open the results/basin.arrow you do see output (all 0.0 for this model).

Here is a screenshot from @Fati-Mon from Crystal model:
<img width="811" height="412" alt="Image" src="https://github.com/user-attachments/assets/e10ec156-f132-4a9c-a4e7-dc34e1d7cf11" />

visr:
Also basic fails on load now, possibly related to https://github.com/Deltares/Ribasim/issues/2933.

```
IndexError: index 4 is out of bounds for axis 2 with size 4 
Traceback (most recent call last):
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\ribasim_qgis.py", line 39, in handleFileDrop
    self.parent.open_model(path)
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\ribasim_qgis.py", line 206, in open_model
    self.ribasim_widget.open_model(path)
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\ribasim_widget.py", line 87, in open_model
    self.__dataset_widget.open_model(path)
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 256, in open_model
    self._open_model(path)
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 264, in _open_model
    self.refresh_results()
  File "C:\Users/visser_mn/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\ribasim_qgis\widgets\dataset_widget.py", line 544, in refresh_results
    self._load_netcdf_results()
[...]

PR #2959, opened 2026-03-03
catch case with zero links

verheem:
Fixes https://github.com/Deltares/Ribasim/issues/2688

### DIFF
--- a/ribasim_qgis/core/netcdf.py
+++ b/ribasim_qgis/core/netcdf.py
@@ -5,7 +5,6 @@ Return lightweight ``NetCDFResult`` dataclasses that keep data in their
 natural 2-D NumPy shape (n_times x n_ids).
 """
 
-import logging
 from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


ribasim_qgis/core/netcdf.py
```diff
@@ -102,7 +102,19 @@ def _read_2d_nc(path: Path, id_name: str) -> NetCDFResult | None:
...
+
+    # Skip empty dimensions (e.g. flow.nc without links) to avoid GDAL Read errors.
+    dim_sizes = [d.GetSize() for d in id_array.GetDimensions()]
+    if any(size == 0 for size in dim_sizes):
+        return None
+
+    try:
+        ids = id_array.ReadAsArray()
+    except RuntimeError:
+        return None
```
visr: Do you know when we would run into this? If we don't perhaps better to remove. Or does this help with PermissionDenied errors?

PR #2975, opened 2026-03-10
QGIS fix water balance plot storage

visr:
I noticed the rc2 water balance plot has issues with plotting the `storage_rate`. This variable, together with `balance_error`, are the only two terms that can appear on either side of the balance. This was not handled well causing the storage rate to sometimes appear on the wrong side leading to an imbalanced figure. So this PR:

- splits the storage_rate into two terms: storage_increase and storage_decrease
- plots the balance_error as a black residual line instead of another balance component
- adjusts the tick and hover formatting to prefer scientific notation over the nano/micro style formatting
- no longer shows balance components that are 0 for the whole duration

**New look:**

image

**Old look:**

Screenshot 2026-03-05 140423

### DIFF
--- a/ribasim_qgis/widgets/plot_widget.py
+++ b/ribasim_qgis/widgets/plot_widget.py
@@ -676,7 +676,7 @@ class PlotWidget(QWidget):
         for row, unit in enumerate(units, start=1):
             for trace in traces_by_unit[unit]:
                 fig.add_trace(trace, row=row, col=1)
-            self._update_yaxis_format(fig, row, unit)
+            self._update_yaxis_format(fig, row=row, title_text=unit)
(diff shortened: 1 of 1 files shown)


ribasim_qgis/widgets/plot_widget.py
```diff
@@ -745,18 +826,29 @@ def _redraw_basin_water_balance(
             if file_units.get(term, "")
         }
         yaxis_title = unit_values.pop() if len(unit_values) == 1 else "(no unit)"
-        fig.update_yaxes(title_text=yaxis_title, row=1, col=1)
+        self._update_yaxis_format(fig, 1, yaxis_title)
```
verheem: magic number 1, please explain

PR #2978, opened 2026-03-10
remove dangling listen links

### LINKED ISSUE #2976: Auto remove listen links
Does removing a control link also remove all listen links that were automatically added?

_Originally posted by @evetion in https://github.com/Deltares/Ribasim/issues/2969#issuecomment-4023951104_

verheem:
Since we implicitly add the required listen links to control links, we should also implicitly remove them if a control node has no control link anymore.  fixes #2976.


verheem:
> This tests the effect of `remove_node`, but using `remove_link` on the control link would be the more precise test.

Right, that functionality did not work yet, but now it does! :)

PR #2995, opened 2026-03-24
Make `model.<node_type>.node.df` read-only

visr:
I don't intend for this to be merged, but want to use it to find remaining errors in Ribasim-NL.

Blocks: .df = ..., df["col"] = ..., del df["col"], df.loc[...] = ..., df.iloc[...] = ..., df.at[...] = ..., df.iat[...] = ..., df.columns = ...
Allows: all read operations (column access, indexing, .loc/.iloc reads, len(), repr(), iteration, etc.)
Error message: directs the user to modify model.node.df instead

### DIFF
--- a/python/ribasim/ribasim/geometry/node.py
+++ b/python/ribasim/ribasim/geometry/node.py
@@ -44,6 +44,90 @@ from .base import _GeoBaseSchema
 
 __all__ = ("NodeTable",)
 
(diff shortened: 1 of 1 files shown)


visr:
Closing the draft PR since it wasn't meant to be merged. I'll keep the branch a bit longer in case people in Ribasim-NL want to still use it.

Issue #2509, opened 2025-08-01
Adaptive timestepping for linearized allocation


simulutions:
Lets say we linearize the storage level relationship $S(h)$ around the beginning of the time step $S^n$. 
This means we approximate $S(h)$ with a tangent line, which we call $S^{\*}(h)$. We can then choose a maximum allowed change in storage such that the linearisation holds within an error $|S(h) - S^*(h)| \leq \epsilon$. 

Then we can put bounds on $S$ for the next allocation and we get adaptive time stepping for free if:

$$
\begin{aligned}
\text{maximize} & \Delta t \\
\text{subject to} & S^n - \epsilon \leq S^{\*} \leq S^n + \epsilon \\
 & 
\end{aligned}
$$

This way, at times of large flows (a lot of rain for example) allocation takes small time steps, while at times of drought, not a lot of interesting things happens and we take larger time steps, which potentially gives significant performance boosts over simulation periods of one year or more.

If we notice small basins connected to large flow edges enforce too small time steps on allocation, we get different time stepping per subnetwork for free, since they are separate optimization problems anyway, and only need to add synchronization on right timestep between the main network and the subnetwork

verheem:
# A Priori Adaptive Allocation Timestep — Design and Derivation

## Motivation

The allocation LP in Ribasim is a **linearization** of the underlying nonlinear water-balance problem. Two relationships get linearized around the current state $(h^n, S^n)$:

1. **Basin storage profile** — $S(h)$ linearized to $S^*(h) = S^n + A^n (h - h^n)$
2. **Connector node flow laws** — $Q(h_a, h_b)$ linearized to $Q^* = Q^n + \partial_a Q\,\Delta h_a + \partial_b Q\,\Delta h_b$

Both linearizations are only trustworthy for small $\Delta h$. A fixed allocation timestep forces the user to pick conservatively (short Δt, wasted compute during quiet periods) or aggressively (long Δt, linearization errors blow up during dynamic events).

The adaptive timestep solves this by deriving the largest Δt for which **all** linearization errors in the network remain below a user-configured tolerance. Δt becomes large during drought and small during floods, automatically.

## Second-Order Error Derivation

Both error bounds fall out of the same Taylor argument.

### Basin profile

The basin profile gives area as a piecewise-linear function of level, $A(h)$. Storage is its integral:
$$S(h) = \int_0^h A(h')\,dh'$$
[...]

PR #3026, opened 2026-04-09
Adaptive timestepping for allocation

verheem:
## Summary
Adds adaptive timestepping for the allocation LP, where Δt is bounded by the linearization errors of **all** linearized relationships in the network — both basin profiles and connector node Q(h) relationships (Deltares/Ribasim#2509).

## Approach
The allocation LP linearizes:
1. **Basin profiles** S(h) — error = ½|dA/dh|·Δh²
2. **Connector node** Q(h) (TabulatedRatingCurve, ManningResistance, LinearResistance) — error = ½|d²Q/dh²|·Δh²

For each, we derive a maximum allowed level change Δh_max from a relative error tolerance ε:

    Δh_max^basin     = sqrt(2 · ε · S_max / |dA/dh|)
    Δh_max^connector = sqrt(2 · ε / |d²Q/dh²|)

The global Δh_max is the minimum across all basins and connector nodes. It is converted to a per-basin time bound:

    Δt_i = A_i · Δh_max / |dS_i/dt|

The connector node second derivatives are computed numerically via nested ForwardDiff, which handles all node types uniformly (including the PCHIP-interpolated rating curves and the nonlinear Manning equation).

Based on the explicit storage change prediction, we increase the bounds of the basins if they are likely to overflow, to allow overflowing basins in allocation

## New config options
- `adaptive_timestep::Bool = true` — enable/disable
- `timestep_tolerance::Float64 = 0.01` — relative tolerance for linearization errors
- `min_timestep::Float64 = 3600.0` — floor to prevent runaway small steps
- `timestep` (existing) — acts as the maximum timestep when adaptive is enabled

## Key changes
- `allocation_init.jl`: Factor Δt out of volume conservation constraint init
- `allocation_optim.jl`: Update flow coefficients with current Δt in `set_simulation_data!` and increase upper bound of basin storage based on predicted in `set_simulation_data!`
- `allocation_util.jl`: New `get_area_slope`, `get_max_flow_curvature`, and `compute_adaptive_Δt` functions
- `model.jl`: New `solve_with_allocation!` with adaptive while-loop
- `parameter.jl`: `AllocationModel` is now mutable (for updating `Δt_allocation`)
- `docs/concept/allocation.qmd`: New "Adaptive timestepping" section with derivations

## Test plan
- [x] Verify existing allocation tests pass (reference LP files regenerated)
- [x] New `adaptive_allocation` test model runs successfully
- [x] Compare allocation results between fixed and adaptive modes
- [ ] Benchmark on longer simulations with varying flow regimes


docs/reference/usage.qmd
```diff
@@ -1,7 +1,5 @@
 ---
 title: "Usage"
-filters:
-  - include-code-files
```
visr, line 4: Docs build just fine, let's check if the toml that is included like this doesn't disappear after render.

```
{.toml include="../../core/test/docs.toml"}
```

The `docs.toml` needs to be updated though, the allocation section.

python/ribasim/ribasim/config.py
```diff
@@ -92,10 +92,20 @@ class Allocation(ChildModel):
...
     timestep : float
-        The simulated time in seconds between successive allocation calls (Optional, defaults to 86400)
+        The simulated time in seconds between successive allocation calls (Optional, defaults to 86400).
+        When adaptive_timestep is enabled, this acts as the maximum timestep.
+    adaptive_timestep : bool
+        Enable adaptive allocation timestepping based on linearization error bounds (Optional, defaults to True)
+    timestep_tolerance : float
+        Relative tolerance for linearization errors of basin profiles and connector node Q(h) relationships (Optional, defaults to 0.01)
+    min_timestep : float
+        Minimum allocation timestep in seconds for adaptive timestepping (Optional, defaults to 3600.0)
```
visr: Shall we align these with the solver settings?
timestep -> dt
If not provided, it is adaptive.
min_timestep -> dtmin

Not sure about timestep_tolerance, there is `reltol` in solver, but perhaps reltol_linearization would be best here?

https://ribasim.org/reference/usage.html

docs/concept/allocation.qmd
```diff
@@ -96,6 +96,72 @@ $$Q(x_1,x_2) \approx Q(x_1^n,x_2^n) + \frac{\partial Q}{\partial x_1}(x_1^n,x_2^
...
+
+gives a maximum level change
+
+$$\Delta h_{\max}^{\text{basin}} = \sqrt{\frac{2\,\varepsilon\, S_{\max}}{|m|}}$$
+
+When $|m| \approx 0$ (basin with constant area), the linearization is exact and imposes no constraint.
+
+### Connector node linearization error
+
+Connector nodes ([LinearResistance](../reference/node/linear-resistance.qmd), [ManningResistance](../reference/node/manning-resistance.qmd) and [TabulatedRatingCurve](../reference/node/tabulated-rating-curve.qmd)) have nonlinear flow-level relationships $Q(h)$ that are also linearized:
```
visr:
```suggestion
Connector nodes ([LinearResistance](/reference/node/linear-resistance.qmd), [ManningResistance](/reference/node/manning-resistance.qmd) and [TabulatedRatingCurve](/reference/node/tabulated-rating-curve.qmd)) have nonlinear flow-level relationships $Q(h)$ that are also linearized:
```

I prefer to keep all references absolute, easier to update when moving things around.

docs/concept/allocation.qmd
```diff
@@ -96,6 +96,72 @@ $$Q(x_1,x_2) \approx Q(x_1^n,x_2^n) + \frac{\partial Q}{\partial x_1}(x_1^n,x_2^
...
+
+Converting the level change bound to a time bound using $\Delta S \approx A^n \,\Delta h$ and the current net storage rate $\dot{S}$:
+
+$$\Delta t_i = \frac{A_i^n \cdot \Delta h_{\max}}{|\dot{S}_i|}$$
+
+The adaptive timestep for the subnetwork is:
+
+$$\Delta t_{\text{adaptive}} = \min_i \Delta t_i$$
+
+clamped between a configured minimum and maximum.
```
visr: The maximum is not configured right?

core/src/allocation_util.jl
```diff
@@ -305,6 +307,177 @@ end
 get_Δt_allocation(allocation::Allocation) =
     first(allocation.allocation_models).Δt_allocation
 
+"""
+Compute the slope dA/dh of the basin profile at a given level.
+The basin profile is piecewise-linear in A(h), so dA/dh is piecewise-constant.
+Returns the max absolute slope of the current and adjacent segments for safety.
+"""
+function get_area_slope(basin::Basin, state_idx::Int, level::Float64)::Float64
```
SouthEndMusic, line 314: I feel like you should not have to work this out explicitly: what you want is the second derivative of $S(h)$, and I think second derivative computation is supported by `DataInterpolations.jl` without the need for AD.

core/src/allocation_util.jl
```diff
@@ -305,6 +307,177 @@ end
...
+    max_curvature = 0.0
+
+    for node_id in connector_ids
+        inflow_id = connector_node.inflow_link[node_id.idx].link[1]
+        outflow_id = connector_node.outflow_link[node_id.idx].link[2]
+
+        h_a = get_level(p, inflow_id, t)
+        h_b = get_level(p, outflow_id, t)
+
+        # d²Q/dh_a² via nested ForwardDiff
```
SouthEndMusic: Does `DifferentiationInterface` supply a cleaner API for second order AD?

core/src/util.jl
visr, line 1: Could you add unit tests for these new utility functions?

core/src/model.jl
```diff
@@ -298,27 +298,78 @@ function step!(model::Model, dt::Float64)::Model
...
+
+    Δt = min(Δt, time_to_next_saveat(t, saveat, tspan_end))
+    Δt = min(Δt, tspan_end - t)
+
+    on_saveat =
+        iszero(saveat) ||
+        isinf(saveat) ||
+        isapprox(t % saveat, 0.0; atol = 1.0e-9) ||
+        isapprox(t % saveat, saveat; atol = 1.0e-9) ||
+        isapprox(t, tspan_end; atol = 1.0e-9)
```
visr: Is this not what `is_saveat_time` is for?

visr (approved):
Great to have this, left a few comments.

core/src/model.jl
verheem: Yes, was not aware of this function

core/test/docs.toml
```diff
@@ -25,7 +25,9 @@ flow_boundary = "block"   # optional, default "block", can otherwise be "linear"
 block_transition_period = 0  # optional, default 0
 
 [allocation]
-timestep = 86400         # optional (required if experimental.allocation = true), default 86400
+# dt = 86400               # optional, fixed allocation timestep in seconds; if omitted, adaptive timestepping is used
+dtmin = 3600.0             # optional, minimum allocation timestep in seconds (default 3600.0)
+reltol_linearization = 0.01 # optional, relative linearization error tolerance (default 0.01)
```
visr:
```suggestion
dt = 86400                  # optional, remove for adaptive time stepping
dtmin = 3600.0              # optional, 3600.0
reltol_linearization = 0.01 # optional, default 0.01
```

This keeps it more in line with the rest. Explanations are below on the same page.

visr (approved):
Looks good!

PR #3032, opened 2026-04-14
[Delwaq] Change system timer, and verbosity control.

evetion:
Fixes https://github.com/Deltares/Ribasim-NL/issues/534

Also adds verbosity control based on the verbosity of the Ribasim model. Currently it is set to debug, which becomes very slow for large models, as it prints all individual boundaries + timeseries.

### DIFF
--- a/python/ribasim/ribasim/delwaq/generate.py
+++ b/python/ribasim/ribasim/delwaq/generate.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
 from ribasim import Model, nodes
+from ribasim.config import Verbosity
 from ribasim.utils import (
     MissingOptionalModule,
     _concat,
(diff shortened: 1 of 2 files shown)


visr (approved):
Nice. I assume this system timer won't change anything about flows being interpreted in m3/s, it is only for the processes?

evetion:
> Nice. I assume this system timer won't change anything about flows being interpreted in m3/s, it is only for the processes?

Correct, you're looking for line 10 (1s). From the documentation (verbatim):

> D-Water Quality uses for its input processing 2 different timers. The first timer is de model timer and is specified as part of the 4th title string in the first group of the input file. It is usually set to 1 second. The second timer is the auxiliary timer that is used for water quality processes. This timer is generally expressed in days (a decay coefficient in 1/day). D-Water Quality supports freedom of choice in this respect, but it is not excluded that some processes routines lean on the fact that the processes timer is generally expressed in days. Because the unit of time is free, only the factor between the two timers need to be provided. This is generally 86400, the number of seconds in a day. Should you chose however a system time of days (and thus flows in m3/day) and processes expressed per year, then you could take 365 as factor.

PR #3035, opened 2026-04-15
Support UserDemand with multiple inflow links

### LINKED ISSUE #2725: Conjunctive use: Allow (and control) water abstraction from multiple sources
StLucia (Uruguay, Min.Water, Patricia ) and Hermana (Columbia, SITO-IS, Betsy/Martha) use the groundwater node of Ribasim. Both projects consider moving to the new Ribasim.

When studying conjunctive use, one UserDemand would be connected to multiple sources (e.g. a river, a Managed Aquifer Recharge system or a pristine aquifer)

Dependent on water availability or time of year, the actual abstraction might be distributed over multiple sources at the same time (like a ratio per source).

Currently abstracting water from multiple sources is not facilitated. Particularly, the allocation module can update the allocated flow rate on the UserDemand node, but it cannot prescribe the link to abstract from.

Associated with #2705 and #2724

### LINKED ISSUE #2776: User demand abstraction from multiple reservoirs (no allocation preference)
**What**
I wanted to create a schematic reservoir network with multiple reservoirs collectively supplying water demand in the basin. The demand is then proportionally allocated to reservoirs based on their storage levels (in m³).

**Why**
In the basin I have no preference to how water is allocated from which reservoirs, as long as they try to satisfy the demand. This circumvents the issue of having to calibrate each reservoir in a schematic study.

**How**
I was not able to create such a system in ribasim so I create my own python script. I has inflows into each reservoirs, a reservoir maximum storage, and a downstream demand at each timestep. Based on all reservoir storages over time I can allocate which reservoir should supply what amount.

**Additional context**
I am using this in a scientific study across the whole of Spain, a country with many reservoirs. So I needed a really basic model.

verheem:
Fixes https://github.com/Deltares/Ribasim/issues/2725
Fixes https://github.com/Deltares/Ribasim/issues/2776

## Summary

This PR extends `UserDemand` to support **multiple inflow links**, allowing a single demand node to draw water from more than one source basin. Previously, `UserDemand` was restricted to exactly one upstream link (`1:1`), making it impossible to model demands supplied by multiple sources.

### Data model (`core/src/parameter.jl`)
- Replaced the single `inflow_link::Vector{LinkMetadata}` with `inflow_links::Vector{Vector{LinkMetadata}}` — each `UserDemand` node now stores a vector of inflow link metadata.
- Added `inflow_link_offsets` to map each node's inflow links into the flat `user_demand_inflow` state vector (one state per link, not per node).
- Added `inflow_link_allocated` to store per-link allocation results from the LP, so the physics layer can split inflow across source basins exactly as allocation decided.
- Added `link_to_state_idx::Dict` to `ParametersIndependent` for direct link → state index lookups, needed when a destination node has multiple inflow-link states.

### Allocation LP (`core/src/allocation_optim.jl`)
- The `user_demand_allocated_sum_constraint` now sums flow over **all** inflow links instead of referencing a single one.
- Return-flow constraints use the total inflow across all links.
- `set_simulation_data!` iterates over all inflow links to set return-factor coefficients.
- `parse_allocations!` stores per-link allocated flow rates into `inflow_link_allocated` and sums cumulative supplied volumes across all inflow links.

### Physics / ODE formulation (`core/src/solve.jl`)
- `formulate_flow!`: computes per-link demand (equal split or allocation-driven), applies per-source low-storage and min-level reduction factors independently, and writes one `du` entry per inflow link.
- `limit_flow!`: bounds are now computed and applied per inflow link with source-specific storage/level factors.

### State vector & indexing (`core/src/util.jl`)
- `state_node_ids`: the `user_demand_inflow` component now emits one node ID per inflow link (not per node), expanding the state vector for multi-inflow nodes.
- `get_state_flow_links`: builds one `(inflow, placeholder)` pair per link instead of per node.
- New `build_link_to_state_idx` function creates the link → state index dictionary.
- `get_state_index(state_ranges, link_to_state_idx, link)` overload resolves per-link states via the dictionary before falling back to the old node-based lookup.
- `reduce_state!` uses link-based lookups for correct basin water balance accounting.

### Jacobian (`core/src/sparsity.jl`)
- `update_J_inner!` for `user_demand_inflow` states now uses `state_inflow_link` to identify the correct source basin directly, rather than dispatching through the node type.

### Concentration / water quality (`core/src/concentration.jl`)
- `mass_inflows_from_user_demand!` mixes return-flow concentrations as a mass-weighted average across all inflow links, proportional to each link's cumulative flow contribution.

### Validation (`core/src/validation.jl`)
- Relaxed `UserDemand` neighbor bounds from `(1,1,1,1)` to `(1,M,1,1)` — multiple inflow links are now valid.

### Python side (`ribasim_python/`)
- Updated `can_connect` to allow up to `M` inflow links for `UserDemand`.

### Test model & tests
- New `two_basin_user_demand_model` test model: two basins (with different `route_priority`) supplying a single `UserDemand`, with decaying `FlowBoundary` forcing the LP to shift demand between sources.
- New integration test verifying that Basin 4's contribution ramps linearly as Basin 3's supply decays, keeping total UserDemand inflow at the requested demand.
- Added corresponding LP reference file for the allocation problem.


core/src/allocation_optim.jl
```diff
@@ -883,8 +884,31 @@ function parse_allocations!(
...
+    flow = allocation_model.problem[:flow]
+
+
+    supplied_volume = node_id ->
+    if is_user_demand
+        # UserDemand has multiple inflow links, sum them.
+        sum(cumulative_supplied_volume[lm.link] for lm in node.inflow_links[node_id.idx])
+    else
+        cumulative_supplied_volume[node.inflow_link[node_id.idx].link]
+    end
```
visr: Can this be a named top-level function? It is a bit hard to read with `->` and the newline without indentation.

core/src/differentiation.jl
```diff
@@ -191,7 +191,11 @@ function update_J_inner!(
     elseif row in state_ranges.outlet
         update_J_inner!(J_inner, val, node_id, col_reduced, outlet)
     elseif row in state_ranges.user_demand_inflow
-        update_J_inner!(J_inner, val, node_id, col_reduced, user_demand; do_outflow = false)
```
visr, line 194: I guess we can remove `do_outflow` from `update_J_inner!`?

core/test/allocation_physics_test.jl
```diff
@@ -300,3 +300,32 @@ end
...
+    using DataFrames: DataFrame
+    using Dates: DateTime
+
+    toml_path = normpath(
+        @__DIR__,
+        "../../generated_testmodels/two_basin_user_demand/ribasim.toml",
+    )
+    @test ispath(toml_path)
+
+    model = Ribasim.run(toml_path)
```
visr, line 315: Can you also add some unit tests for new fields added in parameter.jl, like `inflow_link_offsets` and `link_to_state_idx`?

core/src/solve.jl
```diff
@@ -357,41 +357,49 @@ function formulate_flow!(
...
-            q += alloc
+            q_total_demand += min(
+                allocated[demand_priority_idx],
+                get_demand(user_demand, id, demand_priority_idx, t),
+            )
+        end
+
+        # When allocation disabled,
+        # fall back to an equal split of the total demand. Each link then applies
+        # its own source basin reduction factors
```
visr:
```suggestion
        # With allocation disabled, fall back to an equal split of the total demand.
        # Each link then applies its own source basin reduction factors.
```

visr (approved):

Issue #2657, opened 2025-10-20
Look into chunksize latency and performance


visr:
There's a finetuning option as well, most of the specialize time comes from introducing a chunk-like number (forgot the exact name), that is determined automatically based on the size of the problem. We could hardcode it for all/specific models? Would be good to print that number for different problem sizes.

Not sure if this is all that specialize influences, but this is the chunksize argument: https://github.com/Deltares/Ribasim/pull/2114/files#diff-50a3f887d6d4b8517e8faf1ceaddb02c23149fe7bf4b9199c8f56d52bed6c6f2R303

_Originally posted by @evetion in https://github.com/Deltares/Ribasim/issues/2652#issuecomment-3421185052_
            
See also https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/#Configuring-Chunk-Size

PR #3039, opened 2026-04-16
Farewell, cumulative flow states

### LINKED ISSUE #3037: Core refactor: basin storages as ODE states instead of cumulative flows
# Core refactor: basin storages as ODE states instead of cumulative flows

## Context

While implementing support for UserDemand with multiple inflow links ([#3035](https://github.com/Deltares/Ribasim/pull/3035)), the complexity cost of the cumulative-flow state formulation became very apparent. Adding per-link states required changes across `reduce_state!`, `get_state_flow_links`, `update_J_inner!`, the state vector layout, and a new `link_to_state_idx` dictionary — none of which would have been needed if the ODE states were basin storages. This prompted a broader evaluation of the current formulation.

## What

Reformulate the ODE problem so that the states are **basin storages** (and PID integral terms) instead of cumulative flows per link. This evolves the decision made in [#1808](https://github.com/Deltares/Ribasim/issues/1808), which moved from storage states to cumulative-flow states.

The motivations behind #1808 were valid — accurate flow output and a clean water balance — but the same goals can be achieved with storage states and a trapezoidal integration callback for flow output, while gaining significant performance, simplicity, and numerical robustness.

A trapezoidal callback approach was proposed in [#1473](https://github.com/Deltares/Ribasim/issues/1473) and prototyped in [#1468](https://github.com/Deltares/Ribasim/pull/1468) but was closed as "not planned" when the team chose to keep cumulative flows as states instead. The technique itself — trapezoidal integration at accepted integrator steps — is straightforward and avoids the Gaussian quadrature problem that ruled out SciML's `IntegratingSumCallback` (which requires evaluating flows at intermediate points where the integrator doesn't stop).

### 1. Numerical robustness — the growing-state problem

Cumulative flows grow without bound over time. This causes two compounding issues that are documented in [#1897](https://github.com/Deltares/Ribasim/issues/1897):

- **Floating-point precision loss**: as $u_i(t) = \int_0^t q_i \, dt'$ grows large, new contributions $q_i \cdot \Delta t$ lose significant digits.
- **Effective tolerance degradation**: the ODE solver's error estimate uses $\text{reltol} \times |u|$. As $|u|$ grows, the solver accepts larger absolute errors.

This was observed in practice: water balance errors growing over time in test models, suspected instabilities in the GRM model after 9 months, and LHM hangs after 8 months of simulation. Various workarounds were explored — resetting states ([#2290](https://github.com/Deltares/Ribasim/issues/2290), never implemented), setting `reltol = 0` (doubles compute time), custom error norms, sign-flipping — but none fully address the fundamental issue.

Unless a basin's storage grows unbounded over time (which is a silly simulation and probaly not in scope of Ribasim) the relative tolerance stays effective for the entire simulation, even over decades.

### 2. Performance
The dominant performance bottleneck — Jacobian evaluation — was already solved by the $f(u) = g(Au)$ decomposition in [#2624](https://github.com/Deltares/Ribasim/pull/2624), which reduced Jacobian cost from >80% to <10% of LHM runtime by computing AD only on the basin-sized $J_g$. This was an excellent optimization. However, it adds ~500 lines of custom linear algebra (`HalfLazyJacobian`, `RibasimLinearSolve`, `calc_J_inner!`, `update_J_inner!`, `dolinsolve` overload) to work around a formulation that doesn't natively operate in basin space. With storage states, the solver operates in basin space directly and standard SciML Jacobian handling works out of the box — achieving the same Jacobian performance without the custom machinery.

The remaining per-step overhead from the larger state vector will be modest:

- **`reduce_state!`**: called at every RHS evaluation to reconstruct storages from cumulative flows. Eliminated with storage states, but individually cheap (~microseconds).
- **Back-substitution**: expanding linear solve results from basin space back to full state space after each solve. Also eliminated, also individually cheap.
- **Solver vector operations**: ~10–15 BLAS-1 operations (norms, scaling, axpy) per step on 100k vs 15k entries. Real savings in memory bandwidth, but not dominant.
- **Step size control**: the solver computes error norms over all states. With 100k states, flows on links that don't materially affect the dynamics can force unnecessarily small steps. With 15k basin states, the solver only tracks what drives the dynamics. The impact on step count is hard to predict without benchmarking — it could be negligible or significant.

**Performance impact**: May improve significantly for large models, primarily from reduced per-step overhead and potentially fewer/larger steps.

### 3. Simplicity

The cumulative-flow formulation requires significant bookkeeping that would not exist with storage states:

| Machinery | Purpose | Needed with storage states? |
|---|---|---|
| `reduce_state!` | Implicit $A$ matrix: sum flows → storages | No |
| `formulate_storages!` | Add initial storage, precipitation, drainage | No (storage is the state) |
| `HalfLazyJacobian` | Wrap $J_g$ with implicit $A$ multiplication | No |
| `calc_J_inner!` / `update_J_inner!` | Compute $A \cdot J_g$ sparse product | No |
| `RibasimLinearSolve` + `dolinsolve` | Custom linear solve in reduced space + back-substitution | No |
| `CVector` / `StateTuple` | Index into state vector by component | Simplified to basin indexing |
| `link_to_state_idx` | Map links to states for multi-inflow nodes | No |
| `inflow_link_offsets` | Map nodes to flat state positions | No |
| `state_inflow_link` / `state_outflow_link` | Per-state link metadata | No |
| Water balance error diagnostic | Verify cumulative flows are consistent | Optional (solver guarantees it) |

The impact on development velocity was directly felt in [#3035](https://github.com/Deltares/Ribasim/pull/3035): adding multiple inflow links to UserDemand required threading changes through all of the above layers, plus adding new data structures (`link_to_state_idx`, `inflow_link_offsets`, `inflow_link_allocated`).

The effort to reduce the state count within the flow formulation ([#2614](https://github.com/Deltares/Ribasim/issues/2614) — combining evaporation/infiltration states) is also unnecessary with storage states, since those are already part of the basin storage derivative.

### 4. Water balance

The original issue ([#1808](https://github.com/Deltares/Ribasim/issues/1808)) argued that the water balance is "closed by definition" with cumulative flows. But with storage states the solver directly solves:

$$\frac{dS_b}{dt} = \sum q_{\text{in},b} - \sum q_{\text{out},b}$$

This is equally closed by construction — the solver enforces that storage change matches net flows to within solver tolerance. The water balance error diagnostic becomes optional.

### State vector

Replace the current state vector (one entry per flow link + evaporation + infiltration + PID integral) with:
- One storage state per Basin
- One PID integral state per PidControl

### RHS function

The RHS already computes all flow rates from storages/levels. The change: `du` entries become storage derivatives ($dS_b/dt = \sum q_{\text{in},b} - \sum q_{\text{out},b}$) instead of individual flow rates. The `formulate_flow!` methods compute the same flow rates — they're just summed into basin `du` entries rather than written to per-link state entries.

### Jacobian

The Jacobian becomes a standard $N_{\text{basins}} \times N_{\text{basins}}$ sparse matrix. Standard SciML sparse AD (`SparseConnectivityTracer` + `GreedyColoringAlgorithm`) works directly. The entire `differentiation.jl` custom machinery can be replaced with the default OrdinaryDiffEq Jacobian handling.

### Flow output

Per-link cumulative flows for output, allocation input, and pollutant transport can be accumulated via a trapezoidal callback. This approach was proposed in [#1473](https://github.com/Deltares/Ribasim/issues/1473) / [#1468](https://github.com/Deltares/Ribasim/pull/1468) but was set aside when the team chose cumulative-flow states instead. The technique is straightforward — it only evaluates flows at accepted integrator steps, avoiding the Gaussian quadrature issue that ruled out SciML's `IntegratingSumCallback`:

```julia
function accumulate_flows!(integrator)
    dt = integrator.t - integrator.tprev
    for (i, link) in enumerate(all_links)
        q_prev = flow_rate_prev[i]
        q_curr = compute_flow_rate(integrator, link)
        cumulative_flow[i] += 0.5 * (q_prev + q_curr) * dt  # trapezoidal
        flow_rate_prev[i] = q_curr
    end
end
step_cb = FunctionCallingCallback(accumulate_flows!)
```

This gives second-order accurate flow integration at negligible cost (one extra flow computation per accepted step). The cumulative flow values are accumulated in regular `Float64` vectors outside the state — they can be reset at each save point without affecting the solver, completely avoiding the growing-state problem that plagues the current formulation.

### What can be removed

- `reduce_state!` and the `RibasimReducedCVector` type
- `formulate_storages!` (storage is the state directly)
- `differentiation.jl` in its entirety (~500 lines): `HalfLazyJacobian`, `RibasimLinearSolve`, `RibasimLinearSolveCache`, `calc_J_inner!`, `update_J_inner!`, `calc_W_inner!`, custom `dolinsolve`, custom `SciMLBase.init`
- `build_link_to_state_idx`, `link_to_state_idx`, `inflow_link_offsets`
- `state_inflow_link`, `state_outflow_link` vectors
- `CVector` / `StateTuple` complexity (state vector becomes a plain vector indexed by basin)
- `check_new_input!` complexity (no need to track `u_reduced_prev_call`)
- Water balance error computation (or simplify to a lightweight optional diagnostic)

### What stays largely the same

- All `formulate_flow!` methods (they compute the same flow rates)
- Basin profile / level / area lookups
- Allocation logic
- Concentration / water quality
- Python side, QGIS plugin

## Validation

- Compare basin storage, level, and per-link flow results against the current formulation on the full test suite
- Verify water balance closure to solver tolerance
- Benchmark on a large model (LHM or similar) to quantify speedup
- Run long simulations (multi-year) to confirm no growing errors

## Related issues and PRs

| Reference | Relevance |
|---|---|
| [#1808](https://github.com/Deltares/Ribasim/issues/1808) | Original refactor to cumulative flows — this issue proposes evolving that decision |
| [#1897](https://github.com/Deltares/Ribasim/issues/1897) | Documents the growing-state numerical problem and various failed workarounds |
| [#2290](https://github.com/Deltares/Ribasim/issues/2290) | Proposed continuous flow resetting — complex, never implemented |
| [#2614](https://github.com/Deltares/Ribasim/issues/2614) | Proposed combining evaporation/infiltration states — unnecessary with storage states |
| [#2624](https://github.com/Deltares/Ribasim/pull/2624) | The $f = g \circ A$ Jacobian decomposition — great optimization, but unnecessary with storage states |
| [#1473](https://github.com/Deltares/Ribasim/issues/1473) | Proposed trapezoidal integration callback — closed as "not planned" when cumulative-flow states were chosen, but the technique is valid |
| [#1468](https://github.com/Deltares/Ribasim/pull/1468) | Prototype implementation of the trapezoidal callback |
| [#1444](https://github.com/Deltares/Ribasim/pull/1444) | Original PR adding flows as states — worked but was "pretty bad for performance" before #2624 fixed the Jacobian cost |
| [#3035](https://github.com/Deltares/Ribasim/pull/3035) | Multi-inflow UserDemand — directly motivated this proposal due to complexity encountered |

verheem:
Fixes https://github.com/Deltares/Ribasim/issues/3037

convergence was removed from flow.nc, only basin.nc now has convergence (breaking change)

Cosmetic storage removal was done in commit 351936b2cb5c9013af097d2b6c3cdfbf994ebd22.


core/src/allocation_optim.jl
```diff
@@ -67,8 +66,10 @@ function set_simulation_data!(
 
         # Check whether the storage in the physical layer is within the maximum storage bound
         if storage_now > storage_max
-            @error "Maximum basin storage exceeded (allocation infeasibility)" storage_now storage_max basin_id
-            errors = true
+            # double the maximum storage
+            storage_max *= 2.0
+            # @error "Maximum basin storage exceeded (allocation infeasibility)" storage_now storage_max basin_id
+            # errors = true
```
verheem: Decide better heuristic for max bound determining (could be based on the adaptive timestep feature which estimates the largest acceptable error in level, from which we can determine an upperbound (and maybe lower bound for the storage)

core/src/config.jl
```diff
@@ -189,7 +189,7 @@ end
     autodiff::Bool = true
     evaporate_mass::Bool = true
     depth_threshold::Float64 = 0.1
-    level_difference_threshold::Float64 = 0.02
+    level_difference_threshold::Float64 = 1.0e-4
```
visr, line 192: This would break Ribasim-NL, I'd prefer to keep it at 2 cm if possible.

visr:
I just tried a binary from this branch on a large part of the LHM model, and whereas the release finished in 2 hours, I killed this after a night, with an ETA of around 2 days. We'd need to profile to see if anything stands out.

verheem:
> I just tried a binary from this branch on a large part of the LHM model, and whereas the release finished in 2 hours, I killed this after a night, with an ETA of around 2 days. We'd need to profile to see if anything stands out.

Probably due to the level difference threshold of 1e-4. This also made HWS extremely slow. Lets try again with 2cm

visr:
You were planning to set QNDF as the default algorithm again right? I get unstable errors on Rosenbrock23, but on e.g. Aa en Maas on this branch on QNDF this runs in 15 minutes instead of an hour.

Having QNDF as the default again would make it easier to test this in Ribasim-NL.

visr:
Would be great if this can be rebased on top of main, because it cannot run LHM now since fixes from main are needed.

visr:
Thanks for the update. I use the core binaries from this branch now in Ribasim-NL by default.
Still running a more comprehensive benchmark, but speedups of the waterboard models are 2-3x.

Here is a profile of one of the slow models. Interesting that it spends 64% of time under `relax!` backtracking.
[aa-en-maas-storage-formulation.html](https://github.com/user-attachments/files/27808106/aa-en-maas-storage-formulation.html)

visr:
I pushed bd62c2c9f1c423a3c6d583da27d3aea0fbff4fc3, because I noticed that `BrabantseDelta_fix_model` did effectively hang due to the large abstol set here. (That commit also shows that Python had the wrong defaults compared to Julia.)

A quick search on that model gave this noisy result.

```
[solver]
# abstol = 1.0e-5  # 34 seconds, but also once 1 minute, 43 seconds, 4 milliseconds
# abstol = 2.0e-4  # 36 seconds, 305 milliseconds
abstol = 1.0e-4  # 30 seconds, 916 milliseconds (same timing twice)
# abstol = 2.0e-3  # 1 minute, 6 seconds, 388 milliseconds
# abstol = 1.0e-3  # 41 seconds, 261 milliseconds
# abstol = 1.0e-1  # stuck at the beginning
```

So I moved it down only one order from 1e-5 to 1e-4, instead of to 1e-1.

evetion:
In the Ribasim-NL LHM run I found some odd flow spikes:

Screenshot 2026-05-25 at 08 37 41

I've send you a link to the complete model results, you might have to rerun it with saveat=0 for a more specific time range (the whole run took 2 days).

verheem:
Weird, I analysed the situation. Findings so far:
In the entire waterboard, this is the only unrealistic flow.

Screenshot 2026-06-01 at 15 53 37


In the separate waterboard model, that flow link/basin are not present, so it seems generated in the LHM
Screenshot 2026-06-01 at 15 52 29


The sitatuon is 2 basins with a manning in between. 

(Bemaling gemaal Boezemland/Snelle Polder_129 #3900144)
The generated basin has an area of 20.000 m2 and connects to a basin with 3.900.000 m2 with a maning of length = 10 m, resistance = 0.01, profile width = 2 and slope = 3.

Screenshot 2026-06-01 at 16 04 30

However, in the neighborhood, there is almost exactly the same situation (see screenshot).

(Bemaling Kroes_16 #3900019)
The generated basin has an area of 115.000 m2 and connects to a basin with 200.000 m2 with a maning of length = 10 m, resistance = 0.01, profile width = 2 and slope = 3.

Screenshot 2026-06-01 at 16 03 54

So exactly the same manning, but the strange results occur with probably unstable numbers. The first situation connects 2 basins that differ by a factor of two order of magnitude in size, but the exact same manning parameters as in the stable situation.

.vscode/settings.json
```diff
@@ -69,6 +69,7 @@
         ".pixi"
     ],
     "julia.lint.run": true,
+    "julia.NumThreads": "1",
```
SouthEndMusic: Why is this needed?

core/src/allocation_optim.jl
```diff
@@ -766,22 +767,22 @@ function warm_start!(allocation_model::AllocationModel, integrator::DEIntegrator
     flow = problem[:flow]
     storage_change = problem[:basin_storage_change]
     du = get_du(integrator)
-    (; link_to_state_idx) = p.p_independent
 
     # Extrapolate the current instantaneous flow rates from the physical layer
+    # Flow rates are now stored in cache vectors, look up by link
```
SouthEndMusic: I suggest to have a look at these added comments; some are phrased in a way that only makes sense in the context of this PR

core/src/util.jl
```diff
@@ -690,237 +707,194 @@ low_storage_factor_resistance_node(::Parameters, q::GradientTracer, ::NodeID, ::
...
-        inflow ? :user_demand_inflow : :user_demand_outflow
-    else
-        snake_case(id)
-    end
+"Create a NamedTuple of the node IDs per state component in the state order"
+function state_node_ids(
+        p::Union{ParametersIndependent, NamedTuple},
+    )::StateTuple{Vector{NodeID}}
+    return (;
+        basin = p.basin.node_id,
```
SouthEndMusic: I think storage instead of basin is more clear

core/src/config.jl
```diff
@@ -388,41 +388,25 @@ function function_accepts_kwarg(f, kwarg)::Bool
 end
 
 function get_ad_type(solver::Solver)
-    chunksize = solver.specialize ? nothing : 1
```
SouthEndMusic, line 391: Why was this removed?

core/src/allocation_util.jl
```diff
@@ -440,7 +440,8 @@ function compute_adaptive_Δt(
             continue
         end
 
-        dstorage = formulate_dstorage_wrt_time(du, p.p_independent, t, basin_id)
+        # du.basin = dS/dt
```
SouthEndMusic: Here storage instead of basin would make more sense

core/src/callback.jl
```diff
@@ -66,11 +66,6 @@ function create_callbacks(
     discrete_control_cb = FunctionCallingCallback(apply_discrete_control!)
     push!(callbacks, discrete_control_cb)
 
-    toltimes = get_log_tstops(config.starttime, config.endtime)
```
SouthEndMusic, line 69: Also nice that this can be removed

core/src/callback.jl
```diff
@@ -82,69 +77,189 @@ function create_callbacks(
     return callback, saved
 end
 
+function sync_flow_rates!(node_sym::Symbol, current_flow_cache::Vector{T}, p_independent::ParametersIndependent) where {T}
```
SouthEndMusic: This needs some documentation. Also, it would be nice if this didn't need `getproperty`; I see that the symbol passed to this function is hardcoded, so you could just as well pass the node struct.

Is the goal of this function to translate from flows per node to flows per link? Since it's only copying, is this really needed?

core/src/callback.jl
```diff
@@ -82,69 +77,189 @@ function create_callbacks(
...
+    sync_flow_rates!(:tabulated_rating_curve, cache.current_flow_rate_tabulated_rating_curve, p_independent)
+    sync_flow_rates!(:manning_resistance, cache.current_flow_rate_manning_resistance, p_independent)
+    sync_user_demand_flow_rates!(cache.current_flow_rate_user_demand, time_dependent_cache.user_demand.current_return_factor, p_independent)
+
+    return nothing
+end
+
+"""
+Update cumulative flows.
+Also updates cumulative forcings and allocation flows.
```
SouthEndMusic: I would suggest to add a bit more explanation about what is being updated and why in the docstring.

core/src/callback.jl
```diff
@@ -82,69 +77,189 @@ function create_callbacks(
...
+    t0 = p_mutable.tprev
+    dt = t - t0
+
+    # Sync per-node-type flow rates to per-link vector (end-of-step rates)
+    sync_flow_rates!(p)
+
+    # Compute per-step flow volumes via Simpson's rule:
+    #   ∫ f dt ≈ (Δt/6)(f₀ + 4 f_mid + f₁)
+    # f₀ = start-of-step rate (flow_rate_prev), f₁ = end-of-step rate (current), and
+    # f_mid is evaluated from the solver's dense output at the step midpoint.
```
SouthEndMusic: This can be explained more clearly. As I understand it now, `integrator(u_tmp, τ)` calls the integrator as an interpolation at time `τ`, doing so in place in the provided `u_tmp`. If you decide to keep the allocations in the code for now, I would suggest to change it to `u_tmp = integrator(τ)` for clarity (and maybe give `u_tmp` a bit more descriptive name).

core/src/parameter.jl
```diff
@@ -1148,32 +1152,40 @@ the object itself is not.
...
+    cumulative_infiltration::Vector{Float64} = Float64[]
+    cumulative_evaporation_saveat::Vector{Float64} = Float64[]
+    cumulative_infiltration_saveat::Vector{Float64} = Float64[]
+    # Running totals for BMI (never reset)
+    cumulative_infiltration_total::Vector{Float64} = Float64[]
+    # Running total of UserDemand inflow per node for BMI (never reset)
+    cumulative_user_demand_inflow::Vector{Float64} = Float64[]
+    # Convergence tracking: accumulated normalized Newton residual per basin
+    convergence::Vector{Float64} = Float64[]
+    convergence_ncalls::Vector{Int} = [0]
```
SouthEndMusic: This is quite a large amount of added fields, so I suggest to bundle them in a substruct

core/src/callback.jl
```diff
@@ -82,69 +77,189 @@ function create_callbacks(
...
+    sc = state_and_time_dependent_cache
+
+    t0 = p_mutable.tprev
+    dt = t - t0
+
+    # Sync per-node-type flow rates to per-link vector (end-of-step rates)
+    sync_flow_rates!(p)
+
+    # Compute per-step flow volumes via Simpson's rule:
+    #   ∫ f dt ≈ (Δt/6)(f₀ + 4 f_mid + f₁)
```
SouthEndMusic, line 150: It would be nice if the code was formulated in such a way that this formula is represented in the code explicitly

core/src/callback.jl
```diff
@@ -82,69 +77,189 @@ function create_callbacks(
...
+        infil_end = copy(sc.current_infiltration)
+        # A separate du buffer (not get_du) so we don't clobber the integrator's du history.
+        du_tmp = similar(integrator.u)
+        u_tmp = similar(integrator.u)
+
+        # At a control-state transition the stored f₀ (flow_rate_prev) was invalidated with
+        # NaN because it held the pre-transition rate. Re-evaluate it under the current
+        # control state at the start-of-step state so Simpson's rule is valid for that step.
+        if any(isnan, p_independent.flow_rate_prev)
+            water_balance!(du_tmp, integrator.uprev, p, t0)
```
SouthEndMusic: This does more work than necessary; only the flow rates of the nodes whose parameters were altered by control have to be reformulated. Consider whether that refactor and bookkeeping effort is worth it.

core/src/callback.jl
```diff
@@ -182,7 +289,7 @@ function update_cumulative_flows!(u, t, integrator)::Nothing
 end
 
 function update_concentrations!(u, t, integrator)::Nothing
```
SouthEndMusic, line 257: Rather out of scope for this PR, but I'd like to hear you opinion on solving for the concentrations with a single Euler backward step. I think it highly simplifies the logic here. For a single tracer mass per basin u the system of equations is

du / dt = Q u / S(t)

where Q is the average flow matrix over a timestep and S(t) the storage per basin, where we interpret the division u / S(t) elementwise to obtain concentrations.

core/src/model.jl
```diff
@@ -232,7 +161,7 @@ function Model(config::Config)::Model
         dtmax = something(config.solver.dtmax, t_end),
         config.solver.force_dtmin,
         config.solver.abstol,
-        reltol,
+        reltol = config.solver.reltol,
```
SouthEndMusic: This is not necessary

core/src/util.jl
```diff
@@ -690,237 +707,194 @@ low_storage_factor_resistance_node(::Parameters, q::GradientTracer, ::NodeID, ::
...
-                _ in 1:length(p.user_demand.inflow_links[i])
-        ],
-        user_demand_outflow = p.user_demand.node_id,
-        linear_resistance = p.linear_resistance.node_id,
-        manning_resistance = p.manning_resistance.node_id,
-        evaporation = p.basin.node_id,
-        infiltration = p.basin.node_id,
-        integral = p.pid_control.node_id,
-    )
+@kwdef struct BackTracking
```
SouthEndMusic, line 711: I know we had a custom Backtracking implementation before, is that still needed?

SouthEndMusic (commented):
Overall it looks good, I added some questions and comments.

core/src/util.jl
verheem: Yes

core/src/config.jl
verheem: With AutoSparse + greedy coloring, the number of AD passes equals the number of colors, and the sparse backend batches the seeded directions per color. Pinning chunksize = 1 would block that batching and make the colored Jacobian much slower.

verheem: Also, solver.specialize is now a dead config field, removing it is a breaking change

visr: Besides the chunksize, why do we always use FullSpecialize now? Last time I checked, this only started paying off on LHM scale models, but perhaps this formulation changes that. Having it off by default helped in quickly starting models.

core/src/parameter.jl
SouthEndMusic: This doesn't have to be in this cache anymore as it is part of the state

core/src/callback.jl
```diff
@@ -573,14 +776,14 @@ end
...
     du = get_du(integrator)
     water_balance!(du, u, p, t)
 
     errors = false
     for id in basin.node_id
-        if state_and_time_dependent_cache.current_storage[id.idx] < 0
+        if u.basin[id.idx] < 0
             @error "Negative storage detected in $id"
             errors = true
         end
```
SouthEndMusic, line 608: This whole callback can be replaced by the old out of bounds check passed to the integrator

visr:
I tried e32b1baa6894d9a6b77c40a844d11bdf171b8957 in Ribasim-NL, and compared to an old build from this branch (5096b05) we ran into several water balance check failures of this kind:

```
┌ Error: Too large water balance error
│   id = Basin #1823
│   balance_error = 0.004613639761876393
│   relative_error = 0.056053436965884516
└ @ Ribasim /opt/teamcityagent/work/ecd2b8f9b25b1609/ribasim/core/src/callback.jl:652
```

I think that old commit had temporarily disabled this check.

In any case we should look more into these issues before we release Ribasim on this formulation. Perhaps we can tie up the few open comments and work those out on main, since it is quite clear we are heading this way. Then ideas like https://github.com/Deltares/Ribasim/issues/3108 can be tested on top of that.

Huite:
At the risk of sounding like I'm harping on: I still think that a (DAE) formulation in which we solve for both storage and cumulative flow is the most straightforward. This requires a custom linear solver, but that seems to be relatively straightforward under current DiffEq.jl, with a working version here for another case (where I reduce the linear system from 2N to N):
https://github.com/Huite/HydroDaemonDemo/issues/8#issuecomment-4508337244
(Finding and storing the indices should be done in a cleaner, more transparent manner.)

I need to finish some other things this week, and I'll have three weeks of leave, but afterwards I'd hope to create an accessible example which translates easily to Ribasim (e.g. a cascade of reservoirs). In that setting, it should be easy to show the limitations of quadrature methods, dense interpolation etc., and the (hopefully...!) relative simplicity of what I'm suggesting.

core/src/callback.jl
SouthEndMusic: Superseded by https://github.com/Deltares/Ribasim/issues/3125

core/src/config.jl
visr: I restored `specialize` as it is on main. I understand that chunksize 1 makes it slower, but last time we checked it also made it much quicker to start. Since a different chunksize needs different code compilation. I'd want to check the impact on both first before removing it, which is https://github.com/Deltares/Ribasim/issues/2657.

SouthEndMusic:
I realized a few remaining problems:
- The flow integration doesn't work with control by allocation, because the changes in flow due to allocation control are not taken into account in the $f_0$ term of the integration scheme. As a quick fix, at the end of `update_allocation!`, call `water_balance` and update `flow_rate_prev`
- The way flows are cached I think is a bit messy; caching them in `p_independent` in `water_balance!` and then copying them to `state_and_time_dependent_cache` in the callbacks; that is not what the `state_and_time_dependent_cache` is for. I propose:
  - Introducing a `CVector` for the different types of flows in `state_and_time_dependent_cache` and caching flows there when formulating;
  - formulating `du` at the end of `water_balance!` based on that vector
- I don't know whether we also did this before, but I think that doing `DiscreteControl` based on the latest computed instantaneous flows isn't great, because that's highly dependent on with which arguments `water_balance!` happened to be called last. It would be much better to do `DiscreteControl` based on the average flow of the last timestep.

I think that with https://github.com/Deltares/Ribasim/issues/3113 these things can be handled more cleanly.

visr:
Thanks for pushing on this. If you think these remaining points are better addressed in or after #3113, that is fine with me.

SouthEndMusic:
@visr I prefer to do it now, to have a proper state of the main branch. I also don't know when #3113 will be planned, there's some time in the current sprint to work on this.

SouthEndMusic:
Some more notes:

- I'm refactoring some things around multiple inflows for UserDemand nodes. That doesn't work properly at the moment because the individual inflows aren't stored. At the flow integration equal distribution is assumed
- The flow correction in the flow integration I think is not needed. If I understand correctly, this is about taking into account that we can integrate flow boundary inflow and infiltration exactly. The reason we integrated as many flows exactly as possible in the flow formulation is that it reduced the number of states, but with the storage formulation (and with https://github.com/Deltares/Ribasim/issues/3113) that is no longer a concern. The integration scheme in this PR is probably accurate enough. I also discuss whether we should use exact integration or not in https://github.com/Deltares/Ribasim/issues/3113 under the water balance error section.


The multiple inflows for UserDemand feature actually adds a bit of complexity for https://github.com/Deltares/Ribasim/issues/3113 because with that feature some flows do depend on more than (at most) an upstream and downstream storage.

SouthEndMusic:
I refactored a lot, and I was able to bring the number of failing tests down to a small amount. Still, I am not happy with this direction. Fundamentally we are back at the same problem we had before the cumulative flow states; the post-solver flow quadrature via Simpson's rule doesn't close the water balance sufficiently (at least to avoid negative tracer masses), and any subsequent step to correct that is either rather expensive (i.e. find a minimal correction of the flows that respects things like non-negativity constraints and return factors) or doesn't take parts of the physics into account (e.g. the correction invents non-existent infiltration to close the water balance). 

https://github.com/Deltares/Ribasim/issues/3113 really is the way to go; we should not have to guesstimate what flows the solver (implicitly) used internally.

I think that in the new formulation all Basin forcings should be treated as flow states again. I don't think it would significantly drop performance with the custom Jacobian and linear solve (which was the original reason to do exact integration), it cleans up the code, and most importantly it gets rid of all remaining headaches of matching exact integration with the solver results. If we really care about matching the exact integration we can inlcude it in `isoutofdomain`.

SouthEndMusic:
Closing this as superseded by https://github.com/Deltares/Ribasim/pull/3150.

PR #3042, opened 2026-04-21
Accept plain pd.DataFrame in TableModel.df typing

visr:
This avoids typing issues like:

```
  ERROR `pandas.core.frame.DataFrame` is not assignable to attribute `df` with type `pandera.typing.pandas.DataFrame[ManningResistanceStaticSchema] | None` [bad-assignment]
     --> src\ribasim_nl\ribasim_nl\profiles\implement.py:299:50
      |
  299 |       ribasim_model.manning_resistance.static.df = pd.concat(
      |  __________________________________________________^
  300 | |         [ribasim_model.manning_resistance.static.df, manning_static], ignore_index=True
  301 | |     )
      | |_____^
```

### input_base.py
- Widen `TableModel.df` type from `DataFrame[TableT] | None` to `DataFrame[TableT] | pd.DataFrame | None` so that plain `pd.DataFrame` values are accepted by type checkers
- Add `arbitrary_types_allowed=True` to `model_config` since Pydantic can't generate a schema for bare `pd.DataFrame` without it
- Update `tableschema()` to find the pandera type by iterating over the union args, since the annotation is no longer a simple two-member union

Pandera validation is preserved because `DataFrame[TableT]` comes first in the union — Pydantic tries it before falling through to `pd.DataFrame`.

### test_schemas.py
- `test_df_accepts_plain_dataframe`: runtime test that `pd.concat` result can be assigned to `.df` and pandera validation still runs (missing columns added, index coerced)
- `test_df_accepts_plain_dataframe_typing`: runs pyrefly on a snippet to assert no `bad-assignment` error

### DIFF
--- a/python/ribasim/ribasim/input_base.py
+++ b/python/ribasim/ribasim/input_base.py
@@ -403,11 +403,14 @@ class TableModel[TableT: _BaseSchema](FileModel, ChildModel):
     are represented by a filepath, and are ChildModels, as they're contained in a NodeModel.
     """
 
-    df: DataFrame[TableT] | None = Field(default=None, exclude=True, repr=False)
+    df: DataFrame[TableT] | pd.DataFrame | None = Field(
+        default=None, exclude=True, repr=False
+    )
     _sort_keys: list[str] = PrivateAttr(default=[])
 
     model_config = ConfigDict(
         extra="allow",
+        arbitrary_types_allowed=True,
     )
 
     def __eq__(self, other: object) -> bool:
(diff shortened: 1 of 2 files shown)


visr:
**Why it might not be needed long-term:**
- Pandera could make `DataFrame[T]` a proper type alias for `pd.DataFrame` (it's already just a wrapper at runtime)
- Pyrefly could add special handling for pandera's typing wrappers, similar to how some type checkers handle SQLAlchemy
- This is a known gap in pandera's type system — other type checkers (mypy, pyright) have the same issue

**What the change costs you:**
- `arbitrary_types_allowed=True` weakens Pydantic's type safety on `TableModel`
- `tableschema()` became more fragile (iterating union args instead of a direct lookup)
- The union `DataFrame[TableT] | pd.DataFrame` is subtle — ordering matters for correctness

**The simpler alternative:** The downstream code (ribasim-nl) can just add `# pyrefly: ignore[bad-assignment]` at the few `pd.concat` assignment sites. That's a one-line fix per call site, keeps ribasim's internals clean, and has zero risk of breaking pandera validation.

My recommendation: **don't merge this PR.** The complexity isn't worth it for a type checker annotation issue. The `# pyrefly: ignore` approach at call sites is more pragmatic and easier to remove later if/when pandera or pyrefly fixes this upstream.

Issue #3040, opened 2026-04-20
Make tracers non-experimental


evetion:
In order to make our tracer computation non-experimental:
- [ ] Implement subtimestepping (#3014)
- [ ] Move the tracer config to its own group
- [ ] Document the feature (possibly by @Fati-Mon)
- [ ] Validate the implementation (that subtimesteps are required, and compared to Delwaq).
- [ ] Consider implementing #1969 (or at least how the tables would look like to make it non-breaking in the future).
- [ ] Consider implementing time tracers.
- [ ] Possibly rename concentration to tracer?

@JohnRushKucharski

gijsber:
renaming concentration to tracer seems a good idea to me, as it reduces expectations on the functionality and makes clear that you (basically) need a water quality model (like Delwaq) if you want to do substance/concentration modelling

PR #3052, opened 2026-04-24
Improve order of tracers in plot/legend

evetion:
- Adds default (sorted) ordering (but Initial always comes first) in plot
- Flips legend order to have the same order as the plot
- Allow user color with default for Initial

This should make the fraction plot a bit more useful. As suggested by @DanielTollenaar, and inspired by https://github.com/d2hydro/ribasim_delwaq_aam/blob/main/src/ribasim_tools/ribasim_tools/plot_fractions.py.

Automatic grouping of specific tracers requires more input (something to tackle in #3040). I can make a PR for easier access to the results.

image


visr:
For the default tracers this changes the preferred order set in https://github.com/Deltares/Ribasim/pull/3007. That is not only Initial first, but Precipitation on top etc.

Good to force Initial first for non-default tracers. Could you add the same edit to the QGIS plugin so that it remains consistent?

visr (approved):

PR #3060, opened 2026-05-04
Upload MSIX with release assets

### LINKED ISSUE #3012: Automate the MSIX release
All normal builds are green, but the signing agent doesn't seem to cooperate.
The "Make GitHub Release" pipeline reports:

> Snapshot dependency failed to start: Signing_and_certificates / Ribasim / Signing Ribasim Release

I assume also the MSIX is not fully integrated into this release pipeline yet? The unsigned MSIX was created, but I cannot trigger the signing myself. Once the GitHub release is made and we have a signed MSIX we should add it to the release assets.

The Conda recipes are already merged, and the QGIS plugin is pending approval.

@evetion are you able to help out with this?

visr:
Should fix https://github.com/Deltares/Ribasim/issues/3012


.teamcity/Ribasim/buildTypes/Ribasim_MakeGitHubRelease.kt
```diff
@@ -97,6 +97,17 @@ object Ribasim_MakeGitHubRelease : BuildType({
...
+            snapshot {
+                reuseBuilds = ReuseBuilds.NO
+                onDependencyFailure = FailureAction.FAIL_TO_START
+            }
+
+            artifacts {
+                id = "ARTIFACT_DEPENDENCY_158"
+                artifactRules = "ribasim_windows.msix"
+            }
+        }
```
visr, line 110: This copies the dependency on SigningAndCertificates_Ribasim_SigningRibasimRelease for the signed Windows ZIP above. AFAICT the artifact ID just needs to be unique, so I just did +1 of the previous.

.teamcity/Ribasim_Windows/RibasimWindowsProject.kt
```diff
@@ -80,9 +80,6 @@ object Windows_BuildMsix : BuildType({
     templates(WindowsAgent, GithubCommitStatusIntegration)
     name = "Build MSIX Package"
 
-    // Disable this until signtool is updated so we can sign the MSIX
-    paused = true
-
```
visr, line 85: Not sure what effect this really has. We don't need to trigger this on every commit to main / daily, but I don't think we do, right?

evetion: We should merge the patches first, this was already unpaused in https://github.com/Deltares/Ribasim/blob/main/.teamcity/patches/buildTypes/Windows_BuildMsix.kts.

Also:
```
Please change the [configs version](https://www.jetbrains.com/help/teamcity/2025.11/?Upgrading+DSL#UpgradingDSL-configsVersion) to "2025.11" in projects:
```

About patches
> It is implied that you move the changes from the patch file to the corresponding .kt or .kts file and delete the patch file. Patches generation allows you to continue using the user interface for editing of the project settings and at the same time use the benefits of Kotlin DSL scripts.

Issue #3034, opened 2026-04-14
Ribasim-DELWAQ crash


DanielTollenaar:
Having this DELWAQ parsed with ribasin.delwaq.generate: https://we.tl/t-WOtiXP2z8kmPLhur

Running with:
```
"c:\Program Files\Deltares\D-HYDRO Suite 2026.01 1D2D\plugins\DeltaShell.Dimr\kernels\x64\bin\delwaq.exe" delwaq.inp  delwaq.inp
```

Doesn't want to run with this message (see `delwaq.lst`):

<img width="1071" height="472" alt="Image" src="https://github.com/user-attachments/assets/c74658cf-ee3e-4d83-83a6-cbdc19e5bba2" />

I've had this message before, back then it was related shared boundary-nodes (https://github.com/d2hydro/ribasim_delwaq_aam/issues/5#issuecomment-3452077749). I've checked ribasim_bndlist.inc and do not see one.

DanielTollenaar:
Update, there are levelboundary nodes shared by > 1 connector-node. These are: [8, 50, 55, 81, 115, 133, 136, 1652]. See, e.g.:

<img width="575" height="481" alt="Image" src="https://github.com/user-attachments/assets/3bc2c4b2-df68-4ffe-9c19-d2c2ac207cca" />

Not sure if it's related as I don't see duplicated boundary nodes in the ribasim_bndlist.inc

evetion:
@DanielTollenaar can you provide the Ribasim model so I can investigate?

evetion:
After some investigation, we create a Delwaq edge from a boundary to a boundary without data (thus crashing Delwaq with an off by -one boundary list), because I didn't code for such an occurence. 

A boundary-boundary you say? Yes, this model has a FlowBoundary to a LevelBoundary, and it is something we allow in Ribasim 👀. 

<img width="503" height="537" alt="Image" src="https://github.com/user-attachments/assets/88c1553b-924f-4851-8fdd-578910de1ae4" />

I think this is is mistake in the model (the FlowBoundary doesn't do anything, it doesn't influence the level of the LevelBoundary), and it is is mistake in Ribasim. FlowBoundaries can connect to LevelBoundaries and Terminals, which don't make sense to me.

gijsber:
Maybe time to tighten the validation rules ?
--> It should not be allowed to connect Boundary nodes to other Boundary nodes (including Terminals)

gijsber:
[...]

PR #3062, opened 2026-05-05
Flowboundary should not connect to Levelboundary

evetion:
This will hopefully prevent #3034 from occurring.


visr:
Closing, see discussion in https://github.com/Deltares/Ribasim/issues/3034

PR #3088, opened 2026-05-22
[delwaq] Allow for isolated basin nodes.

evetion:
As they're present in the LHM. I've disabled erroring out on too long names, now they will be cut off (which is only problematic if they become non-unique). Currently running Delwaq, which takes a loooong time.

### DIFF
--- a/python/ribasim/tests/test_delwaq.py
+++ b/python/ribasim/tests/test_delwaq.py
@@ -6,6 +6,7 @@ import pytest
 import xarray as xr
 from ribasim import Model
 from ribasim.delwaq import add_tracer, generate, parse, run_delwaq
+from ribasim.delwaq.util import check_substance_uniqueness
 
 delwaq_dir = Path(__file__).parent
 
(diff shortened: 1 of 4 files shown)


DanielTollenaar:
Hi, I think it is better practice to validate substance name length at delwaq generate. I've done this for Aa en Maas. Reason; I expect to extract concentrations by substances I've defined, not by it's first 20 characters.

python/ribasim/tests/test_delwaq.py
```diff
@@ -65,7 +66,7 @@ def test_offline_delwaq_coupling(tmp_path):
 
     # Ensure concentration.nc uses the expected dimension order
     with xr.open_dataset(results_dir / "concentration.nc") as uds:
-        assert uds["concentration"].dims == ("node_id", "substance", "time")
+        assert uds["concentration"].dims == ("time", "substance", "node_id")
```
visr, line 69: This seems correct. How did this not fail before?

visr (approved):

python/ribasim/tests/test_delwaq.py
evetion: Because the concentration.nc had the same (incorrect) dimensions. And thus failed to load in QGIS.

See line 60 in python/ribasim/ribasim/delwaq/parse.py in this PR.

PR #3090, opened 2026-05-28
QGIS test return code improvements

visr:
Aligning with https://github.com/Deltares/QGIS-Tim/pull/188, which is based off the Ribasim approach.

### DIFF
--- a/ribasim_qgis/scripts/qgis_testrunner.py
+++ b/ribasim_qgis/scripts/qgis_testrunner.py
@@ -102,9 +102,9 @@ writeOut.write = _write
 sys.path.append(QDir.current().path())
 
 
-def __exit_qgis(exit_code: int):
+def __exit_qgis(return_code: int):
     """Exit the QGIS application with the given exit code."""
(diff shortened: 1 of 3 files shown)


visr:
I checked, and both main and this branch return 1 when introducing a failing test. AI summary:

The **mechanism** is different:

**On main:** QGIS always exits with code 0 (the `run_all()` return value is discarded). The failure is caught by string-matching in run_qgis_ui_tests.py:
```python
if any(s in qgis_process.stdout for s in ["QGIS died on signal", "FAILED"]):
    exit(1)
```

**On this branch:** QGIS exits with pytest's actual return code, and `check_returncode()` raises `CalledProcessError` directly.

So the branch is a valid cleanup — it replaces fragile stdout parsing with proper exit code propagation. The difference in output is just that you now get the `CalledProcessError` traceback instead of a silent `exit(1)`.

One minor observation: the removed `"QGIS died on signal"` check is fine because a signal-killed process would have a non-zero exit code anyway, caught by `check_returncode()`.

PR #3103, opened 2026-06-09
Agricultural module prototype

### LINKED ISSUE #3100: Agricultural module prototype
**What**
A prototype agricultural/irrigation module that computes water demand based on cropping needs (evapotranspiration) and soil moisture status. At a later stage (not now) this needs to be extended with a soil salinity balance to computes salt flushing needs
Each timestep the status is updated with actual rain/evap and allocated irrigation water

**Why**
The following information needs to obtained from the prototype:
1. how do we relate this landbased-module to the UserDemand node and the Basin node
2. which variables need to be exchanged on a timestep basis between the library and Ribasim allocation core
3. which variables need to be supplied via the Ribasim data model
4. how will timestepping work
5. what needs to be adjusted on the library side to enable smooth integration
6. where are the risks or complications when going for a full implementation
7. how much effort is foreseen to establish the data model with associated Julia IO
8. how much effort is foreseen to actually have a good demand-allocation-update implementation
9. does this still look like a good and achievable choice ?

**How**
Using a shared library from wflow
Using the Piave model as a testcase with two or three cells representing different cropping areas

**Additional context**
The prototype does not need to be a fully working implementation as long as we can answer the questions

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3100

linked PR in Wflow: https://github.com/Deltares/Wflow.jl/pull/935


SouthEndMusic:
Superseded by https://github.com/Deltares/Ribasim/pull/3148

PR #3110, opened 2026-06-15
[delwaq] Drop required graph and substance state variables

evetion:
So users can just do `generate`, run Delwaq, and run `parse`, instead of having to keep their session open, or rerun `generate` to have the mandatory variables for `parse`.

I have dropped the visualization from the docs, but the description is still there, as is the printer substances (just a bit later). I'm currently testing this approach on the latest LHM, I can confirm the presence of (good looking) variables in the `ribasim.nc`.

CC @DanielTollenaar

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -118,6 +118,7 @@ upload-node-icons = "python utils/write_node_icons.py --upload"
 tests = { depends-on = ["lint", "test-ribasim-python", "test-ribasim-core"] }
 delwaq = { cmd = "pytest --basetemp=python/ribasim/ribasim/delwaq/model python/ribasim/tests/test_delwaq.py" }
 gen-delwaq = { cmd = "python python/ribasim/ribasim/delwaq/generate.py" }
+parse-delwaq = { cmd = "python python/ribasim/ribasim/delwaq/parse.py" }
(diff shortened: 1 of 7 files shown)


visr (approved):
Looks good to me!

pixi.toml
```diff
@@ -118,6 +118,7 @@ upload-node-icons = "python utils/write_node_icons.py --upload"
 tests = { depends-on = ["lint", "test-ribasim-python", "test-ribasim-core"] }
 delwaq = { cmd = "pytest --basetemp=python/ribasim/ribasim/delwaq/model python/ribasim/tests/test_delwaq.py" }
 gen-delwaq = { cmd = "python python/ribasim/ribasim/delwaq/generate.py" }
+parse-delwaq = { cmd = "python python/ribasim/ribasim/delwaq/parse.py" }
```
visr, line 121: Slight improvement?

```
python -m ribasim.delwaq.generate
python -m ribasim.delwaq.parse
```

And probably we need to mention these CLI options in the module docstrings?

evetion: Docstrings are done, but the -m nets
> <frozen runpy>:128: RuntimeWarning: 'ribasim.delwaq.parse' found in sys.modules after import of package 'ribasim.delwaq', but prior to execution of 'ribasim.delwaq.parse'; this may result in unpredictable behaviour

So I'd skip it for now.

Issue #2247, opened 2025-04-17
Add an optional time interval to `DiscreteControl` nodes


SouthEndMusic:
Currently the `DiscreteControl` run logic is applied after every solver timestep. However, sometimes it is more realistic human behavior to apply the logic for instance once a day. This interval can be different between different `DiscreteControl` nodes.

Eventually we also want this update interval to be dependent on the state of the model. This is not necessary for this first implementation, but it could be helpful to keep in mind. This could also be done with `DiscreteControl`, but that might get a bit meta.

visr:
I think `DiscreteControl / condition` would make most sense. That also is already optionally time-dependent via the `time` column. We could add an optional `interval` column. We could for now restrict to only having a constant interval per node_id, and then the data model wouldn't change when we start allowing it to change.

https://ribasim.org/reference/node/discrete-control.html#condition

Your implementation suggestion with split callbacks makes sense to me. Also worth considering is adding tstops at the start, keeping one `FunctionCallingCallback`, and then skipping all controls that don't need to be run at the given `t % interval`. If that can be done cheaply enough that may be simpler.

PR #3150, opened 2026-07-14
Storage and flow states with custom Jacobian + linear solve: the formulation to rule them all

### LINKED ISSUE #3113: Combined storage and cumulative flow states: a highly efficient solver algorithm independent approach
## Method

Say we have the state vector $u = [s, Q, I]^\top$ containing the storages $s$, the cumulative flows $Q$ and the PID control integral states $I$ . For this we have the following ODE system:

$$
\frac{\text{d}}{\text{d}t}
\begin{bmatrix}
s \\\ Q \\\ I
\end{bmatrix} = 
\begin{bmatrix}
M\frac{\text{d}Q}{\text{d}t} \\\ q(s, I, t) \\\ E(s,t)
\end{bmatrix} =
\begin{bmatrix}
0 & M & 0 \\\
0 & 0 & 0 \\\
0 & 0 & 0 \\\
\end{bmatrix}
\frac{\text{d}}{\text{d}t}
\begin{bmatrix}
s \\\ Q \\\ I
\end{bmatrix} +
\begin{bmatrix}
0 \\\ q(s, I, t) \\\ E(s, t)
\end{bmatrix},
$$

and rearranging yields

$$
\begin{bmatrix}
I_n & -M & 0 \\\
0 & I_m & 0 \\\
0 & 0 & I_\nu
\end{bmatrix}
\frac{\text{d}}{\text{d}t}
\begin{bmatrix}
s \\\ Q \\\ I
\end{bmatrix} =
\begin{bmatrix}
0 \\\ q(s, I, t) \\\ E(s, t)
\end{bmatrix}.
$$

Here $\nu$ is the number of PID control nodes and

$$
E(s, t) = r(t) - S_\text{PID} h(s)
$$

is the vector of PID control error terms, with setpoints $r(t)$, PID controlled Basin selection matrix $^{1}$ $S_\text{PID}$ and storage to level relations $h(s)$.

Now we have a linear system with a non-trivial but constant mass matrix, lets call it $\mathcal{A}$. The Jacobian of the rhs is given by

$$
J = 
\begin{bmatrix}
0 & 0 & 0\ \\
J_s & 0 & J_I \\\
-S_\text{PID} \text{diag}\left(h'(s)\right) & 0 & 0
\end{bmatrix},
$$

where $J_s$ and $J_I$ are the Jacobian of $q(s,I,t)$ w.r.t. $s$ and $I$ respectively. The general form of the linear solve for Newton iterations is

$$
(-\gamma^{-1}A + J)x = c,
$$

where thus

$$
-\gamma^{-1}A + J =  
\begin{bmatrix}
-\gamma^{-1}I_n & \gamma^{-1}M & 0 \\\
J_s & -\gamma^{-1}I_m & J_I \\\
-S_\text{PID} \text{diag}\left(h'(s)\right) & 0 & -\gamma^{-1}I_\nu
\end{bmatrix}.
$$

Writing $x = [x_s, x_Q, x_I]^\top, c = [c_s, c_Q, c_I]^\top$, this linear system is solvable in 3 steps:

$$
\begin{align}
      \left[-\gamma^{-1}I_n + M\left(J_s - \gamma J_I S_\text{PID}\\text{diag}\left(h'(s)\right)\right)\right]x_s &= c_s + M\left(c_Q + \gamma J_I c_I\right) \quad &\text{(linear system solve in storage space)} \\
                                      x_I &= -\gamma\left[c_I + S_\text{PID}\left(h'(s) \odot x_s\right)\right] \quad &\text{(explicit computation)} \\
                                      x_Q &= \gamma \left[-c_Q + J_s x_s + J_I x_I\right] &\text{(explicit computation)}
\end{align}
$$

Here we use $v \odot w := diag(v)w$ as notation for elementwise multiplication.

## Evaluating the Jacobian

The nontrivial parts of the Jacobian are $J_s$ and $J_I$. To see how we can compute these efficiently, we write 

$$
q(s, I, t) = q\left(s_\text{up}(s), \ s_\text{down}(s), \ I, \ c\left(s, \ q(s, t)\right), \ t\right) 
$$

reflecting that flows can depend on:
- the level uplink: $s_\text{up}(s) = S_\text{up}s$;
- the level downlink: $s_\text{down}(s) = S_\text{down}s$;
- the PID control integral term $I$;
- the continuous control compound variable $c(s, q(s, t))$.

By the first expression looks like q is implicitly or recursively defined because of the continuous control compound variable term, but it is not; we do not (or at least should not) allow Continuous Control to listen to a flow that is itself controlled by ContinuousControl (https://github.com/Deltares/Ribasim/issues/3128).

For $J_s$ we then obtain

$$
\begin{align}
J_s &=& \frac{\partial q}{\partial s_\text{up}}S_\text{up} + \frac{\partial q}{\partial s_\text{down}}S_\text{down} + \frac{\partial q}{\partial c}\left[\frac{\partial c}{\partial s} + \frac{\partial c}{\partial q}J_s\right] \\
&=& \left[I_m + \frac{\partial q}{\partial c}\frac{\partial c}{\partial q}\right]
\left[\frac{\partial q}{\partial s_\text{up}}S_\text{up} + \frac{\partial q}{\partial s_\text{down}}S_\text{down} + \frac{\partial q}{\partial c}\frac{\partial c}{\partial s}\right]
\end{align}
$$

and $J_I = \frac{\partial q}{\partial I}$. 

### Flow derivatives

So to compute the Jacobian we first of all need the terms $\frac{\partial q}{\partial s_\text{up}}, \frac{\partial q}{\partial s_\text{down}}$, $\frac{\partial q}{\partial c}, \frac{\partial q}{\partial I}$. And here is the neat thing: ***these can all be computed with a single*** $q$ ***call by clever use of `DifferentiationInterface.pushforward!`***. We can bypass all the sparsity machinery: sparsity detection, matrix coloring and result decompression. That is because every flow depends on at most one upstream level, one downstream level, one area, one PID integral term and one continuous control compound variable.  A code example:

*This code is out of date with the above*

```julia
using DifferentiationInterface: prepare_pushforward, pushforward!
using ForwardDiff
using SparseArrays: SparseMatrixCSC, spzeros

# q_unreduced: CVector with component per connector node and vertical outflow
function formulate_unreduced_flows!(q_unreduced, unreduced_flows_input, p, t)
    # ...
end

# UnreducedFlowsInputCVector: CVector with components h_up, h_down, evaporation_area, PID_area, integral, continuous_control_compound

function make_tangents_in(unreduced_flows_input::UnreducedFlowsInputCVector)
    tangents_in = ntuple(i -> zero(unreduced_flows_input), Val(5))
    tangents_in[1].h_up .= 1
    tangents_in[2].h_down .= 1
    tangents_in[3].evaporation_area .= 1
    tangents_in[3].PID_area .= 1
    tangents_in[4].integral .= 1
    tangents_in[5].continuous_control_compound .= 1
    return tangents_in
end

# This is a lazy representation of the Jacobian; this data will be used to formulate the linear solve
@kwdef struct RibasimJac <: AbstractMatrix{Float64}
    n_basin::Int
    unreduced_flows_input::UnreducedFlowsInputCVector
    tangents_in::NTuple{4, UnreducedFlowsInputCVector} = make_tangents_in(unreduced_flows_input)
    derivative_cache::NTuple{4, UnreducedFlowsInputCVector} = ntuple(i -> zero(unreduced_flows_input), Val(5))
    q_unreduced::Vector{Float64} = zeros(length(unreduced_flows_input.h_up))
    ∂q_unreduced_∂h_up::Vector{Float64} = zeros(length(unreduced_flows_input.h_up))
    ∂q_unreduced_∂h_down::Vector{Float64} = zeros(length(unreduced_flows_input.h_dow))
    ∂q_unreduced_∂integral::Vector{Float64} = zeros(length(unreduced_flows_input.integral))
    ∂q_unreduced_∂continuous_control_compound::Vector{Float64} = zeros(length(unreduced_flows_input.continuous_control_compound))
    ∂continuous_control_compound_∂storage::SparseMatrixCSC{Float64} = spzeros(length(unreduced_flows_input.continuous_control_compound), n_basin)
    ∂evaporation_∂area::Vector{Float64} = zeros(n_basin)
    ∂PID_flow_∂area::Vector{Float64} = zeros(length(unreduced_flows_input.PID_area))
end

## Preparation phase

pushforward_prep = prepare_pushforward(
    formulate_unreduced_flows!,
    q_unreduced,
    AutoForwardDiff(),
    unreduced_flows_input,
    Constant(p), # Neatly, formulate_unreduced_flows! doesn't need any caches for derivative propagation.
    Constant(t)
)

## Computation phase

jac(J::RibasimJac, u, p, t) = begin
    (;
        unreduced_flows_input,
        tangents_in,
        derivative_cache,
        q_unreduced,
        ∂q_unreduced_∂h_up,
        ∂q_unreduced_∂h_down,
        ∂q_unreduced_∂integral,
        ∂q_unreduced_∂continuous_control_compound,
        ∂evaporation_∂area,
        ∂PID_flow_∂area,
        ∂continuous_control_compound_∂storage,
    ) = J

    # Compute unreduced_flows_input, c and its Jacobians
    # ...

    # Compute flow derivatives
    pushforward!(
        formulate_unreduced_flows!,
        q_unreduced,
        derivative_cache,
        pushforward_prep,
        unreduced_flows_input,
        tangents_in,
        Constant(p),
    )

    ∂q_unreduced_∂h_up .= derivative_cache[1].h_up
    ∂q_unreduced_∂h_down .= derivative_cache[2].h_down
    ∂q_unreduced_∂integral .= derivative_cache[3].integral
    ∂q_unreduced_∂evaporation_area .= derivative_cache[4].evaporation_area
    ∂q_unreduced_∂PID_area .= derivative_cache[4].PID_area
    ∂q_unreduced_∂continuous_control_compound .= derivative_cache[5].continuous_control_compound

    return nothing
end
```

### ContinuousControl compound variable derivatives

We furthermore need the terms $\frac{\partial c}{\partial s}$ and $\frac{\partial c}{\partial q}$. These are quite easy to compute analytically, but it's probably cleaner and more general (with generalizations like https://github.com/Deltares/Ribasim/issues/2944 in mind) to do this with sparsity detection + AD. 

## Water balance error

The water balance error over a solver timestep is given by

$$
\Delta s - \left[M\Delta Q + \Delta B\right].
$$

$\Delta B$ can be computed by exact integration of the inflow forcing interpolations. However, the solver does it's own potentially inexact integration of $b(t)$, leading to a discrepancy. I see several ways to handle this:

- Add $B$ as states so that we know the exact solver integration, completely abandoning exact integration;
- Instead of $s$, we have just $v := s_0 + MQ$ as a state. Then everywhere $s$ is needed $B$ can be computed exactly;
- Accept this discrepancy. Depending on the degree of the interpolation and the order of the solver algorithm, the solver integration might even be exact too. This is probably the preferred option, at least to start with.

## Implementation steps
- [ ] Define the new state vector structure; like [this](https://github.com/Deltares/Ribasim/blob/28ba0c91aa187217e25ba2ed87a7e7db149dec8d/core/src/parameter.jl#L11-L26) but including storage
- [ ] Create a new function `formulate_compound_variables!(compound_variables, input, storage, p, t)` for both DiscreteControl and ContinuousControl, where `input` is a CVector with components of levels and flows. Storage is not part of `input`, because ContinuousControl doesn't listen to storage so we don't have to differentiate with respect to it
- [ ] Create a new function `formulate_unreduced_flows_input!(unreduced_flows_input, p, t)`
- [ ] Create a new function `formulate_unreduced_flows!(q_unreduced, unreduced_flows_input, p, t)`
- [ ] Define `RibasimJac` as above, with any required overloads for it
- [ ] Update `water_balance!` and `jac` for the new formulation
- [ ] Implement the custom linear solve. Here some care has to be taken:
  - to obtain the proper sparsity pattern for the linear solve at the start, so it's best to apply sparsity detection to the matrix construction
  - If the Jacobian wasn't updated, many steps (primarily matrix factorization) can be skipped
- [ ] Update flow result parsing in the callbacks
- [ ] Investigate how best to only compute the solver error based on the storages; via a custom `internalnorm` or per-state tolerances
- [ ] Formulate `isoutofdomain` for the storage and flow states
- [ ] Since the Jacobian is now so much cheaper to compute, investigate whether
- *To be continued*

## Advantages
- Solver provided integrated flows
- Extremely efficient Jacobian computation (this is possible in whatever state configuration we choose)

## Disadvantages
- We might have to fiddle with the tolerances again, although for this formulation it might work to just ignore the $Q$ part in the `internalnorm`.

## `isoutofdomain`

With all these states there's a lot we can enforce on the solver level with `isoutofdomain`:
- Non-negative storages: $s \ge 0$;
- Non-negative flows where known: $\Delta Q_i \ge 0$;
- Non negative tracer mass: $T \ge 0$.

## Footnotes
$^{1}$ Generally these selection matrices should not explicitly be used in the code; they are a notational convenience
$^{2}$ Note that $s$ and $s_0 + MQ + B$ are interchangable (up to solver accuracy). The latter yields nicer Jacobian expressions which is why they are used here, but $s_0 + MQ + B$ never has to be explicitly computed because $s$ is available.

SouthEndMusic:
fixes https://github.com/Deltares/Ribasim/issues/3113.

I did little to no vibe development for this branch, to make things as explainable as possible.

### File structure changes
- What was `solve.jl` where the rhs for the ODE lives is renamed to `formulate_flows.jl`
- A new `solve.jl` is introduced which contains the custom mass matrix + Jacobian + linear solve implementation

### Discontinuity handling
Handling discontinuities of the rhs deserves some attention. Now that I made all positive forcings part of the state, I noticed that the solver isn't able to integrate them exactly, even though they are constant over the timestep. This is because of the memory QNDF uses of previous timesteps, where the forcing is sometimes different. To fix this and related problems, I had to set the `integrator.derivative_discontinuity` flag when callbacks change parameters which gets most of the way there, and  add a `step_limiter!` to get there exactly. @huite this of course relates to our discussion about performance and what you care about to see in the result, but I think I remember discussing with @visr a while back that we should care about preserving the forcing volume.

It could be interesting to investigate whether setting `integrator.derivative_discontinuity` as a significant impact on performace. If it does, it's worth putting in some more effort to catch also the discontinuities not caused by  callbacks but by non-differentiable time interpolations. There are also still hard clamps for certain flows, see https://github.com/Deltares/Ribasim/issues/1918.

### How to interpret the solver config options
The implementations in this PR hook into the solver at a quite low level, so they are less flexible when it comes to certain solver options than we had before. More precisely:
- `sparse` doesn't mean much for the AD, because the custom implementation represents the Jacobian in a lazy way, and doesn't use sparsity detection and coloring etc. on the rhs. The only sparse arrays still present are `J_inner` and the Jacobian of the continuous control compound variables with respect to the storages and flows, and don't think it makes sense to keep the sparse option just for those.
- `autodiff`: The way I implemented the Jacobian computation is locked into the `ForwardDiff` API. I could refactor it so that it is compatible with `DifferentiationInterface` again as mentioned [here](https://github.com/Deltares/Ribasim/issues/3113#issuecomment-4777753794), which requires defining a new `CVector` type for all the inputs into a single vector.

I think it should be possible to bypass my whole implementation, but we need to think about what we want to support there and how that fits into the config. 

#### Update

I discussed this with @visr and we concluded that:
- It is good to refactor the current implementation so that the AD is implemented in `DifferentiationInterface` terms. That's a bit more readable and can then make use of the `autodiff` flag.
- It should indeed be possible to bypass all the custom solve logic. Then the old sparsity detection + sparse matrix coloring + default linear solve will be used. Let's use a new flag `optimized_solve = true` for this. 
- To avoid confusion and because it is not really necessary anymore the `sparse` flag can be deprecated (will warn and not do anything). 

### The state: `Carrays` $\rightarrow$ `CVectors`
The use of vector components in this implementation is more complex because they are now nested: `(storage, (flows...), pid_integral)`. This was rather broken in the `CArrays` implementation, so I reworked it into `CVectors` which additional constraints like that the ranges must be contiguous. I also added some pretty printing:

```julia
julia> model.integrator.u
33-element Ribasim.CVectors.CVector{Float64, Vector{Float64}, @NamedTuple{storage::UnitRange{Int64}, flow::@NamedTuple{pump::UnitRange{Int64}, outlet::UnitRange{Int64}, flow_boundary::UnitRange{Int64}, tabulated_rating_curve::UnitRange{Int64}, linear_resistance::UnitRange{Int64}, manning_resistance::UnitRange{Int64}, user_demand_inflow::UnitRange{Int64}, user_demand_outflow::UnitRange{Int64}, evaporation::UnitRange{Int64}, infiltration::UnitRange{Int64}, drainage::UnitRange{Int64}, surface_runoff::UnitRange{Int64}, precipitation::UnitRange{Int64}}, pid_integral::UnitRange{Int64}}}
  storage: [775.2364363043691, 775.2342825678, 572.6029998376725, 1130.004999999995]
  flow:
    pump: [4375.716386283238]
    outlet: Float64[]
    flow_boundary: [3162.2399999999943, 3162.2399999999943]
    tabulated_rating_curve: [426.97568814891827, 1280.9263593105832, 0.0]
    linear_resistance: [3614.272680166237, -676.5253671870286]
    manning_resistance: [2773.3327467955037]
    user_demand_inflow: Float64[]
    user_demand_outflow: Float64[]
    evaporation: [346.6708169001252, 346.67104958117375, 227.84697318966766, 364.4387061170138]
    infiltration: [0.0, 0.0, 0.0, 0.0]
    drainage: [0.0, 0.0, 0.0, 0.0]
    surface_runoff: [366.0, 366.0, 366.0, 366.0]
    precipitation: [366.0, 366.0, 366.0, 366.0]
  pid_integral: Float64[]
```

### The linear solve construction

What I like about the matrix construction in `build_J_inner!` is that it very explicitly reveals the structure of the linear solve.  

### PID control
The PID control values are no longer cached, they are computed on-demand in `formulate_pump_and_outlet_flow!`.

### Continuous control

- AD...
- The code now supports continuous control based on flow, it just has to be allowed in the input and tested

### Greedy UserDemand inflow

I think the greedy UserDemand inflow when allocation is not active can be cleaned up a bit: now on main the equal split logic lives both in the flow formulation and in the step limiter, but I think it can be done in a callback where the equal split is assigned to allocated so that other places in the code do not have to know whether it was done by the allocation algorithm or by the simple split.

### Latency

~~As discussed with @visr, precompiling Ribasim takes quite a while because of our custom state vector; a lot of what's precompiled from `OrdinaryDiffEq.jl` in their precompile workload is only valid for plain vector states. We could get around this by passing a plain vector state to the solver and only wrapping it for indexing convenience where needed.~~

Latency is still a concern. In this branch it helps both latency and performance to make `specialize = true` the default, see https://github.com/Deltares/Ribasim/issues/3170#issuecomment-5071042461.

### Allocation

Allocation is outside the scope of this PR, but I noticed some problems with `optimize_multi_objective!` while debugging:
- Infeasibility is only checked after all objectives have been optimized for. If one of them was infeasible, the subsequent goal programming constraints are uninformative garbage. So infeasibility should be checked after each `JuMP.optiomize!`, not just after all optimizations are done
- The route priority objective seems to be implemented in an odd way: not only does it not make use of `allocation_model.objectives` where all objectives should live, it's also optimized for for each other objective, not just once at the end.

### DiscreteControl

The only model that is still misbehaving on this branch is [`continuous_concentration_condition`](https://ribasim.org/reference/test-models.html#continuous_concentration_condition) (confusing name btw, as it doesn't use `ContinuousControl`). This is a typical 'flipper' model because the controlled `LinearResistance` directly influences the salinity in the Basin being conditioned on. By how `DiscreteControl` is currently set up (at a per solver timestep basis), how quickly this oscillation happens is fully dependent on the adaptive time stepper, which apparently behaves differently now than with only the flow states. To me the cleanest solution is https://github.com/Deltares/Ribasim/issues/2247.

### Convergence output
Make new issue to make convergence output optional and put it in separate file.


SouthEndMusic:
@visr I did some profiling with the `hws_transient` model you sent me:
- This branch (without any work on tolerances):
  - With backtracking: 2 min 42 s
  - No backtracking: 1 min 20 s

`water_balance!` and callbacks dominate at least at this scale, not at all the AD or linear solve. This indicates that a lot of performance can be gained from threading.

- Main (no backtracking):   1 min 56 s

visr:
That is promising! I'd leave backtracking and threading out of this PR, and mainly check the performance on the full LHM, which I just sent by email.

Although there main hardly runs, so we'd compare to the version of Lex' branchbthat we ran with.

SouthEndMusic:
There was backtracking in this branch because Lex added it, I removed it again. I did a quick experiment with running the `transient_hws` model ignoring the flow residual, but that didn't speed it up. I'll have a look at the full LHM

SouthEndMusic:
I've been looking at some LHM profiles. The runtime at the start is really inconsistent over simulated time so I'm not sure what a good benchmark is. I did notice that:
- Updating the Jacobian is negligible
- Linear solves take up about 15% of the time, where most of the time is spent on assembling the matrix. There is a simple optimization opportunity here, because most of the assembly time is from terms that only depend on the Jacobian, so if the Jacobian didn't refresh, those terms can be maintained
- `water_balance!` takes up about 50% of the time, very threadable
- Callbacks take up about 25% of the time, also threadable

SouthEndMusic:
Tasks left to do:
- [ ] Complete the technical documentation of the new implementation
- [ ] Complete the in-code documentation of the new implementation
- [ ] Add/modify tests specifically for the new implementation (e.g. regarding which solver algorithms to keep supporting)
- [ ] Write follow-up issues regarding certain findings during development

visr:
@copilot+claude-sonnet-5 resolve the merge conflicts in this pull request

.vscode/settings.json
```diff
@@ -69,6 +69,7 @@
     ],
     "julia.lint.run": true,
     "julia.numTestProcesses": 6,
+    "julia.NumThreads": 1,
```
visr: I guess this can go again now that multi threading is removed?
I am afraid this causes extra waiting time for instance if multiple dependencies need to be precompiled.

```suggestion
```

python/ribasim/ribasim/config.py
```diff
@@ -135,11 +135,14 @@ class Solver(ChildModel):
...
+        The absolute tolerance for adaptive timestepping (Optional, defaults to 1e-5)
     reltol : float
-        The relative tolerance for adaptive timestepping (Optional, defaults to 1e-7)
+        The relative tolerance for adaptive timestepping (Optional, defaults to 1e-6)
     maxiters : int
         The total number of linear iterations over the whole simulation. (Defaults to 1e9, only needs to be increased for extremely long simulations)
+    optimized_implicit_solve : bool
+        Whether to use the implicit solve optimized for Ribasim. Setting this to false bypasses much of the custom numerical code,
+        using a less optimized but potentially more robust algorithm. (Optional, defaults to true)
     sparse : bool
```
visr: Perhaps `optimized` is a bit too general a name, and can become confusing if we introduce more options in the future. How about something more specific like:

- `flow_and_storage_implicit_solve`
- `combined_implicit_solve`

python/ribasim/ribasim/config.py
```diff
@@ -151,7 +154,7 @@ class Solver(ChildModel):
     evaporate_mass : bool
         Whether mass is lost due to evaporation in water quality calculations. (Optional, defaults to true)
     specialize : bool
-        Trades initialization speed for simulation speed, useful for long-running simulations. (Optional, defaults to false)
+        Code generation option, true has best performance and initialization speed (Optional, defaults to true)
```
visr, line 160: Shall we make a follow-up issue to properly deprecate and `specialize`?
And I think we should also make clear there that `sparse` and `autodiff` only do something if the custom solve is not used.

docs/changelog.qmd
```diff
@@ -5,6 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
+- Solver setting `optimized_implicit_solve` (default `true`) to fall back to a less optimized but potentially more robust implicit solve. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+
+### Changed
+- Flows are integrated by the solver itself rather than reconstructed afterwards, so reported flow volumes close the water balance up to floating point precision. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Error control scales the residual of every state component by its change over the timestep, instead of by its magnitude. This avoids tolerances that loosen as cumulative states grow, or that depend on Basin size rather than on the dynamics. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Solver default `reltol` is now `1e-6` (was `1e-5`), and `specialize` now defaults to `true`. Setting `specialize = false` is no longer recommended and may be removed. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Python: Correct the documented and default `abstol` to `1e-5`, matching the core. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
```
visr: I want to keep the changelog brief and only add changes relevant for users. This looks like an important fix, but note that the Python defaults don't matter much, since defaults are not written to the TOML, leaving it up to the core defaults.

```suggestion
```

docs/changelog.qmd
```diff
@@ -5,6 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
+- The `convergence` column of the Basin results is now a genuine storage convergence measure, instead of being derived from the convergence of the evaporation and infiltration flows. The `convergence` column of the flow results remains flow based. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Both convergence columns are computed from the residuals accumulated over all Newton iterations in the saveat interval, instead of from only the final residual of each converged Newton solve. This gives a better indication of how much work the solver spent on a node or link. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Because the convergence output is derived from these residuals, the change in residual scaling described above also changes the reported convergence values. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
```
visr, line 18: These 3 bullets about the convergence output can be condensed into one bullet I think.

docs/changelog.qmd
```diff
@@ -5,6 +5,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
+- Cumulative flows and PID integral terms are now solver states alongside the Basin storages, solved as a mass matrix system with a custom Jacobian and linear solve. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+- Solver setting `optimized_implicit_solve` (default `true`) to fall back to a less optimized but potentially more robust implicit solve. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
+
+### Changed
+- Flows are integrated by the solver itself rather than reconstructed afterwards, so reported flow volumes close the water balance up to floating point precision. [#3150](https://github.com/Deltares/Ribasim/pull/3150)
```
visr, line 13: The reference point for the changelog is the last release v2026.1.2, not Lex' branch.

docs/reference/usage.qmd
```diff
@@ -18,7 +18,7 @@ Ribasim expects the GeoPackage database [database.gpkg](/reference/usage.qmd#sec
 
 The solver section in the configuration file is entirely optional, since we aim to use defaults that will generally work well.
 Common reasons to modify the solver settings are to adjust the calculation or result stepsizes: `dt`, and `saveat`.
-If your model does not converge, or your performance is lower than expected, it can help to adjust other solver settings as well.
+If your model does not converge, or your performance is lower than expected, it can help to adjust other solver settings as well. For example, setting `optimized_implicit_solve = false` will bypass much of the custom numerical code written for Ribasim, using a less optimized but potentially more robust algorithm.
```
visr: I'd leave this out here, having it below is enough, and 'potentially more robust' is hard to grasp for users I think.

```suggestion
If your model does not converge, or your performance is lower than expected, it can help to adjust other solver settings as well.
```

docs/reference/usage.qmd
```diff
@@ -409,7 +412,27 @@ ds
 
 The `link_id` value is the same as the `fid` written to the Link table, and can be used to directly look up the Link geometry.
 Flows from the "from" to the "to" node have a positive sign, and if the flow is reversed it will be negative.
-- The `convergence` is the scaled residual of the solver, giving an indication of which nodes converge the worst (are hardest to solve).
+The `convergence` indicates how hard the flow over this link was to solve compared to the other flows, see [convergence output](#sec-convergence).
+
+## Convergence output {#sec-convergence}
+
+Both the Basin and the flow results contain a `convergence` column, which indicates which parts of the model are hardest for the solver to solve.
```
visr, line 419:
```suggestion
Both the Basin and the flow results contain a `convergence` column, which indicates which parts of the model are hardest on the solver.
```

docs/dev/numerics.qmd
```diff
@@ -23,29 +23,29 @@ Ribasim builds on several packages from the [SciML](https://sciml.ai/) ecosystem
...
 
 The simulation is then advanced with `step!(integrator)` in a loop until `t_end`.
 
 # The right-hand side: `water_balance!`
 
-The RHS function `water_balance!` in `core/src/solve.jl` has the in-place signature `water_balance!(du, u, p, t)` expected by OrdinaryDiffEq. It writes the time derivatives of all cumulative flow states into the pre-allocated output vector `du` (the same length as the state vector `u`); no allocations or return value are involved. Its execution follows a fixed order:
+The RHS function `water_balance!` in `core/src/solve.jl` has the in-place signature `water_balance!(du, u, p, t)` expected by OrdinaryDiffEq. The state vector `u` holds the **Basin storages** and the **PID integral terms**; `water_balance!` writes their time derivatives — `dS/dt` per Basin and `d(integral)/dt` per PID control node — into the pre-allocated output vector `du`. No allocations or return value are involved. Its execution follows a fixed order:
```
visr, line 35: > The state vector `u` holds the **Basin storages** and the **PID integral terms**;

Also the flow states.

docs/dev/numerics.qmd
```diff
@@ -23,29 +23,29 @@ Ribasim builds on several packages from the [SciML](https://sciml.ai/) ecosystem
...
 6. **Continuous control** — evaluate [continuous control](/reference/node/continuous-control.qmd) rules that depend on current flows and levels
 7. **Controlled flows** — flows whose parameters are set by continuous control
-8. **PID control** — evaluate [PID control](/reference/node/pid-control.qmd) (integral/proportional/derivative)
+8. **PID control** — evaluate [PID control](/reference/node/pid-control.qmd), writing the error into `du.integral`
 9. **PID-controlled flows** — flows whose parameters are set by PID control
 
 ::: {.callout-note}
-The state vector $\mathbf{u}$ contains cumulative flows, not instantaneous flows or storages. See [why this formulation](/concept/equations.qmd#why-this-formulation) for the rationale.
+The state vector $\mathbf{u}$ contains the Basin storages and PID integral terms — *not* flows. Horizontal and vertical flows are diagnostic quantities computed from the current state inside `water_balance!`. Their time-integrated volumes (needed to report flows and to close the water balance) are reconstructed separately in a callback, see [flow integration](#sec-flow-integration).
```
visr, line 48: This still describes storage stages and reconstructed flows.

docs/dev/numerics.qmd
```diff
@@ -65,50 +65,25 @@ This split is driven by AD requirements — see [Jacobian computation](#jacobian
...
-# State reduction {#sec-state-reduction}
-
-Not every component of the state vector $\mathbf{u}$ influences every flow independently. Many flow states depend on basin levels, which in turn depend on basin storages. Ribasim exploits this by working with a **reduced state** $\mathbf{u}_\text{reduced}$ that contains only the basin storages and PID integral terms.
+## Damped Newton iteration
 
-The relationship is:
-$$
-\mathbf{u}_\text{reduced} = A \mathbf{u}
-$$
+The nonlinear system at each step is solved with `NLNewton`, configured with a custom backtracking line search (`Ribasim.BackTracking`, set in `algorithm` in `core/src/config.jl` and implemented via `relax!` in `core/src/util.jl`). After computing a Newton direction the line search scales it by a factor $\alpha \in (0, 1]$ that achieves sufficient decrease of the (scaled) residual. The full step $\alpha = 1$ is accepted whenever it does not increase the residual, which is the common case; backtracking only engages when the full step would overshoot. This keeps the solve robust on the smoothed-but-nonlinear flow relations without unnecessarily damping easy steps.
```
visr, line 72: Relaxation is not used right now.

docs/dev/numerics.qmd
```diff
@@ -65,50 +65,25 @@ This split is driven by AD requirements — see [Jacobian computation](#jacobian
...
 
-where $A$ is a linear operator implemented by `reduce_state!` in `core/src/util.jl`. All of its entries lie in $\{-1, 0, +1\}$, and each Basin row simply sums the cumulative flow states that change that Basin's storage:
+## The Jacobian
 
-- $+1$ for every horizontal flow state whose link enters the Basin,
-- $-1$ for every horizontal flow state whose link leaves the Basin,
-- $-1$ on the Basin's own evaporation and infiltration columns.
+The Jacobian $J$ is an $N \times N$ matrix where entry $J_{i,j}$ describes how a small change in state component $j$ affects the rate of change of component $i$. Here $N$ is the number of Basins plus the number of PID integral terms — the length of $\mathbf{u}$ assembled in `state_node_ids` (`core/src/util.jl`). The solver uses this sensitivity information to take stable, accurate steps through stiff regions.
```
visr, line 76: > number of Basins plus the number of PID integral terms

This section and below on Jacobian computation, AD and sparsity all need to be updated.

docs/dev/numerics.qmd
```diff
@@ -153,48 +125,38 @@ When `DifferentiationInterface.jacobian!` calls the RHS, it needs to know how to
 
 This distinction is a performance optimization (fewer allocations), but it introduces a subtle invariant: any cached value in a `Cache()` argument must be freshly computed during the AD call, because the copy starts uninitialized.
 
-## The `HalfLazyJacobian` operator
-
-Rather than materializing the full $N \times N$ Jacobian, Ribasim uses a custom `AbstractSciMLOperator` called `HalfLazyJacobian`. It stores:
+# Linear solves
```
visr, line 128: Needs to be updated, still mentions Simpson's rule.
Callbacks section below as well probably.

docs/dev/allocation.qmd
```diff
@@ -684,12 +684,12 @@ function linearize_connector_node!(
         )
 
         # Set coefficients for basin storage variables if applicable
-        if upstream_id.type == NodeType.Basin
+        if upstream_id.tis_basin
```
visr, line 687:
```suggestion
        if upstream_id.is_basin
```

docs/concept/numerics.qmd
```diff
@@ -2,6 +2,208 @@
...
+- TabulatedRatingCurve flow is non-negative;
+- Linear resistance flow is in the interval `[-max_flow, max_flow]`;
+- UserDemand inflow is bounded by the allocated amount;
+- Positive Basin forcings can be integrated exactly since they are piecewise constant.
+
+Correcting $Q$ this way makes it out of sync with $s$, so after these corrections $s$ is recomputed as $s = MQ + s_0$.
+
+<!-- ---
+title: "Numerical considerations"
+---
```
visr, line 205: Did you mean to comment out the rest here? Perhaps best to remove, and if something still needs to happen, make a follow up issue?

We have concept/numerics and dev/numerics. This one (concept) is updated but the dev one not. Perhaps we need to reconsider this split, because maintaining in two places may not be so convenient. We could move over missing parts from the dev numerics here and remove that?

docs/concept/core.qmd
```diff
@@ -129,7 +129,8 @@ If you're interested in using this experimental feature, please [contact us](/co
 :::
 
 Ribasim can calculate concentrations of conservative tracers (i.e. substances that are non-reactive).
-It does so by calculating the mass transports by flows for each timestep, in the `update_cumulative_flows!` callback.
+It does so by calculating the mass transports by flows for each timestep, in the `update_concentrations!` callback.
+It uses the per-step flow volumes computed in the `update_cumulative_flows!` callback (see [flow integration](/dev/numerics.qmd#sec-flow-integration)), so the transported mass is consistent with the reported flows and the change in Basin storage.
```
visr, line 133: This is outdated.

docs/concept/equations.qmd
```diff
@@ -3,42 +3,148 @@ title: "Equations"
...
-	u_b^\text{forcing}(t) = \int_{t_0}^t q_b^\text{forcing}\text{d}t' \quad \forall b \in B.
+	E_r = \frac{E_b}{r_\text{in} + r_\text{out}}.
 $$
 
-Because of this definition, the initial conditions of all states are simple:
+In the [flow correction](numerics.qmd#correcting-the-solver) the new storages are computed directly from the cumulative flows, so in theory the water balance should always be closed up to floating point precision.
+
+<!--
+# Formal model description
+In this section we give a formal description of the problem that is solved by Ribasim. The problem is of the form
```
visr, line 113: Rather than commenting out perhaps also delete and make a follow up issue, or perhaps some parts are still relevant, like the performance section.

core/test/io_test.jl
```diff
@@ -93,23 +93,20 @@ end
...
     path = results_path(config, RESULTS_FILENAME.flow)
     @test isfile(path)
     NCDataset(path) do ds
-        @test "convergence" in keys(ds)
         @test "flow_rate" in keys(ds)
+        @test "convergence" in keys(ds)
         @test ds.attrib["ribasim_version"] == RIBASIM_VERSION
         convergence = ds["convergence"][:]
-        @test all(isfinite, skipmissing(convergence))
+        @test all(x -> ismissing(x) || isfinite(coalesce(x, 0)), convergence)
```
visr, line 109: Is `missing` expected for the non-state flows?

core/test/io_test.jl
```diff
@@ -93,23 +93,20 @@ end
     path = results_path(config, RESULTS_FILENAME.basin)
     @test isfile(path)
     NCDataset(path) do ds
-        @test "convergence" in keys(ds)
         @test "level" in keys(ds)
         @test "storage" in keys(ds)
         @test ds.attrib["ribasim_version"] == RIBASIM_VERSION
-        convergence = ds["convergence"][:]
-        @test all(isfinite, convergence)
     end
```
visr, line 99: I think this change can now be reverted right?

```suggestion
    NCDataset(path) do ds
        @test "convergence" in keys(ds)
        @test "level" in keys(ds)
        @test "storage" in keys(ds)
        @test ds.attrib["ribasim_version"] == RIBASIM_VERSION
        convergence = ds["convergence"][:]
        @test all(isfinite, convergence)
    end
```

core/test/io_test.jl
```diff
@@ -176,9 +173,7 @@ end
         @test "time" in keys(ds)
         @test "link_id" in keys(ds)
         @test "flow_rate" in keys(ds)
-        @test "convergence" in keys(ds)
         @test ds["flow_rate"].attrib["units"] == "m3 s-1"
-        @test ds["convergence"].attrib["units"] == "1"
         ntime = length(ds["time"])
```
visr, line 177:
```suggestion
        @test "flow_rate" in keys(ds)
        @test "convergence" in keys(ds)
        @test ds["flow_rate"].attrib["units"] == "m3 s-1"
        @test ds["convergence"].attrib["units"] == "1"
        ntime = length(ds["time"])
```

core/src/util.jl
```diff
@@ -40,7 +40,9 @@ end
 Compute the level of a basin given its storage.
 """
 function get_level_from_storage(basin::Basin, state_idx::Int, storage::T)::T where {T}
-    return basin.storage_to_level[state_idx](storage)
+    # Clamp storage positive, since Rosenbrock methods can overshoot to negative
+    s = ifelse(storage > zero(T), storage, zero(T))
+    return basin.storage_to_level[state_idx](s)
```
visr: This is only called in allocation, where this should be positive.

```suggestion
    return basin.storage_to_level[state_idx](storage)
```

Perhaps we can use a single `get_level`.

core/src/util.jl
```diff
@@ -671,6 +625,8 @@ function basin_areas(basin::Basin, state_idx::Int)
     return basin.level_to_area[state_idx].u
 end
 
+get_fixed_area(basin::Basin, state_idx::Int) = basin_areas(basin, state_idx)[end]
```
visr, line 620:
```suggestion
"Get the area at the top of the profile"
get_fixed_area(basin::Basin, state_idx::Int) = basin_areas(basin, state_idx)[end]
```

core/src/util.jl
```diff
@@ -685,242 +641,95 @@ function relaxed_root(x, threshold)
...
+        n_user_demand_inflow,
+        n_user_demand_outflow,
+        n_basin, # evaporation,
+        n_basin, # infiltration,
+        n_basin, # drainage,
+        n_basin, # surface_runoff,
+        n_basin, # precipitation
+    ]
+    ns_flow_cumsum = pushfirst!(cumsum(ns_flow), 0)
+    n_flow = sum(ns_flow)
```
visr: Unused.

```suggestion
```

core/src/util.jl
```diff
@@ -685,242 +641,95 @@ function relaxed_root(x, threshold)
...
+function build_state_vector(p_independent::ParametersIndependent)
+    (; u_prev_saveat, basin) = p_independent
+    u = zero(u_prev_saveat)
+    u.storage .= basin.storage0
+    return getdata(u)
 end
 
 """
 Check whether any storages are negative given the state u.
+Storage states are directly in u.storage.
```
visr:
```suggestion
Check whether any storages are negative in the state u.
```

core/src/util.jl
```diff
@@ -685,242 +641,95 @@ function relaxed_root(x, threshold)
...
 """
-function isoutofdomain(u, p, t)
-    (; current_storage) = p.state_and_time_dependent_cache
-    (; u_reduced) = p.p_independent
-    reduce_state!(u_reduced, u, p.p_independent)
-    formulate_storages!(u_reduced, p, t)
-    return any(<(0), current_storage)
+function isoutofdomain(u_raw, p, t)
+    u = CVector(u_raw, p.p_independent.state_ranges)
+    return any(<(0), u.storage)
```
visr: Unless this automated latency evaluation is wrong, it appeared to me that CVector actually does not increase latency much, meaning we don't have to unwrap it: https://github.com/Deltares/Ribasim/issues/3170#issuecomment-5071042461

core/src/cvectors.jl
```diff
@@ -0,0 +1,170 @@
...
+
+# Recursively compute the flat range covered by axes
+_flat_range(loc::Integer) = loc:loc
+_flat_range(loc::AbstractUnitRange{<:Integer}) = loc
+_flat_range(loc::NamedTuple) =
+    minimum(first ∘ _flat_range, values(loc)):maximum(last ∘ _flat_range, values(loc))
+
+_component_length(loc::Integer) = 1
+_component_length(loc::AbstractUnitRange{<:Integer}) = length(loc)
+_component_length(loc::NamedTuple) = sum(_component_length, values(loc))
```
visr: I don't really want to start using the Python convention of leading underscores for private functions. They are all private unless marked public.

core/src/bmi.jl
```diff
@@ -39,6 +39,7 @@ function BMI.update(model::Model)::Nothing
 end
 
 function BMI.update_until(model::Model, time::Float64)::Nothing
+    model.integrator.derivative_discontinuity = true
```
visr: This is not the public interface.
The public one is slightly uglier, but probably more future proof.

[derivative_discontinuity!(integrator, true)](https://docs.sciml.ai/DiffEqDocs/stable/basics/integrator/#SciMLBase.derivative_discontinuity!)

core/src/callback.jl
```diff
@@ -571,24 +378,21 @@ function save_solver_stats(u, t, integrator)
...
 
-    errors = false
-    for id in basin.node_id
-        if state_and_time_dependent_cache.current_storage[id.idx] < 0
-            @error "Negative storage detected in $id"
-            errors = true
-        end
-    end
+    # Do this here so the cache is up to date for subsequent callbacks
+    set_current_basin_properties!(u, p, t)
```
visr: Is this callback still needed with `isoutofdomain` in use as well?
Perhaps only the `set_current_basi_properties!` call needs to be preserved, possibly in the next callback.

Then we can also remove `has_negative_storage` from the parameters.

core/src/parameter.jl
```diff
@@ -677,17 +668,13 @@ end
 
 """
 node_id: node ID of the FlowBoundary node
-outflow_link: The outgoing flow link metadata
-cumulative_flow: The exactly integrated cumulative boundary flow since the start of the simulation
-cumulative_flow_saveat: The exactly integrated cumulative boundary flow since the last saveat
+outflow_link: The outgoing flow link metadatation
```
visr:
```suggestion
outflow_link: The outgoing flow link metadata
```

core/src/parameter.jl
```diff
@@ -930,21 +861,25 @@ record: Namedtuple with discrete control information for results
         truth_state = String[],
         control_state = String[],
     )
+    extend_record_lock::ReentrantLock = ReentrantLock()
```
visr, line 864: This needs to be added to the docstring.
I assume this is only needed if we decide to do DiscreteControl in parallel?

core/src/parameter.jl
```diff
@@ -1119,14 +1059,11 @@ const ModelGraph = MetaGraph{
...
-- `new_state_and_time_dependent_cache`: Whether the `t` and/or `u_reduced` with which `water_balance!` are called are
-   considered new, and thus whether caches that (only) depend on `u_reduced` must be updated
-- `tprev`: The previous `t` before the latest time step
 """
 @kwdef mutable struct ParametersMutable
     new_time_dependent_cache::Bool = true
-    new_state_and_time_dependent_cache::Bool = true
-    tprev::Float64 = 0.0
+    refresh_jac::Bool = true
+    ad_active::Bool = false
```
visr, line 1071: The new fields need to be documented.

python/ribasim/ribasim/config.py
Huite: `reduced_implicit_solve` might also be an option. Or `custom_linear_solve`?

Maybe just something with custom?

That feels like the most direct description, since I'd say it's probably the biggest part that is getting overloaded.

core/test/io_test.jl
SouthEndMusic: In this formulation there currently are no non-state flows. `missing` is expected when no residuals were collected

PR #3176, opened 2026-07-27
Put Julia directly in the Pixi environment

visr:
## Summary

- install Julia 1.12.6 directly from the `julia-forge` Pixi channel on all supported platforms
- explicitly pin Julia to `julia-forge` so future solves cannot select another channel
- remove redundant JuliaUp preparation from CI workflows
- let VS Code use Julia from the activated Pixi environment

This follows the approach from Deltares/Wflow.jl#1017.

Linux ARM64 packages were enabled upstream in https://github.com/wolfv/julia-forge/pull/7, allowing Ribasim to use one Julia dependency across all four platforms without a JuliaUp fallback.

## Validation

- `pixi search julia -c https://prefix.dev/julia-forge --platform linux-aarch64`
- `pixi lock` across all four supported platforms
- `pixi run --locked julia --version` (`1.12.6`)
- verified every locked Julia package comes from `julia-forge`
- `pixi run prek run --files pixi.toml pixi.lock`
- `git diff --check`

### DIFF
--- a/.github/workflows/core_testmodels.yml
+++ b/.github/workflows/core_testmodels.yml
@@ -36,8 +36,6 @@ jobs:
       - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
         with:
           pixi-version: "latest"
-      - name: Prepare pixi
-        run: pixi run install-ci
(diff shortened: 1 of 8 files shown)


SouthEndMusic (approved):
LGTM

PR #3180, opened 2026-07-30
Accelerate initializing large models

SouthEndMusic:
These changes accelerate initializing the LHM from ~300 s to ~30 s.

### DIFF
--- a/core/src/read.jl
+++ b/core/src/read.jl
@@ -123,7 +123,7 @@ function get_parameter_value(
             @assert !isnothing(data)
             data_group = first(
                 IterTools.groupby(
-                    i -> ntuple(j -> getfield(data, take_first[j])[i], length(take_first)),
+                    i -> ntuple(j -> getproperty(data, take_first[j])[i], length(take_first)),
                     data_group,
                 ),
             )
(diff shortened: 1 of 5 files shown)


core/src/read.jl
```diff
@@ -142,6 +147,39 @@ function get_parameter_value(
...
+        lookup[first(group).compound_variable_id] = StructVector(group)
+    end
+    return lookup
+end
+
+"""Collect StructVector groups by compound_variable_id while preserving input order."""
+function pregroup_by_compound_variable_id(sv::StructVector)
+    isempty(sv) && return StructVector[]
+    return StructVector.(IterTools.groupby(row -> row.compound_variable_id, sv))
+end
```
visr, line 181: Would be good to add some return type annotations and unit tests for these new functions, so it is easier to catch mistakes in the more complicated functions where these are used.

visr (approved):
Nice speedup!

Huite:
Seems like it may be worthwhile to document this in an issue for posterity's sake?

Or a note in the developer documentation? (I'm sure Claude will generate an acceptable paragraph with some nudging.)

PR #3181, opened 2026-07-30
Pin ObjectFile to v0.5.0

visr:
To see if this fixes the Linux binaries.

libribasim.so depends on libjulia-internal.so.1.12. The packaged bin/ribasim/lib/julia/ directory is empty.


visr:
See https://github.com/Deltares/Ribasim/pull/3182

PR #3184, opened 2026-07-31
Add an optional update interval to DiscreteControl nodes

### LINKED ISSUE #2247: Add an optional time interval to `DiscreteControl` nodes

visr:
Fixes #2247. The optional interval column on DiscreteControl / condition makes
the control logic run only at multiples of that interval instead of after every
solver timestep. Solver tstops are added at those times.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


core/src/callback.jl
```diff
@@ -408,17 +408,31 @@ Apply the discrete control logic. There's somewhat of a complex structure:
...
     u = get_u(integrator)
 
     # Loop over the discrete control nodes to determine their truth state
     # and detect possible control state changes
     for idx in eachindex(node_id)
+        interval = update_interval[idx]
+        if !iszero(interval)
+            (t < t_next_update[idx]) && continue
+            t_next_update[idx] = (floor(t / interval) + 1) * interval
+        end
```
SouthEndMusic, line 434: This logic looks sound to me, though it might be a bit cleaner to store an integer `n` such that `n * interval = t_next_update` instead of `t_next_update` itself.

SouthEndMusic:
Close as not planned for now

PR #3191, opened 2026-08-06
Remove storage state again

### LINKED ISSUE #3189: Keep flow formulation, improve performance
A brief history:
- The current formulation of the physical layer on main, in terms of cumulative flow states, was deemed to scale too poorly in the number of states for the lhm, making it too slow
- Lex made https://github.com/Deltares/Ribasim/pull/3039, going back to the storage state formulation which has fewer states, scaling better. However, this formulation has fundamental water balance problems, which was the reason we went for the cumulative flow formulation in the first place
- Now we have https://github.com/Deltares/Ribasim/pull/3150 combining storage and flow states with the hope to combine high performance and a closing water balance.

It however turns out that this last approach has a few problems of its own:
- With the combination of storage and flow states error control is hard to configure, because the storages and cumulative flow residuals need to be somehow combined into a single scalar error
- The performance is not as good as Lex's storage formulation on the lhm
- The water balance seems to be closed only up to solver tolerance, instead of up to floating point precision by construction as in the formulation with only cumulative flows
- Handling the storage term in the solve is a  only a hassle
- The mass matrix adds unnecessary complexity

Because of all this, I propose to keep the cumulative flow formulation, and to use the good parts from #3150. This is motivated by:
- On main the custom solve implementation is very inefficient; ~1/3 of the time is spent on the Jacobian and ~1/3 of the time is spent on *formulating* the reduced linear system. On #3150 the Jacobian time is negligible and formulating the linear system is implemented much more efficiently and cleanly. That might not get us at the same performance that https://github.com/Deltares/Ribasim/pull/3039 achieved, but the remaining time is mostly spent on discrete control, computing levels from storages and formulating flows, which are all fully parallelizable.
- Keep the residual `abstol` scaling with $|u_1 - u_0|$ instead of $\max\\{|u_1|,|u_0|\\}$, getting rid of the need for the adaptive `abstol` hack (but keeping the requirement to recalibrate the default tolerances).
- Keep the bottleneck logging based on accumulated nonlinear solve residuals instead of based on only the residual after convergence

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3189.

This PR removes the storage states again relative to https://github.com/Deltares/Ribasim/pull/3150, but keeps the improvements to the Jacobian evaluation and linear solve.


SouthEndMusic:
### Interpretable tolerances and residuals in a cumulative flow formulation

Here I outline a way to define the tolerances and residuals in a way that I think is hydrologically both interpretable and relevant.

In the cumulative flow formulation the nonlinear solver error vector has values $\tilde{Q}_j$, which is per flow an error in the transported volume over the timestep ($\text{m}^3$). As far as I know, hydrologist care the most about levels, so we have to translate $\tilde{Q}_j$ to a level rate error ($\text{m} \text{s}^{-1}$). We can do that by computing

$$
\tilde{r}_j := \frac{\tilde{Q}_j}{\Delta t A_{j,\min}},
$$

where $A_{j,\min}$ is the minimum of:
- The upstream Basin area at the proposed new state;
- The downstream Basin area at the proposed new state;
- The upstream Basin area at the last accepted $t$;
- The downstream Basin area at the last accepted t.

That way, $A_{j,\min}$ represents a good estimation of the maximum level sensitivity of the basins affected by $\tilde{Q}_j$ to storage change.  It must be noted that $\tilde{r}_j$ is a level rate error per flow, not per Basin. 

We can then Define some $r_\text{tol} > 0$, which is the maximum global level rate error we allow. Then we can define the local scaled error

$$
\frac{\tilde{r}_j}{r_\text{tol}} = \frac{\tilde{Q}_j}{r_\text{tol}\Delta t A_{j,\min}},
$$

which is a quantity that we want to be $< 1$ in absolute value for all flows. Note that this quantity can get large as the area gets small, so we can also similarly add an absolute error term $\Delta Q_\text{tol}$, which yields

$$
	\frac{\tilde{Q}_j}{\Delta Q_\text{tol} + r_\text{tol} \Delta t A_{j,\min}}.
$$

The standard RMS internal norm applied to this quantity for all flows gives then the scalar error

$$
\text{EEst}_\text{RMS} = \sqrt{\frac{1}{m}\sum_{j=1}^m \left(\frac{\tilde{Q}_j}{\Delta Q_\text{tol} + r_\text{tol} \Delta t A_{j,\min}}\right)^2},
$$

which actually does allow for the individual terms to be $> 1$ in absolute value for $\text{EEest}_\text{RMS} < 1$, so a good alternative might be the Inf norm:

$$
	\text{Eest}_\text{Inf} = \max_{j = 1, \ldots, m} \left|\frac{\tilde{Q}_j}{\Delta Q_\text{tol} + r_\text{tol} \Delta t A_{j,\min}}\right|.
$$

This error formulation also doesn't suffer from cumulative terms growing without bound. The flow to storage translation might not be very intuitive though.

@visr @Huite what tolerances should I try this with?

PR #3206, opened 2026-08-19
Allocation fixes

### LINKED ISSUE #3200: Route priority optimization happens at the wrong moment
As far as I know, route priority optimization should be the last goal to optimize for, only using up any degrees of freedom that are left after optimizing for the demands. However, that is not how it is implemented now, if we look here:

https://github.com/Deltares/Ribasim/blob/dee0ef83a0def82cb4059291a36d832ace6b50cc/core/src/allocation_optim.jl#L802-L838

we see that route priorities are optimized for after every other goal, which most importantly is too early, but is also redundant. This means that later priorities potentially get allocated less than they should get.

### LINKED ISSUE #3207: Allocation optimization termination status aren't handled properly
There are 2 ways in which the allocation optimization termination status are not handled properly:
- The termination status is only checked after all goals have been optimized for, but if optimization failed for some goal the rest of the optimizations are garbage. Therefore we should check the termination status after every individual optimization.
- We only check for the termination status `INFEASIBLE`, but other options can also indicate poor results

### LINKED ISSUE #3208: Allocation controlled TabulatedRatingCurve should fall back to uncontrolled behavior is allocation is not active
This is a bug found by @Fati-Mon; allocation controlled TabulatedRatingCurve nodes now always have zero flow when allocation is not active.

### LINKED ISSUE #3209: Improve/fix variable scaling in allocation
Inspecting the code, I can see at least 2 ways in which the variable scaling can be improved:
1. Currently the scaling factors are fixed at the start of the simulation. They should however be updated at the start of the allocation run based on the state of the physical layer and the allocation timestep.
2. It seems that some variables (like error terms) are never scaled at all

SouthEndMusic:
- Fixes when route priority optimization is done; after all other objectives as intended. I refactored the allocation objectives a bit to do this. Fixes https://github.com/Deltares/Ribasim/issues/3200
- Allocation now checks the optimization termination status after every optimization instead of only after the last one: fixes https://github.com/Deltares/Ribasim/issues/3207
- Fixes naming inconsistency of `route_priority` versus `default_route_priority`, I updated in to the latter in all relevant places
- Fixes a bug where allocation controlled TRC has zero flow when allocation is off: fixes https://github.com/Deltares/Ribasim/issues/3208
- Improve scaling (it turns out there weren't bugs here, but with these changes the `allocation_training` model runs again): fixes https://github.com/Deltares/Ribasim/issues/3209
- I also found a bug where the estimation of the storage change over the allocation timestep per Basin was taken as the storage rate at the start, instead of the storage rate at the start multiplied by the allocation timestep

I also changed the goal programming formulation a bit: after each goal an inequality constraint is added instead of an equality constraint (inequality in the direction of better results), which might be a bit easier on the solver.


core/test/docs.toml
```diff
@@ -29,7 +29,7 @@ dt = 86400                  # optional, remove for adaptive time stepping
 dtmin = 3600.0              # optional, 3600.0
 reltol_linearization = 0.01 # optional, default 0.01
 
-[allocation.route_priority]
+[allocation.default_route_priority]
```
visr: I'm not sure this is worth it. Allocation is still experimental, so I guess a breaking change is fine, but we'd at least have to put it in the changelog explicitly under "changed".

This can clarify that these are fallback values if specific nodes don't override them. But the term default can still be confusing, considering these defaults have defaults as well. The usage docs is already quite clear:

> If you wish to set the route priority for specific nodes rather than a fallback per node type, these can be set in the Node table

core/src/solve.jl
```diff
@@ -511,7 +511,7 @@ function formulate_flow!(
         h_b = get_level(p, outflow_id, t)
 
         q_h = tabulated_rating_curve_flow(tabulated_rating_curve, id, h_a, h_b, p, t)
-        q = if tabulated_rating_curve.allocation_controlled[node_idx]
+        q = if allocation_active && tabulated_rating_curve.allocation_controlled[node_idx]
```
visr: Did you also consider always keeping `allocation_controlled` false if allocation is not active? Then you only need to check during initialization, and we cannot forget this later on.

core/src/model.jl
```diff
@@ -309,21 +309,11 @@ function compute_and_set_adaptive_Δt!(model, saveat, tspan_end)::Float64
 
     water_balance!(du, u, p, t)
 
-    Δt = Inf
+    Δt = min(time_to_next_saveat(t, saveat, tspan_end), tspan_end - t)
```
visr:
```suggestion
    Δt = time_to_next_saveat(t, saveat, tspan_end)
```

`time_to_next_saveat` already does this `min`, hence you need to pass `tspan_end`.

core/src/parameter.jl
```diff
@@ -191,31 +191,35 @@ const IndexLookup = ConstantInterpolation{
...
-    objective_expressions_all::Vector{JuMP.AffExpr} = JuMP.AffExpr[]
-    objective_metadata::Vector{AllocationObjectiveMetadata} = AllocationObjectiveMetadata[]
+function Base.show(io::IO, objective::AllocationObjective)
+    (; type, demand_priority) = objective
+    out = "AllocationObjective of type $type"
+    if startswith(string(type), "demand")
+        out *= " with demand priority $demand_priority"
+    end
+    return print(io, out)
 end
```
visr:
```suggestion
function Base.show(io::IO, objective::AllocationObjective)::Nothing
    (; type, demand_priority) = objective
    print(io, "AllocationObjective of type $type")
    if startswith(string(type), "demand")
        print(io, " with demand priority $demand_priority")
    end
    return nothing
end
```

Nit, but I think show should return nothing, and string building with `*=` allocates new strings each time (not that that matters here).

core/src/parameter.jl
```diff
@@ -191,31 +191,35 @@ const IndexLookup = ConstantInterpolation{
...
 
+"""
+Factors with which the variables in the allocation optimization problem are scaled.
+This doesn't change the results, it just makes the problem easier for the optimizer.
+"""
 @kwdef mutable struct ScalingFactors
     flow::Float64 = 1.0e3
     storage::Float64 = 1.0e6
+    mean_half_storage::Float64 = 0.0
 end
```
visr, line 223: The dev docs do have info per field, would be nice to have that here as well.

```suggestion
"""
Factors with which the variables in the allocation optimization problem are scaled.
This doesn't change the results, it just makes the problem easier for the optimizer.
"""
@kwdef mutable struct ScalingFactors
    flow::Float64 = 1.0e3  # Typical storage value (m³)
    storage::Float64 = 1.0e6  # Typical flow rate (m³/s)
    mean_half_storage::Float64 = 0.0  # Initialization-time reference storage (m³)
end
```

core/src/parameter.jl
```diff
@@ -244,27 +248,29 @@ Store information for a subnetwork used for allocation.
...
-Δt_allocation: The time interval between consecutive allocation solves
 Δt_since_last_record: Time elapsed since the last saveat-aligned LP solve
     (i.e., since the last call that pushed records and reset cumulative_supplied_volume).
     Updated after every LP solve and reset to 0 when records are emitted;
     used to divide cumulative_supplied_volume into a rate for the records.
 has_demand_priority: Per demand priority in the whole model whether a demand of this priority is present in this
     subnetwork
 objectives: The objectives (goals) in the order in which they will be optimized for
+explicit_positive_forcing_volume: The computable cumulative forcing over the coming allocation timestep
+implicit_negative_forcing_volume: The predicted cumuative forcing over the coming allocation timestep (assuming no reduction)
```
visr:
```suggestion
implicit_negative_forcing_volume: The predicted cumulative forcing over the coming allocation timestep (assuming no reduction)
```

core/src/parameter.jl
```diff
@@ -316,6 +321,7 @@ allocation_models: The allocation models for the primary network and subnetworks
 primary_network_connections: (from_id: pump or outlet in the primary network, to_id: node in the subnetwork, generally a basin)
     per subnetwork
 demand_priorities_all: All used demand priority values from all subnetworks
+dt_allocation: Used as the timestep if the timestepping is not adaptive
 record_demand: A record of demands and allocated flows for nodes that have these
```
visr, line 326:
```suggestion
dt_allocation: Used as the timestep if the timestepping is not adaptive
adaptive: True for adaptive timestepping, false for fixed timestep of dt_allocation
record_demand: A record of demands and allocated flows for nodes that have these
```

core/src/parameter.jl
```diff
@@ -327,6 +333,8 @@ record_control: A record of all flow rates assigned to pumps and outlets by allo
     primary_network_connections::OrderedDict{Int32, Vector{Tuple{NodeID, NodeID}}} =
         OrderedDict()
     demand_priorities_all::Vector{Int32} = []
+    dt_allocation::Float64 = 86400.0
+    adaptive::Bool = false
```
visr: Why is the default false here when the overall default is adaptive allocation?

core/src/model.jl
```diff
@@ -309,21 +309,11 @@ function compute_and_set_adaptive_Δt!(model, saveat, tspan_end)::Float64
...
     for am in allocation.allocation_models
         Δt_sub = compute_adaptive_Δt(am, p, du, t, config.allocation)
-        am.Δt_allocation = Δt_sub
         Δt = min(Δt, Δt_sub)
```
visr: The old AllocationModel mutable struct had the `Δt_allocation` as a field, and you removed it. I'm not against, but why did you do that? Because I see in `allocation_optim.jl` that this gets passed around a lot, and all these functions already get `AllocationModel`.

In any case the function name `compute_and_set_adaptive_Δt!` is not correct anymore (remove `_and_set`), and the return type in the docstring was already wrong, good to fix that as well.

core/src/allocation_optim.jl
```diff
@@ -789,90 +802,108 @@ function warm_start!(allocation_model::AllocationModel, integrator::DEIntegrator
...
         primary_network_connections = [],
     )::Nothing
-    (; problem, objectives, temporary_constraints, route_priority_expression) =
-        secondary_model
+    (; problem, objectives, temporary_constraints) =
+        model
 
     # Lexicographic goal programming: optimize objectives in sequence
-    # After optimizing objective i, add constraint: obj_i ≤ optimal_i + ε
+    # After optimizing objective i, add constraint: obj_i = optimal_i
```
visr: Should this not be

```suggestion
    # After optimizing objective i, add constraint: obj_i ≤ optimal_i
```

core/src/allocation_optim.jl
```diff
@@ -789,90 +802,108 @@ function warm_start!(allocation_model::AllocationModel, integrator::DEIntegrator
...
-
-    optimize_multi_objective!(allocation_model)
+function parse_termination_status(
+        model::AllocationModel,
+        objective::AllocationObjective,
+        expression::JuMP.AffExpr,
+        latest_constraint,
+        config::Config,
+        t::Number
+    )
```
visr:
```suggestion
    )::Nothing
```

core/src/allocation_optim.jl
```diff
@@ -789,90 +802,108 @@ function warm_start!(allocation_model::AllocationModel, integrator::DEIntegrator
...
     return nothing
 end
 
-function optimize!(allocation_model::AllocationModel, model)::Nothing
-    (; config, integrator) = model
-    (; t) = integrator
-    (; problem, subnetwork_id) = allocation_model
-
-    optimize_multi_objective!(allocation_model)
+function parse_termination_status(
```
visr, line 860: Could use a oneliner docstring. Perhaps `parse` in the function name should be replaced with `check`, since it doesn't do anything when it is optimal?

core/src/parameter.jl
visr: I am not sure that having a separate `adaptive` and `dt_allocation` field here is a good idea.

It makes it easy to use `dt_allocation` when it shouldn't be used, because `adaptive`.
One field, a `Union{Float64, Nothing}` would solve that. A downside is that checking for `nothing` is not so obvious when reading. We could get fancy:

```julia
struct Timestep
    value::Union{Float64,Nothing}
end

is_adaptive(t::Timestep)::Bool = isnothing(t.value)

function dt(t::Timestep)::Float64
    is_adaptive(t) && throw(ArgumentError("timestep is adaptive, no fixed value"))
    return t.value
end
```

core/src/allocation_optim.jl
```diff
@@ -1229,18 +1282,20 @@ function update_allocation!(model, Δt = 0.0; record::Bool = true)::Nothing
 
     du = get_du(integrator)
     water_balance!(du, u, p, t)
+    Δt = isnothing(Δt) ? allocation.dt_allocation : Δt
```
visr: I don't really understand this logic. If `isnothing(Δt)`, don't we use the default adaptive timestepper and should we avoid using `allocation.dt_allocation`? See also my other comment about perhaps using a `struct Timestep` for clarity.

visr (commented):
Overall this reads like a good cleanup, on top of providing several fixes. I have several comments for your consideration.

core/test/docs.toml
SouthEndMusic: `default_route_priority` is the name in `config.py`:

https://github.com/Deltares/Ribasim/blob/1423077ff2c889e25070401b0d39777156e6a791/python/ribasim/ribasim/config.py#L106

So there is an inconsistency on main anyway, because the kernel expects `route_priority`:

https://github.com/Deltares/Ribasim/blob/1423077ff2c889e25070401b0d39777156e6a791/core/src/config.jl#L229

I only came across this problem when trying to change the values in Ribasim Python.

If you think `route_priority` is better than `default_route_priority`, let me know.

core/src/parameter.jl
SouthEndMusic: I refactored it with `dt_allocation::Union{Float64, Nothing}`

core/src/model.jl
SouthEndMusic: I didn't like that each allocation model has the allocation timestep as a field as it implies that the different allocation models run at different timesteps.

core/src/allocation_optim.jl
SouthEndMusic: Ah you are right. I have been experimenting with this quite a bit and in the end I didn't update the docstring properly.

visr (approved):

PR #3214, opened 2026-08-24
Don't set external filepath in TOML `if df is None`

visr:
If I have a model with this external file:

```toml
[basin]
time = "basin_time.nc"
```

And I then remove that table in Python:

```python
model.basin.time.df = None
```

It will leave the reference in the TOML, but not write the NetCDF file. That crashes the core with:

```
ERROR: NetCDF error: Opening path input/basin_time.nc: No such file or directory (NetCDF error code: 2)
```

What I want is to remove the TOML reference when I clear the table in Python.

Ran into this in Ribasim-NL.

### DIFF
--- a/docs/reference/usage.qmd
+++ b/docs/reference/usage.qmd
@@ -233,6 +233,9 @@ To revert a table back to being stored in the GeoPackage database instead of an 
 model.basin.profile.filepath = None
 ```
 
+If a table is absent, its external filepath is omitted from the TOML.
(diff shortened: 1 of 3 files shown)


evetion (approved):
Would be good to document our expected behaviour here. The midcase is an empty table, as removing it is a bit harsh (do we document that?).

I remember playing around these settings, not sure why I didn't this fix in (as we have this check in other places).

visr:
Yeah good to document this. I added de31d33793ad6004ff341ea1fe7915af5ca2d755

PR #3218, opened 2026-08-26
Refactor to single main loop and allocation call

### LINKED ISSUE #3158: Facilitate running allocation (and irrigation) via a callback
There are several ways to run Ribasim:

- Stand-alone or via the BMI
- With allocation active or inactive
- With irrigation active or inactive

Ideally we want to support every combination of these possibilities in the same way instead of maintaining custom code for each one. The best way to do this is to run both allocation and irrigation using callbacks, as callbacks are run regardless of how Ribasim is run (stand-alone or via BMI). However, that is currently not possible because of an order of operations problem. When running Ribasim coupled, wat we want to be able to do is:

1. Ribasim gets demands via the BMI
2. Ribasim runs allocation
3. Ribasim runs the physical layer

When running allocation via a callback, this always happens directly after the physical layer update, so it's not possible for the coupler to update demands in between. To solve this, we need the possibility in `OrdinaryDiffEq.jl` to have callbacks that run directly before the ODE step instead of directly after it. I made an upstream issue for this: https://github.com/SciML/OrdinaryDiffEq.jl/issues/3977.

### LINKED ISSUE #3213: Support adaptive allocation timestep when calling Ribasim from BMI
Allocation is called in different ways when running Ribasim stand-alone versus via the BMI, which is already addressed in https://github.com/Deltares/Ribasim/issues/3158. One of the ways in which these differ is that the BMI case doesn't support adaptive time stepping for allocation. If you tried to do that it silently defaulted to a fixed timestep of 1 day, which will be re placed by an error with https://github.com/Deltares/Ribasim/pull/3206. This issue is mainly to raise awareness of this consequence of having different ways to call allocation, and might be fixed by https://github.com/Deltares/Ribasim/issues/3158.

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3213
Fixes https://github.com/Deltares/Ribasim/issues/3158

The approach of overloading `OrdinaryDiffEqCore.loopheader!` in combination with `invoke` is not the prettiest, but I think unifying the way allocation is called in different scenarios is worth it.

Now that we support the adaptive timestep for allocation when calling Ribasim from the BMI, we have to be careful about synchronization. It would make sense from the side of a coupler that Ribasim calls allocation each time the demands are updated from the outside, but that is not necessarily the case. The only thing a coupler can rely on is that allocation is called at each multiple of `saveat`, but that seems brittle / not what it was meant for. I think it would be the most clear if allocation updates on multiples of the supplied allocation timestep even if it is adaptive, but that is not compatible with the current config (and also enforces coupling on a fixed timestep). We could also design it such that the `update_until` `t` also becomes an update time for allocation, meaning that allocation has to keep track of its own `tstops`.

### DIFF
--- a/core/src/model.jl
+++ b/core/src/model.jl
@@ -278,71 +278,44 @@ function is_finished(model::Model)::Bool
 end
 
 """
-    step!(model::Model, dt::Float64)::Model
-
-Take Model timesteps until `t + dt` is reached exactly.
+Wrap the loopheader! call from the OrdinaryDiffEq.jl internals to inject the allocation call
+before a timestep when needed.
 """
-function step!(model::Model, dt::Float64)::Model
-    (; config, integrator) = model
-    (; t) = integrator
-    if config.experimental.allocation
-        dt_allocation = config.allocation.dt
-        if isnothing(dt_allocation)
-            error(
-                """
-                Ribasim was called with allocation active and adaptive timestepping from the BMI, which is currently not supported.
-                Supply a fixed allocation timestep in the Ribasim config.
-                """
-            )
-        end
-        ntimes = t / dt_allocation
-        if round(ntimes) ≈ ntimes
-            update_allocation!(model, dt_allocation)
-        end
+function OrdinaryDiffEqCore.loopheader!(integrator::DEIntegrator{<:Any, <:Any, RibasimCVectorType{Float64}})
+    (; p, t) = integrator
+    (; allocation) = p.p_independent
+    t_end::Float64 = integrator.sol.prob.tspan[end]
+    if is_active(allocation) && (t == allocation.time.t_allocation_next)
+        allocation_Δt = compute_adaptive_allocation_Δt(integrator, t_end)
+        update_allocation!(integrator, allocation_Δt, record = is_saveat_time(t, allocation.time.saveat, t_end))
+        allocation.time.t_allocation_next += allocation_Δt
+        add_tstop!(integrator, t + allocation_Δt)
     end
-    SciMLBase.step!(integrator, dt, true)
-    return model
+    return invoke(loopheader!, Tuple{DEIntegrator}, integrator)
 end
 
 """
 Compute and return  the adaptive allocation Δt if the timestepping is adaptive, otherwise return
 the fixed timestep.
 """
-function compute_allocation_Δt(model, saveat, tspan_end)::Float64
-    (; config, integrator) = model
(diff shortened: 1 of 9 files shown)


core/src/model.jl
```diff
@@ -278,71 +278,44 @@ function is_finished(model::Model)::Bool
 end
 
 """
-    step!(model::Model, dt::Float64)::Model
-
-Take Model timesteps until `t + dt` is reached exactly.
+Wrap the loopheader! call from the OrdinaryDiffEq.jl internals to inject the allocation call
+before a timestep when needed.
```
visr, line 282:
```suggestion
before a timestep when needed.

Workaround for pre-step callbacks: https://github.com/SciML/OrdinaryDiffEq.jl/issues/3977
```

core/src/read.jl
```diff
@@ -1683,10 +1683,17 @@ function Subgrid(db::DB, config::Config, basin::Basin)
 end
 
 function Allocation(db::DB, config::Config, graph::MetaGraph)::Allocation
+    dt_fixed = config.allocation.dt
+    time = if isnothing(dt_fixed)
+        AllocationTime(; adaptive = true, config.solver.saveat)
+    else
+        AllocationTime(; adaptive = false, dt_fixed, config.solver.saveat)
+    end
```
visr: Might be a bit cleaner to handle this in the AllocationTime constructor?

core/src/parameter.jl
```diff
@@ -313,6 +313,19 @@ struct AllocationControlRecordDatum
...
+"""
+@kwdef mutable struct AllocationTime
+    const adaptive::Bool = true
+    # The fixed timestep if applicable
+    const dt_fixed::Float64 = 0.0
+    # The time (s) when allocation is called next
+    t_allocation_next::Float64 = 0.0
+    # The output writing interval
+    const saveat::Float64 = 0.0
+end
```
visr, line 340: Might be good to use an internal constructor to ensure that e.g. `dt_fixed` is 0.0 when `adaptive`.

visr (approved):
Of course not ideal to wrap `loopheader!` like this, but it is definitely a good improvement over the current situation.

SouthEndMusic:
@visr you already approved the changes but I made some substantial changes to the allocation timestep handling afterwards

visr (approved):
Looks good. I added an inner constructor to check the 0.0 when adaptive.

PR #3221, opened 2026-08-26
Fix allocation + Junction bugs

### LINKED ISSUE #2390: Map links in allocation flow results similar to the physical layer (due to Junction)
Internally in Ribasim junction nodes are removed, so that's a different graph than what is schematized. Therefore when writing flow outputs this is mapped back. This must also be done for the allocation flow output.

### LINKED ISSUE #3220: Allocation with adaptive timestep gives KeyError for Junction
When running the Meuse model with allocation enabled and the default adaptive allocation timestep, the simulation fails with:
```
Warning: The following experimental features are enabled: allocation
ERROR: KeyError: key Junction #3 not found
```
Junction ID 3 does exist in the model.

According to @SouthEndMusic, Junctions should normally be filtered out internally and should not cause errors during allocation.

The same model was being used to test the difference between adaptive and fixed allocation timesteps (#3196).

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3220
Fixes https://github.com/Deltares/Ribasim/issues/2390

### DIFF
--- a/core/src/Ribasim.jl
+++ b/core/src/Ribasim.jl
@@ -66,7 +66,7 @@ using SparseConnectivityTracer: GradientTracer, TracerSparsityDetector
 using SparseMatrixColorings: GreedyColoringAlgorithm, sparsity_pattern
 
 # For efficient sparse computations
-using SparseArrays: SparseMatrixCSC, sparse, nzrange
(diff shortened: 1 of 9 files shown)


visr (approved):
Nice!

Issue #3046, opened 2026-04-23
Make debugging unstable models easier


evetion:
Depending on the method of automation, sometimes models fail to run, erroring very quickly with unstable, seemingly because of too small a timestep. Any suggestions on how to tackle such a problem?

Note that before I had -9999 values in among others, ManningResistance / Static length and that didn't error (neither in Python or Julia).

[test_model_pfafstetter.zip](https://github.com/user-attachments/files/27004690/test_model_pfafstetter.zip)

```
┌ Debug: Setup integrator.
└ @ Ribasim /Users/evetion/code/Ribasim/core/src/model.jl:240
┌ LogLevel(-1): Simulating
│   message = dt=2.00000000000001e-105
t=0.0
max u=4.572991727564276e-37
│   progress = 0.0
└ @ OrdinaryDiffEqCore /Users/evetion/.julia/packages/OrdinaryDiffEqCore/tusSd/src/integrators/integrator_utils.jl:589
┌ LogLevel(-1): Simulating
│   message = dt=2.00000000000002e-205
t=0.0
max u=4.572991727564276e-37
│   progress = 0.0
└ @ OrdinaryDiffEqCore /Users/evetion/.julia/packages/OrdinaryDiffEqCore/tusSd/src/integrators/integrator_utils.jl:589
┌ LogLevel(-1): Simulating
│   message = dt=2.000000000000032e-305
t=0.0
max u=4.572991727564276e-37
│   progress = 0.0
└ @ OrdinaryDiffEqCore /Users/evetion/.julia/packages/OrdinaryDiffEqCore/tusSd/src/integrators/integrator_utils.jl:589
┌ LogLevel(-1): Simulating
│   message = dt=5.0e-324
t=0.0
max u=4.572991727564276e-37
│   progress = done
└ @ OrdinaryDiffEqCore /Users/evetion/.julia/packages/OrdinaryDiffEqCore/tusSd/src/integrators/integrator_utils.jl:410
┌ LogLevel(-1): Simulating
│   message = dt=5.0e-324
t=0.0
max u=4.572991727564276e-37
│   progress = done
└ @ OrdinaryDiffEqCore /Users/evetion/.julia/packages/OrdinaryDiffEqCore/tusSd/src/integrators/integrator_utils.jl:410
```

visr:
I am just looking at the parameters in QGIS a bit, and sorting to see the extreme values. That shows some issues. Don't know if that's it but it would be good to address those anyway.

[...]

PR #3236, opened 2026-08-31
Switch default solver to NordsieckBDF

### LINKED ISSUE #3212: A new best BDF: NordsieckBDF
Title quotes Chris in his JuliaCon 2026 talk: https://www.youtube.com/live/1r6F9dIZ_gM?si=3gYl_ohjmAVsg0dz&t=27885

We should compare to QNDF to NordsieckBDF, added in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4017

In https://github.com/SciML/OrdinaryDiffEq.jl/pull/4150 its default is set to `nlsolve = NLNewton(max_iter = 3)`, which might be a good idea for QNDF as well, worth exploring.

But if it is faster, we should consider making it the new default.

visr:
Fixes https://github.com/Deltares/Ribasim/issues/3212

Docstring here: https://github.com/SciML/OrdinaryDiffEq.jl/blob/b7a983f325087c34d92c77e795bab337662894f3/lib/OrdinaryDiffEqBDF/src/algorithms.jl#L770-L817

I ran this over several larger models, and saw a decent ~10% computation time decrease. Here is a summary of [solver stats](https://ribasim.org/reference/usage.html#solver-statistics---solver_stats.nc) differences for a Zuiderzeeland model run:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/PROGRA~3/DevDrive/cache/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/PROGRA~3/DevDrive/cache/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

Metric | QNDF Average | NordsieckBDF Average | Nordsieck vs QNDF
-- | -- | -- | --
Computation time | 256.99 | 229.98 | -10.5%
RHS calls | 117.88 | 98.85 | -16.1%
Linear solves | 116.85 | 97.81 | -16.3%
Accepted   timesteps | 48.73 | 53.63 | 10.1%
Rejected   timesteps | 4.76 | 1.54 | -67.6%
dt | 4,544.01 | 4,334.34 | -4.6%

</body>

</html>

The `dt` does go down a bit. Though note that this is not the average dt, but the last one before the end of the saveat.

### DIFF
--- a/core/src/config.jl
+++ b/core/src/config.jl
@@ -19,7 +19,7 @@ using OrdinaryDiffEqNonlinearSolve: NLNewton
 using OrdinaryDiffEqLowOrderRK: Euler, RK4
 using OrdinaryDiffEqTsit5: Tsit5
 using OrdinaryDiffEqSDIRK: ImplicitEuler, KenCarp4, TRBDF2
-using OrdinaryDiffEqBDF: FBDF, QNDF
+using OrdinaryDiffEqBDF: FBDF, QNDF, NordsieckBDF
 using OrdinaryDiffEqRosenbrock: Rosenbrock23, Rodas4P, Rodas5P
 import OrdinaryDiffEqDifferentiation
 using LinearSolve:
@@ -178,7 +178,7 @@ for (node_type, kinds) in pairs(node_kinds)
 end
 
 @option struct Solver <: TableOption
-    algorithm::String = "QNDF"
+    algorithm::String = "NordsieckBDF"
     saveat::Float64 = 86400.0
     dt::Union{Float64, Nothing} = nothing
     dtmin::Float64 = 0.0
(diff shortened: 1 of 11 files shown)


SouthEndMusic:
I'm planning to fix (or at least investigate) the model failure(s) here with the same tooling as mentioned [here](https://github.com/Deltares/Ribasim/issues/3046#issuecomment-5480262743).

SouthEndMusic (approved):
LGTM

PR #3237, opened 2026-08-31
Improved residual and convergence formulation

### LINKED ISSUE #3231: Split improved convergence data collection from #3191

### LINKED ISSUE #3232: Split improved residual formulation from #3191

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3231
Fixes https://github.com/Deltares/Ribasim/issues/3232

In this PR:
- The `reltol` decreasing mechanism was removed, and the residual formulation was changed to one better suited to the cumulative flow formulation; from basing the relative tolerance term on the absolute value of the state to basing it on the change change in the value of the state over the timestep.
- The residual data for the convergence output is now collected for all Newton iterations instead of only for the last one, giving a better indication of which states caused slow convergence.
- The `level_range` model (which is defined separately in the docs and as a testmodel) was adjusted with histeresis because with this new formulation `DiscreteControl` started oscillating very rapidly (see the point below).
- A mechanism was added to `DiscreteControl` where it is checked whether subsequent control state changes of a particular `DiscreteControl` node are less than ` minimal_discrete_control_update_dt` (default 1 second) apart.


docs/reference/node/discrete-control.qmd
```diff
@@ -109,3 +109,7 @@ column         | type     | unit | restriction
 node_id        | Int32    | -    |
 truth_state    | String   | -    | Consists of the characters "T" (true), "F" (false) and "*" (any)
 control_state  | String   | -    | -
+
+::: {.callout-note}
+It is possible to configure a DiscreteControl node such that it very rapidly oscillates between control states. To make you aware of this, Ribasim will throw an error if successive control state changes of a DiscriteControl node in simulated time are closer than `minimal_discrete_control_update_dt`, which is configurable.
```
visr:
```suggestion
It is possible to configure a DiscreteControl node such that it very rapidly oscillates between control states. To make you aware of this, Ribasim will throw an error if successive control state changes of a DiscreteControl node in simulated time are closer than `minimal_discrete_control_update_dt`, which is configurable.
```

core/test/docs.toml
```diff
@@ -56,6 +56,7 @@ evaporate_mass = true  # optional, default true to simulate a correct mass balan
 depth_threshold = 0.1  # optional, default 0.1
 max_depth = 2000.0 # optional, default 2000.0
 level_difference_threshold = 0.02 # optional, default 0.02
+minimal_discrete_control_update_dt = 1.0 # optional, default 1.0
```
visr: In other options we use `min` instead of `minimal`. How about `min_discrete_control_interval`?

docs/reference/node/discrete-control.qmd
visr: It might be better to put this under the condition section. There the thresholds are also set. It would be good to mention adding a separate `threshold_low` as a possible solution.

Providing an example of a simple condition that will hit this is also valuable to make users understand what is going wrong, perhaps even with a picture. So perhaps this goes beyond a simple callout. A mention in https://ribasim.org/getting-started/examples.html#model-with-discrete-control might also be good.

SouthEndMusic: I added a simple example model that runs the risk of having rapid oscillations. If you want to add how to fix it with histeresis I think it's better if you explain that.

visr:
Converting to draft since there are still some large performance regressions we don't understand, for instance in `hws_transient`.

SouthEndMusic:
I burned some credits on it:

## Re-tuning `abstol`/`reltol` for the new residual formulation

### Motivation

The new `calculate_residuals!` scales each state by the magnitude of its **change** over the timestep rather than its absolute magnitude:

```
abstol + |u₁ - u₀|·reltol        (new)
abstol + max(|u₀|,|u₁|)·reltol   (default SciML)
```

This is the right thing to do — the states are cumulative volumes, so their absolute value carries no information about the local error, and the old behaviour meant the tolerance kept loosening as the simulation progressed (which is exactly what the now-removed `decrease_tolerance!` callback was hacking around).

But it silently changes the **units** of both tolerances. They no longer mean "tolerance on the cumulative volume" but "tolerance on the volume passed per timestep". Keeping the old numeric defaults therefore makes the criterion drastically stricter, which is the slowdown we observed.

### Quantifying the change in strictness

On `hws_transient`, sampling the residual denominator over 200 steps:

| | median | p90 | max |
|---|---|---|---|
| denominator, new | 1.0e-5 | 2.3e-2 | 1.6e0 |
| denominator, old | 1.0e-5 | 1.4e0 | 9.1e1 |
| **ratio old/new** | **1.0** | **97×** | **8300×** |

Two separate effects:

1. **The `reltol` term** is ~100× tighter at p90, because `|Δu| ≪ |u|`.
2. **`abstol` has become the binding constraint.** 68% of residuals are now `abstol`-floored. On main the floor essentially never bound, since `|u|·reltol ≈ 1e7 · 1e-5 = 100 m³`; now the solver is chasing a 1e-5 m³ absolute volume tolerance.

The effect on the stepsize is large — after 200 steps main has reached t = 8736 s (dt = 180 s) while the branch is only at t = 4728 s (dt = 5.8 s).

### Choosing new defaults

To avoid just trading accuracy for speed, I ran an **iso-accuracy** study: each configuration was compared against a tight reference solution (`abstol = reltol = 1e-8`), measuring the max relative error across the states carrying actual flow. Two models were used, spanning ~4 orders of magnitude in flow volume: `hws_transient` and `basic_transient`.

**Recommendation: `abstol = 1e-4`, `reltol = 1e-4`.**

| model / horizon | config | rhs calls | max rel. error |
|---|---|---|---|
| hws_transient, 10 h | main default | 780 | 1.08e-3 |
| | branch @ 1e-5/1e-5 | 1097 | 2.27e-5 |
| | **branch @ 1e-4/1e-4** | **646** | **3.11e-4** |
| hws_transient, 2 d | main default | 1266 | 2.67e-5 |
| | branch @ 1e-5/1e-5 | 1795 | 5.96e-6 |
| | **branch @ 1e-4/1e-4** | **1011** | **7.68e-5** |
| basic_transient, 7 d | main default | 303 | 5.93e-5 |
| | branch @ 1e-5/1e-5 | 396 | 1.60e-5 |
| | **branch @ 1e-4/1e-4** | **279** | **8.34e-5** |

So `1e-4 / 1e-4` lands within a factor ~3 of main's accuracy on all three cases (better on hws at 10 h, slightly worse elsewhere) while being **cheaper than main** in every case, and ~1.6× cheaper than the branch with the old defaults.

### Why not simply raise `abstol` further

This is the main trap, and worth calling out. Raising `abstol` looks very attractive on `hws_transient` (`abstol = 1e-2` gives 859 rhs calls at 8.3e-5 error), but `abstol` is now an **absolute** tolerance on per-step volume and therefore does **not** scale with model size. The same setting on `basic_transient` gives a max relative error of 7.1e-3 — roughly 100× worse than main. `1e-4` is the largest value that is safe across both scales.

Raising `reltol` past `1e-3` is also risky: `hws_transient` at `reltol = 1e-2` failed to converge (`EEst ≈ 4.2e8`, aborted on `dtmin`).

### Changes in this commit

- config.jl: `abstol`, `reltol` `1e-5` → `1e-4`, with a comment on what they now scale against.
- config.py: same defaults (these were also stale: `1e-6`/`1e-5`, and the docstrings claimed `1e-7`).
- numerics.qmd: the "Compensating for cumulative flows" section still documented the removed `decrease_tolerance!` callback. Rewritten to describe the new residual scaling and to warn that `abstol` no longer scales with model size.

### Caveat

The spreadsheet that triggered this investigation shows a **27×** slowdown (170816 vs 6151 rhs calls), while `hws_transient` only reproduces ~1.4× in rhs calls. The mechanism is the same, but that magnitude suggests the model in question hits the `abstol` floor far harder — likely one with small per-step flows. It would be worth re-running that specific model with the proposed defaults to confirm they are sufficient there too.

visr (approved):

PR #3238, opened 2026-08-31
Lock the Runic formatter version

visr:
Runic v1.10 was just released, and we didn't lock the version in our pre commit hook so far, leading to CI failures. We could do that, but the whole setup was unnecessarily complicated, and can be made simpler. The biggest downside was that we relied on a global, non-locked install of Runic, and that we needed the JETLS extension so we could ask that to use Runic as a formatter. Luckily the current julia-vscode pre-release now also support runic as a formatter, via [JuliaFormat.toml](https://www.julia-vscode.org/docs/dev/userguide/configuration/#Formatting:-JuliaFormat.toml).

That means we can drop both the app install and JETLS recommendation (it can still be useful).

I added some extra new options from Runic to make it display the diff using git, and format julia code in docstrings as well.

One downside is that the Runic version bundled with julia-vscode is not controlled by us and lags behind. But at least prek fixes is right.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,7 +12,6 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "aviatesk.jetls-client",
                 "charliermarsh.ruff",
                 "GitHub.copilot",
                 "julialang.language-julia",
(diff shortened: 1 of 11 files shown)


SouthEndMusic (approved):
LGTM

PR #3241, opened 2026-09-01
Use prek.toml

visr:
This aligns better with Wflow, and the TOML is recommended by Prek: https://prek.j178.dev/configuration/#file-format


pixi.toml
```diff
@@ -64,8 +64,10 @@ docs = { cmd = "PYTHON_JULIAPKG_PROJECT=$PIXI_PROJECT_ROOT quarto preview docs/{
 ], env = { "PYTHON_JULIACALL_CHECK_BOUNDS" = "yes", "PYTHON_JULIACALL_HANDLE_SIGNALS" = "yes", "PYTHON_JULIAPKG_OFFLINE" = "yes" } }
 # Lint
 typecheck = "ty check"
-prek-run = "prek run --all-files"
```
SouthEndMusic, line 67: I feel like the names `check` and `fix-check` can be a bit more descriptive about what is being checked. I also don't really understand the need for and the logic of `fix-check`.

visr: In a way it is intentionally undescriptive about what is being checked, because we'd basically have to list all the prek hooks, and change it each time. I wanted to avoid using the name `prek` or `pre-commit`, since that is an implementation detail. One other name could be `hook`, though that may be less descriptive than `check`.

The `fix-check` is meant for automated PRs. If they for instance write files without newlines, or a new formatter version with slightly different formatting, we don't want CI to fail because it didn't run the git hooks. Also running it once will give a nonzero exit code if it makes any changes, causing it to report as failed. So we run it once, and if the return code is nonzeto, run it again. If say it had to fix newlines, it will have fixed it on the first one, and verify on the second one that all checks pass.

SouthEndMusic (approved):

PR #3248, opened 2026-09-03
Reduce memory footprint of delwaq generate

visr:
@jvleeuwen00 ran into this MemoryError

image


evetion (approved):
So xarray provides two dimensions (time and linkid) with it? Bit implicit, but great.

visr:
No those two are coordinates: https://ribasim.org/reference/usage.html#flow---flow.nc

Issue #16, opened 2022-10-28
coupling Wflow SBM


visr:

visr:
This is the oldest open Ribasim issue. It would be nice to have a working prototype for this, without committing to anything stable yet. The demand is there, and it might not be where we want to go in the end, but it will enable some experimental usage, which will also inform us on next steps. Right now we are thinking about the irrigation module, [[#3159](https://github.com/Deltares/Ribasim/issues/3159)](https://github.com/Deltares/Ribasim/issues/3159). There is overlap with this approach, but this is more complete, because Wflow would cover the entire land domain, and Ribasim the entire water domain.

Here are some suggestions for a minimal prototype:
- Support only fully overlapping model domains, where Wflow takes care of the land and Ribasim of the water.
- Wflow and Ribasim both use as much of their normal code paths as possible, both are initialized from their own TOML, and keep their own data model
- Should work with models generated by `hydromt_wflow` and `hydromt_ribasim`.
- Need something like [`[node_basin_mapping.py](https://github.com/Deltares/imod_coupler/blob/main/pre-processing/primod/mapping/node_basin_mapping.py)`](https://github.com/Deltares/imod_coupler/blob/main/pre-processing/primod/mapping/node_basin_mapping.py) to generate the coupling file.
- The coupling code can live in a separate module in the Ribasim (or Wflow) code for now.
- Disable Wflow surface-water and river routing (river, lake, reservoir, but keep e.g. overland and subsurface routing). Wflow produces locally generated hydrological fluxes, which are spatially aggregated and supplied to Ribasim Basins. Ribasim is solely responsible for routing and storage in the surface-water network.
- Irrigation would be done in Wflow, but water demand goes to Ribasim
	- Irrigation is excluded from the first prototype, but the coupling driver and mapping format should allow a later Ribasim → Wflow exchange of allocated irrigation water.
- The coupled code accepts a TOML like:

```toml
wflow_toml = "path/to/wflow.toml"
ribasim_toml = "path/to/ribasim.toml"
coupling = "coupling.gpkg"
```

Out of scope for the prototype:
- Since both codes are in Julia, it is faster to develop the prototype in Julia as well. Porting to [[iMOD coupler](https://github.com/Deltares/imod_coupler)](https://github.com/Deltares/imod_coupler) once it works well can be discussed afterwards. That would also need e.g. https://github.com/Deltares/Wflow.jl/issues/998
- Upstream Wflow-only domain with river routing.
[...]

PR #3251, opened 2026-09-07
Add BMI support for setting FlowBoundary flow rates

visr:
Adds minimal BMI support for setting FlowBoundary flow rates, exposed as `flow_boundary.flow_rate` (settable) and `flow_boundary.cumulative_flow` (read-only). This is needed to support online coupling (#16).

## Why a separate `flow_rate_bmi` vector

The other BMI-writable variables already *are* plain `Vector{Float64}` "current value" arrays:

- `basin.vertical_flux.{precipitation, drainage, infiltration, surface_runoff}` — the docstring of `VerticalFlux` even says *"Current forcing is stored as separate array for BMI access. These are updated from BasinForcing at runtime."* A `PresetTimeCallback` refreshes them from the interpolations at forcing tstops, and the physics in `set_current_basin_properties!` consumes them as `vertical_flux.x * dt` (piecewise constant).
- `user_demand.demand` — same pattern, refreshed from `demand_interpolation`.

So for those, the interpolation is *not* on the hot path; the `Float64` array is the canonical runtime state, and a BMI write survives until the next callback. There is nothing to add.

`FlowBoundary` is structurally different: it has no current-value array. `formulate_flow_boundary!` and `flow_update_on_link` use `integral(flow_rate[idx], tprev, t)` — the **exact** integral of the interpolation over the step. To make it look like Basin we would have to introduce a current-value array plus a callback and switch to `rate * dt`, which degrades exact integration to piecewise-constant sampling and changes results (it visibly shifted the allocation `.lp` snapshots).

`flow_rate_bmi` is therefore an *override layer* rather than a replacement representation: per index, `NaN` means unset, no new callback, and a strict no-op when unset. It composes with the exact-integration path, since an externally supplied rate is constant in time (`flow_rate_bmi * (t2 - t1)`). `NaN`-as-unset is already the idiom here (`Pump.flow_rate`, `Outlet.flow_rate`, `FlowDemand.demand`).

All reads of the flow boundary rate now go through two small helpers in `parameter.jl`, `boundary_flow_rate` and `boundary_flow_integral`, so the override is honoured consistently by the ODE right hand side, the flow/concentration bookkeeping, discrete control look-aheads and the allocation LP.


SouthEndMusic (approved):
Good stuff. As discussed, there are some subtle cases where the behavior for `FlowBoundary`  and basin forcing is different:
- The BMI-specified boundary flow is leading once it has been set, irrespective of the Ribasim static/time input for `FlowBoundary`
- When Basin forcings are set from the BMI, they can be overwritten by forcing updates from `Basin / time`.

This only leads to potentially unexpected behavior if there are multiple data sources for these flows, which can be considered poor schematization practice

PR #3252, opened 2026-09-07
Store the logger on the Model instead of setting it globally

visr:
## Problem

`BMI.initialize` installed its `FileLogger` via `global_logger`. That has two consequences:

1. A task-local logger takes precedence over the global one, so any caller running Ribasim under `with_logger` (the VS Code Julia test runner does this) silently got an empty `ribasim.log`.
2. Initializing a second Model clobbered the first one's logger, so both models logged to the last opened file.

`BMI.finalize` also had to recover the log stream by reaching into the logger internals (`logger.logger.loggers[1].logger.logger.stream`) to close it.

## Change

The `Model` now carries the `logger` it should log to, plus the `log_io` it owns, and every BMI entry point runs its body under `with_logger(model.logger)`. The logger defaults to `current_logger()` at construction, so building a Model outside BMI keeps behaving as before.

- `Model` gains `logger::AbstractLogger` and `log_io::Union{IOStream, Nothing}` fields. `logger` is deliberately left abstract: `Base.CoreLogging.LogState` stores an `AbstractLogger` too, so a type parameter would only add specializations without helping dispatch.
- `global_logger` is gone from `bmi.jl`; `finalize` closes `model.log_io` directly.
- Removed `logger_stream`, which depended on the internal layout of the composed logger.

The CLI path in `main.jl` was already correct and is unchanged; the BMI path now mirrors it.

## Tests

`"BMI logging"` no longer depends on the ambient logger and asserts the log file contents. Added `"BMI logging of multiple models"`, which initializes two models and checks each log file names its own TOML — this fails on `main`.


SouthEndMusic (approved):
Looks good. Just a thought: if we have to wrap a lot of function bodies with `with_logger(model.ligger) do ... end`, it might be worth adding a simple macro for that.

visr:
Ha I had a macro, but didn't feel like it was worth it. It also becomes harder to read.

PR #3253, opened 2026-09-07
Report average timestep size in solver stats

visr:
The `dt` column in `results/solver_stats.nc` was the size of the last timestep before the `saveat` time, which is not very representative of the interval.

It is now the average timestep size over the `saveat` interval, computed as the interval length divided by the number of accepted timesteps in it. Since `dt` is no longer needed per save, it is dropped from the `SolverStats` named tuple.

Docs and changelog updated.


SouthEndMusic (approved):
Good stuff

PR #3254, opened 2026-09-07
Small fixes

### LINKED ISSUE #3198: Allocation timestep behaviour is difficult to interpret (documentation/error message)
For one of my intermediate conversion models (Script 5), I compared the same three configurations as in #3196.

In this case, however, the behaviour is different:

1. allocation off: runs successfully
2. allocation on (adaptive): runs successfully (although about 22 minutes slower than allocation off)
3. allocation on with Allocation(dt=86400): the simulation reaches 100%, writes the results, but finishes with an Unstable return code.

I wanted to understand why. At first I assumed the instability itself was the problem, but after digging into the results I found that the two simulations already start to differ much earlier.

By comparing the outputs step by step, I found that:
- both simulations start from the same initial conditions;
- the first differences already appear after the first allocation update;
- allocation_flow.nc already contains different optimized allocation results;
- these differences are reflected in allocation_control.nc;
- only after that do the hydraulic results (basin.nc) start to diverge.

While tracing this back, the first difference I found was on an allocation link connected to an allocation-controlled pump (Pump_Be_NavGenkSluice). I'm not saying this pump is causing the instability, only that this was the first place where I could see the two runs behaving differently. During the investigation I also noticed something I couldn't interpret. The optimized allocation flow on that link was about 59418 m³/s in one run, while the pump itself has control states of 0 m³/s and 8 m³/s. I couldn't determine from the documentation how allocation_flow.nc should be interpreted for allocation-controlled pumps, or how this optimized allocation flow relates to the physical pump capacity. Some clarification on this would also help when debugging allocation behaviour.

What confused me as a user is understanding why changing only the allocation timestep already changes the allocation results from the very first allocation update. As a modeller, I would normally expect changing a timestep to mainly affect performance or numerical accuracy. I didn't expect it to immediately lead to different allocation decisions and eventually a different hydraulic solution.

**Error messages**

Another thing I found difficult is understanding why the model became unstable.
The convergence warning reports:

```
UserDemand #490 = 100.00%
UserDemand #1046 = 100.00%
UserDemand #509 = 93.75%
TabulatedRatingCurve #13 = 91.93%
TabulatedRatingCurve #1192 = 84.37%
```

The warning tells me where the solver eventually struggled right? But not necessarily where the divergence started or where I should begin investigating. This took me a while, but still nothing conclusive. Also, the Allocation(dt=86400) run finishes with an Unstable return code, the allocation analysis is not written either, making it a bit harder to investigate.

It would be helpful if the warning messages or the documentation gave a bit more guidance from a modeller's perspective. For example:
- When should I use adaptive allocation versus a fixed dt?
- Should I expect changing only dt to produce different allocation results, or mainly affect performance?
- How sensitive are the results expected to be to this choice?
- If two runs start to diverge immediately, is that expected behaviour?
- What exactly do the reported "convergence bottlenecks" represent?
As a modeller, I initially interpreted them as hydraulic or allocation bottlenecks (for example a shortage somewhere in the network). However, I assume they actually are meant in a numerical sense?
- When an unstable run reports 'convergence bottlenecks', are these always the cause of the instability? 
- Should I start by investigating those reported nodes, or is it better to first inspect outputs such as allocation_flow.nc and allocation_control.nc?

I think a bit more guidance on how to interpret these warnings and where to start debugging would make it much easier for users to understand and troubleshoot unstable allocation runs.

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3198

This PR:
- Adds a `dtmax` for the adaptive allocation timestep alongside the `dtmin` which already existed
- Clarifies that the instability logging is for the physical layer
- Fixes that errors with the demand input parsing did not throw, allowing invalid model runs

We chose for now not to add the linearization error output as suggested in the issue, because it is probably hard for the modeler to interpret and would not often be used.


visr (approved):
Looks good!

PR #3259, opened 2026-09-09
Require and bound allocation controlled flow with `max_flow_rate`

### LINKED ISSUE #3258: Better allocation flow bounds
Allocation works best when variable bounds are as strict as possible. Currently the bounds on flows are only determined during initialization, whereas they can be updated at each run of the allocation algorithm with up to date data for strict bounds. This will probably make the optimization problems better conditioned. 

To achieve good bounds, validation will also be added that allocation controlled Pumps and Outlets have finite `max_flow_rate`.

This issue was inspired by @Fati-Mon's failing Meuse model.

SouthEndMusic:
Fixes https://github.com/Deltares/Ribasim/issues/3258

The large diff is mainly due to the changes in the checked-in allocation optimization problems.


SouthEndMusic:
I let Claude update the allocation docs. I don't think `docs/dev/allocation.qmd` is worth maintaining, because it copies a lot of code which is bound to get stale, unless we introduce references or checks to prevent that.

utils/write_allocation_problems.jl
```diff
@@ -34,6 +35,7 @@ function (@main)(ARGS)::Cint
                     if !isempty(allocation_models)
                         model_dir = normpath(results_path, model_name)
                         mkdir(model_dir)
+                        BMI.update(model)
```
visr: Would the .lp files be more stable if we would do an `update_until(model, 86400.0)` here? Right now the size of the step is left up to the solver.

Or even run the whole simulation period.

docs/concept/allocation.qmd
```diff
@@ -417,7 +419,13 @@ $$
...
+When Pumps and Outlets are part of a subnetwork, they can be controlled by allocation. To accomplish this, they must have `allocation_controlled` set to `true` in the static table for [Pump](/reference/node/pump.qmd#static) or [Outlet](/reference/node/outlet.qmd#static). When a Pump or Outlet is allocation controlled, the flow through the node is only bounded by its capacity and Pump or Outlet specific constraints based on the difference between upstream and downstream level (see [Pump equations](/reference/node/pump.qmd#equations) or [Outlet equations](/reference/node/outlet.qmd#equations)).
+
+:::{.callout-important}
+Because the allocation problem must be bounded, an allocation controlled Pump or Outlet is required to have a finite `max_flow_rate`. This is validated at the start of the simulation, for the initial parameters as well as for all DiscreteControl states of the node. Note that Pumps and Outlets that connect the primary network to a secondary network are automatically made allocation controlled, so they also require a finite `max_flow_rate`.
```
visr: > automatically made allocation controlled

This seems a bit implicit, wouldn't it be better to just require this as well?

core/test/data/allocation_problems/two_basin_user_demand/allocation_problem_1.lp
```diff
@@ -1,40 +1,45 @@
...
 average_flow_unit_error_constraint_2_: -1 user_demand_error_UserDemand_#5,2,first_ + 1 average_flow_unit_error_2_ = 0
-average_storage_unit_error_constraint_1,upper_: -1 level_demand_error_Basin_#3,1,upper,first_ - 1 level_demand_error_Basin_#4,1,upper,first_ + 1 average_storage_unit_error_1,upper_ = 0
-average_storage_unit_error_constraint_1,lower_: -1 level_demand_error_Basin_#3,1,lower,first_ - 1 level_demand_error_Basin_#4,1,lower,first_ + 1 average_storage_unit_error_1,lower_ = 0
+average_storage_unit_error_constraint_1,upper_: -1 level_demand_error_Basin_#3,1,upper,first_ - 1 level_demand_error_Basin_#4,1,upper,first_ + 2 average_storage_unit_error_1,upper_ = 0
+average_storage_unit_error_constraint_1,lower_: -1 level_demand_error_Basin_#3,1,lower,first_ - 1 level_demand_error_Basin_#4,1,lower,first_ + 2 average_storage_unit_error_1,lower_ = 0
 Bounds
--0.001 <= basin_storage_change_Basin_#3_ <= 0.004
--0.001 <= basin_storage_change_Basin_#4_ <= 0.004
+-1 <= basin_storage_change_Basin_#3_ <= 1
+-1 <= basin_storage_change_Basin_#4_ <= 1
```
visr, line 30: All bounds are tighter, except the basin storage changes. Do you know why?

core/test/data/allocation_problems/switch_allocation_control/allocation_problem_1.lp
```diff
@@ -1,23 +1,28 @@
...
-0 <= flow_(LevelBoundary_#5,_Outlet_#4)_ <= 0.001
-0 <= flow_(Outlet_#4,_Basin_#1)_ <= 0.001
-0 <= flow_(Basin_#1,_Outlet_#3)_ <= 8.0e-5
-0 <= flow_(Outlet_#3,_Terminal_#6)_ <= 8.0e-5
-boundary_level_LevelBoundary_#5_ = 0
+flow_(LevelBoundary_#5,_Outlet_#4)_ = 0
+flow_(Outlet_#4,_Basin_#1)_ = 0
+0 <= flow_(Basin_#1,_Outlet_#3)_ <= 0.00032914285714285717
+0 <= flow_(Outlet_#3,_Terminal_#6)_ <= 0.00032914285714285717
+boundary_level_LevelBoundary_#5_ = 1
```
visr, line 25: I see the level is static and set to 1, so that seems good. The old constraint was just wrong then?

core/test/data/allocation_problems/multi_level_demand/allocation_problem_2.lp
```diff
@@ -1,31 +1,37 @@
 minimize
-obj: 
+obj: -1 low_storage_factor_Basin_#2_
```
visr, line 2: Hmm the objective is to maximize the low storage factor? Interesting.

visr (approved):

core/test/data/allocation_problems/multi_level_demand/allocation_problem_2.lp
SouthEndMusic: This is after all demand objectives, there is only the route priority objective after this. The low storage factor is applied as `q * low_storage_factor`, where `q` is fixed (e.g. negative forcing). The low storage factor maximization makes sure as much as possible of `q` is abstracted. That's different from the physical layer, but this is what's possible in the linear allocation world.

core/test/data/allocation_problems/two_basin_user_demand/allocation_problem_1.lp
SouthEndMusic: The Basin storage changes are in a different category because they have units of volume instead of flow

core/test/data/allocation_problems/switch_allocation_control/allocation_problem_1.lp
SouthEndMusic: I think this is due to the change in `write_allocation_problems.jl` where the allocation algorithm is actually run before writing the problems, whereas before the allocation problems were only initialized, partly with dummy data.

PR #3260, opened 2026-09-10
Update to Julia v1.13.0

visr:
https://julialang.org/blog/2026/09/julia-1.13-highlights/

Also updates all Julia dependencies.

See the separate commits, because a few failures had to be fixed.


SouthEndMusic (approved):
Good stuff

PR #3267, opened 2026-09-20
Fix numerical instability logging for states without an inflow link

visr:
Independent of #3237.

`SciMLBase.log_numerical_instability` looked up a node name with
`state_inflow_link[state_idx].link[2]` for every state index. Only the horizontal flow states,
which come first, have an inflow link; the evaporation, infiltration and PID integral states do
not, so indexing past the end throws. Logging a numerical instability could therefore itself
error, replacing the diagnostic with an unrelated stack trace exactly when the user needed it.

This adds a `state_label` helper that falls back to naming the state after its component
(`evaporation 12`, `integral 3`) and uses it at both call sites.

Not covered by an automated test: reproducing it needs a model that goes unstable *through* one
of those states. Suggestions welcome if it is worth constructing one.


core/src/solve.jl
```diff
@@ -1005,6 +1005,19 @@ end
...
 
+"""
+Describe the state at the given index for logging. Only the horizontal flow states, which
+come first, are associated with a link; the remaining states are named after their component.
+"""
+function state_label(u::CVector, state_inflow_link, state_idx::Int)::String
+    state_idx <= length(state_inflow_link) &&
+        return string(state_inflow_link[state_idx].link[2])
+    for (name, range) in pairs(getaxes(u))
+        state_idx in range && return "$name $(state_idx - first(range) + 1)"
```
HendrikKok, line 1016: Is it guaranteed that the state layout in u always consists of contiguous ranges per component?

HendrikKok (approved):

core/src/solve.jl
visr: Yes, the constructor enforces this:
https://github.com/Deltares/Ribasim/blob/f965a3266a4685bf10f3458aaa1855d09fa45a7a/core/src/cvectors.jl#L30

